Issue #46, opened 2017-01-28
Ubuntu 14.04, Debian 8.7, CentOS 7 launch errors (libstdc++.so.6)


SergiySW:
Cannot launch rai_wallet at Ubuntu 14.04 LTS. No errors shown
rai_node launch errors
./rai_node: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./rai_node)
./rai_node: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./rai_node)

Debian 8.7 rai_node launch error
./rai_node: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./rai_node)

CentOS 7 rai_node launch errors
./rai_node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./rai_node)
./rai_node: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by ./rai_node)

PR #47, opened 2017-01-28
Link statically with libgcc and libstdc++ #46

drizzt:


clemahieu:
Will this cause issues if we load any shared libraries?  If we load a .so and it uses the c++ lib .so, are file handles etc going to work between them?

drizzt:
it ONLY links libgcc and libstdc++ statically, any other library is dynamic

clemahieu:
Ok sounds good, I guess those all build on libc.

PR #98, opened 2017-03-08
Fix build on linux i686

drizzt:
CRYPTOPP_DISABLE_ASM should be defined for any targets since it's also
used in some includes files used by non-cryptopp targets


SergiySW:
Looks like with https://github.com/clemahieu/raiblocks/commit/113689f076dc612b71e13b6ecb324208c9af14f2 these changes are no longer actual

Issue #103, opened 2017-04-11
[Suggestion] Minor suggestions for wallet units


RaiBlocksDisciple:
Set Mrai units as default option in the wallet, as Mrai is used throughout the site, current default can cause confusion for the inexperienced and is rarely used.

Also, consider adding Grai unit.

SergiySW:
Support Mrai as default in desktop wallet

PR #106, opened 2017-04-14
Default unit in wallet switch from krai to Mrai

jamescoxon:
Simple change to default unit (I find that whenever I open the wallet I automatically click to Mrai)


jamescoxon:
#103

Issue #60, opened 2017-02-09
Add block count to wallet GUI header


SergiySW:
to make sync check easier for people

HostFat:
Good idea.

PR #108, opened 2017-04-15
Status block

jamescoxon:
Adds the block count to the status bar (with a tool tip to explain what it means), might make it easier for new users and people trying to support them getting their wallet running. Following up #60


RaiBlocksDisciple:
Maybe add the tooltip to where blockcount is viewed under 'Advanced' too?

i.e. is it currently 'Blockcount', maybe change it to 'Wallet Block count (blocks remaining to sync)'

maybe also add the current block from the website too somewhere under 'Advanced', because when syncing early on, figures like 11(443000) can appear which can be confusing.

Issue #111, opened 2017-04-18
Add text and descriptions to loading screen


clemahieu:
Add 'initializing' class which can pump information to the loading screen.

clemahieu:
[https://github.com/clemahieu/raiblocks/issues/48](url)

Issue #48, opened 2017-01-30
Initialize GUI before node


clemahieu:
So users know the program has started even if initializing takes a while.

jamescoxon:
Not sure if people have thoughts on what image might be best for a splash screen?

clemahieu:
6b540ee92e40ab5da80691f2b49f2ffd8ad541e7

PR #112, opened 2017-04-18
Added Splash screen to wallet start with logo and text

jamescoxon:
Compiles logo.png into the binary using a qt resource file, now wallet boots with (large) logo and can display some text - currently reminds people to save their wallet seed but could also be used to show Loading...

#48 
#111


clemahieu:
6b540ee92e40ab5da80691f2b49f2ffd8ad541e7

Issue #79, opened 2017-02-17
--datadir launch option


SergiySW:
lauch option to move RaiBlocks data directory to other location
With HDD it can make Disk Queue 5 or more and slow down OS

clemahieu:
209d5293620ea088ae76f61ef3990a950b81f391

PR #114, opened 2017-04-19
Added argument option to --daemon to set custom path for data directory

jamescoxon:
#79 
Only set for rai_node right now


clemahieu:
If this is added to "add_node_options" instead it will be added with any usage of rai::node.  We should probably call it something like "data_directory".

PR #150, opened 2017-06-30
GUI: change representative & password with locked wallet

SergiySW:
\+ show representative of current account


SergiySW:
+ disallow lock with empty password field

PR #164, opened 2017-08-04
RPC: Fix republish genesis with sources attempt

SergiySW:


SergiySW:
'+ optional "destinations"

PR #173, opened 2017-08-26
Proposal: Unchecked head (frontier) for bootstrap

SergiySW:
Should be useful in some cases with bootstrap restart after partial sync  
 +read-only transactions


SergiySW:
More testing required for possible stucks

PR #192, opened 2017-09-18
Config setting to tune unchecked_cache_max

SergiySW:
to improve 1st bootstrap performance at HDD


rai/node/node.cpp
```diff
@@ -948,6 +956,7 @@ bool rai::node_config::deserialize_json (bool & upgraded_a, boost::property_tree
 		callback_address = tree_a.get <std::string> ("callback_address");
 		auto callback_port_l (tree_a.get <std::string> ("callback_port"));
```
SORIN340: VREAU OBTINEREA ADRESEI

PR #199, opened 2017-10-21
Burning changes for open block + available supply

SergiySW:


soliury:
I think we should merge some code, and then release a new version.
Now the 8.0 rai_node sometimes will stop running on ubuntu 16.04, I should restart it manual, and it takes too much memory.

soliury:
@clemahieu

PR #207, opened 2017-11-15
Increase transactions processing speed with blocks vector

SergiySW:
& Testing signature generation time

Draft test vectors for processing (with test_network rules changes): https://github.com/SergiySW/raiblocks/tree/debug_processing


SergiySW:
Additional fix: blocks from local accounts should be processed & only then returned to RPC/GUI
https://github.com/clemahieu/raiblocks/issues/204

PR #211, opened 2017-12-07
Processing vector of block-action pairs

SergiySW:


SergiySW:
Replaced with https://github.com/clemahieu/raiblocks/commit/cf32e1677e51b2d1e712483c5f7df75712b5aa96

PR #215, opened 2017-12-09
Add dockerfile

zjeraar:


zjeraar:
When doing:

`docker run -d -p 7075:7075 -p 7075:7075/udp -p 127.0.0.1:7076:7076 --name xrb_node raiblocks/rai_node`

`rai_node --daemon` will be executed

It's also possible to execute other commands like this:

`docker run -d -p 7075:7075 -p 7075:7075/udp -p 127.0.0.1:7076:7076 --name xrb_node raiblocks/rai_node "rai_node --wallet_list"`

After this has been merged, we can setup the automated build on Docker Hub, which always builds when the repo is updated. Also need to change the Wiki after that.

PR #222, opened 2017-12-14
Flushing unchecked blocks & votes to disk DB only above min amount

SergiySW:
To not fill disk with microchanges of non-critical data


rai/secure.hpp
```diff
@@ -426,6 +426,7 @@ class block_store
 	rai::store_iterator unchecked_end ();
 	size_t unchecked_count (MDB_txn *);
 	std::unordered_multimap <rai::block_hash, std::shared_ptr <rai::block>> unchecked_cache;
+	static size_t const cache_min = 256;
```
im95888: 1

SergiySW:
Rebased

PR #225, opened 2017-12-14
Add CMake and Qt artifacts to .gitignore

devinus:
There may be more autogenerated artifacts I've missed, but this seems to cover macOS 10.3 and Ubuntu 17.10.


androm3da:
Can you investigate why the CI fails and/or rebase the changes?

argakiig:
resolved conflicts

PR #250, opened 2017-12-21
Upgrade our Boost.Beast dependency to version 144

devinus:
**This requires Boost >= 1.66.**

Along with bug fixes and performance improvements, it seems to have fixed the zombie bind issues (at least on macOS 10.13).

### DIFF
--- a/rai/core_test/rpc.cpp
+++ b/rai/core_test/rpc.cpp
@@ -26,19 +26,19 @@ public:
 				boost::property_tree::write_json (ostream, request);
 				req.method (boost::beast::http::verb::post);
 				req.target ("/");
-				req.version = 11;
+				req.version (11);
 				ostream.flush ();
-				req.body = ostream.str ();
+				req.body() = ostream.str ();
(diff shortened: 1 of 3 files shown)


androm3da:
This PR creates a dependency on >= 1.66, or 1.66 was necessary in order to fix the bind issue seen on macOS?

If we increase the required boost version, we should update https://github.com/clemahieu/raiblocks/blob/master/ci/bootstrap_boost.sh#L3 and I guess we would have to change all of the Ubuntu targets to bootstrap.

lukealonso:
The submodule can be deleted as well, since Beast is included in boost 1.66. I did this locally (and made all of the same small fixes before discovering this PR), seems to work fine.

clemahieu:
I'll add this shortly.

androm3da:
I believe that this PR has been obsoleted by another one. @devinus please confirm.

devinus:
Rebased into https://github.com/clemahieu/raiblocks/pull/380.

PR #260, opened 2017-12-22
fix platform specific builds

Dekoze:


androm3da:
Thanks, I was just about to look into this.

androm3da:
Seems like github should flag this as conflicting with https://github.com/clemahieu/raiblocks/pull/261/ now that it was merged but it says "no conflicts".  Hmm.  

In any case this PR is more complete at fixing the issues so @clemahieu or @SergiySW please take it when you get a chance.

PR #266, opened 2017-12-23
 provide error message when a port is already in use

Dekoze:


androm3da:
Is `BOOST_LOG (node.log)` not available in the spots where you're writing to stderr?

Dekoze:
The stderr message I added is outside the scope of the node.log resource so I don't think so. I just changed a couple messages that were within the scope of the node class though.

androm3da:
I like the concept of this feature but it looks like it's changing enough that it's worth waiting until after the regression tests are passing before adding this feature.

On that note, it's a good idea to have an explicit test case for this one.

Dekoze:
This PR depends on PR #298 to properly free resources tied to the node object

PR #284, opened 2017-12-26
Accessing vote_cache needs to be mutex protected.

gentlejack:
file changed.


gentlejack:
Files changed.

androm3da:
I don't think we are planning another release from the `V9.0` branch, but if we do then we can reopen this PR.

PR #289, opened 2017-12-27
Add code formatter script for consistency

rymo4:
Looks like raiblocks has a mix of 4 spaces and tabs. Switch to all 4 spaces instead tabs, which should lead to cleaner diffs. Happy to make the switch the other way if that is preferred (tabs instead of spaces).

If there are concerns about this diff being large and potentially having changes hidden within, I'm happy to break this up into small chunks of files.

PS first PR against this project, starting small :)

EDIT: This now includes a full formatter script

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,114 @@
+---
+# This is the WebKit style, with minor modifications.
+Language:        Cpp
+AccessModifierOffset: -4
+AlignAfterOpenBracket: DontAlign
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   false
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:   
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: All
+BreakBeforeBraces: Allman
+BreakBeforeInheritanceComma: false
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeComma
+BreakAfterJavaFieldAnnotations: false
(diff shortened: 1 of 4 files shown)


androm3da:
Better still would be doing a `clang-format`-ting of the entire project and adding a `.clang-format` file with the project's preferred style.  @clemahieu do you have a preference?  Or is `llvm` sufficient?

I'm on the fence in general about running these things.  It's good to have uniformity but these are really disruptive to diffs after they're made globally.  So IMO it's never do it or do it as early as possible and create mechanisms to preserve or enforce it.

rymo4:
Happy to to go all the way and use a legit formatter. 

> It's good to have uniformity but these are really disruptive to diffs after they're made globally.

Yes, I definitely agree, so IMO better to get it over with now than to have extra lines in diffs forever because a contributor's editor switched the tabs to spaces

androm3da:
Ok, great attitude @rymo4 then let's give `clang-format` a go.

Unless we hear differently from @clemahieu on the matter I would vote for just trying the various styles (llvm, chromium, google, webkit, etc).  Whichever produces the minimal diff is ideal.  We can see if we need to customize beyond that.

We can talk about adding a check to the CI config -- maybe that would help preserve the effect of this change.

androm3da:
Note also that `clang-format-diff.py` exists -- that strikes a good balance here that is not disruptive to diffs.

clemahieu:
I've been trying to slowly change to tab indents as I'm making commits to try and minimize large confusing diffs.  I usually go forward/back a few functions and stomp out inconsistent formatting.

I've been mostly using Allman style indentation https://en.wikipedia.org/wiki/Indentation_style
I haven't been having a column limit and instead letting the editor make that decision.  I usually add a space before argument or parameter definitions.

rymo4:
Here is a "oneliner" to see which of the default styles is closest to the project:

```
> for style in LLVM Google Chromium Mozilla WebKit ; do git reset --hard -q ; for file in `find . -type f  -name *.cpp` ; do clang-format --style=$style -i $file ; done ; echo $style ; git diff --shortstat ; done
LLVM
 51 files changed, 29127 insertions(+), 31345 deletions(-)
Google
 51 files changed, 29327 insertions(+), 31371 deletions(-)
Chromium
 51 files changed, 29686 insertions(+), 31360 deletions(-)
Mozilla
 51 files changed, 31752 insertions(+), 29844 deletions(-)
WebKit
 51 files changed, 24706 insertions(+), 28394 deletions(-)
```

So, seems like WebKit is closest, but it uses spaces instead of tabs. My suggestion is:

1. Make a `.clang-format` file with the desired target formatting options (perhaps WebKit with `UseTab: ForContinuationAndIndentation` instead of `UseTab: Never`)
2. (optional) Format the entire project, with PRs file by file if needed. Seems like you are hesitant to do this because of the large/confusing/meaningless diff it will create, so this can be skipped.
3. Make formatting code diffs as easy as `./format-diff` (just calls `clang-format-diff.py` the right way) and make a git pre commit hook that runs this part of the project setup.
4. Add `clang-format-diff.py` to CI to check new PRs for style guidelines

If that seems good to you both, I can start but just dumping a `.clang-format` file in this PR.

androm3da:
Allman and BS_Allman are supported directly AFAICT.

https://clang.llvm.org/docs/ClangFormatStyleOptions.html

rymo4:
Ok, I changed this commit to instead add a proper `.clang-format` based on WebKit, but with the Allman + tabs changes. Then I included the `clang-format-diff.py` (adding a note in the license file about it) and wrote a small bash script to run this on the git diff that can eventually be used as a pre commit hook if all the devs find it useful.

clemahieu:
Does the format file work?  Can we run it once and get it all formatted?

rymo4:
@clemahieu yeah it works, but I see @androm3da added a different `.clang-format` now. I can delete that file from this PR now and (as a different commit) format the entire repo if that SG

androm3da:
I think this PR may be obsolete now.  Please rebase it if there are new changes.

PR #304, opened 2017-12-28
switch CI to use testnet

Dekoze:
Many tests in core_test operate under the assumption that genesis and other account data is populated with testnet values. This means core_test requires being built targeting the testnet. Whether or not this may cause coverage issues I don't know.


Dekoze:
@androm3da

Issue #288, opened 2017-12-26
Cannot lock wallet


ghost:

SergiySW:
Press "Lock" without password. People often tried to set password with that field

Vladlak:
Ahh this seems to work, if you press lock without password it locks it
if you press it with a password you get an error message

dannyvankooten:
Duplicate of #279, but as @SergiySW most likely just an UX issue and not a bug.

Issue #279, opened 2017-12-25
OSX lock wallet returns "Error"


dannyvankooten:
When inputting the correct password in the OSX wallet and clicking the "Lock" button, the button text changes to "Error" (in red). The only way for me to lock my wallet is by clicking the "Unlock" button without entering a password.

I can properly unlock my wallet with my password once it's locked though (so my password is definitely correct).

I'm running 9.0.0 by the way.

dannyvankooten:
Going through the [interface code for this](https://github.com/clemahieu/raiblocks/blob/master/rai/qt/qt.cpp#L1435), it seems this is not a bug but more of an UX issue.  The "Lock" button works just fine when I leave the password field empty, although I don't think it is really clear that it works like this with the way those fields are displayed.

![image](https://user-images.githubusercontent.com/885856/34335705-35959454-e951-11e7-9640-9f53938e6523.png)

Since the wallet can never be locked & unlocked at the same time, perhaps use a single button and change its text instead?

PR #311, opened 2017-12-28
combine lock & unlock button into a single button + password field

### LINKED ISSUE #237: [Feature] Disable "Unlock" button when unlocked and "Lock" button when locked
### Details

There are 2 buttons in the desktop Mac wallet, Unlock / Lock. Both seem enabled at all times. They could be enabled or disabled depending on the lock state. It is confusing to try to unlock the wallet when it's already unlocked.

### Snapshot

screen shot 2017-12-16 at 8 02 11 pm

### Version
8.0.0

dannyvankooten:
This pull request combines the "Lock" and "Unlock" button into a single button which is toggled depending on the lock state.

Additionally, it will disable the password field when the wallet is unlocked. 

Unlocked wallet:
image

Locked wallet:
image

This fixes #237, #279, #288

### DIFF
--- a/rai/qt/qt.cpp
+++ b/rai/qt/qt.cpp
@@ -1267,10 +1267,7 @@ rai_qt::settings::settings (rai_qt::wallet & wallet_a) :
 window (new QWidget),
 layout (new QVBoxLayout),
 password (new QLineEdit),
-lock_window (new QWidget),
-lock_layout (new QHBoxLayout),
-unlock (new QPushButton ("Unlock")),
-lock (new QPushButton ("Lock")),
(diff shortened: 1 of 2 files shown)


citizen2077:
I like the idea but if you disable the password field when the wallet is unlocked what if a user changes their password while the wallet is unlocked?

dannyvankooten:
Theoretically, nothing should happen as locking the wallet does not require a password. So people can change their password when the wallet is unlocked, press the "Lock" button and it just works.

When the wallet is locked and people try to change their password, this message shows.

image

PR #321, opened 2017-12-29
Improve balance formatting in the UI.

lukealonso:


lukealonso:
screen shot 2017-12-29 at 12 04 56 am
screen shot 2017-12-29 at 12 03 53 am
screen shot 2017-12-29 at 12 06 11 am

rai/core_test/uint256_union.cpp
```diff
@@ -13,6 +13,33 @@ TEST (uint128_union, decode_dec)
...
 
+struct test_punct : std::moneypunct<char> {
+    pattern do_pos_format () const { return { {value, none, none, none} }; }
+    int do_frac_digits () const { return 0; }
+    char_type do_decimal_point () const { return '+'; }
+    char_type do_thousands_sep () const { return '-'; }
+    string_type do_grouping () const { return "\3\4"; }
+};
+
+TEST (uint128_union, balance_format)
```
androm3da: Can you add tests for `uint128_t` with all of the bits set, all of them except for the LSB, all of them except the MSB?

rai/lib/numbers.cpp
```diff
@@ -579,6 +587,114 @@ bool rai::uint128_union::decode_dec (std::string const & text)
...
+			stream << decimal_point;
+			format_frac (stream, frac_part, scale, precision);
+		}
+	}
+	return stream.str();
+}
+
+std::string rai::uint128_union::format_balance(rai::uint128_t scale, int precision, bool group_digits)
+{
+	auto thousands_sep = std::use_facet< std::numpunct<char> >(std::locale ()).thousands_sep ();
```
androm3da: Thanks, good move including locale-specific separators/radix.

rai/lib/numbers.hpp
```diff
@@ -35,6 +35,8 @@ union uint128_union
 	bool decode_hex (std::string const &);
 	void encode_dec (std::string &) const;
 	bool decode_dec (std::string const &);
+	std::string format_balance (rai::uint128_t scale, int precision, bool group_digits);
```
androm3da: Please make these methods `const` if it's possible.

rai/core_test/uint256_union.cpp
lukealonso: Done, the 0xFF... case actually found an overflow - thanks :)

rai/lib/numbers.hpp
lukealonso: Done

rai/lib/numbers.cpp
```diff
@@ -311,6 +311,10 @@ bool rai::uint256_union::decode_dec (std::string const & text)
 		{
 			stream >> number_l;
 			*this = number_l;
+			if (!stream.eof ())
```
lukealonso: This is to prevent parsing things like "1.1" as "1" until we implement it properly.

PR #341, opened 2017-12-30
All x64_86 CPU should support SSE2

SergiySW:


CMakeLists.txt
```diff
@@ -34,9 +34,11 @@ else (WIN32)
     endif()
 
 	IF (CMAKE_SYSTEM_PROCESSOR MATCHES "^(i.86|x86(_64)?)$")
-        if (RAIBLOCKS_SIMD_OPTIMIZATIONS)
-		    set (PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} -msse4")
-        endif()
+		if (RAIBLOCKS_SIMD_OPTIMIZATIONS)
```
androm3da: This conditional `MATCHES "^(i.86|x86(_64)?)$"` includes non-`x86_64` targets.

SergiySW:
In fact 32-bit CPU aren't supported anymore

lukealonso:
Ideally we would check for compiler support for SSE2/4 and AVX2 in CMakeLists (not hard), and refactor blake2b and ed25519 to compile all of the various supported backends instead of just one, then do a runtime check a la:

https://github.com/twitter/vireo/blob/master/imagecore/image/internal/platform_support.cpp

CMakeLists.txt
```diff
@@ -34,9 +34,11 @@ else (WIN32)
...
+			set (PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} -msse4")
+		elseif(CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
+			set (PLATFORM_COMPILE_FLAGS "${PLATFORM_COMPILE_FLAGS} -msse2")
```
clemahieu: Do we need to set -msse2 here if it's set already above?

SergiySW: Where?

CMakeLists.txt
```diff
@@ -21,7 +21,7 @@ if(NOT CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 
 if (WIN32)
-	set (PLATFORM_COMPILE_FLAGS "-DBOOST_SPIRIT_THREADSAFE -D_WIN32_WINNT=0x0600 -DWINVER=0x0600 -DWIN32_LEAN_AND_MEAN -DMINIUPNP_STATICLIB")
+	set (PLATFORM_COMPILE_FLAGS "-DBOOST_SPIRIT_THREADSAFE -D_WIN32_WINNT=0x0600 -DWINVER=0x0600 -DWIN32_LEAN_AND_MEAN -DMINIUPNP_STATICLIB -msse2")
```
clemahieu: -msse2 here

SergiySW: For Windows

PR #342, opened 2017-12-30
Adding data_path support for CLI operations

SergiySW:
To resolve issue https://github.com/clemahieu/raiblocks/issues/328

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -2971,6 +2971,15 @@ void rai::add_node_options (boost::program_options::options_description & descri
 bool rai::handle_node_options (boost::program_options::variables_map & vm)
 {
 	auto result (false);
+	boost::filesystem::path data_path;
+	if (vm.count ("data_path"))
+	{
+		data_path = boost::filesystem::path (vm ["data_path"].as <std::string> ());
+	}
+	else
+	{
+		data_path = rai::working_path ();
+	}
 	if (vm.count ("account_create"))
 	{
 		if (vm.count ("wallet") == 1)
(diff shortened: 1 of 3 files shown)


rai/node/node.cpp
```diff
@@ -2971,6 +2971,15 @@ void rai::add_node_options (boost::program_options::options_description & descri
 bool rai::handle_node_options (boost::program_options::variables_map & vm)
 {
 	auto result (false);
+	boost::filesystem::path data_path;
+	if (vm.count ("data_path"))
+	{
+		data_path = boost::filesystem::path (vm ["data_path"].as <std::string> ());
```
androm3da: This would be better as a ternary IMO.

rai/rai_node/entry.cpp
```diff
@@ -131,26 +131,26 @@ int main (int argc, char * const * argv)
 	boost::program_options::store (boost::program_options::parse_command_line(argc, argv, description), vm);
 	boost::program_options::notify (vm);
 	int result (0);
+	boost::filesystem::path data_path;
+	if (vm.count ("data_path"))
```
androm3da: same comment here regarding ternary

cryptocode:
@SergiySW I'm looking forward to getting this merged. Seems like a rebase is required though.

rai/node/node.cpp
brightmonkey: I agree with @androm3da, this would be cleaner as a ternary. This change also makes the data_path assignment under the 'vacuum' option redundant, and it should be removed.

SergiySW:
Replaced with https://github.com/clemahieu/raiblocks/pull/586

PR #362, opened 2018-01-02
Add block reconfirmation and election history.

lukealonso:
The next step would be making the node itself track this automatically, and not need to be asked.

This is the absolute minimum required for a user of the node to cobble together a fix for various network problems.

`curl -H "Content-Type: application/json" -X POST -d '{"action":"request_reconfirmation","hash":"791AF413173EEE674A6FCF633B5DFC0F3C33F397F0D
A08E987D9E0741D40D81A"}' http://[::1]:7076`

You can check the result of any recent election with:

`curl -H "Content-Type: application/json" -X POST -d '{"action":"check_election_results","hash":"791AF413173EEE674A6FCF633B5DFC0F3C33F397F0DA08E987D9E0741D40D81A"}' http://[::1]:7076`

```
{
    "confirmed": true,
    "tally": "35ABEE023B214C987F8E1A86DC9E6751",
    "percent_circulating_approved": "53.5"
}
```

### DIFF
--- a/rai/core_test/node.cpp
+++ b/rai/core_test/node.cpp
@@ -1428,3 +1428,198 @@ TEST (node, vote_replay)
 		ASSERT_GT (400, iterations);
 	}
 }
+
+TEST (node, reconfirm_mitm)
+{
+	rai::system system (24000, 3);
+	auto & sender (*system.nodes[0]);
+	auto & receiver (*system.nodes[1]);
+	auto & rep (*system.nodes[2]);
+	auto sender_wallet (system.wallet (0));
+	auto receiver_wallet (system.wallet (1));
+	auto rep_wallet (system.wallet (2));
+	rai::keypair sender_key;
+	rai::keypair receiver_key;
+	sender_wallet->insert_adhoc (sender_key.prv);
+	receiver_wallet->insert_adhoc (receiver_key.prv);
+	rep_wallet->insert_adhoc (rai::test_genesis_key.prv);
+
+	// Genesis -> Sender
+	auto block1 = rep_wallet->send_action (rai::test_genesis_key.pub, sender_key.pub, 10 * rai::Mxrb_ratio);
+	int iterations = 0;
+	while (true)
+	{
+		system.poll ();
+		auto balanceOnSender = sender.balance (sender_key.pub);
+		auto balanceOnReceiver = receiver.balance (sender_key.pub);
+		if (balanceOnSender == 10 * rai::Mxrb_ratio && balanceOnReceiver == 10 * rai::Mxrb_ratio)
+		{
+			break;
+		}
+		ASSERT_GT (400, iterations);
+		iterations++;
+	}
+
+	// Simulate MITM.
+	receiver.block_processor.stop ();
+	receiver.network.packet_filter = [](uint8_t * buffer, size_t size) {
+		// In the real attack the filter would be a lot more sophisticated, but this works for this test.
+		return true;
+	};
+
+	// Sender -> Genesis (not seen by Receiver)
+	auto block2 = sender_wallet->send_action (sender_key.pub, rai::test_genesis_key.pub, 10 * rai::Mxrb_ratio);
+	iterations = 0;
+	while (true)
+	{
+		system.poll ();
+		auto balanceOnSender = sender.balance (sender_key.pub);
+		auto balanceOnRep = rep.balance (sender_key.pub);
+		auto balanceOnReceiver = receiver.balance (sender_key.pub);
+		if (balanceOnSender == 0 * rai::Mxrb_ratio && balanceOnRep == 0 * rai::Mxrb_ratio && balanceOnReceiver == 10 * rai::Mxrb_ratio)
+		{
+			break;
+		}
+		ASSERT_GT (400, iterations);
+		iterations++;
+	}
+
+	// Fork block.
+	rai::send_block fork (block2->root (), receiver_key.pub, 0 * rai::Mxrb_ratio, sender_key.prv, sender_key.pub, system.work.generate (block2->root ()));
+	{
+		rai::transaction transaction (receiver.store.environment, nullptr, true);
+		ASSERT_EQ (rai::process_result::progress, receiver.ledger.process (transaction, fork).code);
+	}
+
+	iterations = 0;
+	while (true)
+	{
+		system.poll ();
+		auto balance = receiver.balance_pending (receiver_key.pub);
+		if (balance.second == 10 * rai::Mxrb_ratio)
+		{
+			break;
+		}
+		ASSERT_GT (400, iterations);
+		iterations++;
+	}
+
+	// Request confirmation.
+	auto fork_block = std::shared_ptr<rai::block> (&fork, [](rai::block *) {});
+	receiver.request_confirmation (fork_block);
+	iterations = 0;
+	while (iterations < 100)
+	{
+		system.poll ();
+		iterations++;
+	}
+
+	// Heard nothing, we know we may be compromised.
+	rai::election_result result;
+	ASSERT_FALSE (receiver.check_election_results (fork_block, result));
+	ASSERT_EQ (0, result.tally.number ());
+	ASSERT_FALSE (result.confirmed);
+
+	// Remove MITM.
+	receiver.network.packet_filter = [](uint8_t * buffer, size_t size) {
+		return false;
+	};
+
+	// Ask again.
+	receiver.request_confirmation (fork_block);
+	iterations = 0;
+	while (iterations < 100)
+	{
+		system.poll ();
+		iterations++;
+	}
+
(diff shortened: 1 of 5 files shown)


lukealonso:
@androm3da feedback welcome

rai/node/node.cpp
```diff
@@ -2288,6 +2288,22 @@ void rai::node::process_confirmed (std::shared_ptr <rai::block> confirmed_a)
     confirmed_a->visit (visitor);
 }
 
+void rai::node::request_reconfirmation (std::shared_ptr <rai::block> block_a)
+{
+	std::cout << "requesting reconfirmation for block: " << block_a->hash().to_string() << " root: " << block_a->root().to_string() << "\n";
```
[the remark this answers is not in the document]
lukealonso: It should just be removed entirely.

rai/node/node.cpp
```diff
@@ -2870,6 +2900,28 @@ void rai::active_transactions::start (MDB_txn * transaction_a, std::shared_ptr <
...
+{
+	std::lock_guard <std::mutex> lock (mutex);
+    auto root (block_a->root ());
+    auto existing (roots.find (root));
+    if (existing != roots.end ())
+    {
+		existing->election->get_progress(transaction_a, result);
+		return false;
+    }
+	else
```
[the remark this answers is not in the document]
lukealonso: Indentation is inconsistent throughout the file, I think it was confusing my editor. I'll fix it, thanks.

lukealonso:
Added tests for the MITM attack and added an LRU for election history.

@clemahieu I think this is ready to be looked at. Next step would be adding support in the wallet, if we thought that was a good idea.

rai/node/node.cpp
```diff
@@ -428,21 +429,24 @@ void rai::network::receive_action (boost::system::error_code const & error, size
     {
         if (!rai::reserved_address (remote) && remote != endpoint ())
         {
-            network_message_visitor visitor (node, remote);
-            rai::message_parser parser (visitor, node.work);
-            parser.deserialize_buffer (buffer.data (), size_a);
-            if (parser.error)
+            if (!packet_filter || !packet_filter (buffer.data (), size_a))
```
lukealonso: This is only used by the test.

rai/node/node.cpp
```diff
@@ -2890,6 +2922,45 @@ bool rai::active_transactions::start (MDB_txn * transaction_a, std::shared_ptr <
...
+	}
+	return true;
+}
+
+void rai::active_transactions::add_election_history (const rai::block_hash & root, const election_result & result)
+{
+	rai::election_history history;
+	history.root = root;
+	history.result = result;
+	auto existing_history (election_history.get <1>().find (root));
```
lukealonso: I'm not a boost expert, so some extra eyes here would be nice. This is supposed to be an LRU.

rai/core_test/node.cpp
```diff
@@ -1426,3 +1426,190 @@ TEST (node, vote_replay)
...
+		}
+		ASSERT_GT (400, iterations);
+		iterations++;
+	}
+
+	// Install MITM.
+	receiver.block_processor.stop ();
+	receiver.network.packet_filter = [] (uint8_t* buffer, size_t size)
+	{
+		return true;
```
lukealonso: This could be fancier and check the block hash before blocking the packet, but this is fine for this test.

rai/node/node.cpp
```diff
@@ -2739,9 +2758,15 @@ rai::uint128_t rai::election::minimum_threshold (MDB_txn * transaction_a, rai::l
...
+rai::uint128_t rai::election::reconfirm_threshold (MDB_txn * transaction_a, rai::ledger & ledger_a)
+{
+	// Minimum number of votes required to consider safe for reconfirmation
+    return ledger_a.supply (transaction_a) / 3;
+}
+
 void rai::election::confirm_once (MDB_txn * transaction_a)
 {
-	if (!confirmed.test_and_set ())
+	if (!finished.exchange(true))
```
lukealonso: Quite certain this is exactly equivalent, but please sanity check.

rai/node/node.cpp
```diff
@@ -2834,6 +2872,9 @@ void rai::active_transactions::announce_votes ()
 				i->election->confirm_cutoff (transaction);
 				auto root_l (i->election->votes.id);
 				inactive.push_back (root_l);
+				election_result result;
+				election_l->get_progress(transaction, result);
+				add_election_history(root_l, result);
```
lukealonso: Note to reviewer: We could only track this for explicitly requested blocks, but then it becomes purely an appendage for the reconfirmation feature vs a more general recent election history system. Easy to change, either way.

androm3da:
LGTM.  @clemahieu please scrutinize this feature.

rai/node/node.cpp
```diff
@@ -2739,9 +2758,15 @@ rai::uint128_t rai::election::minimum_threshold (MDB_txn * transaction_a, rai::l
 	return ledger_a.supply (transaction_a) / 16;
 }
 
+rai::uint128_t rai::election::reconfirm_threshold (MDB_txn * transaction_a, rai::ledger & ledger_a)
+{
+	// Minimum number of votes required to consider safe for reconfirmation
+	return ledger_a.supply (transaction_a) / 3;
```
lukealonso: Another magic number. Have to be careful with this one, since a block won't be considered reconfirmed unless it meets this threshold. That being said, a block not being reconfirmed has no impact on the node itself, it's purely information for some other system talking to the node over RPC.

androm3da: I think low/medium/high is in the eye of the beholder.  This is a user-visible knob like bitcoin's "# of confirmations" that individuals can customize to meet the needs of their users/transactions.

rai/node/rpc.cpp
```diff
@@ -2869,6 +2869,64 @@ void rai::rpc_handler::search_pending_all ()
...
+		auto block_l (node.store.block_get (transaction, hash_l));
+		if (block_l != nullptr)
+		{
+			rai::election_result result;
+			if (!node.check_election_results (std::shared_ptr <rai::block>(move (block_l)), result))
+			{
+				boost::property_tree::ptree response_l;
+				response_l.put ("verified", result.met_reconfirm_threshold && result.confirmed);
+				response_l.put ("tally", result.tally.to_string());
+				response_l.put ("percent", (double)(result.tally.number() / (node.ledger.supply (transaction) / 1000)) / 10.0);
```
androm3da: Please qualify this as "percent_approved" or "percent_of_in_circulation_approved".

rai/node/rpc.cpp
```diff
@@ -2869,6 +2869,64 @@ void rai::rpc_handler::search_pending_all ()
...
+	{
+		rai::transaction transaction (node.store.environment, nullptr, false);
+		auto block_l (node.store.block_get (transaction, hash_l));
+		if (block_l != nullptr)
+		{
+			rai::election_result result;
+			if (!node.check_election_results (std::shared_ptr <rai::block>(move (block_l)), result))
+			{
+				boost::property_tree::ptree response_l;
+				response_l.put ("verified", result.met_reconfirm_threshold && result.confirmed);
```
androm3da: I worry that the presence of this field will negate the flexibility benefit of the "percent" below.  We should assume that clients are the lowest common denominator and won't take the time to understand what "verified" means for their particular transaction.

rai/node/rpc.cpp
```diff
@@ -4436,6 +4494,14 @@ void rai::rpc_handler::process_request ()
 		{
 			search_pending_all ();
 		}
+		else if (action == "request_reconfirmation")
```
androm3da: Should we call it "request_confirmation" instead?  I know that the protocol might make this look like a "reconfirmation" but RPC clients don't necessarily understand that part.

rai/node/rpc.cpp
lukealonso: I agree, that's how I had it in a previous version of the patch, then was worried about not providing some sort of official view on the level of risk. I'll revert.

rai/node/rpc.cpp
lukealonso: Sure.

rai/node/rpc.cpp
lukealonso: Done.

rai/node/node.cpp
androm3da: It seems crufty but ifdef seems cruftier.  Please give it a name that stands out like `_TESTING_packet_filter` so no one mistakes it for normal functionality.

androm3da:
@lukealonso can you rebase this change?

slact:
Is this not getting merged? I'm following this PR with _intense interest_.

PlasmaPower:
I'm currently rebasing this PR and adding some stuff 😄.

PlasmaPower:
#450

PR #371, opened 2018-01-02
Add .travis.yml

drizzt:
TODO:
 - Fix and enable OSX build support
 - Enable Windows (mingw) build support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,41 @@
+language: cpp
+
+os:
+  - linux
+
+cache:
+  apt: true
+  ccache: true
+  directories:
+    - $HOME/boost
+
+env:
+  global:
+    - BOOST_URL=http://ftp.osuosl.org/pub/blfs/conglomeration/boost/boost_1_63_0.tar.bz2
+
+compiler:
+  - gcc
+  - clang
+
+addons:
+  apt:
+    packages:
+      - qtbase5-dev
+
+before_install:
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew update; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew install qt5; fi
+
+before_script:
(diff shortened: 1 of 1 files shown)


drizzt:
You still need to enable travis and caching support on https://travis-ci.org

androm3da:
Does it actually work?  When I tested building raiblocks on travis I found that the builds would all fail because of resource limits.  I added gitlab CI instead.  

Travis is great if it works.

drizzt:
I had to change slightly the commit since plain "make" is actually broken due to broken qt_test as you can see in [the log](https://api.travis-ci.org/v3/job/324171995/log.txt) and so I changed the travis script to only use `make rai_node rai_wallet`.

I think think should be changed when qt_test is fixed since plain make should always work.

Example of working travis job: https://travis-ci.org/drizzt/raiblockscommunity/builds/324215243

lukealonso:
We should make a docker image with boost and the other dependencies already installed, but this is a great first step.

androm3da:
I'll bet I had trouble because I used ninja and it was spawning too many parallel processes. 

Can you adapt CI/build.sh to use unix makefiles and invoke it here?

androm3da:
Let's make and run the full test suite but have it contingent on the other PR already queued that fixes qt_test

androm3da:
The mingw build would use a cross compiler?  Could you really make executables too?   Is there a linker that you can use on linux that makes COFF binaries?

neilsh:
For reference, here's the other PR that fixes qt_test: https://github.com/clemahieu/raiblocks/pull/369

lukealonso:
@androm3da we could make binaries with some other CI, as long as this is running tests for us per-PR

androm3da:
yes that's fine

androm3da:
Now that #369 is merged please include the full scope

androm3da:
@clemahieu to follow through we need to enable this in travis: http://travis-ci.org/clemahieu/

Issue #367, opened 2018-01-02
UX issue: wallet address split onto two lines is potentially confusing


charlie-axsy:
I just tried to withdraw my raiblocks from bitgrail to my desktop wallet. When I input my account address on the receive I thought only the top line was the address as, when you full screen the wallet, it still condenses the address down to two lines even though there is plenty of room to fit the whole address into one line.

This was my fault and a noob error, luckily I only did a test withdrawal of 1 rai, however it is a reasonably easy fix.

My suggestion is that you put the address in a box to make it more obvious that you should use the entire string. Additionally, when the window is enlarged or full screened the address could reformat to go on one line (perhaps do it as a 2 stage reformatting, so it's either 2 equal strings or one full string, so you never have a big string on top and a little string below).

Alternatively the label could be un-selectable, to force the user to press the copy button and remove the change of a highlighting error.

cbrunnkvist:
"Potentially"? I'd say _totally_ :)

rkeene:
@cryptocode has verified this is done.  Closing this issue.

PR #372, opened 2018-01-02
use readonly QLineEdit for wallet address

dannyvankooten:
This PR switches the multi-line QLabel for a single line read-only QLineEdit so that the user can only ever select their entire wallet address. This also plays a little nicer when resizing the wallet to a larger size (increasing the window width).

Improves #367 

image

Or, when the window size is increased.

image

Additionally, this adds visual feedback after the "Copy" button is pressed by temporarily changing the button text to "Copied".


charlie-axsy:
Closing the loop https://www.reddit.com/r/RaiBlocks/comments/7nlvhi/raiblocks_desktop_wallet_account_address_on_one/ds2qqtu/?context=3

lukealonso:
Code LGTM

PR #376, opened 2018-01-02
stretch last column of all tableview instances (so table uses entire avail. width)

dannyvankooten:
Stretches the last section (column) of each QTableView instance so that the entire table width is always utilized.

image


lukealonso:
LGTM

PR #380, opened 2018-01-03
Upgrade boost to 1.66 and remove beast.

lukealonso:


lukealonso:
@clemahieu This has everything ready to go, except re-creating the Docker image, if it saves you any time.

PR #391, opened 2018-01-04
Allow static build of rai_node

devinus:
~~Observes the CMake BUILD_SHARED_LIBS variable when building rai_lib.~~

~~See: https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS~~

**Edit:** I've updated the build definition to continue to build all executables static and build our separate `rai_lib` shared library.

### DIFF
--- a/rai/secure.cpp
+++ b/rai/secure.cpp
@@ -10,8 +10,6 @@
 
 #include <ed25519-donna/ed25519.h>
 
-thread_local CryptoPP::AutoSeededRandomPool rai::random_pool;
-
 // Genesis keys for network variants
 namespace
 {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,7 +177,7 @@ if (CRYPTOPP_CUSTOM OR NOT RAIBLOCKS_SIMD_OPTIMIZATIONS)
 else ()
 	set (CRYPTOPP_LIBRARY cryptopp-static)
 	add_subdirectory (cryptopp)
-endif (CRYPTOPP_CUSTOM)
+endif ()
 
 add_library (argon2
 	phc-winner-argon2/src/argon2.c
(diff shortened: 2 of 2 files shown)


androm3da:
But we don't want to build all of them shared, just `rai_lib`, right?

Does the build and test work as expected with both BUILD_SHARED_LIBS=ON and BUILD_SHARED_LIBS=OFF?

devinus:
@androm3da You're absolutely right. Using `BUILD_SHARED_LIBS` for this isn't the right call.

rai/secure.cpp
```diff
@@ -10,8 +10,6 @@
 
 #include <ed25519-donna/ed25519.h>
 
-thread_local CryptoPP::AutoSeededRandomPool rai::random_pool;
```
cryptocode: This is removed because it's unused?

devinus: @cryptocode It's in ria_lib, so it's a duplicate symbol.

lukealonso:
Can we make this optional, since it'll double the already slow build times?

rai/secure.cpp
clemahieu: This is here because of how windows does symbols in a shared library.  Need to make sure Windows still builds if this is removed.

androm3da:
We're only adding this for the sake of distributing binaries?  What if we solved this problem by building distro-specific tarball releases?

devinus:
@lukealonso It only links the already built object files twice, not rebuilds the sources twice.

devinus:
Note: Travis failures are unrelated to this PR.

`/home/travis/build/clemahieu/raiblocks/rai/node/rpc.hpp:7:27: fatal error: boost/beast.hpp: No such file or directory`

@drizzt

lukealonso:
@devinus could you rebase this on master? CI should work now.

PR #396, opened 2018-01-04
Set callback Content-Type to application/json

PlasmaPower:
This is required for some servers to properly parse the request.


cryptocode:
Should it be possible to configure this, in case existing integrations for some reason depend the current behavior? Unlikely to be a problem. On the other hand, some libs/frameworks will surely bug out if Content-Type is wrong, so maybe there are some workarounds out there that will stop working once the correct value is in place.

PlasmaPower:
Previously, the `Content-Type` was missing altogether (not incorrect), so I doubt this will break anything.

lukealonso:
@PlasmaPower Could you rebase this on master so CI runs, please?

PlasmaPower:
Done

PR #412, opened 2018-01-06
Attempt to obfuscate the source of votes

PlasmaPower:
Warning: untested, only compiles. I'm not sure how to effectively test this change.

Also, the `confirm_req` code might/will need changing.

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -183,7 +183,28 @@ bool confirm_block (MDB_txn * transaction_a, rai::node & node_a, T & list_a, std
 			}
 			for (auto j (list_a.begin ()), m (list_a.end ()); j != m; ++j)
 			{
-				node_a.network.confirm_send (confirm, bytes, *j);
+				// We'd prefer if attackers didn't know that our IP is a representative.
+				// This makes DDOS attacks harder.
+				// To help accomplish this, we use a delay. The delay is procedurally generated
+				// based off our private key and the peer's address. This makes timing attacks
+				// to determine a representative's IP signficantly harder. It may appear to an
+				// attacker that several peers adjacent to the representative which got
+				// low delays are the representative itself, when in fact they are not.
+				auto endpoint = *j;
+				blake2b_state hash;
+				blake2b_init (&hash, 2);
+				blake2b_update (&hash, prv_a.data.bytes.data (), prv_a.data.bytes.size ());
+				assert (endpoint.address ().is_v6 ());
+				rai::uint128_union address;
+				address.bytes = endpoint.address ().to_v6 ().to_bytes ();
+				blake2b_update (&hash, address.bytes.data (), address.bytes.size ());
+				uint16_t peer_delay;
+				blake2b_final (&hash, &peer_delay, sizeof (peer_delay));
+				peer_delay &= (1 << 10) - 1; // maximum of 1023 milliseconds delay
+				boost::asio::deadline_timer t (node_a.service, boost::posix_time::millisec (peer_delay));
+				t.async_wait ([&node_a, &confirm, bytes, endpoint] (boost::system::error_code const & ec) {
+					node_a.network.confirm_send (confirm, bytes, endpoint);
+				});
 			}
 		});
 	}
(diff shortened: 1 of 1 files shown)


slact:
It's s start, but as discussed, this breaks the distance metric used to route `confirm_req`s efficiently.

rai/node/node.cpp
```diff
@@ -183,7 +183,28 @@ bool confirm_block (MDB_txn * transaction_a, rai::node & node_a, T & list_a, std
...
+				blake2b_update (&hash, prv_a.data.bytes.data (), prv_a.data.bytes.size ());
+				assert (endpoint.address ().is_v6 ());
+				rai::uint128_union address;
+				address.bytes = endpoint.address ().to_v6 ().to_bytes ();
+				blake2b_update (&hash, address.bytes.data (), address.bytes.size ());
+				uint16_t peer_delay;
+				blake2b_final (&hash, &peer_delay, sizeof (peer_delay));
+				peer_delay &= (1 << 10) - 1; // maximum of 1023 milliseconds delay
+				boost::asio::deadline_timer t (node_a.service, boost::posix_time::millisec (peer_delay));
+				t.async_wait ([&node_a, &confirm, bytes, endpoint] (boost::system::error_code const & ec) {
```
lukealonso: Good idea, if the impact on confirm_req routing can be solved. 

There's probably a better implementation of this (most likely a priority queue of absolute times for the sends), this won't be as efficient.

lukealonso:
Good idea, if the impact on confirm_req routing can be solved.

There's probably a better implementation of this (most likely a priority queue of absolute times for the sends), this won't be as efficient.

PlasmaPower:
> There's probably a better implementation of this (most likely a priority queue of absolute times for the sends), this won't be as efficient.

I don't know much about boost async, but boost should be doing this internally. Is it not?

PlasmaPower:
Another possible solution, that would solve the `confirm_req` problem, is a threshold ring signature scheme. However, that would be computationally expensive, and the implementation would be difficult.

rai/node/node.cpp
```diff
@@ -200,7 +200,7 @@ bool confirm_block (MDB_txn * transaction_a, rai::node & node_a, T & list_a, std
 				blake2b_update (&hash, address.bytes.data (), address.bytes.size ());
 				uint16_t peer_delay;
 				blake2b_final (&hash, &peer_delay, sizeof (peer_delay));
-				peer_delay &= (1 << 10) - 1; // maximum of 1023 milliseconds delay
+				peer_delay &= (1 << 7) - 1; // maximum of 127 milliseconds delay
```
cryptocode: This edit made me wonder if the lower and upper limits should be configurable, in case there turns out be problems in either direction (speed vs security) ? Default limits, overridable from config.json.

Something like:

`peer_delay = boost::algorithm::clamp (peer_delay, min_delay, max_delay)`

PlasmaPower: `clamp` wouldn't work, as the delay would be way more likely to be either the min or the max. However, it's probably a good idea to let users configure a power of 2 for the max.

cryptocode: @PlasmaPower but isn't clamping to max basically what you're doing anyway with `peer_delay &= (1 << 7) - 1;`  ?

PlasmaPower: No the `&=` means it will be uniformly distributed across the range 0 to 123 (assuming blake2 has a uniform distribution).

The lowest 7 bits will be kept, and the others will be discarded. If, say, the 8th bit was set, it wouldn't clamp it to 123, but instead just unset the bit and keep the other bits as they are (resulting in a random number between 0 and 123).

cryptocode: @PlasmaPower Yeah, I see that now, clamping would be completely wrong. Thanks for explaining.

PlasmaPower:
This might not completely solve the issue. You could still doing a timing attack by creating blocks, only sending them to some peers, measuring the time until you get a vote, and then performing a binary search (or similar) from there.

androm3da:
Closing this PR for now, let's discuss and work out the design and then reopen it.

PR #422, opened 2018-01-06
Support the HTTP OPTIONS verb for RPC

devinus:
Useful for [preflighting requests when using CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS#Preflighted_requests_in_CORS).

### DIFF
--- a/nano/node/bootstrap.cpp
+++ b/nano/node/bootstrap.cpp
@@ -3,11 +3,13 @@
 #include <nano/node/common.hpp>
 #include <nano/node/node.hpp>
 
+#include <algorithm>
 #include <boost/log/trivial.hpp>
 
 constexpr double bootstrap_connection_scale_target_blocks = 50000.0;
 constexpr double bootstrap_connection_warmup_time_sec = 5.0;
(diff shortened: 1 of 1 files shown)


androm3da:
Can you create a test case that uses `options`?

devinus:
@androm3da Right now the RPC tests aren't testing the actual HTTP handler, so that would be part of a larger endeavor. I'm looking into how we can do that now.

rai/node/rpc.cpp
```diff
@@ -4116,34 +4116,53 @@ void rai::rpc_connection::parse_connection ()
...
+				auto method (this_l->request.method ());
+				switch (method) {
+					case boost::beast::http::verb::post:
+					{
+						auto handler (std::make_shared<rai::rpc_handler> (*this_l->node, this_l->rpc, this_l->request.body (), response_handler));
+						handler->process_request ();
+						break;
+					}
+					case boost::beast::http::verb::options:
+					{
```
androm3da: From what I understand the `OPTIONS` method would be clearest if it returned a documentation object.  What do you think?

PR #425, opened 2018-01-07
Requeue pull with last received expected block

SergiySW:
If bulk pull client fails in the middle of large pull

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -451,9 +451,14 @@ rai::bulk_pull_client::~bulk_pull_client ()
 		--connection->attempt->pulling;
 		connection->attempt->condition.notify_all ();
 	}
-	if (!pull.account.is_zero ())
+	// If received end block is not expected end block
+	else if (expected != pull.end)
 	{
-		connection->attempt->requeue_pull (pull);
+		connection->attempt->requeue_pull (rai::pull_info (pull.account, expected, pull.end));
+		if (connection->node->config.logging.bulk_pull_logging ())
+		{
+			BOOST_LOG (connection->node->log) << boost::str (boost::format ("Bulk pull end block is not expected %1% for account %2%") % pull.end.to_string () % pull.account.to_account ());
+		}
 	}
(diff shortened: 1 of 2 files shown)


SergiySW:
Requesting pull from frontier peer after 4 attempts

SergiySW:
Print remaining sync queue in logs

rai/node/bootstrap.cpp
```diff
@@ -554,10 +559,6 @@ void rai::bulk_pull_client::received_type ()
 		case rai::block_type::not_a_block:
 		{
 			connection->attempt->pool_connection (connection);
```
lukealonso: We might also want to stop talking to this peer if they sent incomplete data.

rai/node/bootstrap.hpp
```diff
@@ -79,6 +79,7 @@ class bootstrap_attempt : public std::enable_shared_from_this<bootstrap_attempt>
 	void requeue_pull (rai::pull_info const &);
 	bool still_pulling ();
 	std::deque<std::weak_ptr<rai::bootstrap_client>> clients;
+	std::weak_ptr<rai::bootstrap_client> connection_frontier_request;
```
cryptocode: Could this be renamed to indicate it's a lock? Would make the usage code a bit clearer.

rai/node/bootstrap.cpp
```diff
@@ -1022,6 +1024,19 @@ void rai::bootstrap_attempt::requeue_pull (rai::pull_info const & pull_a)
 		pulls.push_front (pull);
 		condition.notify_all ();
 	}
+	else if (pull.attempts == 4 && !connection_frontier_request.expired ())
+	{
+		std::lock_guard<std::mutex> lock (mutex);
+		auto connection_shared = connection_frontier_request.lock ();
```
cryptocode: I believe the  `weak_ptr#lock` result should be checked against nullptr here, since *if (!expired) + mutex lock* isn't atomic. So it may technically have expired at this point?

Issue #430, opened 2018-01-07
Add RPC method to get list of all wallet ids


kennell:
Hi there, would it be possible to add a RPC method that works similar to the `./rai_node wallet_list` command?

So, e.g. client sends

```json
{
  "action": "wallet_list"
}
``` 

Node returns

```json
{
  "wallets": [
     "AAAABBBBCCCCDDDDEEEEFFFF",
     "GGGGHHHHJJJJ....."
  ]
}
```

cryptocode:
@kennell I'll take a look

kennell:
Yes this looks fine - thank you for the fast response.

androm3da:
This feature is abandoned, at least for now.  It would be contingent on more significant design changes.

PR #432, opened 2018-01-07
Support wallet_list in RPC

cryptocode:
Implements `wallet_list` in RPC.

See https://github.com/clemahieu/raiblocks/issues/430

### DIFF
--- a/rai/core_test/rpc.cpp
+++ b/rai/core_test/rpc.cpp
@@ -587,6 +587,34 @@ TEST (rpc, wallet_destroy)
 	ASSERT_EQ (system.nodes[0]->wallets.items.end (), system.nodes[0]->wallets.items.find (wallet_id));
 }
 
+TEST (rpc, wallet_list)
+{
+	rai::system system (24000, 1);
+	rai::rpc rpc (system.service, *system.nodes[0], rai::rpc_config (true));
+	rpc.start ();
+	rai::keypair key2;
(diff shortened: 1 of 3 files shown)


cryptocode:
@androm3da Test case added

androm3da:
@clemahieu any reason not to disclose the wallet IDs to the RPC peer?  Sounds like a reasonable request from #430 and RPC seems like it's designed to be trusted.

clemahieu:
The only reason I haven't added this is as a layer of security, basically the wallet IDs serve as an API key.  Right now even if someone gets access to an exposed RPC port they would need the wallet IDs to actually make transactions.

The only current way to get the wallet ids is on the command line, with this they could get it through the RPC system.

cryptocode:
@clemahieu So either add a config flag, or ditch this PR? I'm fine with both :)

PlasmaPower:
Be careful with this - CORS is enabled for the RPC. That means if you have rai_node running locally, and wallet_list enabled, any website you visit in your browser can steal your XRB.

clemahieu:
Yea I think we have to deal with the inconvenience for now until we get better security on the RPC port.  If we had a mapping of certificate -> wallet ID and could validate with https that would be ideal but it's only http at the moment.

cryptocode:
@PlasmaPower right, the eth guys are having this problem. Closing.

PR #439, opened 2018-01-07
Replace deep nesting with result short-circuit idiom

cryptocode:
This PR is a request for discussion regarding deep nesting more than anything else. These pyramids appear a few places.

I find the short-circuit idiom easier to read, and it nicely mirrors the corresponding `write_header` function.

`read_header` was low-hanging fruit (and I assume it's already covered by tests, seems to work fine)

### DIFF
--- a/rai/node/common.cpp
+++ b/rai/node/common.cpp
@@ -54,36 +54,18 @@ void rai::message::write_header (rai::stream & stream_a)
 
 bool rai::message::read_header (rai::stream & stream_a, uint8_t & version_max_a, uint8_t & version_using_a, uint8_t & version_min_a, rai::message_type & type_a, std::bitset<16> & extensions_a)
 {
+	uint16_t extensions_l;
 	std::array<uint8_t, 2> magic_number_l;
 	auto result (rai::read (stream_a, magic_number_l));
+	result = result || magic_number_l != magic_number;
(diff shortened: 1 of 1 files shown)


PlasmaPower:
There are a couple more functions utilizing `rai::read` like this that could be changed.

Personally, I'd propose using exceptions instead of `result`, but this is still an improvement IMO.

cryptocode:
@PlasmaPower Yeah I definitely agree on exceptions, but I think some discussions about stuff like the exception hierarchy (if any) and "who catches what" is in order before taking that step.

lukealonso:
-1 to exceptions in the packet parsing code, where failures don't need to be needlessly slow,  but +1 for this change

cryptocode:
@lukealonso Yeah, I agree exceptions are of little benefit in this specific instance. As for slowness and exceptions, I'll make a comment in the issue regarding exceptions.

androm3da:
"needlessly slow"?  Just how much of a penalty are we talking about to unwind the stack?

I think exceptions make more sense here and it looks like they would preserve the semantics of the original code better.

pocesar:
the exceptions can be 10-20x slower @androm3da and in performance critical parts, it shouldn't be throwing, ever

PR #440, opened 2018-01-08
Fix wallet race condition.

lukealonso:
The rekey() operation writes to memory and to LMDB, but the visibility of the two operations is totally different, leading to a number of race conditions.

I had another version of this patch that made the visibility of the password match that of the LMDB transaction through a new transaction finalizer concept, and a transaction read barrier, but it was much more complex. This version just duplicates the wallet key in memory, where visibility can be serialized with the password.

### DIFF
--- a/rai/node/wallet.cpp
+++ b/rai/node/wallet.cpp
@@ -27,10 +27,12 @@ rai::uint256_union rai::wallet_store::salt (MDB_txn * transaction_a)
 
 void rai::wallet_store::wallet_key (rai::raw_key & prv_a, MDB_txn * transaction_a)
 {
-	rai::wallet_value value (entry_get_raw (transaction_a, rai::wallet_store::wallet_key_special));
+	std::lock_guard<std::recursive_mutex> lock (mutex);
+	rai::raw_key wallet_l;
+	wallet_key_mem.value (wallet_l);
 	rai::raw_key password_l;
 	password.value (password_l);
-	prv_a.decrypt (value.key, password_l, salt (transaction_a).owords[0]);
+	prv_a.decrypt (wallet_l.data, password_l, salt (transaction_a).owords[0]);
 }
 
 void rai::wallet_store::seed (rai::raw_key & prv_a, MDB_txn * transaction_a)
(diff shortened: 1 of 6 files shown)


rai/node/wallet.hpp
```diff
@@ -113,6 +114,7 @@ class wallet_store
 	rai::kdf & kdf;
 	rai::mdb_env & environment;
 	MDB_dbi handle;
+	std::recursive_mutex mutex;
```
lukealonso: Note: This isn't ideal, and indicative of a bigger restructuring that needs to happen in the wallet API.

rai/node/wallet.cpp
```diff
@@ -27,10 +27,13 @@ rai::uint256_union rai::wallet_store::salt (MDB_txn * transaction_a)
 
 void rai::wallet_store::wallet_key (rai::raw_key & prv_a, MDB_txn * transaction_a)
 {
-	rai::wallet_value value (entry_get_raw (transaction_a, rai::wallet_store::wallet_key_special));
+	mutex.lock();
```
androm3da: Can you use a `std::lock_guard<recursive_mutex>` here instead?

rai/node/wallet.cpp
```diff
@@ -126,15 +129,18 @@ bool rai::wallet_store::valid_password (MDB_txn * transaction_a)
...
 	rai::uint256_union check_l;
 	check_l.encrypt (zero, wallet_key_l, salt (transaction_a).owords[0]);
-	return check (transaction_a) == check_l;
+	bool ok = check (transaction_a) == check_l;
+	return ok;
 }
 
 bool rai::wallet_store::attempt_password (MDB_txn * transaction_a, std::string const & password_a)
 {
+	mutex.lock();
```
androm3da: ...and here

rai/node/wallet.cpp
```diff
@@ -152,6 +158,7 @@ bool rai::wallet_store::attempt_password (MDB_txn * transaction_a, std::string c
 bool rai::wallet_store::rekey (MDB_txn * transaction_a, std::string const & password_a)
 {
 	bool result (false);
+	mutex.lock();
```
androm3da: ditto

rai/node/wallet.cpp
```diff
@@ -689,18 +710,20 @@ node (node_a)
 
 void rai::wallet::enter_initial_password ()
 {
+	rai::transaction transaction (store.environment, nullptr, true);
+	store.mutex.lock();
```
androm3da: another case for `std::lock_guard<mutex>`

lukealonso:
@androm3da done

Issue #384, opened 2018-01-03
RFC: New GUI implementation


rainbyte:
I have been writing an unofficial new GUI implementation for RaiBlocks wallet using Qt Quick, code is published [here](https://github.com/rainbyte/raiblocks/tree/qml_gui).

Recently I uploaded it, after decent level of supported features was reached (reddit post [here](https://www.reddit.com/r/RaiBlocks/comments/7moghd/implementing_new_gui/)).

Some parts remain missing, but I'm adding them and rebasing from official repo.

If this could be reviewed and merged, maybe it could improve user experience.

These are some issues which could be solved with new implementation: #78 #254 #132 #1 #383 

Thanks in advance for reading this!

soliury:
good job

cryptocode:
@rainbyte Cool. The work I'm going to do on QR code support should be self contained enough to work in your case as well.

dannyvankooten:
I'm not a Raiblocks maintainer or contributor but I really like how this looks. Great work @rainbyte! 

I reckon the devs would like a pull request for this with details on what exactly is still missing? That will make it a lot more likely for this to get merged, I think.

rainbyte:
Only the functionality inside "advanced" and "import" submenus is missing, but old submenus are shown in a separate window as workaround.

I mean, I have ported piece by piece keeping the missing parts intact in a 2nd window, always available.

This was the only way, because QtWidgets cannot be used inside Qml ApplicationWindow.

rainbyte:
@cryptocode I tried to integrate [this](https://github.com/nayuki/QR-Code-generator) qr lib, but I had no luck with it yet. Great to know you are working on that :D

cryptocode:
@rainbyte yeah, I have a working prototype

rainbyte:
@cryptocode , I have found [this](https://github.com/Acce0ss/qqr.js) qrcode library and to my surprise it worked!

Maybe not for the long term, but it is useful to fill the gap while you finish your version.

Also, [this](https://github.com/neocotic/qrious) is a rewrite of the library above, but not yet available for Qml.

pocesar:
I was going to do it using electron, but you beat me to it hahaha, looks much better

donutloop:
That looks pretty good!

Syd:
I thought about doing the same, but wondered if work had already begun to port it to quick for the mobile client?

rainbyte:
[...]

PR #441, opened 2018-01-08
New Qml GUI implementation

rainbyte:
This is new Qml GUI implementation (see #384 ), rebased to include last changes from main repo.

I'm opening a pull request to not let it diverge to much from main codebase, and because even if not 100% complete almost all functionality is implemented.

Missing parts (advanced tools) are available by showing old GUI via a button inside settings panel.


pocesar:
if you rebase and squash your commits is easier to review

androm3da:
@pocesar  These ones might not necessarily deserve squashing.  In any case the "files changed" sums the changes for review.

@rainbyte I will recommend that you create a new PR targeting a new branch where we can all evaluate and consider this change since its impact appears significant.  What do you think?

pocesar:
@androm3da I was talking about reviewing locally, but if it's done on github, indeed

rainbyte:
@androm3da , I will make it pass these CI tests and target new branch

rainbyte:
It seems that missing QtQuick deps is the reason why build fails.

I have tried adding those deps (look [here](https://github.com/rainbyte/raiblocks/commit/aea51cb3ee88ccd41ab1ec17577a4ebc5f32504c)), but it didn't work.

Should they be added to other place?

bonrry:
Reading DailyDevUpdate number 6, I guess this PR is not relevant anymore.
Still, kuddos to @rainbyte for putting it together! I keep using your branch on my laptop man ;)
screenshot from 2018-01-24 11-39-50

If the core team choose to go with QML for the new Wallet (would be a good choice ;)) I'm in the starting blocks for reviewing/contributing!

rainbyte:
Maybe that's the case. After reading dev update, I think that this PR is not relevant if there is already other GUI implementation. 

These days I have been updating qml_gui branch to follow official repo (mainly rebase and minor changes), but I couldn't make it compatible with old Qt version used by CI Ubuntu image yet, because Qt 5.6 doesn't have many components available on Qt 5.10.

This is qml gui today, rabased to latest main repo commit f70efdc7c3d71ddd0e1604505adfcb089ff566fb

captura de pantalla de 2018-01-24 08-39-28

rainbyte:
@CathalT , yeah, you are right... it would be better to support more Qt versions, maybe Qt 5.10 or Qt 5.9 LTS are too new for most users (even if those are better versions). The thing is that Qt 5.2.1 support is not needed anymore if there is other GUI implementation. We don't know even if that impl uses Qt or other lib. Maybe we should wait for core dev comments before continue with this effort.

About Qml vs Qwidgets, I think Qml is easier from design POV and it makes UI code more independent from backend. If you read the code you can see that app behaviour can be encapsulated in the C++ part and Qml UI state is updated via signals, in that way UI doesn't have so much logic and can concentrate on layout, widget design (colours, effects, etc), and user experience. Qwidgets are more complex, and UI code ends up mixed with logic implementation, eg. why should account or key handling code mixed with buttons and colours code? That's not good, it is a mess. It seems Qml gives more opportunities for modularization, encapsulation of logic, better user experience, and also more platforms (eg. Android, iOS, etc). And, as you said, embedded devices is other market which could be interesting for RaiBlocks, and could be supported with Qml too. Obviously Qml is different, so devs need to get used to a new approach of tcoding.

bonrry:
Hey, I found over time that Qt is pretty good to help handle a modular native desktop app for Win+MacOs+Linux, and QML is really powerful for a modern UI that just works across OSes... But let's not troll here :trollface: 

I'm just concerned the new UI may need the local node's RPC service to be ON (on localhost interface, but still a no-go for me), seeing @devinus 's background with JS 😜 I hope my feeling is wrong... ^^

rainbyte:
@bonrry , Qml gui doesn't use RPC (and neither Qwidgets), because it is connected directly to C++ backend via Qt signals & slots. I don't know if the UI mentioned in dev update uses RPC.

bonrry:
Sure I know that, and that's what I like about them ;)
I was just saying that I hope the new wallet stick to it and doesn't require exposing the RPC service on the local machine

CathalT:
You shouldn't mix your business code logic with your UI logic you're correct! Thats just poor design, not a QWidget vs Qml issue. Qml makes it easier to do so , that's a fair point!

Yeah looks like they are moving away from Qt and to node.js going from the reddit comments. (Take that with a pinch of salt)

Supporting Qt 5.2.1 came from : https://github.com/clemahieu/raiblocks/issues/527
I wonder if its still a valid issue...

rainbyte:
@CathalT , I have written a comment there. CI build is using other Ubuntu version, which has Qt 5.5 in the repo. Travis log is a bit misleading about this, I was also confused about Ubuntu version.

I think best solution would be to decide and document minimal Qt version, maybe 5.9 LTS is good choice. Qt 5.9 will be supported lot of time, and it has a good balance between new features, availability and testing.

androm3da:
I tried the changes from https://github.com/rainbyte/raiblocks `qml_gui` and got `QQmlComponent: Component is not ready` when starting up.  Is it a bug?

I found this but didn't look any further.  https://stackoverflow.com/a/18025964/489590

rainbyte:
This happens with old Qt version because some widgets are not supported (mainly qtquick controls 2), it should not happen with new version. I think better error reporting could be added.

androm3da:
Wow, really?  This is on ubuntu 17.10 which seems fairly recent.

androm3da:
> I wonder what will hppen with this effort after dev update gui is published.

Good question, I'm not sure.  But for now I'd love to give it a try.  What platform are you building it on?

rainbyte:
I'm using Archlinux here. I appreciate suggestions to continue improving this. Thanks!

PlasmaPower:
Unfortunately this might go to waste with the new Electron UI 😢. I wish we would've had some warning about that.

androm3da:
I don't know how hard it would be to target the Qt release that comes with ubuntu 17.10 (5.9.1?), but I would definitely use this if it could.  Or is there some way to soft/gradually de-feature it if the libqt found  is missing some newer capabilities?

rainbyte:
Maybe it is time to close this PR? It seems to useless because electron ui release is near...

PR #450, opened 2018-01-08
Add an RPC to request and track votes, along with a callback

PlasmaPower:
I rebased #362 then added a callback. The original commit is the same thing as in #362, and the next commit is my changes.

Unfortunately callbacks aren't currently tested, and I'm not 100% sure how to setup the infrastructure for that. I might look into it later.

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -1432,17 +1432,18 @@ store (init_a.block_store_init, application_path_a / "data.ldb", config_a.lmdb_m
 gap_cache (*this),
 ledger (store, config_a.inactive_supply.number (), config.state_block_parse_canary, config.state_block_generate_canary),
 active (*this),
-wallets (init_a.block_store_init, *this),
 network (*this, config.peering_port),
 bootstrap_initiator (*this),
 bootstrap (service_a, config.peering_port, *this),
 peers (network.endpoint ()),
 application_path (application_path_a),
+wallets (init_a.block_store_init, *this),
 port_mapping (*this),
 vote_processor (*this),
 warmed_up (0),
 block_processor (*this),
-block_processor_thread ([this]() { this->block_processor.process_blocks (); })
+block_processor_thread ([this]() { this->block_processor.process_blocks (); }),
+online_reps (*this)
 {
 	wallets.observer = [this](bool active) {
 		observers.wallet (active);
@@ -1570,6 +1571,9 @@ block_processor_thread ([this]() { this->block_processor.process_blocks (); })
 	});
 	observers.vote.add ([this](std::shared_ptr<rai::vote> vote_a, rai::endpoint const &) {
 		this->gap_cache.vote (vote_a);
+	});
+	observers.vote.add ([this](std::shared_ptr<rai::vote> vote_a, rai::endpoint const &) {
+		this->online_reps.vote (vote_a);
 	});
 	observers.vote.add ([this](std::shared_ptr<rai::vote> vote_a, rai::endpoint const & endpoint_a) {
 		if (this->rep_crawler.exists (vote_a->block->hash ()))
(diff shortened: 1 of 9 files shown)


clemahieu:
Cool, I'll take a look.

PlasmaPower:
There might be an issue with the callback when there are multiple votes for a block. After looking at the callback output for a bit, I didn't see any >50%, and I only saw a couple where the total weight agreeing was greater than the weight of that vote.

Might just be chance though, IDK. The node also wasn't running for long before hand, so I might have just been picking up the last couple of votes for a block.

PlasmaPower:
I've rebased these changes and added a confirmation callback, which seems to be working as expected

androm3da:
Is this a regression?

```
[ RUN      ] node_config.serialization
/workspace/rai/core_test/node.cpp:517: Failure
Value of: config1.bootstrap_fraction_numerator
  Actual: 10
Expected: config2.bootstrap_fraction_numerator
Which is: 1
```

Also, please run the clang-format-all.sh

PlasmaPower:
Formatting fixed and some commits squashed. I don't think that's a regression.

rai/node/node.cpp
```diff
@@ -2742,20 +2865,31 @@ rai::uint128_t rai::election::minimum_threshold (MDB_txn * transaction_a, rai::l
...
+		auto node_l (node.shared ());
 		auto exceeded_min_threshold = winner->first > minimum_threshold (transaction_a, node.ledger);
 		if (!(*block_l == *last_winner))
 		{
 			if (exceeded_min_threshold)
 			{
-				auto node_l (node.shared ());
 				node.background ([node_l, block_l]() {
-					node_l->block_processor.process_receive_many (rai::block_processor_item (block_l, true));
+					node_l->block_processor.process_receive_many (rai::block_processor_item (block_l, true), [node_l, block_l](rai::account account, rai::amount amount) {
```
lukealonso: Can we avoid this callback? An alternative would be to let the fork block get processed by process_receive_many normally, then let it either trigger the election normally through the block observer, or if it's not in `recent` anymore, use the election history:

```		
                if (this->block_arrival.recent (block_a->hash ())) {
			rai::transaction transaction (store.environment, nullptr, true);
			active.start (transaction, block_a);
		} else if ( ... check election history .. ) {
                        // trigger callback with results of the election + the params from the block observer
                }
```

rai/node/node.cpp
```diff
@@ -2764,8 +2898,18 @@ void rai::election::confirm_once (MDB_txn * transaction_a)
 				BOOST_LOG (node.log) << boost::str (boost::format ("Retaining block %1%") % last_winner->hash ().to_string ());
 			}
 		}
+		else if (exceeded_min_threshold && (bool)last_winner_callback_info)
```
lukealonso: Personally, I think we should only say a block was confirmed if it exceeded 50%, since that's the only safe number (and even then not completely safe due to reasons I won't write here).

rai/node/node.cpp
PlasmaPower: I understand why having a callback there can be misleading, but your proposed solution sounds too convoluted. I'll think about better ways of implementing this.

PlasmaPower:
Unfortunately this is going to need to wait until a vote republishing rules fix is deployed. The rate limiting per block should be per block per rep.

PlasmaPower:
Note to self: add `final` flag to election results, indicating if the election is over, regardless of whether the queried block won.

rai/node/node.cpp
```diff
@@ -2345,6 +2479,69 @@ bool rai::block_arrival::recent (rai::block_hash const & hash_a)
...
+	auto rep_it (reps.get<1> ().find (rep));
+	auto info (rai::rep_last_heard_info{ now, rep });
+	if (rep_it == reps.get<1> ().end ())
+	{
+		auto old_stake (online_stake);
+		online_stake += node.weight (rep);
+		if (online_stake < old_stake)
+		{
+			// overflow
+			online_stake = ~0;
```
cryptocode: While correct, `std::numeric_limits<rai::uint128_t>::max()` is in line with the rest of the code base and makes the semantics clear.

rai/node/node.cpp
```diff
@@ -2345,6 +2479,69 @@ bool rai::block_arrival::recent (rai::block_hash const & hash_a)
...
+		reps.get<1> ().replace (rep_it, info);
+	}
+}
+
+void rai::online_reps::recalculate_stake (bool initializing)
+{
+	if (initializing)
+	{
+		// set stake to max until we're sure who's online
+		online_stake = ~0;
```
cryptocode: Same as above

rai/node/node.cpp
```diff
@@ -2345,6 +2479,69 @@ bool rai::block_arrival::recent (rai::block_hash const & hash_a)
...
+	std::lock_guard<std::mutex> lock (mutex);
+	auto now (std::chrono::steady_clock::now ());
+	while (!reps.empty () && reps.begin ()->last_heard + std::chrono::seconds (rai::node::cutoff) < now)
+	{
+		auto it (reps.begin ());
+		auto old_stake (online_stake);
+		online_stake -= node.weight (it->representative);
+		if (online_stake > old_stake)
+		{
+			// overflow
```
cryptocode: I think underflow was meant here.

rai/node/node.cpp
clemahieu: I think this also needs to be done because the negation is being done on an int type before conversion to the uint128_t

cryptocode: The integral conversion rules makes it work (~0 is signed). 

But ::max is clearer.

PlasmaPower:
This has been rewritten and the feature is now in master.

PR #452, opened 2018-01-09
Fix wallet tests

PlasmaPower:


rai/core_test/wallets.cpp
```diff
@@ -2,12 +2,28 @@
...
+	}
+}
+
+TEST (wallets, open_create)
 {
 	rai::system system (24000, 1);
 	bool error (false);
 	rai::wallets wallets (error, *system.nodes[0]);
 	ASSERT_FALSE (error);
+	destroy_existing_wallets (wallets);
```
lukealonso: This is a case where the expectation of the test is wrong, you should just change the assertion. The intent is to test the behavior of rai::wallets, not test new destroy code.

PlasmaPower: Makes sense. I just wasn't sure what was creating the wallet, and as such, I didn't want to rely on finicky behavior. I'll update this to the new behavior.

lukealonso: If the wallet has flaky behavior, that should definitely get fixed (and maybe another test added to further expose the bad behavior).

PlasmaPower: I'm pretty sure it doesn't, but at first, I thought another test was creating the wallet. I didn't realize they were independent at first.

PlasmaPower:
Fixed.

PR #466, opened 2018-01-09
Tweak wallet balance display.

lukealonso:
- Right-align the balances in history.
- Make the balance text consistent when viewing other accounts.

### DIFF
--- a/rai/qt/qt.cpp
+++ b/rai/qt/qt.cpp
@@ -243,10 +243,10 @@ void rai_qt::accounts::refresh_wallet_balance ()
 		balance = balance + (this->wallet.node.ledger.account_balance (transaction, key));
 		pending = pending + (this->wallet.node.ledger.account_pending (transaction, key));
 	}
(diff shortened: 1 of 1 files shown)


rai/qt/qt.cpp
```diff
@@ -243,10 +243,10 @@ void rai_qt::accounts::refresh_wallet_balance ()
 		balance = balance + (this->wallet.node.ledger.account_balance (transaction, key));
 		pending = pending + (this->wallet.node.ledger.account_pending (transaction, key));
 	}
-	auto final_text (std::string ("Wallet balance (XRB): ") + wallet.format_balance (balance));
+	auto final_text (std::string ("Balance: ") + wallet.format_balance (balance));
```
androm3da: This didn't require corresponding changes to the test case(s)?

lukealonso: This is for viewing other accounts in the advanced menu, doesn't have a test AFAIK.

PR #467, opened 2018-01-09
Add balance observer.

lukealonso:
This fixes an issue where pending balances won't show up until the user triggers a UI refresh some other way.


androm3da:
I suspected that this was happening. 

I think a lot of people have complained about this. Thanks.

JKJameson:
Awesome work, this has to be the biggest problem with the wallet making people think there are transaction delays, but in fact the UI just doesn't update itself.

pocesar:
funny how changing between Mrxb, and xrb usually would update the UI and show the actual balance

PR #468, opened 2018-01-09
Simplify code; mostly reduced nesting.

cryptocode:
A cleanup PR in the "flat is better than deep" category.

### DIFF
--- a/rai/lib/numbers.cpp
+++ b/rai/lib/numbers.cpp
@@ -257,28 +257,21 @@ void rai::uint256_union::encode_hex (std::string & text) const
 bool rai::uint256_union::decode_hex (std::string const & text)
 {
 	auto result (false);
-	if (!text.empty ())
+	if (!text.empty () && text.size () <= 64)
 	{
-		if (text.size () <= 64)
+		std::stringstream stream (text);
+		stream << std::hex << std::noshowbase;
+		rai::uint256_t number_l;
+		try
 		{
-			std::stringstream stream (text);
-			stream << std::hex << std::noshowbase;
-			rai::uint256_t number_l;
-			try
-			{
-				stream >> number_l;
-				*this = number_l;
-				if (!stream.eof ())
-				{
-					result = true;
-				}
-			}
-			catch (std::runtime_error &)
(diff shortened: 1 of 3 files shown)


rai/lib/numbers.cpp
```diff
@@ -257,28 +257,21 @@ void rai::uint256_union::encode_hex (std::string & text) const
 bool rai::uint256_union::decode_hex (std::string const & text)
 {
 	auto result (false);
-	if (!text.empty ())
+	if (!text.empty () && text.size () <= 64)
```
PlasmaPower: How about:

```cpp
if (text.empty() || text.size() > 64) return true;
```

cryptocode: @PlasmaPower I believe early returns are frowned upon in the code base? @clemahieu ?

PlasmaPower: I've been wondering about that, but I haven't seen a place where objects need cleanup (that isn't done by destructors).

cryptocode: IMO it's about readability, as lots of exit points can be hard to reason about, and are bug prone when refactoring/maintaining.

JKJameson: I agree with OP because functions are intended to be returned, it's much cleaner to have a "if <whatever> then return block" rather than a ton of nested if statements.

cryptocode: @JKJameson As usual in software engineering, there are tradeoffs. At any rate, this patch attempts to reduce nesting while keeping in line with the code base's established practice of avoiding early returns. Someone from core would have to chime in on the topic of changing that (I personally hope not, but I do want reduce nesting in a few places.)

JKJameson: A function's sole purpose is to minimize replicated code. However, logically it's purpose is to return a value, therefore the sooner it returns the better and the cleaner the code.

cryptocode: @JKJameson Interesting debate, I'm sure, but this patch improves nesting, and the preference in the current code base is single exits, so I'm adhering to that.

PR #471, opened 2018-01-10
Fix wallet creation crash, add lmdb_max_dbs config.

lukealonso:
This fixes a problem encountered by some users with many wallets, where the node would crash or behave incorrectly when creating additional wallets. The underlying cause is the max db limit in LMDB (plus a mistake in the wallet store), which is hardcoded to 128. @clemahieu suggested we make it a config param.  

Also fixes the RPC endpoint to correctly report wallet creation failed.

### DIFF
--- a/rai/node/utility.hpp
+++ b/rai/node/utility.hpp
@@ -101,7 +101,7 @@ bool fetch_object (T & object, boost::filesystem::path const & path_a, std::fstr
 class mdb_env
 {
 public:
-	mdb_env (bool &, boost::filesystem::path const &);
+	mdb_env (bool &, boost::filesystem::path const &, int max_dbs = 128);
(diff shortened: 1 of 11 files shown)


rai/node/utility.hpp
```diff
@@ -101,7 +101,7 @@ bool fetch_object (T & object, boost::filesystem::path const & path_a, std::fstr
 class mdb_env
 {
 public:
-	mdb_env (bool &, boost::filesystem::path const &);
+	mdb_env (bool &, boost::filesystem::path const &, int max_dbs = 128);
```
lukealonso: I'm usually allergic to default args, but in this case it means several hundred edits or other refactoring for what will only be a temporary solution. We should remove this when the wallet store is refactored.

PR #479, opened 2018-01-11
Initial doxygen configuration file

cryptocode:
There's some interest in documenting the source base, and it's nice if everyone involved lands on the same format. 

This PR proposes doxygen. The configuration file will probably needs tweaking, but it's a start - enough to get a nice class hierarchy overview and file listings. 

As a starting point, I've set the knobs to allow for both Javadoc and C++ tripleslash style comments with auto-brief (otherwise `\brief` or `@brief ` annotations are required, which is a bit noisy.)

**Usage:** `doxygen doxygen.config` from the project root.

I assume the CI can call doxygen and upload the result somewhere?

I've left all the possible config flags and comments intact. We can do a cleanup once the config has settled.


androm3da:
As discussed offline - please add this to the CI so we can see the results and/or detect bugs in the doc comments.

androm3da:
If this was obsoleted by #483, please close it.

Issue #446, opened 2018-01-08
Should we go for exceptions?


cryptocode:
The topic of exceptions came up in https://github.com/clemahieu/raiblocks/pull/439

There's a lot of code in the source tree that does error handling by checking return codes.

I don't  necessarily think throwing exceptions everywhere is appropriate, but some of the code would certainly benefit from a well designed exception hierarchy/structure. And a lot of boost/std methods throw exceptions anyway.

pocesar:
unless the failure result is not discardable, there's no point of exceptions. for example, in the `read_header` method, does it matter if the magic number doesn't match or the version is wrong? I think that would only matter for analyzing debug logs, but without the payload itself, would be of no use. I guess the short-circuiting is the way to go

PlasmaPower:
With exceptions though, you don't have to do short circuiting. You can just call the functions and only worry about failure where it needs handled.

cryptocode:
@pocesar I don't think the read_header method is a good example. There's a lot more complex methods (and chains thereof) that would benefit more.

pocesar:
I'm not fond of exceptions for code that are executed in tight loops, that's something to be aware of. even if it's better for code maintainability, exceptions shouldn't be thrown in functions that performance is crucial (`read_header` is one of them)

PlasmaPower:
C++ exceptions are only slow for the failure case though. How crucial is performance then? I don't think it'll be the bottleneck.

cryptocode:
@PlasmaPower Right. I think the summary at the end of https://mortoray.com/2013/09/12/the-true-cost-of-zero-cost-exceptions/ is good (at least for Itanium ABI compilers). 

In short, it probably doesn't matter, especially not in code dominated by IO where exception cost is basically noise. The "happy path" is fast, but the concern is definitely something to consider in critical paths.

pocesar:
if you think that the nodes / clients will always receive sane packets, indeed, exceptions will be rare. that's not the case

PlasmaPower:
No, I just think that the vast majority (but not 100%) of packets will be sane. I'm also confident this is not the weakest link for a DDOS attack.

cryptocode:
Not always, but almost always.

lukealonso:
In this particular instance it's a dumb waste of cpu cycles in code where the benefit is zero, minus the time that could have been spent programming something useful.

cryptocode:
[...]

PR #480, opened 2018-01-12
Improve exception handling in the daemon

cryptocode:
I wanted to back up the #446 exception discussion with a PR.

While there aren't really many changes, the diff looks like that since there's wrapping of code into try/catch blocks.

It solves a real problem I've been having: misconfigurations/typos in config.json causes daemon startup to fail, but with no clues as to what might be wrong.

The cause of errors while parsing the config file is lost when turning runtime_error into error flags. @Dekoze mentions a similar problem in https://github.com/clemahieu/raiblocks/issues/446

If I mistype something in config.json, I now get a message like `Configuration error: Invalid json in 'rpc'` The precision can be improved further, but it's a start.

I'm proposing the introduction of rai/lib/exceptions.hpp, where we can define exception types, making it easy to do exception-specific error handling. They should all inherit from runtime_error, allowing catch-all when it makes sense.

Finally, this PR ties into https://github.com/clemahieu/raiblocks/pull/479, as I've documented the exceptions with `@throws` doxygen annotations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,6 +237,7 @@ add_library (secure
 
 SET (RAI_LIB_SOURCES
 	${PLATFORM_LIB_SOURCE}
+	rai/lib/exceptions.hpp
 	rai/lib/blocks.cpp
 	rai/lib/blocks.hpp
 	rai/lib/interface.cpp
--- a/rai/core_test/uint256_union.cpp
+++ b/rai/core_test/uint256_union.cpp
@@ -406,8 +406,7 @@ TEST (json, fetch_object)
 	stream1.close ();
 	rai::open_or_create (stream1, path1.string ());
 	json_upgrade_test object1;
-	auto error1 (rai::fetch_object (object1, path1, stream1));
-	ASSERT_FALSE (error1);
+	ASSERT_NO_THROW (rai::fetch_object (object1, path1, stream1));
 	ASSERT_EQ ("changed", object1.text);
 	boost::property_tree::ptree tree1;
 	stream1.close ();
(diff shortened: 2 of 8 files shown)


lukealonso:
@cryptocode 

```
-#include <string>
 #include <stdexcept>
+#include <string>
 
 namespace rai
 {
Code formatting differs from expected - please run ci/clang-format-all.sh
```

cryptocode:
Rebased again

clemahieu:
Error logging in the daemon definitely needs to be fixed.  I want the railib to be an ANSI C interface so we won't want exceptions at that point.

For getting better information out of the I was thinking of passing in a class that holds a list of strings and error messages can be inserted in there as strings.

cryptocode:
@clemahieu I was thinking it might be a good idea to use exceptions up to the C api (since there's no escaping some runtime_error's anyway, and thus to avoid errors being swallowed.)

For clients of the C api, you would definitely use strings. 

If a C api functions return an error, maybe a call `rai_api_error_string` to get the error? Not an uncommon design, but there are of course many options here.

Is this in line with your thinking, or do you want to avoid exceptions internally?

cryptocode:
Closing this after offline discussions, the code base is going to prefer catching exceptions as close to the source as possible. 

Just some closing thoughts: I think an exception that express JSON misconfiguration is an excellent exception(!)  to the (mostly good) rule of catching locally, since there's nothing sensible you can do but report the error to the user in the most precise way possible.

By catching it early, you must make assumptions on how to report to the user. If the config error is detected in a CLI setting, you want to report to the terminal. If from a daemon service, you want to report to a log file. In other settings, maybe in other ways.

Throwing long is thus fine and simplifies the code base considerably. Just my 2 cents ;)

PR #483, opened 2018-01-13
Doxygen config

PlasmaPower:
#479 + testing that Doxygen builds on Travis. No deploy yet, as there are some unanswered questions there, such as where to put the output.


cryptocode:
@PlasmaPower Great! Maybe push it to GitHub pages, as some do?

PlasmaPower:
Yeah, however, that'll be complicated to test. Travis doesn't give PR builds secrets (otherwise a malicious PR could steal the secrets), so I can't test that deploy is working with a PR.

Issue #399, opened 2018-01-05
Support IPv4 only hosts


androm3da:
Right now rai_lib uses IPv4-mapped region of IPv6.  Hosts w/o a dual-stack will see failures to bind to addresses (`EADDRNOTAVAIL` etc).

androm3da:
Other mitigations include:
* provide clearer error message describing the problem
* offer a no-rpc command line option that won't consider this error to be fatal
* add appropriate details to the troubleshooting guide

bra1n:
Any updates here? Or workarounds? Would love to run a node on an IPv4-only server...

clemahieu:
I don't think it's worthwhile, aside from removing the ipv6 stack intentionally I don't think there are any operating systems capable of running a node that don't support IPv6.

rkeene:
I'm closing this issue based on the information we have at hand that having an IPv6 stack is sufficient to run the node, even with no IPv6 addresses or routes.  Having an IPv6 stack currently seems like a reasonable requirement for a network daemon in the current environment.  If there are further questions, let me know !

PR #498, opened 2018-01-15
Delete redundant calls to ```node.peers.insert```.

Nevsor:
```node.peers.contacted``` does already call ```node.peers.insert``` so the deleted lines are not necessary.

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -350,7 +350,6 @@ public:
 		}
 		++node.network.incoming.publish;
 		node.peers.contacted (sender, message_a.version_using);
-		node.peers.insert (sender, message_a.version_using);
 		node.process_active (message_a.block);
 	}
 	void confirm_req (rai::confirm_req const & message_a) override
@@ -361,7 +360,6 @@ public:
 		}
 		++node.network.incoming.confirm_req;
 		node.peers.contacted (sender, message_a.version_using);
-		node.peers.insert (sender, message_a.version_using);
 		node.process_active (message_a.block);
 		rai::transaction transaction_a (node.store.environment, nullptr, false);
 		if (node.store.block_exists (transaction_a, message_a.block->hash ()))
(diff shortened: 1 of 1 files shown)


cryptocode:
@Nevsor I don't think this patch is correct.

`contacted` switches to a different mapping if the input is ipv4. The insert in `contacted` inserts the local variable, not the argument.

PlasmaPower:
We should always switch to ipv6 mapping, right? Also, if this patch is wrong, then `insert` should be added to keepalive, I think.

cryptocode:
@PlasmaPower My idea is that endpoint_l conditionally changes, hence the insert isn't redundant.

PlasmaPower:
It may not be completely redundant, but it's still incorrect. If the peer is an IPv4 address, then it'll duplicate the peer in its peer list (both IPv4 and IPv6 forms). I think we only want the IPv6 form.

cryptocode:
@PlasmaPower ok, I assumed both forms were registered as you later needed to check for contacted-status from both forms.

cryptocode:
@PlasmaPower I see, so, say, known_peer, will never be called with an IPv4?

PlasmaPower:
It shouldn't be. We don't want to duplicate peers as both IPv4 and IPv6. Otherwise, we might broadcast blocks to the same peer twice, or we might try to bootstrap from the same peer twice.

PlasmaPower:
There might be more changes needed to make sure that peers are completely IPv6, I'm not sure.

cryptocode:
@PlasmaPower does an eventual implementation of https://github.com/clemahieu/raiblocks/issues/399 change anything? Or will there still be an ipv6 *mapping* ?

PlasmaPower:
I think if we fix #399, we'll still use IPv6 internally, but we can use IPv4 externally. A lot of the codebase hardcodes IPv6.

cryptocode:
Closing this PR as the node_id handshake is now the only `insert` callee from the node.

PR #500, opened 2018-01-16
CMake and Boost 1.66

argakiig:
Adds CMake option BOOST_CUSTOM to inline FindBoost.cmake allowing for detection of Boost 1.66 built from source on Windows and Linux. Would be interested to see if it works on Mac as well and if the option could be removed all together

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,11 @@ include_directories (${CMAKE_SOURCE_DIR})
 
 set(Boost_USE_STATIC_LIBS        ON)
 set(Boost_USE_MULTITHREADED      ON)
+
+if (WIN32)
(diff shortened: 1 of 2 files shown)


cryptocode:
@argakiig This will be fixed it seems: https://gitlab.kitware.com/cmake/cmake/issues/17575

I suppose we should flip back to the cmake-provided version once 1.66 is supported, to avoid having to maintain and sync FindBoost.cmake with upstream (to benefit from bug fixes, etc)?

argakiig:
Well once the changes get released we could just move the cmake requirement to a newer version. But with the inline version we are able to cover all older and out of date version. So realistically we could do either.

cryptocode:
I see, thanks for explaining.

PR #507, opened 2018-01-18
Re-enable payment_wait test

PlasmaPower:
Not sure why it was disabled, but it passes consistently on my machine. It takes about 1 second, which I don't think is too long.


androm3da:
How many iters did you attempt?

PlasmaPower:
I just ran it another 100 times to be sure, and it passed all of them.

PR #510, opened 2018-01-19
Apply bootstrap fork handling uniformly.

lukealonso:
Blocks that go into unchecked lose the callback that triggers fork handling. This patch triggers
it directly, instead of relying on the callback.

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -589,32 +589,7 @@ void rai::bulk_pull_client::received_block (boost::system::error_code const & ec
 			{
 				expected = block->previous ();
 			}
-			auto attempt_l (connection->attempt);
-			auto pull_l (pull);
-			attempt_l->node->block_processor.add (rai::block_processor_item (block, [attempt_l, pull_l](MDB_txn * transaction_a, rai::process_return result_a, std::shared_ptr<rai::block> block_a) {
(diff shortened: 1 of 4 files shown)


lukealonso:
@SergiySW

SergiySW:
Guess this way blocks_processing can be deque of simple std::shared_ptr<rai::block> instead of block_processor_item

lukealonso:
Good point, forgot to remove that also.

lukealonso:
@SergiySW Still need the 'force' bool for rollbacks, unless we do that some other way

rai/node/bootstrap.cpp
```diff
@@ -1124,6 +1110,15 @@ void rai::bootstrap_initiator::notify_listeners (bool in_progress_a)
 	}
 }
 
+void rai::bootstrap_initiator::process_fork (MDB_txn * transaction, std::shared_ptr<rai::block> block_a)
+{
+	std::unique_lock<std::mutex> lock (mutex);
+	if (attempt != nullptr)
```
androm3da: Does it really make sense to compare a `shared_ptr` with `nullptr`?  Maybe `if (attempt)` makes more sense?

lukealonso: Thanks, old habits die hard. This is valid, but I agree it's not idiomatic at all.

PR #518, opened 2018-01-21
Improve bootstrap performance.

lukealonso:
This patch reduces initial bootstrap times by searching for higher-performing peers, and improves the overall behavior when the number of bootstrap peers is high.

I recommend setting "bootstrap_connections" to 32, 64 or 128 in the config to test.

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -159,7 +159,11 @@ node (node_a),
 attempt (attempt_a),
 socket (node_a->service),
 endpoint (endpoint_a),
-timeout (node_a->service)
+timeout (node_a->service),
+block_count (0),
+pending_stop (false),
+hard_stop (false),
+start_time (std::chrono::steady_clock::now ())
 {
 	++attempt->connections;
 }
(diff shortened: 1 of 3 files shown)


rai/node/bootstrap.cpp
```diff
@@ -162,9 +166,33 @@ rai::bootstrap_client::~bootstrap_client ()
 	--attempt->connections;
 }
 
+double rai::bootstrap_client::block_rate () const
+{
+	auto elapsed = elapsed_seconds ();
+	if (elapsed > 0.0)
```
androm3da: Consider instead 

```
return (elapsed > 0.0)
  ? (double)block_count.load () / elapsed
   : 0.0
```

rai/node/bootstrap.cpp
```diff
@@ -318,13 +328,27 @@ void rai::frontier_req_client::received_frontier (boost::system::error_code cons
 		rai::bufferstream latest_stream (connection->receive_buffer.data () + sizeof (rai::uint256_union), sizeof (rai::uint256_union));
 		auto error2 (rai::read (latest_stream, latest));
 		assert (!error2);
+		if (count == 0)
+		{
+			start_time = std::chrono::steady_clock::now ();
+		}
 		++count;
 		auto now (std::chrono::system_clock::now ());
 		if (next_report < now)
```
androm3da: Seems like this should be sampling `steady_clock` too.

rai/node/bootstrap.cpp
```diff
@@ -939,29 +968,84 @@ bool rai::bootstrap_attempt::consume_future (std::future<bool> & future_a)
...
+				if (elapsed > 30 && rate < 10)
+				{
+					client->stop (true);
+				}
+			}
 		}
-		else
+	}
+
+	if (sorted_connections.size () >= (node->config.bootstrap_connections * 2 - 1) / 3 && node->config.bootstrap_connections > 1)
```
androm3da: Can you decompose this expression so it's easier to digest, or give it a name with a local `bool` or a function that returns `bool`?

rai/node/bootstrap.cpp
```diff
@@ -939,29 +968,84 @@ bool rai::bootstrap_attempt::consume_future (std::future<bool> & future_a)
...
+	{
+		auto drop = std::max (1, (int)(sqrtf ((float)node->config.bootstrap_connections * 2.0f) + 0.5f) - 1);
```
androm3da: Similar comment here.  What's the signficance of this expression?  `√(2c+½)-1`  Why cull this number of bootstrap peers?

Also, we shouldn't weed anyone out if we're getting really good throughput from everyone, right?  What if they were only thrown out if the aggregate is slower than a threshold?  Or only throw out the outliers -- ones that are N std devs away from the median?

lukealonso: This has the nice property that there's no threshold param to fiddle with, it just uses the bottom N slots for discovering better peers (but doesn't interrupt them mid-pull, it lets the account finish pulling before dropping them).

The expression is arbitrary, but seemed to provide a nice balance for a wide range of peer counts.

rai/node/bootstrap.cpp
lukealonso: Sure, been writing too much Go lately, no ternary :)

PR #520, opened 2018-01-22
Preventing possible requeue_pull loop

SergiySW:
if connection_shared is expired


rai/node/bootstrap.cpp
```diff
@@ -1032,8 +1032,8 @@ void rai::bootstrap_attempt::requeue_pull (rai::pull_info const & pull_a)
 	}
 	else if (pull.attempts == 4)
 	{
+		pull.attempts++;
```
cryptocode: Not specific to this PR, but it seems to me that pull_info#attempts should be atomic.

PR #526, opened 2018-01-23
make empty arrays appear as [] instead of "" after json serialization

### LINKED ISSUE #499: RPC "chain" inconsistent return type
**Description of bug:**

When issuing a "chain" RPC call, the type of the returned "blocks" property is defined as an array of strings in the RPC documentation. When providing a valid block hash, this holds true. However, if you supply an invalid block hash (e.g. the one listed in the documentation example), the block result is an empty string. 

I think it should either return an empty array, or return an error.

1. `curl -v -d '{ "action": "chain", "count": 1, "block": "000D1BAEC8EC208142C99059B393051BAC8380F9B5A2E6B2489A277D81789F3F"  }' [::1]:7076`

 `{ "blocks": "" } `

 `{ "blocks": [] } `

or 

 `{ "error": "Invalid block hash" } `

This is especially problematic for strongly typed languages when they are deserializing.

**Environment**:

**logs**

triamazikamno:
A small workaround for the boost's inability to serialize empty arrays, makes life much easier for clients on strongly typed languages. Fixes #499

### DIFF
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -142,7 +142,8 @@ namespace
 {
 void fix_empty_array (boost::property_tree::ptree & property)
 {
-	if (property.empty ()) {
+	if (property.empty ())
+	{
 		property.put ("", "__empty_array__");
 	}
 }
@@ -915,7 +916,7 @@ void rai::rpc_handler::blocks ()
 			error_response (response, "Bad hash number");
 		}
 	}
-	fix_empty_array(blocks);
+	fix_empty_array (blocks);
 	response_l.add_child ("blocks", blocks);
 	response (response_l);
 }
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -140,6 +140,13 @@ void rai::rpc::observer_action (rai::account const & account_a)
 
 namespace
 {
+void fix_empty_array (boost::property_tree::ptree & property)
+{
+	if (property.empty ()) {
+		property.put ("", "__empty_array__");
+	}
+}
+
 void error_response (std::function<void(boost::property_tree::ptree const &)> response_a, std::string const & message_a)
 {
 	boost::property_tree::ptree response_l;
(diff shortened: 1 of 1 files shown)


cryptocode:
@triamazikamno how does this affect existing RPC clients? An RPC "version" action is probably desired at this point.

rai/node/rpc.cpp
```diff
@@ -4124,6 +4152,7 @@ void rai::rpc_connection::parse_connection ()
 					boost::property_tree::write_json (ostream, tree_a);
 					ostream.flush ();
 					auto body (ostream.str ());
+					boost::replace_all (body, "\"__empty_array__\"", "[]");
```
cryptocode: This is rather expensive on large output (since the entire buffer is scanned regardless of the presence of empty arrays)

androm3da:
I agree that a change like this deserves a protocol rev.

But I personally don't prefer this method.  Although I think the change's effect is a desired one.

I'd personally prefer replacing `property_tree` with a more suitable JSON library.  @clemahieu thoughts?

clemahieu:
Agreed on all points. Is it true boost can’t make that happen?  Is there a way we scan only smaller messages?  Empty results are probably under some size. Is there a better json library out there? 

Long term I want to move the RPC code or of process but that hasn’t been queued to be worked on yet, we probably don’t want to wait since I’m sure this is annoying.

If we can make scanning efficient by a size check and wait until json is moved out of process and maybe in to a different language, we wouldn’t have to bother with a c library change.

androm3da:
From what I've seen boost has no plans to implement this feature.  I saw comments to that effect on a proprerty_tree bug report from ~2014 or so.

A new JSON lib could resolve this issue and make all of the values typed appropriately.

cryptocode:
For the sake of completeness, here are some popular and compliant ones:

http://rapidjson.org/
https://github.com/nlohmann/json
https://github.com/dropbox/json11
https://github.com/open-source-parsers/jsoncpp
https://github.com/sirikata/json-spirit (based on Boost spirit, but unfortunately not a very active project)

clemahieu:
RapidJSON looks good according to this comparison.  https://github.com/miloyip/nativejson-benchmark

cryptocode:
Seems like most people prefer nlohmann/json (single header, nice API), but RapidJSON is definitely very fast. Even when network IO is most of the overhead (small messages), RapidJSON takes less resources.

cryptocode:
This will be fixed when https://github.com/nanocurrency/nano-node/pull/526 is implemented

PR #531, opened 2018-01-23
Use steady_clock and clean up epoch-time use

cryptocode:
This PR deals with time:

* There were two different seconds-since-epoch implementations, one using std::chrono, the other using boost::posix_time. I've factored out the chrono approach into common.hpp and updated all call sites to use that one (both approaches produced the same result)

* Using `system_clock` for measuring durations is problematic, since the system clock may jump (for instance when NTP kicks in, or DST transitions). `steady_clock` is thus used instead.

* Removed an ostream `operator<<` taking a system_clock which was never used.

I don't think any of the code depends on wall time (except seconds_since_epoch), but please verify this.

### DIFF
--- a/rai/core_test/block_store.cpp
+++ b/rai/core_test/block_store.cpp
@@ -1,4 +1,5 @@
 #include <gtest/gtest.h>
+#include <rai/node/common.hpp>
 #include <rai/node/node.hpp>
 #include <rai/versioning.hpp>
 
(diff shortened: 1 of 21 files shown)


androm3da:
This looks great, I've skimmed it a couple times now and it looks like they're all right so far.

I will go through each change in detail to try and make sure, though.

cryptocode:
Rebased since the newly merged https://github.com/clemahieu/raiblocks/commit/e06716f5ee59a6202167e43d23d058597e042de3 would cause conflicts.

cryptocode:
@gautarg now, it needs wall time to compute the epoch distance

PR #535, opened 2018-01-24
Fix QTimer dependency. Fixes #527

### LINKED ISSUE #527: QTimers dependency requires newer Qt?
Since https://github.com/clemahieu/raiblocks/pull/508 we started failing builds on Trusty.

We now have compilation errors like the one below.

@CathalT thanks so much for your contribution.  Would you be willing to rework the change with a CMake feature test or some other way to make sure it still builds on older distros?

```

/usr/include/qt5/QtCore/qtimer.h:82:17: note:   candidate expects 4 arguments, 2 provided
../rai/qt/qt.cpp: In lambda function:
../rai/qt/qt.cpp:1415:6: error: no matching function for call to 'QTimer::singleShot(std::chrono::duration<long int, std::ratio<1l, 1000l> >::rep, rai_qt::settings::settings(rai_qt::wallet&)::__lambda73::__lambda74)'
     });
      ^
../rai/qt/qt.cpp:1415:6: note: candidates are:
In file included from /usr/include/qt5/QtCore/QtCore:77:0,
                 from /usr/include/qt5/QtGui/QtGuiDepends:2,
                 from /usr/include/qt5/QtGui/QtGui:3,
                 from ../rai/qt/qt.hpp:9,
                 from ../rai/qt/qt.cpp:1:
/usr/include/qt5/QtCore/qtimer.h:81:17: note: static void QTimer::singleShot(int, const QObject*, const char*)
     static void singleShot(int msec, const QObject *receiver, const char *member);
                 ^
/usr/include/qt5/QtCore/qtimer.h:81:17: note:   candidate expects 3 arguments, 2 provided
/usr/include/qt5/QtCore/qtimer.h:82:17: note: static void QTimer::singleShot(int, Qt::TimerType, const QObject*, const char*)
     static void singleShot(int msec, Qt::TimerType timerType, const QObject *receiver, const char *member);
                 ^
/usr/include/qt5/QtCore/qtimer.h:82:17: note:   candidate expects 4 arguments, 2 provided
ninja: build stopped: subcommand failed.
```

CathalT:
Fix was a lot simpler than I had original thought.
Just took a look at the Qt source code, they are pretty much doing the exact same thing...

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -999,7 +999,7 @@ void rai::bootstrap_attempt::populate_connections ()
 	if (sorted_connections.size () >= (node->config.bootstrap_connections * 2) / 3 && node->config.bootstrap_connections >= 4)
 	{
 		// 4 -> 1, 8 -> 2, 16 -> 4, arbitrary, but seems to work well.
-		auto drop = (int)roundf(sqrtf((float)node->config.bootstrap_connections - 2.0f));
+		auto drop = (int)roundf (sqrtf ((float)node->config.bootstrap_connections - 2.0f));
 		for (int i = 0; i < drop; i++)
 		{
 			auto client = sorted_connections.top ();
--- a/rai/qt/qt.cpp
+++ b/rai/qt/qt.cpp
@@ -4,6 +4,14 @@
 
 #include <sstream>
 
+namespace QTimerWrapper
+{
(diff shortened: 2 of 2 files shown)


CathalT:
Wait, I need to double check 5.2.1 source code... dammit

CathalT:
@androm3da can you check if this compiles on Qt5.2.1 , going from the docs this should work. I'll try get setup myself in the mean time.

PR #536, opened 2018-01-25
Add load tester

PlasmaPower:
This PR adds a test for very high transaction load, written in Rust.

I'm also going to open a couple of issues for problems identified with this. Edit: #537 #538


PlasmaPower:
For now, I added `apt-get install cargo` to the `build-travis.sh`. Once this PR is ready to be merged, we can add that to the Docker image.

PR #539, opened 2018-01-25
remove unnecessary ledger.representative() call

zhaoguojie2010:


zhaoguojie2010:
Why CI failed?

rai/secure.cpp
```diff
@@ -1990,7 +1990,7 @@ class rollback_visitor : public rai::block_visitor
 		rai::account_info info;
 		auto error (ledger.store.account_get (transaction, destination_account, info));
 		assert (!error);
-		ledger.store.representation_add (transaction, ledger.representative (transaction, hash), 0 - amount);
+		ledger.store.representation_add (transaction, representative, 0 - amount);
```
cryptocode: This change basically implies that

`ledger.representative (transaction, block_a.hashables.previous)`

is the same as

`ledger.representative (transaction, block_a.hash ())`

Is that right?

PlasmaPower: Yeah I don't think this change is correct.

PlasmaPower: Yeah I don't think this change is correct.

zhaoguojie2010: rep is chosen by iterating block's previous until open or change block is found, so I think the change is correct.

zhaoguojie2010:
Another change: There's no need to create a election for every recently arrived block.  A election should be created only when a fork a detected, so this should be removed.

zhaoguojie2010:
By the way, how can I create one PR for each commit?

PlasmaPower:
Both of these changes are incorrect. In addition, there are merge conflicts.

PR #541, opened 2018-01-25
Remove `apt-get install cargo` in build-travis.sh

PlasmaPower:
Once @lukealonso has updated the docker image, this will no longer be necessary.


PlasmaPower:
Rolled into #582

PR #542, opened 2018-01-25
Remove QTimers

CathalT:
As the Qt client is being deprecated I only made the minimum effort to fix the QTimers issue.
This should compile on all versions of Qt since before my original commit.


CathalT:
@androm3da should get your trusty builds working again if you still want to use the Qt client

androm3da:
Instead of reverting, could you instead put the old code inside `#if QT_VERSION < QT_VERSION_CHECK(5, 3, 0)` and the new code in the `#else`?

CathalT:
@androm3da sorry the revert was misleading, it still fixes the original issue, I'm just using:
```
this->wallet.application.postEvent (&this->wallet.processor, new eventloop_event ([this]() {
```
to post on to the GUI thread.

I don't see the point in doing version checks and adding back QTimers if there are new desktop clients on the way.

androm3da:
Can you fix the formatting?  Should be a super small whitespace change.

```

diff --git a/rai/qt/qt.cpp b/rai/qt/qt.cpp
index 87c15ea..3b68371 100644
--- a/rai/qt/qt.cpp
+++ b/rai/qt/qt.cpp
@@ -1282,7 +1282,7 @@ void rai_qt::wallet::push_main_stack (QWidget * widget_a)
 	main_stack->setCurrentIndex (main_stack->count () - 1);
 }
 
-void rai_qt::wallet::pop_main_stack()
+void rai_qt::wallet::pop_main_stack ()
 {
 	main_stack->removeWidget (main_stack->currentWidget ());
 }

```

CathalT:
Uh oh, I wonder is it because I ran clang-format from windows? Different whitespaces, uh I'll see if I can fix it

CathalT:
@androm3da  is that what you were referring to?

androm3da:
Sorry I should've been more clear.  The fix should have been to add a space in `pop_main_stack ()` before the parens.

CathalT:
@androm3da added the space, closed the PR I didn't think it was needed anymore, feel free to reopen and commit

PR #543, opened 2018-01-25
Fixing requeue_pull

SergiySW:
https://github.com/clemahieu/raiblocks/issues/524


cryptocode:
That improved things.

```
else if (pull.attempts == 4)
{
    ....
    if (auto connection_shared = connection_frontier_request.lock ())
```

99% of the time, `connection_frontier_request.lock ()` is nullptr. Is that expected?

Possibly lifetime issues related to the below?

```
		// The bulk_pull_client destructor attempt to requeue_pull which can cause a deadlock if this is the last reference
		// Dispatch request in an external thread in case it needs to be destroyed
```

SergiySW:
Unfortunate

PR #544, opened 2018-01-25
Load-tester: use env_logger

PlasmaPower:
We don't use `log`, but our dependencies do, so this allows us to view their logs (e.g. `RUST_LOG=info ...`).

Also sorted dependencies alphabetically.


PlasmaPower:
Looks like the Travis error is spurious.

PR #545, opened 2018-01-25
Allow skipping building the GUI

elliottneilclark:
Not all linux machines will want qt to be installed, so add a way to skip
building the GUI in ci/build.sh

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -40,6 +40,12 @@ else
     SANITIZERS=""
 fi
 
+if [[ ${SKIP_GUI} -eq 1 ]]; then
+    GUI=""
+else
+    GUI="-DRAIBLOCKS_GUI=ON"
+fi
+
 if [[ "${BOOST_ROOT}" -ne "" ]]; then
     BOOST_CFG="-DBOOST_ROOT='${BOOST_ROOT}'"
 else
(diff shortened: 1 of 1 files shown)


androm3da:
Why would you want to skip the GUI?

elliottneilclark:
One of my dev boxes is Windows 10 with the linux subsystem. It doesn't play well with X11 without jumping through some hoops.

androm3da:
The `ci/build.sh` is designed for the continuous integration build, though.

elliottneilclark:
Yeah that's fair, however I've been getting link errors with cmake(using plain makefiles rather than ninja) so using the ci/build.sh has been a nice way to get a build working without diving too deep.

cryptocode:
@eliaswinnard your get Qt link errors with cmake even after flipping off the GUI flag and regenerating the makefile?

elliottneilclark:
Nope, I get link errors with boost when using cmake + Makefiles, so instead of learning ninja I've been building with the ci/build.sh. Since I don't have x11 or qt on some of my dev boxes adding a flag to disable gui made that possible.

CMake + Makefiles seems to link with -lboost_log and no link directories for /usr/local/boost, so any box with ubuntu's 1.58 boost fails linking rai_node, because the headers are found from the 1.66 in /usr/local/boost/include while the .a's are from ubuntu's.

PR #550, opened 2018-01-27
Tool to facilitate randomness testing

cryptocode:
This PR proposes a tool to help test the strength of AutoSeededRandomPool on various OS's. The entropy source is obviously critical for generation of seeds, wallet keys, work attempts, pwd fan-out, etc.

Using a randomness test suite may catch bugs in OS's and libraries, or usage thereof, as well as malicious commits.

This tool basically makes it easy to feed output from the pool into randomness test suites, such as [dieharder](https://webhome.phy.duke.edu/~rgb/General/dieharder.php) (available at least for macOS and Linux). In other words, it behaves exactly like `cat /dev/random`, but with AutoSeededRandomPool.

Example:

```
brew install dieharder
. /random_test | dieharder -a -g 200

   rng_name    |rands/second|   Seed   |
stdin_input_raw|  7.24e+06  |3948269495|
#=============================================================================#
        test_name   |ntup| tsamples |psamples|  p-value |Assessment
#=============================================================================#
   diehard_birthdays|   0|       100|     100|0.55550323|  PASSED
      diehard_operm5|   0|   1000000|     100|0.35161023|  PASSED
  diehard_rank_32x32|   0|     40000|     100|0.76240519|  PASSED
    diehard_rank_6x8|   0|    100000|     100|0.62220366|  PASSED
   diehard_bitstream|   0|   2097152|     100|0.67553992|  PASSED
        diehard_opso|   0|   2097152|     100|0.64977517|  PASSED
        diehard_oqso|   0|   2097152|     100|0.12354025|  PASSED
 ... a lot more tests here ...
```

Running the entire suite takes a long time (for good reasons), but it's possible to pick just a few strong tests.

CI integration would be great IMO. dieharder can format the output for CSV parsing. Maybe a cron job that runs the entire suite periodically, instead of every commit?

As a test, I introduced non-randomness into the stream in various ways and dieharder detected it every time.


cryptocode:
Anyone able to spot why it builds only on macOS? 

Linker doesn't find `ed25519_hash_init` et. al. despite linking against ed25519 and rai_lib_static.

androm3da:
`librai_lib_static.a` defines `ed25519_hash_init` and `libed25519.a` has undefined references to it.  I suspect maybe there's a cycle here, but I don't see it yet.  The libs are listed twice which I think is there in order to avoid problems created by cyclic references.

Just for giggles, try enclosing the libraries with `--start-group` and `--end-group`.  If it works we can dig deeper on the root cause.

cryptocode:
@androm3da good call, grouping on Linux worked, on macOS it works even without it. Any way to avoid these circular dep's? 

And I wonder why this problem doesn't kick in for `slow_test` and `core_test` which links the exact same libs in the exact same order.

androm3da:
I think the macOS linker has that group around the link items implicitly.

It's not clear to me why this shows up for this new executable but I guess it must be something different about the references (in the new source files) themselves.

androm3da:
Hmm maybe we need to look at d62f5279553fab61e0fc72d1e086c266621a7b86

PlasmaPower:
Has someone tested building this on Windows?

cryptocode:
@PlasmaPower there's a weird linker ordering issue on Linux that crops up with this PR. androm3da had some idea, but for now there's an ugly cmake workaround.

cryptocode:
Replacement PR under way

PR #551, opened 2018-01-27
Reverse validate_work return value. Fixes #549

### LINKED ISSUE #549: work.cpp "work_validate" inverse success return value (updated)
https://github.com/clemahieu/raiblocks/blob/master/rai/lib/work.cpp#L8 
~~Should return true if the work provided is valid right?~~

Returns false if work is valid...

~~RPC "work_validate" returns 0 instead of 1 , https://github.com/clemahieu/raiblocks/blob/master/rai/node/rpc.cpp#L4022~~

~~Docs specify "1" but don't elaborate:
https://github.com/clemahieu/raiblocks/wiki/RPC-protocol#work-validate~~

Edit: Ok so the confusion lies with the work_validate function returning "true" for failure and "false" for success.

CathalT:


cryptocode:
If you inverted the logic, you need to update all the tests as well.

Note that returning true on error is a common idiom in the code base.

lukealonso:
The vast majority of the code uses true for failure and false for success, this will just make it inconsistent.

CathalT:
@cryptocode its quite confusing alright, not that 0 signifies success , that's fair enough.
Its the inconsistency. Especially using bools. Bool is true or false, if you want an error code use an int.

My bad I don't seem to have core test in my solution, oops

```
response_l.put ("moved", error ? "0" : "1"); // error = true , return 0
entry.put ("pending", exists ? "1" : "0"); // exists = true, return 1
response_l.put ("changed", error ? "0" : "1");  // error = true , return 0
response_l.put ("valid", error ? "0" : "1");  // error = true , return 0
response_l.put ("valid", valid ? "1" : "0");  // valid = true , return 1
response_l.put ("locked", valid ? "0" : "1"); // valid = false , return 0
response_l.put ("exists", exists ? "1" : "0"); //exists = true , return 1
response_l.put ("valid", error ? "0" : "1"); // error = true , return 0
response_l.put ("exists", exists ? "1" : "0"); // exists = true, return 1
response_l.put ("valid", valid ? "1" : "0"); // valid = true , return 1
response_l.put ("valid", validate ? "0" : "1"); // validate = true , return 0 // Inconsistent with the rest
```

PlasmaPower:
In this code base, errors are always `bool`s.

CathalT:
Fair enough, I was just going by what I saw in rpc.cpp
If that's the pattern there's no point me changing it

PR #552, opened 2018-01-27
Scale bootstrap connections based on load.

lukealonso:
This patch adds a new config setting "bootstrap_max_connections", which solves two problems:

a) There's no need to spin up 64 connections when doing the usual nearly-synced 30-50 frontier bootstraps the node does all the time.
a) "bootstrap_connections" can remain at 4, or whatever the user previously had set. They will benefit from the new default value of "bootstrap_max_connections" regardless of whether they change their config.

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -5,6 +5,12 @@
 
 #include <boost/log/trivial.hpp>
 
+constexpr double bootstrap_connection_scale_target = 50000.0;
+constexpr double bootstrap_connection_warmup_time = 5.0;
+constexpr double bootstrap_minimum_block_rate = 10.0;
+constexpr double bootstrap_minimum_termination_time = 30.0;
+constexpr unsigned bootstrap_max_new_connections = 10;
+
 rai::block_synchronization::block_synchronization (boost::log::sources::logger_mt & log_a) :
 log (log_a)
 {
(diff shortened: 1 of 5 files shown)


lukealonso:
@PlasmaPower There's no way to differentiate someone choosing "4" explicitly, and that just being there because it was written as the default. Ideally the initial config would be empty, and then anything in there is user-specified.

argakiig:
@PlasmaPower I was more referring to setting a default vaule if one was not found in the config

rai/node/bootstrap.cpp
```diff
@@ -5,9 +5,14 @@
 
 #include <boost/log/trivial.hpp>
 
-// Updated 1-27-18
+constexpr double bootstrap_connection_scale_target = 50000.0;
```
androm3da: Can you change these to specify the units in the identifier?

e.g. `bootstrap_minimum_block_rate_blocks_per_sec` or `bootstrap_minimum_block_rate_bps`, `bootstrap_minimum_termination_time_sec`, etc.

rai/node/bootstrap.cpp
```diff
@@ -971,39 +976,55 @@ struct block_rate_cmp
...
+	size_t num_pulls = 0;
 	std::priority_queue<std::shared_ptr<rai::bootstrap_client>, std::vector<std::shared_ptr<rai::bootstrap_client>>, block_rate_cmp> sorted_connections;
 	{
 		std::unique_lock<std::mutex> lock (mutex);
+		num_pulls = pulls.size ();
 		for (auto & c : clients)
 		{
 			if (auto client = c.lock ())
 			{
 				double elapsed = client->elapsed_seconds ();
```
androm3da: ...and then a corresponding change here, `elapsed_sec`.

PR #553, opened 2018-01-27
Avoid bootstrapping from stuck nodes.

lukealonso:
This patch introduces a hardcoded checkpoint on the number of frontiers a node should know about before we pull based on the frontiers it provides. This number should monotonically increase with node releases.

There are many stuck nodes with 8000-2M checked blocks in the network, and it's both a waste of time and a potential source of unresolvable forks to pull from them.

This is the smallest of 4 changes to get 100% reliable bootstrapping from zero in ~30 minutes.

Note: This isn't meant to and will not help with any kind of bootstrap poisoning attack, it simply helps with a real problem on the live network.

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -5,6 +5,9 @@
 
 #include <boost/log/trivial.hpp>
 
+// Updated 1-27-18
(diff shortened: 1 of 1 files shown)


PlasmaPower:
Should this number be configurable?

lukealonso:
@PlasmaPower Personally, I don't think there would be much benefit to that, but I'd like to hear if anyone has thoughts to the contrary.

PR #554, opened 2018-01-27
Force resolution of forks during bootstrap.

lukealonso:
This patch adds unresolved fork tracking to the bootstrapper. It will continue to try to resolve the forked blocks until it succeeds, or the bootstrap finishes (likely stalled if there are any unresolved forks).

A single unresolved fork can mean the bootstrap will stall with a few million unchecked blocks.

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -5,12 +5,14 @@
 
 #include <boost/log/trivial.hpp>
 
-constexpr double bootstrap_connection_scale_target = 50000.0;
-constexpr double bootstrap_connection_warmup_time = 5.0;
-constexpr double bootstrap_minimum_block_rate = 10.0;
-constexpr double bootstrap_minimum_termination_time = 30.0;
+constexpr double bootstrap_connection_scale_target_blocks = 50000.0;
+constexpr double bootstrap_connection_warmup_time_sec = 5.0;
+constexpr double bootstrap_minimum_blocks_per_sec = 10.0;
+constexpr double bootstrap_minimum_frontier_blocks_per_sec = 1000.0;
+constexpr unsigned bootstrap_frontier_retry_limit = 16;
+constexpr double bootstrap_minimum_termination_time_sec = 30.0;
 constexpr unsigned bootstrap_max_new_connections = 10;
-constexpr unsigned bootstrap_peer_frontier_minimum = rai::rai_network == rai::rai_networks::rai_live_network ? 339000 : 0;
+constexpr unsigned bootstrap_peer_frontier_minimum_blocks = rai::rai_network == rai::rai_networks::rai_live_network ? 339000 : 0;
 
 rai::block_synchronization::block_synchronization (boost::log::sources::logger_mt & log_a) :
 log (log_a)
@@ -346,6 +348,15 @@ void rai::frontier_req_client::received_frontier (boost::system::error_code cons
 			start_time = std::chrono::steady_clock::now ();
 		}
 		++count;
+		std::chrono::duration<double> time_span = std::chrono::duration_cast<std::chrono::duration<double>> (std::chrono::steady_clock::now () - start_time);
+		double elapsed_sec = time_span.count ();
+		double blocks_per_sec = (double)count / elapsed_sec;
+		if (elapsed_sec > bootstrap_connection_warmup_time_sec && blocks_per_sec < bootstrap_minimum_frontier_blocks_per_sec)
+		{
+			BOOST_LOG (connection->node->log) << boost::str (boost::format ("Aborting frontier req because it was too slow"));
+			promise.set_value (true);
+			return;
+		}
 		auto now (std::chrono::steady_clock::now ());
 		if (next_report < now)
 		{
@@ -419,7 +430,7 @@ void rai::frontier_req_client::received_frontier (boost::system::error_code cons
 			{
 				try
 				{
-					promise.set_value (count < bootstrap_peer_frontier_minimum);
+					promise.set_value (count < bootstrap_peer_frontier_minimum_blocks);
 				}
 				catch (std::future_error &)
 				{
@@ -608,7 +619,6 @@ void rai::bulk_pull_client::received_block (boost::system::error_code const & ec
 			{
 				expected = block->previous ();
 			}
-			connection->attempt->node->block_processor.add (rai::block_processor_item (block));
 			if (connection->block_count++ == 0)
 			{
 				connection->start_time = std::chrono::steady_clock::now ();
@@ -867,19 +877,21 @@ bool rai::bootstrap_attempt::still_pulling ()
 	auto running (!stopped);
 	auto more_pulls (!pulls.empty ());
 	auto still_pulling (pulling > 0);
-	return running && (more_pulls || still_pulling);
+	auto more_forks (!unresolved_forks.empty ());
+	return running && (more_pulls || still_pulling || more_forks);
 }
 
 void rai::bootstrap_attempt::run ()
 {
 	populate_connections ();
+	resolve_forks ();
 	std::unique_lock<std::mutex> lock (mutex);
 	auto frontier_failure (true);
 	while (!stopped && frontier_failure)
 	{
 		frontier_failure = request_frontier (lock);
 	}
-	// Shuffle frontiers.
+	// Shuffle pulls.
 	for (int i = pulls.size () - 1; i > 0; i--)
 	{
 		auto k = rai::random_pool.GenerateWord32 (0, i);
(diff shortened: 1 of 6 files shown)


clemahieu:
Cool, I'll take a look at it too, sounds like a good fix.

PlasmaPower:
There are a couple of tricks to "read" an external file at compile time. I think it would be worthwhile to use one of these, so we could generate the representatives weights from a script easier.

Alternatively, we could move the weights into a separate minimal C++ file, and generate that from a script.

slact:
The baked-in stake distribution also needs to have some kind of expiration timeframe. This could mos tobviously be a timestamp, or it could be applied only to frontiers within some percentage of the frontier size at time of stake distribution snapshot.

Otherwise, an attacker could make use of an old baked-in stake distribution to bootstrap an old client into a non-consensus, state.

lukealonso:
@PlasmaPower Agree, it could be done by cmake pretty easily.

clemahieu:
I think solving the forks should be done separately than baked-in rep weight. I agree the rep-weight should be external and configurable.  I think part of the issue of rep weight being calculated before the ledger is updated can be fixed by replacing peer_information::rep_weight with peer_information::rep_account then in peer_container::representatives instead of checking "rep_weight.is_zero ()" check "ledger.weight(rep_account).is_zero ()"

PlasmaPower:
I think you might be missing the point. Forks have to be judged by the "current" representative stake, not the stake at the time of the fork. For a node bootstrapping, that means it must look into the future.

clemahieu:
Understood, what do you think about this information being in code versus in a configuration file?

androm3da:
I think it would be ideal if we didn't need this but this does seem like the most expedient way to resolve the problem for now.  But perhaps it still makes sense to follow up with a protocol change to avoid this problem.  It seems like it becomes a new DoS attack vector.

If we hope/assume the hosts with poisoned databases get purged, and find a more robust resolution, then it might make sense to delete this feature later.

I don't have a preference about code vs in config file, especially if this feature is temporary.  Even if it takes two-three-four software releases before we fix it, I would still refer to this fix as "temporary."

PlasmaPower:
I do think we need a copy in the code, so users can just run the program without needing to worry about configuring it. I also think it should be overridable by the config though.

@androm3da I think this has to be a permanent solution. We either have to trust something hard coded like this, or trust something like a majority of nodes. I think the former is the much safer option.

SergiySW:
Seems like centralized solution to me.
Can you provide sample of data.ldb where forks cannot be resolved with official representatives & current representatives set? In any past configurations they had majority.

androm3da:
Foundational trust items like genesis should come from the software.  Absent some other feature, this makes sense to "centralize."

However I believe it's possible to solve this issue long term without adding these items.  What about tracking historical representation?  @lukealonso if that's not an adequate solution for mitigating future occurrences of this problem, would you be willing to go into (copious) detail about the problem?  Maybe write a brief article regarding how you discovered it and the evidence you used (dumps and graph analysis tool)?  

I appreciate that you've already spent a good amount of time on this but I think this would help the whole team understand better.

clemahieu:
In the back of my head I always felt periodic checkpoints would be a solution to this for a couple reasons:
1) I feel like checkpoints are a generalization of the genesis block.  The concept is there could have been transactions before the genesis block that were "pruned" and we just chose the genesis block as our checkpoint.
2) Representatives are supposed to be long-term operations so if we do checkpoints mostly regularly, perhaps monthly, the drift in active representatives between checkpoints would be low.

lukealonso:
@clemahieu I tried the rep_weight.is_zero () thing, it didn't solve the problem. This combination is the only thing that has helped reliably so far.

lukealonso:
@SergiySW It's certainly less centralized than distributing the entire ledger to people.

To reproduce the issue, just try to bootstrap a few times from zero, you'll get stuck with a few million unchecked blocks after the first round of account pulls most of the time. This is always due to improperly resolved forks.

Sometimes it will eventually resolve itself after many more pulls of the same blocks (wasting a ton of bandwidth), if you have enough checked blocks to make some sense of incoming votes. Sometimes it will get stuck permanently. With this change and the frontier min count, you can get to fully synced in one pull.

PlasmaPower:
Here's an example of why we need this. Consider the following blockchain. The first block is the genesis block (Block A), and the representative field is set to Rep A. Next is a fork, with these branches:

- Block B: a change block, setting the representative to Rep B
- Block C: a change block, setting the representative to Rep C

Rep A votes for Block B, Rep B votes for Block B, and Rep C votes for Block C.

Initially, you might think that Block B must be the winner, because Rep A voted for it. However, in the past, Rep B may have voted for Block C. That would mean Rep B gave up its power, and even if it's voting for Block B now, it wouldn't mean anything as Rep C has the stake.

Even if we attempt to keep records of votes, there's no trustless way to verify a vote's timestamp (or even relative ordering of votes from different representatives).

There's no trustless way to bootstrap this situation. This example is purely hypothetical, however, similar things have happened.

lukealonso:
I modified a patch kindly provided by @PlasmaPower to bake the rep weights into the executable in a more reproducible way.

clemahieu:
I'm seeing a lot of these in my log while running this patch:

[2018-01-30 04:15:27.616954]: While bootstrappping, fork between our block: FDD76F7563EB416AC411C35E391CF48165D69C865ABFC9A8D59D8259C114E823 and block BE35CFA899AD90A9D7B69BB0FE1D7695E2797F6EB25402211CE00BB4EA373F4E both with root A0BE2E63929285BC1AEAE3A44EAC83734469556563668C9A25B9364B1D2C7381 3
[2018-01-30 04:15:27.626849]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:27.640691]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.232175]: 4 unresolved forks while bootstrapping
[2018-01-30 04:15:32.245684]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.269781]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.288917]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.333724]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.344539]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.355382]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.373496]: Broadcasted confirm req to 194 representatives
[2018-01-30 04:15:32.436554]: Broadcasted confirm req to 194 representatives

Is it attempting to resolve every fork including those received off the network?  This seems a possible DoS situation where someone could spam a node with fork blocks and get it to attempt to confirm always which sends a lot of packets.

It seems like the situation would resolve if we only attempt to resolve forks if it did not recently arrive off the network via UDP, this means it's likely it was solicited by bootstrap and is the only time we want to attempt to resolve a fork.

lukealonso:
@clemahieu Makes sense - I'll split the patches and add the check for block_arrival.recent() so this doesn't happen for blocks received over UDP.

lukealonso:
@clemahieu This PR is now just the retry logic for forks. I removed the part of the patch where it'll wait for the forks to be resolved before considering the bootstrap complete, since that will hang indefinitely without the initial rep weight patch to help resolve the forks.

rai/node/bootstrap.cpp
```diff
@@ -1115,21 +1191,8 @@ void rai::bootstrap_attempt::stop ()
 
 void rai::bootstrap_attempt::add_pull (rai::pull_info const & pull)
 {
-	static rai::account landing ("059F68AAB29DE0D3A27443625C7EA9CDDB6517A8B76FE37727EF6A4D76832AD5");
```
lukealonso: This doesn't help much at all, I tried doubling down and prioritizing a larger set of accounts to try to have fewer unchecked blocks, but it's not worth it.

rai/node/bootstrap.cpp
```diff
@@ -8,6 +8,7 @@
 constexpr double bootstrap_connection_scale_target = 50000.0;
 constexpr double bootstrap_connection_warmup_time = 5.0;
 constexpr double bootstrap_minimum_block_rate = 10.0;
+constexpr double bootstrap_minimum_frontier_rate = 1000.0;
```
androm3da: Please change these identifiers to include the units for the quantities.

e.g.

`bootstrap_minimum_frontier_rate` -> `bootstrap_minimum_frontier_rate_blocks_per_sec`
`bootstrap_connection_warmup_time` -> `bootstrap_connection_warmup_time_sec`

rai/node/bootstrap.cpp
```diff
@@ -346,6 +347,15 @@ void rai::frontier_req_client::received_frontier (boost::system::error_code cons
 			start_time = std::chrono::steady_clock::now ();
 		}
 		++count;
+		std::chrono::duration<double> time_span = std::chrono::duration_cast<std::chrono::duration<double>> (std::chrono::steady_clock::now () - start_time);
+		double elapsed = time_span.count ();
```
androm3da: `elapsed` -> `elapsed_sec`

rai/node/bootstrap.cpp
```diff
@@ -346,6 +347,15 @@ void rai::frontier_req_client::received_frontier (boost::system::error_code cons
...
+		double rate = (double)count / elapsed;
```
androm3da: `rate` -> `rate_blocks_per_sec`

rai/node/bootstrap.cpp
```diff
@@ -950,12 +960,78 @@ bool rai::bootstrap_attempt::consume_future (std::future<bool> & future_a)
...
+	if (unresolved_forks.size () > 0)
+	{
+		BOOST_LOG (node->log) << boost::str (boost::format ("%1% unresolved forks while bootstrapping") % unresolved_forks.size ());
+		rai::transaction transaction (node->store.environment, nullptr, false);
+		for (auto & it : unresolved_forks)
+		{
+			try_resolve_fork (transaction, it.second, false);
+		}
+	}
+	if (!stopped)
```
androm3da: Does it make more sense for `stopped` to guard the entire function?

rai/node/bootstrap.cpp
```diff
@@ -950,12 +960,78 @@ bool rai::bootstrap_attempt::consume_future (std::future<bool> & future_a)
...
+		}
+	}
+}
+
+void rai::bootstrap_attempt::resolve_forks ()
+{
+	std::unique_lock<std::mutex> lock (mutex);
+	if (unresolved_forks.size () > 0)
+	{
+		BOOST_LOG (node->log) << boost::str (boost::format ("%1% unresolved forks while bootstrapping") % unresolved_forks.size ());
```
androm3da: Does `BOOST_LOG` ever result in I/O?  Say, when buffers are exhausted or special characters are present in the output?

rai/node/bootstrap.cpp
```diff
@@ -950,12 +960,78 @@ bool rai::bootstrap_attempt::consume_future (std::future<bool> & future_a)
 
 void rai::bootstrap_attempt::process_fork (MDB_txn * transaction_a, std::shared_ptr<rai::block> block_a)
 {
+	std::unique_lock<std::mutex> lock (mutex);
+	try_resolve_fork (transaction_a, block_a, true);
+}
+
+void rai::bootstrap_attempt::try_resolve_fork (MDB_txn * transaction_a, std::shared_ptr<rai::block> block_a, bool from_processor)
+{
+	assert (!mutex.try_lock ());
```
androm3da: These locks are recursive, why not just add another one?  Or for that matter just move it from `process_fork()` into here?

rai/node/bootstrap.cpp
```diff
@@ -950,12 +960,78 @@ bool rai::bootstrap_attempt::consume_future (std::future<bool> & future_a)
...
+			if (auto this_l = this_w.lock ())
+			{
+				if (resolved)
+				{
+					{
+						std::unique_lock<std::mutex> lock (this_l->mutex);
+						this_l->unresolved_forks.erase (block_a->hash ());
+						this_l->condition.notify_all ();
+					}
+					rai::transaction transaction (this_l->node->store.environment, nullptr, false);
```
androm3da: IIUC this is almost certain to trigger I/O.

rai/node/bootstrap.cpp
lukealonso: Sure, makes sense.

rai/node/bootstrap.cpp
lukealonso: Done

rai/node/bootstrap.cpp
lukealonso: Done

rai/node/bootstrap.cpp
lukealonso: std::mutex isn't recursive

androm3da: Yes, you're right, I misunderstood TFM.

rai/node/bootstrap.cpp
androm3da: Ok, Luke showed me I was missing the fact that this block is all deferred work in a lambda.

rai/node/bootstrap.cpp
lukealonso: I reduced the scope of the mutex.

rai/node/bootstrap.cpp
lukealonso: I reduced the scope of the mutex anyways, it wasn't ideal.

androm3da:
LGTM

PR #558, opened 2018-01-28
Rename result to more descriptive names

cryptocode:
A very boring renaming patch. People seem to trip up on false-is-success, etc, so more descriptive names may help. @androm3da we talked about this one. I picked *error* instead of *failure*, since *error* is already used in quite a few places.

I'll do another round for the node subdir later on.

### DIFF
--- a/rai/lib/blocks.cpp
+++ b/rai/lib/blocks.cpp
@@ -10,11 +10,11 @@ std::string rai::to_string_hex (uint64_t value_a)
 
 bool rai::from_string_hex (std::string const & value_a, uint64_t & target_a)
 {
-	auto result (value_a.empty ());
-	if (!result)
+	auto error (value_a.empty ());
+	if (!error)
 	{
-		result = value_a.size () > 16;
-		if (!result)
+		error = value_a.size () > 16;
+		if (!error)
 		{
 			std::stringstream stream (value_a);
 			stream << std::hex << std::noshowbase;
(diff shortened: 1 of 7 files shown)


rai/lib/blocks.cpp
```diff
@@ -196,29 +196,29 @@ bool rai::send_block::deserialize_json (boost::property_tree::ptree const & tree
...
+						error = signature.decode_hex (signature_l);
 					}
 				}
 			}
 		}
 	}
 	catch (std::runtime_error const &)
 	{
-		result = true;
+		error = true;
```
androm3da: What exception are we eating here?  failures from `tree.get<>()`?  Besides "the node at that path was not present" what other failure modes get thrown out here?  It would be great if we could record these in a log.  Or at least leave a comment describing what we expect to catch here.

Thanks for cleaning this up, but no good deed goes unpunished. ;)

rai/lib/interface.cpp
```diff
@@ -37,24 +37,21 @@ int xrb_uint256_from_string (const char * source, xrb_uint256 destination)
...
+	return error ? 1 : 0;
 }
 
 int xrb_uint512_from_string (const char * source, xrb_uint512 destination)
 {
 	auto & number (*reinterpret_cast<rai::uint512_union *> (destination));
 	auto error (number.decode_hex (source));
-	auto result (error ? 1 : 0);
-	return result;
+	return error ? 1 : 0;
```
androm3da: Do these match the polarity of the stdlib `EXIT_SUCCESS` / `EXIT_FAILURE`  defines ?  Can we use those?

rai/lib/blocks.cpp
cryptocode: Well, you know how I feel about exceptions and eating them ;) In my opinion, making sure errors are logged and not eaten is a separate patch, this is just renaming. 

I think there's a lot of places that happens. There's also the issue of boost logger not being available everywhere (for instance, some io callbacks which now fail silently)

rai/lib/interface.cpp
cryptocode: Those are usually only used for std::exit. 

If we were to use them, we need to be careful since EXIT_SUCCESS is not required by the standard to be 0. It basically excludes the `if (!something())` idiom.

PR #559, opened 2018-01-28
Removing librai_lib.so copy from Dockerfile

SergiySW:
No longer required for default build


PlasmaPower:
Other Dockerfile changes have included this already.

PR #564, opened 2018-01-30
Fix config upgrade path for bootstrap_connections_max.

lukealonso:


PlasmaPower:
While we're on the topic, I'm pretty sure the case-switch shouldn't have all the `break`s.

androm3da:
@PlasmaPower you think they should fallthrough?

androm3da:
Yeah, I see it, that makes sense.

PR #572, opened 2018-02-01
Fix OpenCL printstate

PlasmaPower:
The blake2b state buffer is 128 bytes long, not 256.

### DIFF
--- a/rai/node/openclwork.cpp
+++ b/rai/node/openclwork.cpp
@@ -382,12 +382,23 @@ __kernel void raiblocks_work (__global ulong * attempt, __global ulong * result_
 
 void printstate (blake2b_state * S)
 {
-	printf ("%lu %lu %lu %lu %lu %lu %lu %lu %lu %lu %lu %lu\n", S->h[0], S->h[1], S->h[2], S->h[3], S->h[4], S->h[5], S->h[6], S->h[7], S->t[0], S->t[1], S->f[0], S->f[1]);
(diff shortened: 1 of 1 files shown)


rai/node/openclwork.cpp
```diff
@@ -383,7 +383,7 @@ __kernel void raiblocks_work (__global ulong * attempt, __global ulong * result_
 void printstate (blake2b_state * S)
 {
 	printf ("%lu %lu %lu %lu %lu %lu %lu %lu %lu %lu %lu %lu\n", S->h[0], S->h[1], S->h[2], S->h[3], S->h[4], S->h[5], S->h[6], S->h[7], S->t[0], S->t[1], S->f[0], S->f[1]);
```
cryptocode: I know this is just a debug dump, but... `%lu` isn't exactly portable, maybe use `PRIu64` or replace the whole thing with `std::cout` ?

rai/node/openclwork.cpp
```diff
@@ -383,7 +383,7 @@ __kernel void raiblocks_work (__global ulong * attempt, __global ulong * result_
...
-	for (int i = 0; i < 256; ++i)
+	for (int i = 0; i < 128; ++i)
```
androm3da: It seems like it would be more robust to use `sizeof(S->buf)` right ?

PlasmaPower:
Updated. I was initially just fixing an annoying compiler warning 😄.

Issue #458, opened 2018-01-09
Wallet_create on the command line doesn't work with a running daemon


clemahieu:
The database entry is created but the items map in memory isn't updated when the file changes.

The items map in memory could be removed and just always check the database for wallet IDs.

brightmonkey:
I was going to take a stab at fixing this, but as far as i can tell wallet_create seems to work. What are the steps I need to do to reproduce the symptoms you see?

clemahieu:
I think the issue is if the node is running "rai_node --daemon" and someone runs "rai_node --wallet_create" on the command line, they expect that to be effective in the running node but it's not because the map in rai::wallets isn't updated with the new wallet id.

brightmonkey:
Isn't that the equivalent of running two instances of the node? If so, how would node B influence/change the properties of node A running the daemon? Furthermore, if someone wants to add a wallet to node A, why wouldn't they use the RPC command?

brightmonkey:
Never mind, I think I answered my own question. Looks like the expectation **is** that node B updates the database and node A is unaware of the change. I'll look into removing the items map and checking the database for wallet IDs instead.

brightmonkey:
@clemahieu the majority of command line options don't respect the --data_path flag. Was this a specific design choice? If not I'd like to create and submit a change that allows commands like ./rai_node --wallet_list to use the path defined by --data_path if that's ok with you. The rationale for this is to keep my debug environment entirely separate from the default installation path used by my personal desktop wallet.

cryptocode:
@brightmonkey there's a PR fixing that: https://github.com/clemahieu/raiblocks/pull/342

PlasmaPower:
There was a PR to fix that, but it got abandoned.

brightmonkey:
@cryptocode thanks! 
@PlasmaPower, are you talking about #342 or a different PR?

PlasmaPower:
Same PR.

brightmonkey:
Looks like it's still open, although it needs a rebase. @SergiySW, is #342 still active?

SergiySW:
Updated for recent version https://github.com/clemahieu/raiblocks/pull/586

brightmonkey:
@clemahieu there are ~180 places in the code that reference the wallets.items map. Were you thinking of replacing all of them with direct db reads? I don't know what impact removing the map entirely would have on performance, but I would think accessing a map in memory would always be faster than file IO access to a database on disk. 

[...]

PR #574, opened 2018-02-02
RPC support for listing all wallets in a node

brightmonkey:
In the process of investigating issue #458, I discovered there isn't a simple way to see what wallets a node actually contains. 

So, I added an RPC command named "wallet_list" that lists all wallets controlled by the node. The output looks like this:

mariog$ curl -g -d '{ "action": "wallet_list"}' '[::1]:7076'
{
    "wallets": [
        "D4F5829253E41DCC3AAECF9DE1142AD052BBCCD9F5E4F101D88A8EFE6D3850F6",
        "B04E7807B3F23EC28ECA579DF25A39925815E97078388701EBDC67DF2F1C12B7",
        "92A806356FAFEEFBCA2454770DA47D15EC1730DA1B10DE4BE7089ECA136195AF"
    ]
}

I also added a test that creates two extra wallets in addition to the default wallet and ensures all of them are listed properly. 

This new command is wrapped with the 'if(rpc.config.enable_control)' flag, since I figured we wouldn't want random Joe Cryptos to be able to query a node for the list of wallets it controls. 

Unfortunately, this commit doesn't _fix_ issue #458, but it should help me with debugging the issue.


PlasmaPower:
See https://github.com/clemahieu/raiblocks/pull/432

It causes security problems, among other things.

brightmonkey:
Well shucks. Looks like I just reinvented a wheel.

PlasmaPower:
Well, at least you got to know the RaiBlocks codebase a bit better 😄

PM me on Discord if you want any help with this issue: PlasmaPower#0068

brightmonkey:
Yes, the one benefit of reinventing a wheel is that you end up understanding the wheel very well 😂

On Feb 1, 2018, at 9:29 PM, Lee Bousfield <notifications@github.com> wrote:

Well, at least you got to know the RaiBlocks codebase a bit better 😄

PM me on Discord if you want any help with this issue: PlasmaPower#0068

brightmonkey:
Looks like the Travis build failed because I didn't run clang-format first. Oops. Regardless, I'm going to close this pull request in light of the new information.

brightmonkey:
As per comments in #432, it's clear this feature should not be in any release implementation of a node. But thinking about this some more, there's definitely value in being able to query nodes for a wallet list for debug purposes only.  @clemahieu, what are your thoughts on adding this functionality under a special debug flag?

brightmonkey:
On second thought, forget what I said about adding it under a debug flag. Regardless of how much of a barrier we add, someone somewhere will inevitably deploy it by mistake with potential bad consequences. I'll just keep this patch tucked away and apply it locally when I need it.

cryptocode:
@brightmonkey and for debug needs, you can always use `--wallet_list` on the node

brightmonkey:
@cryptocode true, but I'd have to stop the node in order to use that option. Issue #458 refers to a situation where a node is already running, which is why I needed a way to query it in real time.

cryptocode:
Yeah, that's an issue ;)

PR #576, opened 2018-02-03
process_result::overspend -> process_result::negative_spend

PlasmaPower:
If a send block has a higher balance than the previous, then it's trying to spend a negative amount, not spend more than it has.


androm3da:
So perhaps you've also uncovered a gap in our testing, then?

PlasmaPower:
No, it's not possible to overspend with the nano protocol. The send block's balance is the balance after the send, and it is unsigned, so it cannot be negative.

PR #581, opened 2018-02-03
Watch only accounts

SergiySW:
Draft watch-only accounts for large services

### DIFF
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -1724,6 +1724,12 @@ void rai::rpc_handler::ledger ()
 				error_response (response, "Invalid count limit");
 			}
 		}
+		uint64_t modified_since (0);
+		boost::optional<std::string> modified_since_text (request.get_optional<std::string> ("modified_since"));
+		if (modified_since_text.is_initialized ())
+		{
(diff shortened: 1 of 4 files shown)


rai/node/rpc.cpp
```diff
@@ -3526,6 +3580,94 @@ void rai::rpc_handler::wallet_key_valid ()
...
+void rai::rpc_handler::wallet_ledger ()
+{
+	bool representative (false);
+	boost::optional<bool> representative_optional (request.get_optional<bool> ("representative"));
+	if (representative_optional.is_initialized ())
+	{
+		representative = representative_optional.get ();
+	}
+	bool weight (false);
+	boost::optional<bool> weight_optional (request.get_optional<bool> ("weight"));
```
androm3da: This behavior (`false` if maybe/undefined, `false` if defined `false`, `true` if defined `true` ) can be simplified as 

`const bool weight = *(request.get_optional<bool> ("weight"));`

SergiySW: Ok, will change all similar parts as well

androm3da:
LGTM @clemahieu please review.

PR #588, opened 2018-02-04
Cleanup docker and deploy

PlasmaPower:


PlasmaPower:
I might've fixed the deploy problem, and even if I didn't, it'll no longer cause the build as a whole to fail. It'll also output information about the failure if that does happen.

PR #595, opened 2018-02-06
Stop processing blocks during shutdown

cryptocode:
Exiting the wallet/node can sometimes take minutes. I believe the culprit is that `block_processor::process_receive_many` does not honor the stopped flag (and there can be a large amount of blocks queued up according to my debug dumps)

I've seen no hangs during shutdown so far with this patch. 

Any problems with exiting the process_receive_many loops?

I made be stopped flag atomic, because the function is called without mutex protection from the election::confirm_once path.

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -1113,7 +1113,7 @@ void rai::block_processor::stop ()
 void rai::block_processor::flush ()
 {
 	std::unique_lock<std::mutex> lock (mutex);
-	while (!stopped && (!blocks.empty () || !idle))
+	while (!stopped.load () && (!blocks.empty () || !idle))
 	{
 		condition.wait (lock);
 	}
@@ -1129,7 +1129,7 @@ void rai::block_processor::add (rai::block_processor_item const & item_a)
 void rai::block_processor::process_blocks ()
 {
 	std::unique_lock<std::mutex> lock (mutex);
-	while (!stopped)
+	while (!stopped.load ())
 	{
 		if (!blocks.empty ())
 		{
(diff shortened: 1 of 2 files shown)


PlasmaPower:
At first glance, this looks good. It looks like it also fixes some balance observer problems?

cryptocode:
@PlasmaPower if it does, unintentionally ;) I though the life cycle of block_processor followed the node?

lukealonso:
@PlasmaPower the balance observer stuff looks unchanged, just a weird diff

@cryptocode With this patch you'll lose the blocks in blocks_processing, which could be particularly bad during bootstrap.

cryptocode:
@lukealonso but the stopped flag is set only during exit, no?

lukealonso:
Right, exiting while bootstrapping. When you re-open the node, it won't pick up where it left off, and will likely need to re-download the whole ledger. Not a fatal issue, just pointing it out.

cryptocode:
@lukealonso yeah, I get that. But it's not great to hang for minutes during shutdown either, and bootstrap should be uber-fast now, right? ;)

PlasmaPower:
I think the best solution is to figure out how to pick up where it left off.

lukealonso:
We could write the being-processed blocks to another table in LMDB on exit so they don't get dropped.

cryptocode:
It seems to pick up where it left, but I guess you mean it's downloading more than it needs when exiting during bootstrap?

lukealonso:
@cryptocode the extra downloads won't be visible in the UI, you would have to check the logs for a second frontier request

cryptocode:
@lukealonso @PlasmaPower keeping track is something we should do then. Going to be a fair amount of extra lmdb IO though, a problem? 

Anyone with better knowledge of the bootstrap code volunteer to implement that on top of this PR?

PlasmaPower:
We can just write the unprocessed blocks to lmdb on shutdown.

cryptocode:
Will revisit this later, more thought is needed regarding in-flight blocks.

PR #603, opened 2018-02-11
Fix ledger processor hash2 race condition

### LINKED ISSUE #598: Latest commit of rai_node seg faults in osx
**Bug**

- The latest commit 257af11d under OSX segfaults. 

**Steps**

 1. Compile the latest commit 257af11d under OSX.

 2.  Run rai_node as a daemon.

**Issue**

- Segfaults after a short time. 

**Logs**

- [log_2018-02-08_13-46-12.39.log](https://github.com/clemahieu/raiblocks/files/1707181/log_2018-02-08_13-46-12.39.log)

- [rai_node_2018-02-08-134654_MacPro.crash.txt](https://github.com/clemahieu/raiblocks/files/1707191/rai_node_2018-02-08-134654_MacPro.crash.txt)

### LINKED ISSUE #602: Segfault in latest trunk, block_store representation_add
**Description of bug:**
Segfault somewhere in the representative code.

1. Compile and run
2. Wait for sync to occur for a while
3. Observe segfault

**Environment**:

- OS information 
- (Linux) Kernel (e.g. `uname -a`): Linux nyu 4.4.0-91-generic #114-Ubuntu SMP Tue Aug 8 11:56:56 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
- Node version: latest trunk (10..0) 22c647f2e59c2d3caf7c21e05e856386f62568b7

```
Core was generated by `/opt/raiblocks/rai_node --daemon'.
Program terminated with signal SIGSEGV, Segmentation fault.
---Type <return> to continue, or q <return> to quit---
#0  0x00000000006de930 in rai::block_store::representation_add(MDB_txn*, rai::uint256_union const&, boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<128u, 128u, (boost::multiprecision::cpp_integer_type)0, (boost::multiprecision::cpp_int_check_type)0, void>, (boost::multiprecision::expression_template_option)0> const&) ()
[Current thread is 1 (Thread 0x7f44d0986700 (LWP 20104))]
(gdb) bt
#0  0x00000000006de930 in rai::block_store::representation_add(MDB_txn*, rai::uint256_union const&, boost::multiprecision::number<boost::multiprecision::backends::cpp_int_backend<128u, 128u, (boost::multiprecision::cpp_integer_type)0, (boost::multiprecision::cpp_int_check_type)0, void>, (boost::multiprecision::expression_template_option)0> const&) ()
#1  0x00000000006e42b6 in (anonymous namespace)::ledger_processor::change_block(rai::change_block const&) ()
#2  0x00000000006da479 in rai::ledger::process(MDB_txn*, rai::block const&) ()
#3  0x00000000005e65de in rai::block_processor::process_receive_one(MDB_txn*, std::shared_ptr<rai::block>) ()
#4  0x00000000005e832c in rai::block_processor::process_receive_many(std::deque<rai::block_processor_item, std::allocator<rai::block_processor_item> >&) ()
#5  0x00000000005e9bb5 in rai::block_processor::process_blocks() ()
#6  0x00000000008eac30 in execute_native_thread_routine ()
#7  0x00007f44d64d86ba in start_thread (arg=0x7f44d0986700) at pthread_create.c:333
#8  0x00007f44d5cf03dd in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

PlasmaPower:
Resolves #598, resolves #602.

A block's dependency (`previous` or `source`) was being added to the ledger after the hash2 was calculated, which caused the assert to fail. This fixes the problem.

### DIFF
--- a/rai/secure.cpp
+++ b/rai/secure.cpp
@@ -2660,7 +2660,12 @@ void ledger_processor::change_block (rai::change_block const & block_a)
 				result.code = validate_message (account, hash, block_a.signature) ? rai::process_result::bad_signature : rai::process_result::progress; // Is this block signed correctly (Malformed)
 				if (result.code == rai::process_result::progress)
 				{
-					assert (!hash2.is_zero ());
+					if (hash2.is_zero ())
+					{
+						// recalculate hash2 to prevent race condition
+						hash2 = ledger.store.hash2_calc (transaction, block_a);
+						assert (!hash2.is_zero ());
+					}
 					ledger.store.hash2_put (transaction, hash, hash2);
 					ledger.store.block_put (transaction, hash2, block_a);
 					auto balance (ledger.balance (transaction, block_a.hashables.previous));
@@ -2703,7 +2708,12 @@ void ledger_processor::send_block (rai::send_block const & block_a)
 					result.code = info.balance.number () >= block_a.hashables.balance.number () ? rai::process_result::progress : rai::process_result::negative_spend; // Is this trying to spend a negative amount (Malicious)
 					if (result.code == rai::process_result::progress)
 					{
-						assert (!hash2.is_zero ());
+						if (hash2.is_zero ())
+						{
+							// recalculate hash2 to prevent race condition
+							hash2 = ledger.store.hash2_calc (transaction, block_a);
+							assert (!hash2.is_zero ());
+						}
 						ledger.store.hash2_put (transaction, hash, hash2);
 						auto amount (info.balance.number () - block_a.hashables.balance.number ());
 						ledger.store.representation_add (transaction, info.rep_block, 0 - amount);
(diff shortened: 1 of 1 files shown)


PlasmaPower:
Note that in some cases your database is already corrupted, since it has hash -> 0 hash2 mappings. This happens because the assert only triggers in debug mode.

PlasmaPower:
Also, now that I think about it, there's still a possible race condition with elections, where a block can be removed from the ledger. Fixing that will be much harder.

androm3da:
This change might be effective but it doesn't look like it's addressing the root cause of the race.

PlasmaPower:
Yeah, that's evident in the problem with voting. I thought lmdb transactions would take care of this, but I guess not...

Basically, a block can be added to the ledger between when `ledger_processor::send_block` checks `ledger.store.block_exists (transaction, block_a.hashables.previous)`, and when `hash2_visitor::send_block` checks `!store.hash2_get (transaction, block_a.hashables.previous).is_zero()` (I condensed the expression a bit).

This fixes the problem when a block is added to the ledger between those checks, however, if voting *removes* a block from the ledger between those checks, then there could be a problem. I think that could be a problem even before hash2. Hash2 only amplified the issue because it could happen with both adding blocks and removing blocks. Because of that, I'd propose merging this PR until a more comprehensive locking system can be created (assuming this PR fixes the segfaults that is).

PR #604, opened 2018-02-11
Fix ledger_processor::change_block block_put

PlasmaPower:
Now it matches the other block types. I think this might have also been part of #598, as I got a different segfault after running it for a while.


PlasmaPower:
My wallet has now been running for 30 minutes without problems with this newest commit. Previously, it consistently crashed within 10 seconds.

PR #608, opened 2018-02-12
Logging confirm req block hash

SergiySW:


PlasmaPower:
Any specific reason for this?

SergiySW:
Tracking why sometime pending blocks cannot be auto-received by node.
And this message is useless without info what block node is trying to confirm.

PlasmaPower:
Okay. If you were running the hash2 changes before #603 and #604 were merged, your database probably has bad data, which could be causing problems.

SergiySW:
It doesn't depend on hash2

SergiySW:
in https://github.com/nanocurrency/raiblocks/pull/711

PR #610, opened 2018-02-12
Add RPC send ID for idempotency

PlasmaPower:
Probably needs some more discussion.

Note that currently, send IDs are per wallet. If we decide to keep that behavior, it should be documented.

### DIFF
--- a/rai/node/wallet.cpp
+++ b/rai/node/wallet.cpp
@@ -386,7 +386,8 @@ std::vector<rai::account> rai::wallet_store::accounts (MDB_txn * transaction_a)
 void rai::wallet_store::initialize (MDB_txn * transaction_a, bool & init_a, std::string const & path_a)
 {
 	assert (strlen (path_a.c_str ()) == path_a.size ());
-	auto error (mdb_dbi_open (transaction_a, path_a.c_str (), MDB_CREATE, &handle));
+	auto error (0);
+	error |= mdb_dbi_open (transaction_a, path_a.c_str (), MDB_CREATE, &handle);
 	init_a = error != 0;
 }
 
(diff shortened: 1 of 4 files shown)


rai/node/wallet.cpp
```diff
@@ -386,7 +386,11 @@ std::vector<rai::account> rai::wallet_store::accounts (MDB_txn * transaction_a)
 void rai::wallet_store::initialize (MDB_txn * transaction_a, bool & init_a, std::string const & path_a)
 {
 	assert (strlen (path_a.c_str ()) == path_a.size ());
-	auto error (mdb_dbi_open (transaction_a, path_a.c_str (), MDB_CREATE, &handle));
+	std::stringstream ids_path;
+	ids_path << path_a << "_send_action_ids";
+	auto error (0);
+	error |= mdb_dbi_open (transaction_a, path_a.c_str (), MDB_CREATE, &handle);
+	error |= mdb_dbi_open (transaction_a, ids_path.str ().c_str (), MDB_CREATE, &send_action_ids);
```
cryptocode: I realize this pattern is used elsewhere, but this will try to open the second db even if the first one failed. Maybe we should change the pattern to `error = error || mdb_dbi_open ...` ?

rai/node/wallet.cpp
```diff
@@ -925,43 +929,70 @@ std::shared_ptr<rai::block> rai::wallet::change_action (rai::account const & sou
...
+			{
+				if (auto id = id_a)
+				{
+					std::vector<uint8_t> vector;
+					{
+						rai::vectorstream stream (vector);
+						block->serialize (stream);
+					}
+					auto status (mdb_put (transaction, store.send_action_ids, rai::mdb_val (id->size (), const_cast<char*> (id->data ())), rai::mdb_val (vector.size (), vector.data ()), 0));
+					assert (status == 0);
```
cryptocode: Shouldn't `send_action` propagate errors to the caller?  In release mode, mdb_put errors are swallowed.

rai/node/wallet.hpp
```diff
@@ -5,6 +5,7 @@
 #include <rai/secure.hpp>
 
 #include <mutex>
+#include <optional>
```
cryptocode: boost/optional.hpp is used elsewhere

rai/node/wallet.cpp
PlasmaPower: I don't think there's much harm to opening the second DB, and it's very unlikely to fail. I'd like to stick with precedent here.

rai/node/wallet.cpp
PlasmaPower: Fixed

rai/node/wallet.hpp
PlasmaPower: Fixed

PlasmaPower:
Load tester is failing. This happened before on another recent commit, but it never failed before the hash2 changes. I'm running it locally for some debugging.

rai/node/wallet.cpp
```diff
@@ -925,43 +929,73 @@ std::shared_ptr<rai::block> rai::wallet::change_action (rai::account const & sou
...
 
-std::shared_ptr<rai::block> rai::wallet::send_action (rai::account const & source_a, rai::account const & account_a, rai::uint128_t const & amount_a, bool generate_work_a)
+std::shared_ptr<rai::block> rai::wallet::send_action (rai::account const & source_a, rai::account const & account_a, rai::uint128_t const & amount_a, bool generate_work_a, boost::optional<std::string> id_a)
 {
 	std::shared_ptr<rai::block> block;
+	bool cached_block = false;
 	{
-		rai::transaction transaction (store.environment, nullptr, false);
-		if (store.valid_password (transaction))
+		rai::transaction transaction (store.environment, nullptr, (bool)id_a);
```
lukealonso: Why does this need to be a write transaction if id_a is set?

PlasmaPower: It updates the `send_action_ids` store. Should I add a comment?

lukealonso: It's better to only hold a write transaction for the exact amount of time you need it, since it blocks all other readers. You're not even sure you need it at this point.

lukealonso: (the send queue is serialized, so there's no race)

PlasmaPower: I'm assuming it's better to release the old transaction before creating the write transaction, instead of using a nested transaction?

PlasmaPower:
I'm also planning on returning an `error` when the `wallet::send_action` fails, instead of a zero block hash. Thoughts? Is that too much of a breaking change for the RPC?

rai/node/wallet.cpp
```diff
@@ -925,43 +929,73 @@ std::shared_ptr<rai::block> rai::wallet::change_action (rai::account const & sou
...
 				{
-					rai::account_info info;
-					auto error1 (node.ledger.store.account_get (transaction, source_a, info));
-					assert (!error1);
-					rai::raw_key prv;
-					auto error2 (store.fetch (transaction, source_a, prv));
-					assert (!error2);
-					block.reset (new rai::send_block (info.head, account_a, balance - amount_a, prv, source_a, generate_work_a ? work_fetch (transaction, source_a, info.head) : 0));
+					rai::bufferstream stream (reinterpret_cast<uint8_t const *> (i->second.data ()), i->second.size ());
+					block = rai::deserialize_block (stream, rai::block_type::send);
```
lukealonso: Do we need to store the whole block, or will the just the hash suffice?

If you can't lookup the block by hash to republish it, there's a good chance republishing it won't work anyways since it's probably a fork at that point (the send queue will have moved on, and another block will have taken it's place as the descendant of the parent block at the time).

rai/node/wallet.cpp
lukealonso: Yes, annoyingly enough.

PlasmaPower:
Fixed @lukealonso's comments, and changed the `wallet::send_action` failure into an error instead of a zero hash.

lukealonso:
LGTM - @clemahieu @androm3da This should be merged and released ASAP.

PlasmaPower:
I'd support backporting this change until we're certain the hash2 changes are stable.

lukealonso:
@PlasmaPower Agreed, this is a critical fix.

PlasmaPower:
Hold off for a bit - this fails to restart.

PlasmaPower:
False alarm - it's some other problem, probably related to the testnet.

PlasmaPower:
The test does prove the idempotency with the same id (it asserts the balance and the block returned), and that it'll repeat the action with a different ID.

From a code architecture standpoint, I think it makes sense for the database to be per-wallet.

lukealonso:
It might be better to do it cross-wallet, if only because the number of dbs is limited and doesn't scale well.

PlasmaPower:
That's a good point. One question is concurrency. Are we doing actions for multiple wallets in parallel?

clemahieu:
It's possible for that to happen, yes, though write transactions are exclusive so we have some concurrency protection there.

We could do both per-wallet and also in a single database by making the key {wallet_id, tx_id} -> hash.  I'm not sure if variable keysizes make a performance difference.

lukealonso:
If you resend the same id to a different wallet it should fail, right?

clemahieu:
Honestly I think so.  It's very easy to pick a unique, maybe random ID and it seems like it might be more confusing if multiple wallets can have the same id.

PlasmaPower:
So maybe the mapping should be `action_id -> {wallet_id, block}`?

PlasmaPower:
In which structure should the `MDB_dbi` be stored?

clemahieu:
I'm thinking the rai::wallets class.  The reason is at some point i want to separate the ledger from the wallet data and since TX id tracking is related to a wallet operation it should be associated with that.

PlasmaPower:
I think we should either keep the send ids for multiple wallets independent, or return a block for a different wallet if you pass it the same send id. My latest commit does the latter.

In my opinion, it doesn't make much sense to store the wallet id just to give the user an error when it doesn't match.

clemahieu:
LGTM

lukealonso:
LGTM as well

PR #611, opened 2018-02-12
 Fixing posible DB growth with unchecked blocks

SergiySW:
Unchecked subdatabase allowed duplicates
https://github.com/clemahieu/raiblocks/issues/513
https://github.com/clemahieu/raiblocks/issues/522


lukealonso:
How does this perform? It might be the lesser of two evils to allow duplicates, since with the new bootstrap fixes, it should be much more difficult to get into a state where you keep downloading the ledger over and over and putting it into unchecked.

SergiySW:
There is broadcasting as well (and possible malicious nodes)

PR #622, opened 2018-02-15
Encrypted seed export and import commands

cryptocode:
This PR proposes two new commands, `seed_safe_export` and `seed_safe_import`, for exporting and importing encrypted seeds in a wallet agnostic way. 

This can be used for seed backup and safely importing seeds to other wallets supporting the format.

Sample output:

```
{
    "type_key": "ARGON2-IV256",
    "type_cipher": "AES256-CTR-IV128",
    "type_checksum": "SHA256",
    "key_iv": "E82FF3EF7E49FB42670A72C5C56C10DF9C54CB5F071E64ADF2C086905B3A4B8C",
    "seed_iv": "57931279153FBA7F0205D81BDC9979B1",
    "seed_encrypted": "BF2B32110D95842B2E7BE0DFE81EBDF184C1A218B33877A92A2298B1067DCC35",
    "checksum": "37662A432EAB048F7ADF26571F50EEA2B9890F4DF4E4BE0668550D284F222D21"
}
```

The will allow the following flow: Export seed from any wallet and create a QR code. Use "Scan encrypted seed" and enter passphrase to import. Other wallets can just import the json (like the import command does)

As a future improvement to the node CLI, I suggest introducing support for non-echoing terminal input to avoid passing passwords and other keys/seeds in plaintext to commands.

The `type_` tuples are what they are because that's what the node already use, but this can be changed and extended since the format includes all the info needed to decrypt once handed the passphrase.

### DIFF
--- a/rai/node/cli.cpp
+++ b/rai/node/cli.cpp
@@ -1,3 +1,5 @@
+#include <cryptopp/hex.h>
+#include <cryptopp/sha.h>
 #include <rai/lib/interface.h>
 #include <rai/node/cli.hpp>
 #include <rai/node/common.hpp>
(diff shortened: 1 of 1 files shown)


PlasmaPower:
How does this handle adhoc keys? Have you verified that the data is decryptable by the OpenSSL CLI?

cryptocode:
@PlasmaPower I've verified with external tools, seems to work. Now sure how ad-hoc keys play in here, `seed_safe_export` is only for exporting the seed, e.g. for safe import into other wallets.

cryptocode:
But the command is called `seed_safe_export`, isn't that pretty clear?

PlasmaPower:
Yeah, I guess. We'll be more careful if/when it's exposed in the UI.

cryptocode:
Agreed, it must be obvious that it's the seed.

SergiySW:
Guess it require some rebasing now :)
Also would be good to have some tests for such functions. May be move to wallet.cpp or something like that?

cryptocode:
Closing this as there are good ways to pipe --wallet_decrypt_unsafe into openssl or anyone's favorite tool. If we get requests for such a feature, I can redo the PR after the wallet split.

Issue #628, opened 2018-02-16
Formalize release process


androm3da:
We did `10.0` release and found some hiccups/speed bumps.  During this release cycle, many more developers were involved than were involved in previous releases.

We will need to formalize a process around releases, including steps, roles and responsibilities.

If we had to do a release in a hurry to fix a critical/widespread issue, we would probably face some challenges.  So near the top of our wishlist for release process should be to have a process that's as automated as possible.

androm3da:
Among the roles needed is a Release Manager.  This will be someone who is executive/administrative, facilitating: (1) coordinating release content (rejecting/accepting appropriate features/fixes), (2) the actual steps of the release process itself, (3) preserving overall project schedule goals for releases.

If no one else prefers this role, I will nominate myself.

androm3da:
Items to go in a release checklist:

- testing / evaluation procedures (beyond existing automated test suite)
- don't use the patch version number, just the major and minor
- expected tag format
- expected branching scheme for dev/releases

rkeene:
There is now such a document internally, and it will eventually live on the wiki.

PR #630, opened 2018-02-16
Show the patch version number 

paulmelis:
in places where a version is shown or returned (including RPC).


cryptocode:
This may be a breaking change for clients depending on the old format. Sounds like there should be a request parameter to indicate patch.

paulmelis:
You mean the RPC return value? Yeah, I was wondering about that

cryptocode:
It's possible to imagine exchange scripts checking the cmdline --version too, but in that case the option is probably too new to be an issue. I'm more worried about the output of RPC commands changing.

argakiig:
The CLI version was added recently, but the RPC one, what about adding an optional flag to show patch version, but default would be to maintain current behavior?

paulmelis:
I added a "patch" parameter, so the returned version number stays compatible with the current string. 

Interestingly, the RPC docs are currently not correct, as they show the patch version being returned when it isn't:

```
{  
  "rpc_version" : "1",
  "store_version" : "2",
  "node_vendor" : "RaiBlocks 7.5.0"
}
```

paulmelis:
Crap, the test is now wrong

paulmelis:
Failed tests look unrelated to this commit?

argakiig:
Failed due to formatting,

```
Code formatting differs from expected - please run ci/clang-format-all.sh```

File: rai/node/rpc.cpp [hunk 1]
```diff
@@ -3126,10 +3126,24 @@ void rai::rpc_handler::unchecked_keys ()
 
 void rai::rpc_handler::version ()
 {
+	bool patch (false);
+	boost::optional<bool> patch_optional (request.get_optional<bool> ("patch"));
+	if (patch_optional.is_initialized ())
+	{
+		patch = patch_optional.get ();
+	}
+
```
cryptocode: All of that can be reduced to `bool patch = request.get<bool> ("patch", false);`

argakiig: I agree I think this would be a good idea at least here @clemahieu thoughts?

clemahieu:
A while ago I removed the patch number from a lot of these places. My reasoning was to try and adopt the versioning scheme by Firefox where they stick with increasing major numbers instead of a long string of dot-separated minor and patch numbers. 

What do you think of using the major number and then maybe the first 6 digits of the sha of the commit it’s built on?

paulmelis:
My reason for this pull request was that the 10.0.1 release didn't identify itself as such, possibly leading to confusion w.r.t incorrect balances. I'd say being able to tell exactly which version you're running should be the main goal here, independent of how the version number looks.

androm3da:
My mistake, then -- I should've made it `10.1` instead.

androm3da:
Instead of accepting this PR, I have noted "don't use the patch number, only major/minor" in a set of items to be gathered into a release process checklist (#628 ).

Please reopen if you think it's still valid.

PR #632, opened 2018-02-16
Don’t re-use connections to unsynced peers.

lukealonso:


PlasmaPower:
Good catch! LGTM (though I don't understand the bootstrapping architecture well)

argakiig:
Agreed, does look good, I thought this was already in, but was for blocks not frontiers as well.

PR #634, opened 2018-02-16
Get rid of the frontier minimum, not needed anymore.

lukealonso:
The bootstrap rep weights make it easy to resolve forks, so we don’t need to avoid unsynced clients.


argakiig:
LGTM

PR #640, opened 2018-02-17
Compatibility with Cmake 2.8 (CentOS 7)

SergiySW:
to fix https://github.com/nanocurrency/raiblocks/issues/636


androm3da:
I suspect the wallet still won't build for centos 7 but this is a good solution for the node, thanks.

PR #653, opened 2018-02-20
Add explicit check for openssl 1.0

paulmelis:


cryptocode:
Does this pull in, say, 1.0.2n - not just 1.0 ?

paulmelis:
It only checks `major.minor` to be 1.0, the patch number is ignored as far as I can tell

paulmelis:
On my system I have both openssl 1.1.0g and 1.0.2n. If I configure cmake with

```
    -DOPENSSL_INCLUDE_DIR=/usr/include/openssl-1.0 \
    -DOPENSSL_CRYPTO_LIBRARY=/usr/lib/openssl-1.0/libcrypto.so \
    -DOPENSSL_SSL_LIBRARY=/usr/lib/openssl-1.0/libssl.so \
```

the version check passes.

cryptocode:
@paulmelis Great. @androm3da I think this should go in then, since the RPC-TLS addition depends on the Long Term Support release of OpenSSL.

PR #661, opened 2018-02-21
Exclude total pending amount from total representative stake

PlasmaPower:
Untested. I'm pretty sure this is the reason why it appears we don't have 50% of rep stake online.


PlasmaPower:
After some discussion about this, I think a better solution is to track online voting weight, and use some sort of moving average in place of the total supply. I'll make a new PR for that soon.

PR #662, opened 2018-02-22
Refactoring of secure.*

cryptocode:
secure.cpp is quite large and a bit hard to navigate at this point. 

As a start, I've split it into `common`, `ledger` and `blockstore`, and moved working_path to `utility`, where it was declared anyway. Some doxygen comments have been added as well.

No functional changes have been made.

Probably painful to review, but I think something has to be done before it gets too unwieldy.

If accepted, I hope this one gets merged early since it'll be rather hard to rebase if people start making changes to the secure.cpp ;) 

I think only #222,  #539 will need to be rebased, and those are both fairly small.

All tests pass and the node and wallet seems to run as before.


rai/node/utility.hpp
```diff
@@ -98,6 +98,9 @@ bool fetch_object (T & object, boost::filesystem::path const & path_a, std::fstr
 	return error;
 }
 
+/**
+ * RAII wrapper for MDB_env
```
argakiig: Any particular reason for RAII instead of Rai? Also since you are adding the commenting is there any reason we dont start moving it to NANO? same with like 133

argakiig: Other than this, LGTM

PlasmaPower: RAII here means Resource Acquisition Is Initialization. It's a C++ principle that describes what this wrapper does. It's not rai specific.

cryptocode: http://en.cppreference.com/w/cpp/language/raii

PR #663, opened 2018-02-23
Fix vote republishing rules

PlasmaPower:
The primary change in this PR is to make the cooldown per block per representative, as opposed to the previous per block cooldown.

Once widely deployed, this should really help with the slow receive times since it'll be easy to hit the quorum threshold.

Edit: this now also fixes duplicate sequence number detection.

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -222,26 +222,23 @@ void rai::network::republish_block (MDB_txn * transaction, std::shared_ptr<rai::
 
 // In order to rate limit network traffic we republish:
 // 1) Only if they are a non-replay vote of a block that's actively settling. Settling blocks are limited by block PoW
-// 2) Only if a vote for this block hasn't been received in the previous X second.  This prevents rapid publishing of votes with increasing sequence numbers.
-// 3) The rep has a weight > Y to prevent creating a lot of small-weight accounts to send out votes
-void rai::network::republish_vote (std::chrono::steady_clock::time_point const & last_vote, std::shared_ptr<rai::vote> vote_a)
+// 2) The rep has a weight > Y to prevent creating a lot of small-weight accounts to send out votes
+// 3) Only if a vote for this block from this representative hasn't been received in the previous X second.
+//    This prevents rapid publishing of votes with increasing sequence numbers.
+//
+// These rules are implemented by the caller, not this function.
+void rai::network::republish_vote (std::shared_ptr<rai::vote> vote_a)
 {
-	if (last_vote < std::chrono::steady_clock::now () - std::chrono::seconds (1))
+	rai::confirm_ack confirm (vote_a);
+	std::shared_ptr<std::vector<uint8_t>> bytes (new std::vector<uint8_t>);
 	{
-		if (node.weight (vote_a->account) > rai::Mxrb_ratio * 256)
-		{
-			rai::confirm_ack confirm (vote_a);
-			std::shared_ptr<std::vector<uint8_t>> bytes (new std::vector<uint8_t>);
-			{
-				rai::vectorstream stream (*bytes);
-				confirm.serialize (stream);
-			}
-			auto list (node.peers.list_sqrt ());
-			for (auto j (list.begin ()), m (list.end ()); j != m; ++j)
-			{
-				node.network.confirm_send (confirm, bytes, *j);
-			}
-		}
+		rai::vectorstream stream (*bytes);
+		confirm.serialize (stream);
+	}
+	auto list (node.peers.list_sqrt ());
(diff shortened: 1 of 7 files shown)


PlasmaPower:
Once widely deployed, this should really help with the slow receive times since it'll be easy to hit the quorum threshold.

PlasmaPower:
I'm trying to decide on a good value for the min republish vote weight. Right now I'm using 10k XRB. Previously the limit was 256 XRB, which was way too small in my opinion.

PlasmaPower:
Weight to cooldown distribution rewritten in terms of supply, and changed to be more restrictive (min weight percent for rebroadcast is now 0.1% of supply, currently including 31 reps).

rai/node/node.hpp
```diff
@@ -57,7 +57,7 @@ class election : public std::enable_shared_from_this<rai::election>
 	rai::uint128_t minimum_threshold (MDB_txn *, rai::ledger &);
 	rai::votes votes;
 	rai::node & node;
-	std::chrono::steady_clock::time_point last_vote;
+	std::map<rai::account, std::chrono::steady_clock::time_point> last_votes;
```
androm3da: probably could/should be an `unordered_map`

rai/node/node.cpp
```diff
@@ -2806,9 +2802,31 @@ void rai::election::confirm_cutoff (MDB_txn * transaction_a)
 
 void rai::election::vote (std::shared_ptr<rai::vote> vote_a)
 {
-	node.network.republish_vote (last_vote, vote_a);
-	last_vote = std::chrono::steady_clock::now ();
+	// see republish_vote documentation for an explanation of these rules
 	rai::transaction transaction (node.store.environment, nullptr, true);
+	auto supply (node.ledger.supply (transaction));
```
androm3da: I think it's ideal to put the units here.  These are both raw?

PlasmaPower: Yeah. Units aren't really important here though, as everything is in proportions.

cryptocode:
Before PR 663 and 664 (by skip, I mean republish_vote -> confirm_send not getting called):

```
calls to election_vote : 7000
skip due to time       : 5733
skip due to weight     : 84
```

After:

```
calls to election_vote : 7000
skip due to time       : 0
skip due to weight     : 6227
```

PlasmaPower:
Based on @cryptocode's numbers, this PR seems to *decrease* the number of votes rebroadcasted. In my opinion that's a good sign, as we're properly prioritizing votes by weight.

PlasmaPower:
Despite the decrease in rebroadcasted votes from one node, this PR will still probably increase bandwidth. Currently, the network isn't sending nodes votes with high weight, so very few votes are being rebroadcasted by this PR. After this PR is widely deployed, nodes *will* be getting votes with high weight, which will likely all be rebroadcasted.

PlasmaPower:
Undid `cache_vote`->`vote_max` rename as I came full circle on the design of the function.

PlasmaPower:
Now testing the cooldown: https://github.com/nanocurrency/raiblocks/pull/663/files#diff-045a181bb917e4bc71081327d83285da

PlasmaPower:
It's important to note that vote rebroadcasting rules and vote processing rules have been merged, so if a vote won't be rebroadcasted (e.g. cooldown not up, vote weight too low) it won't be processed.

I'm not entirely sure if that's the right idea, but I think it makes some theoretical attacks harder.

rai/node/node.cpp
```diff
@@ -373,11 +370,10 @@ class network_message_visitor : public rai::message_visitor
 		auto vote (node.vote_processor.vote (message_a.vote, sender));
 		if (vote.code == rai::vote_code::replay)
 		{
-			assert (vote.vote->sequence > message_a.vote->sequence);
 			// This tries to assist rep nodes that have lost track of their highest sequence number by replaying our highest known vote back to them
 			// Only do this if the sequence number is significantly different to account for network reordering
 			// Amplify attack considerations: We're sending out a confirm_ack in response to a confirm_ack for no net traffic increase
-			if (vote.vote->sequence - message_a.vote->sequence > 10000)
+			if (vote.vote->sequence + 10000 > message_a.vote->sequence)
```
clemahieu: I think this needs to be vote.vote->sequence > message_a.vote->sequence + 10000.

PlasmaPower: Good catch. Fixed.

PlasmaPower:
This has been merged with some modifications: https://github.com/nanocurrency/raiblocks/commit/63e15c21ae2b0098db266ef23c6f58f1e40676df

PR #664, opened 2018-02-23
Lower announce interval

PlasmaPower:
I'm not sure why it was so high in the first place, so I'd like someone who knows why to weigh in before merging this PR.

This does not affect the throughput, only latency.


argakiig:
@PlasmaPower does this still need looked into or can it be closed now? I dont know if you had gotten the answer you were trying to find or not

PlasmaPower:
I think so, but I think other changes should come first.

PR #665, opened 2018-02-23
Properly version (tag) docker images

PlasmaPower:
Now, the docker tag `latest` (the default) will refer to the latest commit tagged in git (effectively the latest release).

You can see a test run of this on my fork here: https://travis-ci.org/PlasmaPower/raiblocks/builds/345122239

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,5 +42,10 @@ script:
   - if [ "$TRAVIS_OS_NAME" = "osx" ]; then ci/build-travis.sh "/usr/local/opt/qt5/lib/cmake/Qt5"; fi
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then docker run -v $TRAVIS_BUILD_DIR:/workspace nanocurrency/nano-ci /bin/bash -c "cd /workspace && ASAN=${ASAN} TSAN=${TSAN} ./ci/build-travis.sh /usr/lib/x86_64-linux-gnu/cmake/Qt5 ${PWD}"; fi
 
-after_script:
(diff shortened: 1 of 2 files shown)


PlasmaPower:
Looks like it worked!

image

marvinroger:
It's better, but I am still not comfortable with that approach. The docker repository should contain proper version tags.

If there's a 10.1.0 release, it shoud be tagged `latest`, `10`, `10.1`, `10.1.0`. That way, we can use a semver approach, requiring the `10` version and having backward compatible changes without having to manually require `10.1.0`.

This would definitely help the Nano integration for devs. Using only `latest`, for now, feels too dangerous for a production env.

PlasmaPower:
@marvinroger The network evolves as a whole. If you're still running v10 when universal blocks go live, your node will cease to function. Semver doesn't really apply here.

PlasmaPower:
That said, I have set it to make a docker tag for every git tag. I'll consider expanding it to tagging the latest major version too, but I'm not sure if that's a good idea.

marvinroger:
Semver might not apply completely, but still, 10.x will stay compatible with previous 10.x releases, right?

Say the v11 is not backward compatible with v10 (the database is structured differently, for instance). Doing a git pull would break the node. Upgrading your node (and the network evolves as a whole, I get that) should be an explicit task (changing the version number). If you absolutely want to use the latest version anyway, while developing for example, you can still use `latest`.

It is considered a best practice to tag images with a version number. There are multiple benefits:

* You can target a specific version. Some releases might introduce bugs, and it is important to be able to rollback

* You can quickly see whether or not a release has been pushed to Docker

These best practices apply to software in general. It does not really matter that "the network evolves as a whole", it still requires a manual update to "evolve". If it really must to "evolve" automatically, the node should be able to auto-update anyway.

PR #668, opened 2018-02-25
Improve account history RPC endpoint

PlasmaPower:
This PR merges the `account_history` and `history` endpoint handlers, and adds the following options:

- `head`: Use this instead of the `history` endpoint, which doesn't really make sense.
- `offset`: Useful for pagination. Skip n blocks.
- `raw`: Don't oversimplify the output for wallets. Include all fields, include change blocks, and don't pretend open blocks are receive blocks.

In addition, the RPC will now return a `previous` field when the `count` resulted in the entire history not being returned. This field can be used for the `head` parameter for fast incremental pagination.


rai/node/rpc.cpp
```diff
@@ -1604,101 +1624,109 @@ class history_visitor : public rai::block_visitor
...
 	rai::block_hash const & hash;
 };
 }
 
-void rai::rpc_handler::history ()
+void rai::rpc_handler::account_history ()
 {
-	std::string hash_text (request.get<std::string> ("hash"));
 	std::string count_text (request.get<std::string> ("count"));
+	bool output_raw (*request.get_optional<bool> ("raw"));
```
cryptocode: I don't think this dereference is correct.

PlasmaPower: I'm not completely sure either. It seemed to work, but it might've just skipped an assert in release mode. I've changed it to `== true` which should definitely work, according to http://www.boost.org/doc/libs/1_56_0/libs/optional/doc/html/boost_optional/tutorial/a_note_about_optional_bool_.html

PR #672, opened 2018-02-26
Snapshot

argakiig:
Allow for creation of database snapshots with compaction without requiring the node to be stopped

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -3106,11 +3106,11 @@ bool rai::handle_node_options (boost::program_options::variables_map & vm)
 			std::cerr << "Vacuum failed" << std::endl;
 		}
 	}
-	else if (vm.count("snapshot"))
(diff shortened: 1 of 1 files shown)


rai/node/node.cpp
```diff
@@ -3105,6 +3106,37 @@ bool rai::handle_node_options (boost::program_options::variables_map & vm)
...
+			std::cout << "This may take a while..." << std::endl;
+
+			bool success = false;
+			{
+				inactive_node node (data_path);
+				success = node.node->copy_with_compaction (snapshot_path);
+			}
+			if (success)
+			{
+				std::cout << "Snapshot completed, This can be found at " << snapshot_path << std::endl;
```
androm3da: If this fails for some reason other than an exception, we won't print anything to stderr and will probably give an exitcode of zero to the parent.

PR #682, opened 2018-02-28
Move to C++14

cryptocode:
Sets -std to c++14. Replaces a single `unique_ptr#reset` with `make_unique` to see if things compile in CI. 

Also reverts a v10 release-specific RPATH change which currently prevents the wallet from running on macOS dev machines


PlasmaPower:
Doesn't mvsc not support c++14?

clemahieu:
Looks like a fair amount is supported in 2015 namely make_unique and generic lambda captures.

https://docs.microsoft.com/en-us/cpp/visual-cpp-language-conformance

cryptocode:
@PlasmaPower everything but esoteric SNIFAEness

cryptocode:
Alright, so CI accepted it. Would be great if someone could give it a spin on VC++ (everything in C++14 except the new SFINAE stuff should work)

argakiig:
i can give it a run this evening let you know what I find out

argakiig:
Core tests all pass as c++14 on VS2015
[----------] Global test environment tear-down
[==========] 450 tests from 52 test cases ran. (23384 ms total)
[  PASSED  ] 450 tests.

PR #683, opened 2018-02-28
Fix typos

practicalswift:
Fix typos.

### DIFF
--- a/FindBoost.cmake
+++ b/FindBoost.cmake
@@ -925,7 +925,7 @@ function(_Boost_MISSING_DEPENDENCIES componentvar extravar)
 endfunction()
 
 #
-# Some boost libraries may require particular set of compler features.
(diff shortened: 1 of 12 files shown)


PlasmaPower:
The folders other than `rai` are libraries. I think it's better to leave those alone.

FindCryptoPP.cmake
```diff
@@ -16,7 +16,7 @@
 #
 #   CRYPTOPP_LIBRARIES
 #     Points to the CryptoPP libraries that should be passed to
-#     target_link_libararies.
+#     target_link_libraries.
```
PlasmaPower: This is another library.

practicalswift:
@PlasmaPower Fixed. Please re-review :-)

PlasmaPower:
LGTM though I don't have permission to merge. @androm3da ?

PR #687, opened 2018-03-02
Fix bootstrapping ublocks

PlasmaPower:
Also includes a test for bootstrapping ublocks.


PlasmaPower:
Travis failing is expected because the ublocks branch isn't `clang-format`ed yet.

PR #691, opened 2018-03-02
Fix test for wallet_create_max to allow for utx database addition

argakiig:
update nonWalletDbs to account for utx database


PlasmaPower:
Looks like this PR targeted master instead of the ublocks branch.

argakiig:
Had checked, this doesn't break anything to future proof for ublocks

PlasmaPower:
Okay, though the test is still failing on the ublocks branch.

PR #695, opened 2018-03-03
Support gap_source for ublocks

PlasmaPower:


rai/lib/blocks.cpp
```diff
@@ -1110,7 +1110,7 @@ bool rai::utx_block::valid_predecessor (rai::block const & block_a) const
 
 rai::block_hash rai::utx_block::source () const
 {
-	return 0;
+	return hashables.link;
```
PlasmaPower: I looked through the code and `source ()` isn't used anywhere that could cause problems with an invalid block.

PlasmaPower: Removing in favor of a future PR with a general block source function.

PR #704, opened 2018-03-06
Improved error handling in message parser

cryptocode:
Communicates errors from message_parser to the node more precisely so errors can be logged.


argakiig:
LGTM, This improves the errors generated from message_parser

PR #719, opened 2018-03-09
Modify `wallet_create` CLI to allow seed & password change

s1na:
It creates a new wallet, and changes the seed to the provided `--key`.

However I couldn't think of a good way to check if a wallet with this seed already exists, since it would require all of their passwords.


rai/node/node.cpp
```diff
@@ -3524,6 +3525,25 @@ bool rai::handle_node_options (boost::program_options::variables_map & vm)
...
+		rai::raw_key key;
+		if (!key.data.decode_hex (vm["key"].as<std::string> ()))
+		{
+			inactive_node node (data_path);
+			rai::keypair gen_key;
+			auto wallet (node.node->wallets.create (gen_key.pub));
+			wallet->enter_initial_password ();
+			rai::transaction transaction (wallet->store.environment, nullptr, true);
+			wallet->change_seed (transaction, key);
+			std::cout << gen_key.pub.to_string () << std::endl;
```
PlasmaPower: Could you use `to_account` here? Public keys are meant to be internal where possible.

SergiySW:
@cryptocode it does make more sense with https://github.com/nanocurrency/nano-node/pull/1567
I guess for unification -wallet_create + --wallet_change_seed are better

SergiySW:
Original PR was modified to allow seed & password set inside of --wallet_create command

PR #720, opened 2018-03-10
Disable block_info table for utx blocks

SergiySW:


PlasmaPower:
Where is block_info used? Can we update those uses to support ublocks?

SergiySW:
It's workaround for old blocks for faster calculation of balance/amount & block account
https://github.com/nanocurrency/raiblocks/pull/200

PR #722, opened 2018-03-11
Corrected public key of preconfigured beta-rep

rkeene:
The public key for the beta rep was actually specified as the wallet ID


PlasmaPower:
Just double checking. The equivalent account is `xrb_3kbzg73bjsi85scbwnouj44iinrsqtdqphzay1x3pwmgmhkdwg8yjntxff33`. Is that correct?

rkeene:
Right, this is the account specified in the beta.json as well ( https://github.com/nanocurrency/raiblocks/blob/6673c42f68ae85401859abd6138ee7af13564186/docker/node/config/beta.json#L41 )

PR #729, opened 2018-03-13
Prevent RPC double write results with optional errors

SergiySW:


SergiySW:
Replaced with return; to simplify https://github.com/nanocurrency/raiblocks/pull/730

PR #730, opened 2018-03-13
Prevent RPC double write results with optional errors

SergiySW:


PlasmaPower:
IIRC `block_create` has a double error write when the wallet is locked.

SergiySW:
This is not single possible location. Now each optional error_response is followed by return; to exit function

PlasmaPower:
I think we've been trying to avoid `return` throughout the codebase.

SergiySW:
It's much simpler to read than https://github.com/nanocurrency/raiblocks/pull/729/files
What is the reason to avoid return; in void functions?

PlasmaPower:
Colin started the convention throughout the codebase. I think it's to preserve control flow from the start to the end of the function.

SergiySW:
In this case there is https://github.com/nanocurrency/raiblocks/pull/729 but it's heavier to read IMO

cryptocode:
This problem is one example where I think we should be using function-localized exceptions. Just a single `response (...)` in the try block, throw string on error, and a single `error_response (err)` in the catch.

cryptocode:
It was suggested that we use RPC to get std::error_code/std::unexpected tested, https://github.com/nanocurrency/raiblocks/pull/769

SergiySW:
Replaced with https://github.com/nanocurrency/raiblocks/pull/970

PR #735, opened 2018-03-14
Serialize ublock link as both hex and account

### LINKED ISSUE #734: UTX callback: Link field is always base32-encoded, even if referencing a block hash
I find the RPC callback behavior a little odd. Block hashes in the Link field get encoded with base32. 

E.g. Beta network's block `711366594E5E8D17AA1340CEC4D2C7A354F23179DA6E5CEB7C88FEABEED6D043` is a UTX-style receive block, so the Link field references the sender block hash, but RPC callback sends ` \"link\": \"xrb_3u7go5p597us1eieg7un3t44rcpmnuunztgoxfxcyag48pozpzzkmfue3r8j\"` instead, so you have to decode it first in order to get the block hash `ECAEA8EC3397790320C717740E842C2AD3A6F74FE9D5EB7AAF21C235ABFB7FF2`. 

Is this really intentional? I see more reasons against this behavior than in favor of it.

PlasmaPower:
Fixes #734 

I chose to have `link` be hex, because that's the more direct representation. It's encoded as an account in `link_as_account`. I chose that name to clarify that it's the same field, just a different representation.

This PR also allows for `link` being hex in `process` calls and such, since it's now returned that way from `block_create`.

### DIFF
--- a/rai/lib/blocks.cpp
+++ b/rai/lib/blocks.cpp
@@ -882,7 +882,7 @@ rai::utx_hashables::utx_hashables (bool & error_a, boost::property_tree::ptree c
 					error_a = balance.decode_dec (balance_l);
 					if (!error_a)
 					{
-						error_a = link.decode_account (link_l);
+						error_a = link.decode_account (link_l) && link.decode_hex (link_l);
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


PlasmaPower:
Looks like this branch has some whitespace issues again (not part of this PR), so Travis will fail.

PR #736, opened 2018-03-16
Moving initial bootstrap weights from generated array to ledger table

cryptocode:
Adds a command to produce a redistributable ledger db with initial bootstrap representative weights. The generated rai_bootstrap_weights array is thus gone.

initial_repweight_generate does the same job as the Python script did, except weights end up in the redist ledger.

The command dumps the SHA256 of the redist db to console, which can be used for verification at use-site.

There's an optional `--file` argument which takes a json settings file for redist generation, currently with limit (default 99) and cutoff (default 250K), e.g. same defaults as the old solution. I can make a wiki page for this if the PR gets merged.

We probably want more stuff in the redist ledger in the future, and having tons of params on the cmdline gets unwieldy. Hence the settings file.

The PR proposes a few utility functions/operators to improve code reuse (e.g. got another patch that needs the sha256 function)

There's an existing TEST(ledger, bootstrap_rep_weight) which I believe is still valid.


cryptocode:
fwiw bootstrap_fork_open fails on master as well when I run core_test locally

cryptocode:
Not sure if this is still relevant. If it is, I'll rebase it.

rkeene:
Do we want the weights to be software built-in or external ?

cryptocode:
@clemahieu had the idea of generating ledgers with initial bootstrap weights, this was an attempt at implementing that - not sure if this is still the route to take?

cryptocode:
This doesn't seem to fit in the v18 cleanup release

rkeene:
Yeah, let's defer this until later.  Also, I'm not sure why I assigned this to myself instead of @cryptocode but I suspect that was an accident.

cryptocode:
Closing this oldie, too many conflicts. Can redo if still relevant. An simpler alternative is to let anyone provide json dumps of their repweights for others to import (if e.g a repweight-live.json is present in data_dir, then it'll override the repweights in the binary)

PR #746, opened 2018-03-22
Draft UDP protocol docs

paulmelis:
I started writing up the UDP protocol and block encoding, as far as I can deduce them. The description is not complete yet, might contain errors and it might not be in a form you find acceptable. I do think it would be good to have this specification in the sources repo and not on the wiki, as keeping the spec in sync with the code would be the main priority.

Anyways, I'll keep working on this locally. I'm doing this pull request in case it's judged to be useful for the main repo.

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -29,10 +29,10 @@ The following different types of values are used in the protocol:
     - The 256-bit hash of a block  
     - The hash is computed from a set of "hashables": fields in a block that are hashed
       with the Blake2b hashing function
-    - Unless noted other the hashables for a block are the block-specific fields
+    - Unless noted otherwise the hashables for a block are the block-specific fields
       listed in the table with each block description. They are to be hashed in the
       order they are listed.
-    - The block hash is used (a.o.) to make references among blocks
+    - The block hash is used (amongst other things) to make references among blocks
     - Encoded as a sequence of 32 bytes
   * Balance / Amount
     - A 128-bit unsigned integer
@@ -44,8 +44,11 @@ The following different types of values are used in the protocol:
       - A 512-bit Ed25519 signature
       - Encoded as a sequence of 64 bytes
   * Work
-      - A 64-bit Proof-of-Work value
-      - Encoded as a *big-endian* unsigned integer
+      - A 64-bit unsigned integer Proof-of-Work value
+      - Encoding depends on the type of block: *big-endian* for a State block,
+        *little-endian* for other types of blocks
+
+In general, integer values of up to 8 bytes are encoded as little-endian.
 
 The Ed25519 implementation in Nano does not use the standard
 SHA-512 hashing algorithm, but Blake2b. See [here](https://github.com/nanocurrency/raiblocks/wiki/Design-features)
@@ -198,10 +201,37 @@ The different block types are denoted by a Type value (see `enum class block_typ
   | Change     | 5     |
   | State      | 6     |
 
+## Transfer of funds
+
+For each block in an account's chain a balance value can be determined, either from
+a value stored in the block itself or by using differences between successive
+blocks in the chain. Differences in balance values between successive blocks
+can be used to compute the amount of funds transferred by blocks to other
+accounts.
+
+Only Send and State blocks contain an explicit Balance value. For other block
+types balances and amounts needs to be deducted recursively:
+
+* The balance of an Open block is equal to the amount being sent by the corresponding
+  Send block. An exception to this is the Open block for the Genesis account, which always
+  receives the maximum supply of 2<sup>128</sup>-1 raw.
+* The balance of a Receive block is the balance of its predecessor in the
+  account chain plus the amount sent by the corresponding Send block.
+* The amount of funds being sent by a Send block is defined implicitly as
+  the difference between the balance at the previous block in the account chain
+  and the balance at the Send block. This amount must be > 0.
+* A Change block has no influence on an account's balance, so its balance is
+  the same as the balance of its predecessor.
+
+All State blocks contain a Balance value, hence determining amounts sent or
+received from one State block to the next is straightforward.
 
 ## Send block
 
 Reference: `rai::send_block` in `rai/lib/blocks.cpp`
+
+A Send block transfers funds from the account the block belongs to to a
+different account.
 
 A Send block contains the following block-specific fields:
 
@@ -214,6 +244,8 @@ A Send block contains the following block-specific fields:
 Note that the Balance value in the block is the balance of the sending
 account *after the amount sent in the block has been deducted*.
 
+XXX is it valid to send to yourself?
+
 ## Receive block
 
 Reference: `rai::receive_block` in `rai/lib/blocks.cpp`
@@ -225,7 +257,7 @@ A Receive block contains the following block-specific fields:
   | Previous      | 32    | Hash of the previous block in the account chain |
   | Source        | 32    | Hash of the corresponding send block from which the funds are sent |
 
-The block referenced by Previous must be a Send block.
+The block referenced by Source must be a Send block.
 
 ## Open block
 
@@ -242,13 +274,18 @@ An Open block contains the following block-specific fields:
   | Representative | 32    | The account representative for the newly opened account |
   | Account        | 32    | The account being opened |
 
-The block referenced by Previous must be a Send block.
+The block referenced by Source must be a Send block.
+
+An exception to the previous remark applies to the Open block of the Genesis
+account: its Source value is the public key of the Genesis account. The Open
+blocks for the 3 different networks are defined by the `..._genesis_data` JSON
+blocks in `rai/common.cpp`.
 
 ## Change block
 
 Reference: `rai::change_block` in `rai/lib/blocks.cpp`
 
-A Change block sets a new representative for an account
+A Change block sets a new representative for an account.
 
 A Change block contains the following block-specific fields:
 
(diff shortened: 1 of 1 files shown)


docs/protocol.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Vote
+
+Reference: class `rai::vote` in `rai/common.cpp`
+
+  | Field     | Size (bytes) | Description                          |
+  | --------- | ------------ | ------------------------------------ |
+  | Account   | 32           | Public key of the voting Account     |
+  | Signature | 64           | Signature                            |
+  | Sequence  | 8            | 64-bit sequence number (big-endian?) |
```
PlasmaPower: This is little endian IIRC. In general anything 8 bytes or less is little endian. We are going to change this stuff to big endian eventually though.

docs/protocol.md
```diff
@@ -0,0 +1,286 @@
...
+
+  - If the Representative value has changed then a new representative is set
+    for the account
+  - If the current Balance > previous Balance then the block represent a receive of funds
+  - If the current Balance < previous Balance then the block represent a sending of funds
+
+XXX how are open blocks replaced by state blocks? By setting Link to the sending
+block of the other account?
+
+XXX are 0 amount sends allowed?
```
PlasmaPower: No, but 0 receives are.

docs/protocol.md
```diff
@@ -0,0 +1,286 @@
...
+The difference between two sequential State blocks determines the type of
+state transition from the first block to the next:
+
+  - If the Representative value has changed then a new representative is set
+    for the account
+  - If the current Balance > previous Balance then the block represent a receive of funds
+  - If the current Balance < previous Balance then the block represent a sending of funds
+
+XXX how are open blocks replaced by state blocks? By setting Link to the sending
+block of the other account?
```
PlasmaPower: In that case, yes link is the source, and previous is 0.

docs/protocol.md
```diff
@@ -0,0 +1,286 @@
...
+
+XXX are 0 amount sends allowed?
+
+The Work value for a State block only contains the 48 least significant bits? 
+(as mentioned in `class state_block` in `rai/lib/blocks.hpp`).
```
PlasmaPower: That comment is incorrect or outdated; I'm not sure which.

docs/protocol.md
```diff
@@ -0,0 +1,286 @@
...
+
+Blocks are encoded into a byte stream, with each field directly following
+the previous value. All blocks contain a Signature and a Work value, which follow
+the block-specific fields:
+
+  | Field     | Size (bytes) | Description                                |
+  | --------- | ------------ | ------------------------------------------ |
+  | *Block-specific fields* | |                                           |
+  | Signature | 64           | Ed25519 signature of the block's hashables |
+  | Work      | 8            | Unsigned 64-bit integer (little endian?)   |
```
cryptocode: Work for universal blocks (state_blocks) are big endian.

docs/protocol.md
```diff
@@ -0,0 +1,286 @@
...
+    for the account
+  - If the current Balance > previous Balance then the block represent a receive of funds
+  - If the current Balance < previous Balance then the block represent a sending of funds
+
+XXX how are open blocks replaced by state blocks? By setting Link to the sending
+block of the other account?
+
+XXX are 0 amount sends allowed?
+
+The Work value for a State block only contains the 48 least significant bits?
```
cryptocode: IIRC this comment is outdated.

docs/protocol.md
```diff
@@ -0,0 +1,286 @@
...
+    - This number has unit "raw", where 1 XRB = 1 Mxrb = 10<sup>30</sup> raw.
+      See [here](https://github.com/nanocurrency/raiblocks/wiki/Distribution,-Mining-and-Units)
+      for more information.
+    - Encoded as a sequence of 16 bytes, *big-endian* order
+  * Signature
+      - A 512-bit Ed25519 signature
+      - Encoded as a sequence of 64 bytes
+  * Work
+      - A 64-bit Proof-of-Work value
+      - Encoded as a *big-endian* unsigned integer
```
cryptocode: Depends on block type, state_block big endian, otherwise little endian.

argakiig:
FYI appveyor build failure is a Red Herring, this is due to how pull request are done I need to fix in the future, but for now future PR's shouldnt try and build

docs/protocol.md
```diff
@@ -0,0 +1,282 @@
...
+There are other types of messages denoted by `message_type`, but these are
+only used in the "bootstrapping" procedure when a node needs to synchronize
+itself with all the blocks in the distributed ledger (in a bulk fashion).
+Bootstrapping is done over TCP connections, not with UDP packets.
+
+All UDP messages have a fixed-size header (see `rai::message::read_header()` in `rai/node/common.cpp`):
+
+  | Field         | Size (bytes)  | Description |
+  | ------------- | ------------- | ----------- |
+  | Magic         | 2             | Magic number to indicate this is a Nano packet |
```
termhn: Technically the magic number is only one byte, 'R' in ASCII, and the second byte is another magic number ('A', 'B', 'C' in ASCII) denoting the network (live, beta, test respectively)

PlasmaPower: Yeah that's described a bit below.

docs/protocol.md
```diff
@@ -0,0 +1,282 @@
...
+
+A Keepalive message contains a list of IP addresses of peers, in the form of (IPv6 address, port) pairs.
+Each IP address is stored as 16 bytes, each port as 2 bytes:
+
+  | Field        | Size (bytes) | Description     |
+  | ------------ | ------------ | --------------- |
+  | IPv6 address | 16           | Big-endian?     |
+  | Port         | 2            | Little-endian?! |
+
+The number of pairs in the message can vary, from 1 up to 8 peers.
```
termhn: Evidently there's currently a bug where there must be 8 peers or the reference node does an out of bounds read https://github.com/nanocurrency/raiblocks/issues/673

docs/protocol.md
termhn: Right just saw that hah.

docs/protocol.md
cryptocode: @termhn that's fixed

termhn: Sweet

PlasmaPower:
Personally I think this would be a good fit for the GitHub wiki. Is there a reason you'd prefer to put it in the repo itself?

docs/protocol.md
paulmelis: Why is the work value stored as uint64_t in the node anyway? It seem to be only ever used a input for the blake2b hash and is never used as integer value.

termhn:
@goophile there's actually a native implementation of the Nano protocol as a dissector in the latest development release of Wireshark :D

goophile:
@termhn Just found it, cool!

renesq:
@goophile I heard that slact's lua dissector is a little more detailed than the native one for certain situations. Not sure how they all differ, I haven't used them yet. https://gist.github.com/slact/63571aad31d8f445ac045391a7857ef5

docs/protocol.md
```diff
@@ -0,0 +1,333 @@
...
+  * Balance / Amount
+    - A 128-bit unsigned integer
+    - This number has unit "raw", where 1 XRB = 1 Mxrb = 10<sup>30</sup> raw.
+      See [here](https://github.com/nanocurrency/raiblocks/wiki/Distribution,-Mining-and-Units)
+      for more information.
+    - Encoded as a sequence of 16 bytes, *big-endian* order
+  * Signature
+      - A 512-bit Ed25519 signature
+      - Encoded as a sequence of 64 bytes
+  * Work
```
renesq: You could add that the work is normally computed based upon the previous block's hash, except on open blocks where the new address's public key is used instead (should apply to state blocks accordingly) AFAIK. 

Another interesting detail may be the difference in the work thresholds for the different networks.

docs/protocol.md
renesq: @paulmelis Nodes use the value to validate the work. As long as it's a hexadecimal value with a length of 16 characters, storing it as 64bit binary makes the most sense to me since it's the most compact one i think. What better options do we have?

docs/protocol.md
```diff
@@ -0,0 +1,333 @@
...
+  | Field     | Size (bytes) | Description                          |
+  | --------- | ------------ | ------------------------------------ |
+  | Account   | 32           | Public key of the voting Account     |
+  | Signature | 64           | Signature                            |
+  | Sequence  | 8            | 64-bit sequence number (big-endian?) |
+  | Block     | Varies       | The Block being voted on             |
+
+The Signature of a Vote signs the following list of hashables: `hash(Block)`, `Sequence`
+
+XXX what is the exact function of the sequence number?
```
PlasmaPower: A representative can change its vote for a block in a fork (i.e. if the other block won). This is to ensure that the new vote overrides the old vote.

docs/protocol.md
termhn: Length 8 byte array (`uint8_t work[8]`)

docs/protocol.md
termhn: Was explaining this yesterday to someone; an example:

Say for example, my node originally votes for block X; I send sequence number of 0 and my vote signature for block X. Now, when another node is processing my votes, it sees my vote for X and adds my weight to X in that election, and that my vote had sequence number 0. Later, my node observes > 50% of voting power is now voting for the opposite block and so now I want to change my vote. Well now I vote for block Y and attach a sequence number of 1. Other nodes then see my vote for Y, compare the new sequence number to the sequence number it's already stored for my rep and sees that the new sequence is greater (1 > 0), so it now replaces my vote and adds my stake to block Y. The reason this is required is because without the sequence, it's possible that my vote for block Y could reach this node first and my vote for block X could have reached it second. In that case, without the sequence number, it would then think my latest vote was for block X, which is of course the opposite of what it should be.

docs/protocol.md
```diff
@@ -0,0 +1,333 @@
...
+  | Field         | Size (bytes) | Description |
+  | ------------- | ----- | ----------------------------------------------- |
+  | Previous      | 32    | Hash of the previous block in the account chain |
+  | Destination   | 32    | The destination account being sent to |
+  | Balance       | 32    | The sending account's new balance |
+
+Note that the Balance value in the block is the balance of the sending
+account *after the amount sent in the block has been deducted*.
+
+XXX is it valid to send to yourself?
```
renesq: Yes, with the old "send" block type, you can send to yourself, even 0 value transactions.

PlasmaPower: You can also send to yourself with the new state blocks, unless I forgot about something.

docs/protocol.md
paulmelis: Casting a uint64_t to a uint8_t* as is currently done makes it dependent on the endianness of the system running on (although there aren’t many BE cpus around). 

Also, reading/writing the value as an array of bytes is easier, it saves the current endian handling (which was changed for state blocks)

cryptocode:
> I do think it would be good to have this specification in the sources repo and not on the wiki, as keeping the spec in sync with the code would be the main priority.

The same applies to the RPC document - it must be kept up to date. I think a release process where the wiki is updated could work too - then at least everything is in one place.

docs/protocol.md
```diff
@@ -0,0 +1,333 @@
...
+
+The different block types are denoted by a Type value (see `enum class block_type` in `rai/lib/blocks.hpp`):
+
+  | Block type | Value |
+  | ---------- | ----- |
+  | Send       | 2     |
+  | Receive    | 3     |
+  | Open       | 4     |
+  | Change     | 5     |
+  | State      | 6     |
```
IvanAnishchuk: Maybe some restructuring here is a good idea. To emphasize that state blocks are to be used in all situations and other types are obsolete... Perhaps, "state blocks" and "legacy block types" subsections or something?

zhyatt:
Most of this content is covered in https://docs.nano.org/ today or will be part of future living whitepaper updates. Also protocol message details are here: https://github.com/nanocurrency/protocol/tree/master/reference

PR #758, opened 2018-03-28
Log errors in work peer client

cryptocode:
Also fix log output where port was reported as an error code

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -2176,27 +2176,27 @@ public:
 											}
 											else
 											{
-												BOOST_LOG (this_l->node->log) << boost::str (boost::format ("Work peer %1% responded with an error %2%") % connection->address % connection->port);
+												BOOST_LOG (this_l->node->log) << boost::str (boost::format ("Work peer responded with an error %1% %2%: %3%") % connection->address % connection->port % connection->response.result ());
 												this_l->failure (connection->address);
 											}
 										}
 										else
 										{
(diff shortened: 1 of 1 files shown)


rai/node/node.cpp
```diff
@@ -2176,27 +2176,27 @@ class distributed_work : public std::enable_shared_from_this<distributed_work>
...
-											BOOST_LOG (this_l->node->log) << boost::str (boost::format ("Unable to read from work_peer %1% %2%") % connection->address % connection->port);
+											BOOST_LOG (this_l->node->log) << boost::str (boost::format ("Unable to read from work_peer %1% %2%: %3%") % connection->address % connection->port % ec);
 											this_l->failure (connection->address);
 										}
 									});
 								}
 								else
 								{
-									BOOST_LOG (this_l->node->log) << boost::str (boost::format ("Unable to write to work_peer %1% %2%") % connection->address % connection->port);
+									BOOST_LOG (this_l->node->log) << boost::str (boost::format ("Unable to write to work_peer %1% %2%: %3%") % connection->address % connection->port % ec);
```
clemahieu: Do we want to see the code or ec.message()?

cryptocode: Good question. Not sure if a message is always present, so maybe both? Error message with code in parenthesis sound good?

clemahieu: Oh yea, that's a good idea.

cryptocode: Updated

PR #763, opened 2018-03-30
Disable work generation by default for mass creation of accounts

SergiySW:
Rationale: for 100+ accounts at once even fastest GPU can't handle it


argakiig:
LGTM, good call.

PlasmaPower:
What if these are for deposit accounts, where likely only one or two are being used at once? I'd prefer an explicit work flag.

SergiySW:
It's about creation time when node can start work generation for each new

PlasmaPower:
Oh, I forgot about caching. Never mind then.

PR #764, opened 2018-03-30
Increase change_seed accounts check count

SergiySW:
Check existance of up to 64 accounts from last found + extra check for large wallets


argakiig:
Formatting, otherwise lgtm

PR #765, opened 2018-03-30
RPC wallet_deterministic_check

SergiySW:
to check if there are existing deterministic accounts > deterministic_index


SergiySW:
Doesn't make much sense with RPC wallet_info. Now service can check exact wallet accounts count & restore remaining count later with RPC accounts_create

PR #766, opened 2018-03-30
Disable work generation for change_seed

SergiySW:
if there are more than ~~1 account~~ 4 accounts to insert


PlasmaPower:
How about 4 accounts? I think that's a reasonable upper limit for an average user.

SergiySW:
Considering ~ 13 seconds work generation time for server Atoms. 13 * 4 < 60 seconds. Makes sense

PR #768, opened 2018-03-30
Check for pending blocks in change_seed action

SergiySW:


rai/node/wallet.cpp
```diff
@@ -1313,6 +1314,17 @@ rai::public_key rai::wallet::change_seed (MDB_txn * transaction_a, rai::raw_key
 			// i/64 - Check additional accounts for large wallets. I.e. 64000/64 = 1000 accounts to check
 			n = i + 64 + (i / 64);
 		}
+		else
+		{
+			// Check if there are pending blocks for account
+			rai::account end (pair.pub.number () + 1);
+			for (auto ii (node.store.pending_begin (transaction_a, rai::pending_key (pair.pub, 0))), nn (node.store.pending_begin (transaction_a, rai::pending_key (end, 0))); ii != nn; ++ii)
```
PlasmaPower: Why not just check if begin != end directly here?

SergiySW: Unfortunantely 
if (node.store.pending_begin (transaction, rai::pending_key (pair.pub, 0)) != node.store.pending_end ())
won't work here as we have account+ hash pending key

PlasmaPower: Ohh I see. That makes sense.

PR #769, opened 2018-03-30
WIP: Error handling using the proposed std::expected

cryptocode:
This is a "tracer bullet" PR for improving error handling using an implementation of the upcoming std::expected. This is a Rust-like combined result and error code approach, which
plays nicely with `std::error_code`

It's made in the hopes of getting a discussion going. The PR will be tweaked to improve the ergonomics of error handling, but should be enough to get a feel of the thing.

Example:

```
// Get an account
auto info (store.account_get (transaction, pending_key.account));
if (info)
{
    // We definitely have an account. We use the account_info fields via the arrow operator
    check_account (info->rep_block);
}
else
{
    // Conveniently, the error has a message attached to it.
    status->setText ("Could not query account information: " + info.error ().message ()));

    // And of course error codes
    if (info.error () == rai::block_store::error::missing_account)
    {
        // Do something special
    }
}
```

The PR:

1. Adds the scaffolding to use _std::expected_ + _std::error_code_
2. Implements it on a *single* function, `block_store#account_get`, to get a sense of the effort involved and what the pros and cons might be.

This implementation uses *expected-lite* to get a consistent result across compilers. Once std::expected lands in the C++ specification and is implemented in all compilers we care about, we can switch over.

Overall, the idea is as follows:

1. a) Instead of returning bool or int to indicate errors, *expected* is returned with the error enum defined by the class. b) If there's no error, just return the value.

2. Every class using _expected_ MUST provide an enum enlisting all possible errors. Multiple enums can be used when that makes sense.

3. Every class using expected SHOULD provide an std::error_code implementation.
This is kinda verbose, so a macro is provided to turn the std stuff into a one-liner. The remaining task is thus to implement std::error_category to associate error messages with each error. A "make_error_code" must be added for each class.  All of this is a one-time job per class, and thus not a big deal IMO.

4. Simple classes can use _expected_ with the enum instead of using the more elaborate std::error_code.

Benefits:
- Dereferencing an invalid return value will assert during development.
- A clear way to associate error messages with each error.
- More readable code and improved type safety (for instance, checking enum error codes in switches enforces exhaustive checks unless you specify a default)
- Some people find predicate-looking functions that return false as success confusing. With std::expected, you check if the return value is "truthy" and extract the error code otherwise.
- C++ may get monadic extensions for chaining functions that may fail. For now, the API support *value_or*, where you can chain an immediately invoked lambda.

Disadvantages:
- There's a bit of boilerplate for each class. That said, the scaffolding is a one-time job, after which adding new errors is easy.
- It's a significant effort to convert the current code base, with risk of mistakes being made.


rai/blockstore.cpp
```diff
@@ -838,23 +873,31 @@ bool rai::block_store::account_exists (MDB_txn * transaction_a, rai::account con
...
+	expected<rai::account_info, std::error_code> res;
+	rai::account_info info_a;
 	rai::mdb_val value;
 	auto status (mdb_get (transaction_a, accounts, rai::mdb_val (account_a), value));
 	assert (status == 0 || status == MDB_NOTFOUND);
-	bool result;
 	if (status == MDB_NOTFOUND)
 	{
-		result = true;
+		res = make_unexpected (make_error_code (error::missing_account));
```
cryptocode: This idiom is what people seem to use, but I think better ergonomics should be possible. For a class using just enums (instead of system::error_code), this would be `make_unexpected (error::missing_account);`

clemahieu:
Conceptually I like the idea and I've seen other projects do similar things to give more reliable handling of errors.

Implementing this in a less critical section would probably be a good start, maybe putting this in to the RPC code which could simultaneously solve some of the complexity there around error handling.

cryptocode:
Yeah,  this WIP is just for discussion. I picked a popular function like account_get to see much code was affected.

Is RPC a good candidate considering we want to move that out-of-process soon-ish (once we decide on the C API/shared mem, etc) ? I was thinking doing it on something that would stay in core would be better - the RPC will probably be completely rewritten, using a different JSON lib, etc, right?

cryptocode:
@clemahieu I'll make a variant of this PR for RPC, maybe start it off with a few functions.

cryptocode:
Closing in favor of an error handling PR for RPC

PR #770, opened 2018-04-01
Prevent tests from accumulating thousands of files and directories

cryptocode:
My RaiBlocksTest directory has accumulated tens of thousands of tmp directories and files, totaling 1GB+. Each test run produce hundreds of files using `rai::unique_path`

This PR removes any tmpdirs and lockfiles produced by a test (in the `system` destructor)

The same way you can use environment variables to pick tests (GTEST_FILTER), you can use an environment variable (TEST_KEEP_TMPDIRS=1) to retain the directories, e.g. to view log files when there are test failures.

### DIFF
--- a/rai/node/testing.cpp
+++ b/rai/node/testing.cpp
@@ -1,5 +1,6 @@
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
+#include <cstdlib>
 #include <rai/node/common.hpp>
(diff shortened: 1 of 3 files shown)


PlasmaPower:
LGTM. @clemahieu are you using the tmpdirs for analysis? Are you fine with this change?

Personally I just edit testing.cpp and add `logging.log_to_cerr_value = true` whenever I want to debug something.

PR #771, opened 2018-04-02
Prevent some RPC commands from returning an invalid response

koczadly:
This PR fixes an issue where RPC queries sometimes returned an invalid body because the return function was being called multiple times in instances where an else check wasn't used.

This resulted in responses like this when feeding bad arguments to `block_create` (this is only the body of the response):
```text
{    "error": "Bad private key"}HTTP/1.1 200 OKContent-Type: application/jsonAccess-Control-Allow-Origin: *Access-Control-Allow-Headers: Accept, Accept-Language, Content-Language, Content-TypeConnection: closeContent-Length: 35{    "error": "Bad private key"}HTTP/1.1 200 OKContent-Type: application/jsonAccess-Control-Allow-Origin: *Access-Control-Allow-Headers: Accept, Accept-Language, Content-Language, Content-TypeConnection: closeContent-Length: 35{    "error": "Bad private key"}
```
This fix will now return the valid response: `{"error": "Bad private key"}`

---

The PR additionally fixes the consistency with locked wallet responses which had two different variants; `Wallet locked` and `Wallet is locked`. The former have been replaced with the latter, and follow the similar style of other messages.

### DIFF
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -562,7 +562,7 @@ void rai::rpc_handler::account_remove ()
 				}
 				else
 				{
-					error_response (response, "Wallet locked");
+					error_response (response, "Wallet is locked");
 					return;
 				}
 			}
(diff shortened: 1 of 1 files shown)


SergiySW:
https://github.com/nanocurrency/raiblocks/pull/730

PR #772, opened 2018-04-02
Possible unsynced nullptr exception

SergiySW:


SergiySW:
Shoundn't be possible with correct database

PR #773, opened 2018-04-02
Reorder lock/transaction to start tx only after successful lock

cryptocode:
I measured tx times, and e.g. gap_cache::vote was open for ~30s, probably due to the mutex being locked.

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -1698,8 +1698,8 @@ void rai::gap_cache::add (MDB_txn * transaction_a, std::shared_ptr<rai::block> b
 
 void rai::gap_cache::vote (std::shared_ptr<rai::vote> vote_a)
 {
-	rai::transaction transaction (node.store.environment, nullptr, false);
 	std::lock_guard<std::mutex> lock (mutex);
+	rai::transaction transaction (node.store.environment, nullptr, false);
 	auto hash (vote_a->block->hash ());
 	auto existing (blocks.get<1> ().find (hash));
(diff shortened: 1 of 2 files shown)


clemahieu:
Did we run this for a while, maybe during a full sync to make sure it doesn't cause deadlocks?

cryptocode:
@clemahieu I ran tests but no sync - I think that kind of testing needs to be done. If moving a transaction (which has no locks) above the lock_guard can result in deadlock, then locking in the node is even more subtle than I thought, but who knows.

PlasmaPower:
Write transactions do block other write transactions.

I think I'll revert the write transaction changes in this PR because I forgot that we sometimes pass transactions into functions, which puts the transaction lock first. That said, the read transactions don't have a potential to deadlock (to my knowledge), and in that case this change is much safer.

cryptocode:
Two machines synced fine now, but I guess better be on the safe side. Having a a design where passing around transactions risks deadlocks doesn't sound good - anything we can improve here?

clemahieu:
We can check it with thread sanitizer, that should detect lock order inversions. 

I was also running on a slow machine and it seemed to freeze but I didn’t have a debugger attached so I don’t know what it was actually doing, it was speculation on my part. 

Read transactions shouldn’t cause an issue though it would be nice to have a discipline than minimizes brittleness.

clemahieu:
That's a good point, seems like it would be worth changing this everywhere, I'm trying to think of a case where this could make something worse but I can't think of one.  We really want all transactions to be open as short as possible.

PR #774, opened 2018-04-02
Add sequence to RPC representatives_online

SergiySW:


rai/node/rpc.cpp
```diff
@@ -2816,9 +2816,11 @@ void rai::rpc_handler::representatives_online ()
 	boost::property_tree::ptree response_l;
 	boost::property_tree::ptree representatives;
 	auto reps (node.online_reps.list ());
+	rai::transaction transaction (node.store.environment, nullptr, false);
```
clemahieu: Would this make more sense on the "representatives" call to get all representatives sequence numbers?

SergiySW: And this only returns list of online reps? Yes, excet we should add to release notes info about changes in response

SergiySW:
Replaced

PR #775, opened 2018-04-02
Add accounts parameter to wallet_change_seed

PlasmaPower:
Right now this only affects the RPC, not the CLI. I'd like to figure out what to do with the CLI option before merging this. Untested (though it compiles).


rai/node/wallet.cpp
```diff
@@ -1311,34 +1311,37 @@ void rai::wallet::init_free_accounts (MDB_txn * transaction_a)
...
 
-rai::public_key rai::wallet::change_seed (MDB_txn * transaction_a, rai::raw_key const & prv_a)
+rai::public_key rai::wallet::change_seed (MDB_txn * transaction_a, rai::raw_key const & prv_a, uint32_t count_a)
 {
 	store.seed_set (transaction_a, prv_a);
 	auto account = deterministic_insert (transaction_a);
-	uint32_t count (0);
-	for (uint32_t i (1), n (64); i < n; ++i)
+	uint32_t count (count_a);
+	if (count != 0)
```
SergiySW: Perhaps if count == 0 ? Then some automatic search

rai/node/wallet.cpp
```diff
@@ -1311,34 +1311,37 @@ void rai::wallet::init_free_accounts (MDB_txn * transaction_a)
...
 }
 
-rai::public_key rai::wallet::change_seed (MDB_txn * transaction_a, rai::raw_key const & prv_a)
+rai::public_key rai::wallet::change_seed (MDB_txn * transaction_a, rai::raw_key const & prv_a, uint32_t count_a)
 {
 	store.seed_set (transaction_a, prv_a);
 	auto account = deterministic_insert (transaction_a);
-	uint32_t count (0);
-	for (uint32_t i (1), n (64); i < n; ++i)
+	uint32_t count (count_a);
```
SergiySW: Probably count can be declared already in
rai::public_key rai::wallet::change_seed (MDB_txn * transaction_a, rai::raw_key const & prv_a, uint32_t count)
It's not constant or shared variable

rai/node/wallet.hpp
```diff
@@ -153,7 +153,7 @@ class wallet : public std::enable_shared_from_this<rai::wallet>
 	void init_free_accounts (MDB_txn *);
 	bool should_generate_state_block (MDB_txn *, rai::block_hash const &);
 	/** Changes the wallet seed and returns the first account */
-	rai::public_key change_seed (MDB_txn * transaction_a, rai::raw_key const & prv_a);
+	rai::public_key change_seed (MDB_txn * transaction_a, rai::raw_key const & prv_a, uint32_t count = 0);
```
SergiySW: perhaps without variable name?
, uint32_t = 0);

SergiySW:
@renesq It's not reverting, just add you option to specify count directly (desktop casual users won't do it anyway)

SergiySW:
@renesq thinking about casual desktop users, they won't specify count & want all their 10... 20... 50 accounts with nano to be restored (instead of clicking "Create account" 50 times). This search wasn't intended to be used by "pro", but if they are too lazy to listen usually

PlasmaPower:
Yeah the default is auto-restore.

SergiySW:
QT: originally
CLI: https://github.com/nanocurrency/raiblocks/pull/637
RPC: this
So now everywhere, huh

At last account with frontier (it doesn't depend on balance) or pending block. With 64+(n/64) accounts check after last restored account.
So if there is gap more than 64+(n/64) between, some account won't be restored.

Actually after https://github.com/nanocurrency/raiblocks/commit/b67feea6876499481fd8e1374392e34a7a010966 & https://github.com/nanocurrency/raiblocks/pull/766 default count to check can be increased to several thousands

cryptocode:
@SergiySW What do you think about adding a `create_work_count` parameter (default 2) instead of hard coding it to 2?

If that makes sense, then maybe change the "accounts" parameter name to "create_count" as well.

SergiySW:
Probably better disable work generation here completely
What about "count" ?

PR #785, opened 2018-04-07
Work generation for state blocks in block_create RPC

cryptocode:


PlasmaPower:
We have some tests for block_create with state blocks. Could you modify them to remove the work parameter, so that gets tested?

cryptocode:
@PlasmaPower I ended up adding a test to deal with this specific case.

PlasmaPower:
LGTM

PR #801, opened 2018-04-14
Error handling using std::error_code and the proposed std::expected type

cryptocode:
Docs and rationale: https://github.com/cryptocode/raiblocks/wiki/Error-handling-in-the-node-using-std::error_code-and-std::expected

This PR applies the above error handling ideas to a handful of RPC handlers. A few functions with the double-response-write problem are fixed as well, using std::error_code.

I've put a few error codes in a common enum, which can be used elsewhere later on.

`decode_unsigned` is used to test `std::expected` which allows both values and errors to be returned.   I think `expected` is a good fit for the node, but it's probably a good idea to introduce it gradually to see how well things compose.

Since expected isn't standardized yet (expected in C++20), I've included `expected-light` (a single-header file) which we'll remove once compilers ships with a std implementation.


rai/node/rpc.cpp
```diff
@@ -2945,126 +3062,120 @@ void rai::rpc_handler::representatives_online ()
...
-		{
-			error_response (response, "Invalid sources number");
-		}
+		ec = error_rpc::invalid_sources_number;
 	}
-	boost::optional<std::string> destinations_text (request.get_optional<std::string> ("destinations"));
-	if (destinations_text.is_initialized ())
+
+	auto destinations (decode_unsigned (request.get<std::string> ("destinations", "0")));
+	if (!ec && !destinations)
```
PlasmaPower: Nit: in an above function (I don't remember which) you don't check `ec` when parsing future values. It doesn't matter that much, but we should probably be consistent.

cryptocode:
>It looks like you didn't transform some RPC functions, like block_create. Was there a reason for this?

Yeah, only a few functions are converted. It's a fair amount of work and I only have so much time :/ (as I do the conversion, I test the RPC calls a lot with various inputs) 

Serg offered to chip in and help with the conversion once the path has been set, which is the point of this PR.

If you prefer this PR to do the full conversion, I'm happy to do that. But it'll take some time for me to get around to that.

> Should we return an error code from the RPC functions?

I was thinking that `process_request` can't do much about them anyway since the error response is already sent by the action handler.

One option would be to let process_request do the error response, but I'm not sure the change is worth it.

> Should we use something similar to the Rust try macro (now an operator) for preconditions, and maybe other places?

That would be great. P0779R0 is a proposal to implement this, which will work with std::expected. Not sure what a good C++ idiom is right now. There are also a monadic extension proposal to allow chaining of expected.

> When std::expected is converted to a boolean, it returns true if it was successful. There's not much we can do about this, but it is the opposite of our previous idiom of errors being truthy.

Since std::expected is not an error, but a value _or_ an error, I think it makes sense.

I see your point, though, and it's going to be a bit painful to read the code base as the old idiom gets slowly replaced. 

> Should we keep the success-first pattern in if statements? Personally I favor putting the error first so it's right next to the failing condition.

I have no strong opinions here. Possibly the no-early-returns principle in the code base affected the decision to place errors at the end, since you're then forced to nest statements  ?

rai/node/rpc.cpp
```diff
@@ -1815,102 +1928,99 @@ class history_visitor : public rai::block_visitor
...
 	bool output_raw (request.get_optional<bool> ("raw") == true);
-	auto error (false);
 	rai::block_hash hash;
 	auto head_str (request.get_optional<std::string> ("head"));
 	rai::transaction transaction (node.store.environment, nullptr, false);
 	if (head_str)
 	{
-		error = hash.decode_hex (*head_str);
-		if (!error)
+		if (hash.decode_hex (*head_str))
```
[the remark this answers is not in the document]
cryptocode: See my main comment about this topic.

rai/node/rpc.cpp
```diff
@@ -1815,102 +1928,99 @@ class history_visitor : public rai::block_visitor
 
 void rai::rpc_handler::account_history ()
 {
+	std::error_code ec;
 	std::string account_text;
+
```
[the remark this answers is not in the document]
cryptocode: Yeah, fixed.

rai/node/rpc.cpp
cryptocode: Agreed, though I don't seem to find the missing check.

rai/node/rpc.cpp
```diff
@@ -1970,56 +2079,100 @@ void rai::rpc_handler::key_expand ()
...
 			{
-				error_response (response, "Invalid count limit");
+				ec = error_rpc::invalid_starting_account;
 			}
 		}
-		uint64_t modified_since (0);
-		boost::optional<std::string> modified_since_text (request.get_optional<std::string> ("modified_since"));
-		if (modified_since_text.is_initialized ())
+		auto count (decode_unsigned (request.get<std::string> ("count", ""), std::numeric_limits<uint64_t>::max ()));
+		if (!count)
```
PlasmaPower: Here's the condition that should have a `!ec` if we're going with that style.

PlasmaPower:
> That would be great. P0779R0 is a proposal to implement this, which will work with std::expected. Not sure what a good C++ idiom is right now.

Should we make our own macro for the time being?

cryptocode:
The proposal requires language changes. A decent C-style macro solution would require statement expressions, which is not supported by MSVC. Unless someone has come up with an alternative trick that I'm not aware of.

rai/node/rpc.cpp
cryptocode: Ah, good catch, I'll add that.

PlasmaPower:
Oh, I forgot that statements aren't expressions in C++ (unlike in Rust 😉). We could probably make a macro that takes the name of a variable and sets the value, but that doesn't seem like a good idea.

rai/node/rpc.cpp
cryptocode: Updated

PR #802, opened 2018-04-14
Prevent log deletion by inactive_node

SergiySW:
if logging.max_size > default value


argakiig:
Formatting then lgtm

SergiySW:
Should pass travis now

PR #803, opened 2018-04-15
Adding optional confirm_req answers

SergiySW:
If tthis is request for forked block/account representative knows about

### DIFF
--- a/rai/ledger.cpp
+++ b/rai/ledger.cpp
@@ -851,8 +851,6 @@ void rai::ledger::change_latest (MDB_txn * transaction_a, rai::account const & a
 
 std::unique_ptr<rai::block> rai::ledger::successor (MDB_txn * transaction_a, rai::block_hash const & block_a)
 {
-	assert (store.account_exists (transaction_a, block_a) || store.block_exists (transaction_a, block_a));
-	assert (store.account_exists (transaction_a, block_a) || latest (transaction_a, account (transaction_a, block_a)) != block_a);
 	rai::block_hash successor;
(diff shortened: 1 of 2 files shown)


PlasmaPower:
`node.ledger.successor` should handle a lot of this logic for you.

SergiySW:
Yeah, testing with this

PlasmaPower:
I think you misunderstood what I meant. `node.ledger.successor` will automatically detect if the root is an account or a block, and return said block. You will no longer have to split up the conditions.

SergiySW:
Hm, ok

SergiySW:
This should be valid now

PlasmaPower:
LGTM

PlasmaPower:
Actually, I think you'll have to revert the last two commits :(. Even if a block's root exists, the block itself (or another successor) may not exist, which causes a crash.

SergiySW:
Hm, there is check for nullptr, it what conditions it can cause crash?

PlasmaPower:
`node.ledger.successor` asserts that it's not nullptr, so it'll crash in debug mode. Maybe you should add a function that's like `node.ledger.successor` but that returns an optional?

PlasmaPower:
LGTM again.

rai/ledger.cpp
```diff
@@ -865,9 +863,7 @@ std::unique_ptr<rai::block> rai::ledger::successor (MDB_txn * transaction_a, rai
 	{
 		successor = store.block_successor (transaction_a, block_a);
```
clemahieu: Successor might not be initialized.

PR #808, opened 2018-04-16
Hardening receive_all

SergiySW:
- check minimum confirmation threshold
- check winner to be required account head
- don't receive possible successors of confirmed account head

### DIFF
--- a/rai/node/wallet.cpp
+++ b/rai/node/wallet.cpp
@@ -1232,9 +1232,9 @@ public:
 							rai::transaction transaction (this_l->wallet->node.store.environment, nullptr, true);
 							this_l->wallet->node.active.start (transaction, block_l, [this_l, account, hash](std::shared_ptr<rai::block> winner_l, bool confirmed) {
 								// If there were any forks for this account they've been rolled back and we can receive anything remaining from this account
-								if (confirmed && winner_l->hash () == hash)
+								if (confirmed && this_l->wallet->node.ledger.block_exists (winner_l->hash ()))
 								{
-									this_l->receive_all (account, hash);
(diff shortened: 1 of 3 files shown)


rai/node/wallet.cpp
```diff
@@ -1228,10 +1228,18 @@ class search_action : public std::enable_shared_from_this<search_action>
...
 						wallet->node.background ([this_l, account, block_l] {
+							rai::block_hash hash (block_l->hash ());
 							rai::transaction transaction (this_l->wallet->node.store.environment, nullptr, true);
-							this_l->wallet->node.active.start (transaction, block_l, [this_l, account](std::shared_ptr<rai::block>, bool) {
+							this_l->wallet->node.active.start (transaction, block_l, [this_l, account, hash](std::shared_ptr<rai::block> winner_l, bool confirmed) {
 								// If there were any forks for this account they've been rolled back and we can receive anything remaining from this account
-								this_l->receive_all (account);
+								if (confirmed && winner_l->hash () == hash)
+								{
+									this_l->receive_all (account, hash);
```
PlasmaPower: Why check `winner_l->hash () == hash`? Why not just check `confirmed`, and then call `this_l->receive_all (account, winner_l->hash ());`? Other than that LGTM.

SergiySW: Yeah, seems better

SergiySW: I guess in rare conditions block_processor probably don't process new winner before callback (as there is queue in block_processor). For this this theoretic possibility I add new extra check for winner existance

SergiySW:
Obsolete

PR #815, opened 2018-04-19
Use default values when getting success/error ptree nodes

cryptocode:
This gives accurate gtest assert errors instead of ptree exceptions when tests fail on those lines.

I've avoided one case since https://github.com/nanocurrency/raiblocks/pull/814 removes that line anyway (and which fixes the failing block_create test)


PlasmaPower:
LGTM

Issue #818, opened 2018-04-20
"account_history" RPC-call inconsistent between state blocks and send blocks


eirikrye:
The `account` field in the history list is the sending account for state blocks, instead of the sendee:

```
$ curl -d '{"action":"account_history", "account":"xrb_14zt6dsi3ooamfmeja63xszfpkk35pinysetcdsp7nrh8t9ksx7emph8wozm", "count": "1"}' [::]:7076
{
    "history": [
        {
            "type": "send",
            "account": "xrb_14zt6dsi3ooamfmeja63xszfpkk35pinysetcdsp7nrh8t9ksx7emph8wozm",
            "amount": "73447999999999999999999999999",
            "hash": "5041CE94BD875AA6B2DABB2051F91226F3CFA6467A2C69A761F6F0CDE3569364"
        }
    ],
    "previous": "2C2BBC9EB935343047869BB3134BFEB75F1C34C3A755EDB308B4B808421C56ED"
}
```

However, this is the actual block, which sends to the account `xrb_3g4oh7h4h51398bemj3fixc44dnij633kd4tdic8nxh51j6gsoefub1figad`:

```
$ curl -d '{"action":"block", "hash":"5041CE94BD875AA6B2DABB2051F91226F3CFA6467A2C69A761F6F0CDE3569364"}' [::]:7076
{
[...]

PR #820, opened 2018-04-21
Fix stateblock receive account field to be consistent with legacy blocks

eirikrye:
#818

### DIFF
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -1753,7 +1753,7 @@ public:
 				{
 					tree.put ("type", "receive");
 				}
-				tree.put ("account", block_a.hashables.account.to_account ());
+				tree.put ("account", handler.node.ledger.account (transaction, block_a.hashables.link).to_account ());
 				tree.put ("amount", (balance - previous_balance).convert_to<std::string> ());
 			}
 		}


PlasmaPower:
This was merged.

eirikrye:
Actually, it was not. The previous (imcomplete) pull request only incorrectly fixed the `account` field for the send subtype. This also fixes it for the receive subtype.

PlasmaPower:
Oh oops, you're right.

argakiig:
LGTM

PlasmaPower:
I'm somewhat worried about this breaking things, but I think that's better than the existing behavior.

PlasmaPower:
If the `raw` option is set, could we add the actual account to the output?

eirikrye:
Like so? I'd prefer if the account is returned, even if `raw` is not set.

PlasmaPower:
LGTM

rai/node/rpc.cpp
```diff
@@ -1809,6 +1809,8 @@ void rai::rpc_handler::account_history ()
 				boost::property_tree::ptree history;
 				if (!error)
 				{
+					auto account (node.ledger.account (transaction, hash));
```
cryptocode: This is a really expensive operation. Seems like you can just use the input account for account queries, and only do the expensive lookup for head queries?

PlasmaPower: It's not expensive if the supplied block is a head block though, right?

cryptocode: It looks through up to 5 tables for the block (block_get_raw), then does a block_info_get. All that instead of just setting the output string to the input string.

PlasmaPower: Fair enough.

eirikrye: The `history_visitor` does this call for every single open, receive and state block already, so the overhead would be negligible for history calls with a somewhat high `count`. I do agree this is a reasonable optimization, though.

My rationale for doing the lookup was because I assumed the passed parameter was case insensitive and therefore this would double as a normalization of the account string, but I see now that this is not the case.

PlasmaPower: Theoretically the account parameter could be encoded using the old v1 account format, but realistically whatever. I'm not too concerned about if we optimize this or not.

eirikrye:
Any reason this was not merged for 13.0?

PlasmaPower:
Sorry, looks like we forgot about this! I'll merge it now so it gets into the next version.

PR #824, opened 2018-04-25
ledger::supply is a hotspot, so we memoize and update periodically

cryptocode:


rai/core_test/ledger.cpp
```diff
@@ -1464,11 +1464,11 @@ TEST (ledger, inactive_supply)
...
-	ledger.inactive_supply = 60;
-	ASSERT_EQ (0, ledger.supply (transaction));
-	ledger.inactive_supply = 0;
-	ASSERT_EQ (50, ledger.supply (transaction));
+	ASSERT_EQ (10, ledger.supply.circulating_get (true));
+	ledger.supply.inactive_set (60);
+	ASSERT_EQ (0, ledger.supply.circulating_get (true));
+	ledger.supply.inactive_set (0);
+	ASSERT_EQ (50, ledger.supply.circulating_get (true));
 }
```
PlasmaPower: Could you add a test here testing the cache? e.g.

```c++
ledger.inactive_supply = 60;
ASSERT_EQ (50, ledger.supply.circulating_get (false));
```

PlasmaPower: To clarify, I think it should be an addition to this test.

cryptocode: Good idea, added a separate test for the supply cache

rai/ledger.cpp
```diff
@@ -495,12 +495,52 @@ bool rai::shared_ptr_block_hash::operator() (std::shared_ptr<rai::block> const &
...
+
+void rai::supply::circulating_update ()
+{
+	std::unique_lock<std::mutex> lock (mutex);
+	update_cache ();
+}
+
+void rai::supply::update_cache ()
+{
+	if (ledger.store.environment)
```
cryptocode: fyi: this is due to a test that tests db failure...

clemahieu: This doesn't need to be checked, it should already be guaranteed.  Supply is a part of ledger and store appears before supply in the class definition so it will be initialized first.

cryptocode: It's because of a check that fails if the check isn't there :( Not sure what the "right way" is.

cryptocode: the test in question tests failure to open the db

PR #829, opened 2018-04-26
Resolve breaking changes in Boost 1.67

devinus:


rai/lib/numbers.cpp
```diff
@@ -443,13 +442,13 @@ rai::uint512_union rai::sign_message (rai::raw_key const & private_key, rai::pub
...
 
-void rai::deterministic_key (rai::uint256_union const & seed_a, uint32_t index_a, rai::uint256_union & prv_a)
+void rai::deterministic_key (rai::uint256_union const & seed_a, uint64_t index_a, rai::uint256_union & prv_a)
 {
 	blake2b_state hash;
 	blake2b_init (&hash, prv_a.bytes.size ());
 	blake2b_update (&hash, seed_a.bytes.data (), seed_a.bytes.size ());
 	rai::uint256_union index (index_a);
-	blake2b_update (&hash, reinterpret_cast<uint8_t *> (&index.dwords[7]), sizeof (uint32_t));
+	blake2b_update (&hash, reinterpret_cast<uint8_t *> (&index.dwords[7]), sizeof (index.dwords[7]));
```
PlasmaPower: Why convert index to a `rai::uint256_t`? Are we changing the endianness? If so, I think we should use boost's functions explicitly for that.

rai/node/wallet.cpp
```diff
@@ -465,7 +465,7 @@ rai::key_type rai::wallet_store::key_type (rai::wallet_value const & value_a)
 	}
 	else
 	{
-		if ((number >> 32).convert_to<uint32_t> () == 1)
+		if ((number >> 64).convert_to<uint64_t> () == 1)
```
PlasmaPower: Are you sure that this change is correct? I think this would require a database upgrade.

PlasmaPower:
Also, there are a couple of small formatting problems causing the Travis error.

rai/node/wallet.cpp
devinus: @PlasmaPower No, the PR is a bit premature. I wanted to run Travis tests against it.

clemahieu:
What about boost 1.67 is causing these changes to be needed?  Some types can’t be widened because they’re part of the current wire protocol like the frontier_req 32bit types.

devinus:
@clemahieu Sorry! The PR was premature and included some things I just wanted tested on Travis, I had hoped nobody would see it before I whittled it down to just the changes actually needed. 😳

devinus:
@PlasmaPower @clemahieu Ready for review.

PlasmaPower:
LGTM

PR #831, opened 2018-04-26
Convert dynamic casts to static casts

cryptocode:
dynamic_cast's as used in block comparisons show up in the profiler in a significant way, and it's probably even worse on msvc (which lowers to a mess of string comparisons)

This converts the dyncasts to type()-check + static_cast, which doesn't show up in the profiler at all.

Dynamic casts are used elsewhere too, but in less perf' critical paths, so they can be dealt with later.


PlasmaPower:
LGTM

PR #839, opened 2018-05-02
Removing private addresses from live network peers

SergiySW:
https://tools.ietf.org/html/rfc1918
https://tools.ietf.org/html/rfc6598
https://tools.ietf.org/html/rfc4193


SergiySW:
Fixing Hetzner abuse messages
Portscan detected from host       x.x.x.x

PR #841, opened 2018-05-03
Rewrite unsynced (fixes bulk_push_client)

PlasmaPower:


rai/node/bootstrap.cpp
```diff
@@ -351,14 +200,22 @@ void rai::frontier_req_client::receive_frontier ()
...
 {
-	auto current (ours_a);
-	while (!current.is_zero () && current != theirs_a)
+	if (bulk_push_cost < bulk_push_cost_limit)
 	{
-		connection->node->store.unsynced_put (transaction_a, current);
-		auto block (connection->node->store.block_get (transaction_a, current));
-		current = block->previous ();
+		connection->attempt->add_bulk_push_target (head, end);
+		if (!connection->node->wallets.exists (transaction_a, current))
```
clemahieu: I think this check can be removed, it seems like it's checking if a block hash exists as an account in any of the wallets which will never be true.

PlasmaPower: Yeah that's a mistake. I'll probably pass in the account as a parameter to check this.

PlasmaPower: Actually this was correct. It was using `current` not `head`. It shouldn't be too important though.

PR #852, opened 2018-05-07
Rebroadcast blocks via process RPC

clemahieu:


PlasmaPower:
I think we could check if the block exists after the `process_receive_many` call to see if there's an error. It'd be nice to tell if the `process` went through even if you don't get the exact error.

SergiySW:
Doing same thing, but without any useful output.
@PlasmaPower even if we check if the block exists after we con't see exact error.

rai/node/rpc.cpp
```diff
@@ -2594,89 +2594,10 @@ void rai::rpc_handler::process ()
...
-					rai::transaction transaction (node.store.environment, nullptr, false);
-					auto account (node.ledger.account (transaction, hash));
-					auto amount (node.ledger.amount (transaction, hash));
-					bool is_state_send (false);
-					if (auto state = dynamic_cast<rai::state_block *> (block_a.get ()))
-					{
-						rai::transaction transaction (node.store.environment, nullptr, false);
-						is_state_send = node.ledger.is_send (transaction, *state);
-					}
-					node.observers.blocks (block_a, account, amount, is_state_send);
```
SergiySW: While it can be just replaced here
node.active.start (block_a);

clemahieu:
My concern with doing it another way other than using block_processor is if there's another change in the future we might miss copying the functionality again.
Checking for block existence right after sending it to the block processor is a race condition since it's processed in a background thread.
The closest way I can think of to retain the functionality is to have a processing callback but we've tried that before and the problem is if it's even put in to unchecked and later pulled out we'll loose the callback.

SergiySW:
I guess now all functionality after
https://github.com/nanocurrency/raiblocks/blob/5e0cb86fa144c4b7193a60c25513217f18586df8/rai/node/node.cpp#L1301
can be moved to rai::block_processor::process_receive_one

clemahieu:
Yea that's a good idea, a lot cleaner too I don't know why they're separate in the first place.

I made a patch to do that change but some tests need to be fixed, could you look at those?

SergiySW:
Sure

PR #854, opened 2018-05-07
Supporting nano_ and xrb_ prefixes.

clemahieu:


renesq:
While the `nano_` prefix is not necessarily bad, I had a different solution in mind:
Opting for addresses like nano1111111111111111111111111111111111111111111111111111hifc8npp 
would
1) keep new addresses at the same even length as legacy addresses, hence improve compatibility across new and old style, especially for internal processing (like plausibility checks and truncating to retrieve public keys. It might also affect state block processing, where the link field could represent either an address or a hash, but i'm not quite sure about this). And also keeping compatibility visually (even legacy addresses are hard to fit on smaller displays already).
2) fix the issue that some browsers stop text selection at the underscore when double-clicking

PR #860, opened 2018-05-10
Fix typo, quorom -> quorum

meltingice:
Noticed a typo not only in the code, but also in the node `config.json`. Not the end of the world, but don't want this to trip someone up later due to inconsistency.

### DIFF
--- a/rai/core_test/rpc.cpp
+++ b/rai/core_test/rpc.cpp
@@ -1599,7 +1599,7 @@ TEST (rpc, work_peer_bad)
 	system.wallet (0)->insert_adhoc (key.prv);
 	rai::rpc rpc (system.service, node1, rai::rpc_config (true));
 	rpc.start ();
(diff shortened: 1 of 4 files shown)


clemahieu:
Hah, terrible.

We’ll have to do a config file upgrade to support the text change.

meltingice:
Oh yeah, good call. Added the config upgrade. My C++ is a little, uh, rusty... so you'll want to double check the code 😅

PlasmaPower:
I think that currently this will reset the value on upgrade (though I'm not sure because I'm on mobile right now). If that's the case, you should read the old value and use that for the new one.

meltingice:
Nah I read the old value during the upgrade and use it for the new JSON key.

meltingice:
@cryptocode good to know, wasn't sure. I saw it being done elsewhere so I just followed convention.

Issue #855, opened 2018-05-08
config to enable work_generate to use work_peers


NiFNi:
Calling `work_generate` makes the node not request its work_peers. This might be wanted for some cases. But most current wallets (if not all) pregenerate work using work_generate. For wallet developers it would be easier to just call one node to reach out to multiple work peers. Right now they would need to implement the complete work process with querying multiple peers, cancel work when one returned a value, etc.. 

I think it would be great to have a config value with which you can switch this behaviour. So for example having `enable_work_generate_peers=true` would result in `work_generate` also requesting the `work_peers`.

NiFNi:
Was implemented as an optional param to `work_generate`.

PR #862, opened 2018-05-11
optional use_peers param to generate_work rpc

NiFNi:
To make the pregeneration of work using multiple work peers easier I added the use_peers option to generate_work. With this light wallets (for example) can easily switch their node to use the multiple work peers configured in the node instead of having to query each work-node.

I added this as rpc parameter instead of a config option as proposed in #855 to not change the RPC behaviour if the config is changed.

I think this PR is pretty interesting for Canoe and Nanovault which both use work_generate to pregenerate the work and thought about adding multiple work peers. Also the default behaviour (not using work peers with that rpc) was pretty unexpected for multiple people I told about this.

### DIFF
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -4228,23 +4228,40 @@ void rai::rpc_handler::work_generate ()
 	if (rpc.config.enable_control)
 	{
 		std::string hash_text (request.get<std::string> ("hash"));
+	    bool use_peers (request.get_optional<bool> ("use_peers") == true);
 		rai::block_hash hash;
 		auto error (hash.decode_hex (hash_text));
 		if (!error)
 		{
 			auto rpc_l (shared_from_this ());
-			node.work.generate (hash, [rpc_l](boost::optional<uint64_t> const & work_a) {
-				if (work_a)
-				{
-					boost::property_tree::ptree response_l;
-					response_l.put ("work", rai::to_string_hex (work_a.value ()));
-					rpc_l->response (response_l);
-				}
-				else
-				{
-					error_response (rpc_l->response, "Cancelled");
-				}
-			});
+			if (!use_peers) {
+				node.work.generate (hash, [rpc_l](boost::optional<uint64_t> const & work_a) {
+					if (work_a)
+					{
+						boost::property_tree::ptree response_l;
+						response_l.put ("work", rai::to_string_hex (work_a.value ()));
+						rpc_l->response (response_l);
+					}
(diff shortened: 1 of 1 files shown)


rai/node/rpc.cpp
```diff
@@ -4228,23 +4228,40 @@ void rai::rpc_handler::work_generate ()
...
+						response_l.put ("work", rai::to_string_hex (work_a.value ()));
+						rpc_l->response (response_l);
+					}
+					else
+					{
+						error_response (rpc_l->response, "Cancelled");
+					}
+				});
+			}
+
```
cryptocode: Formatting issues aside, there seems to be some code duplication going here. I'd recommend putting the inner if/else after the outer which should merely stuff the `to_string_hex (...)` result into variable.

NiFNi: Hm yeah I didn't like to do it that way. But I am not that much into cpp and don't really understand how the functor-like coding works. I mean best would be if you could save this function to some value (callback for example):

```
callback = [rpc_l](boost::optional<uint64_t> const & work_a) {
	if (work_a)
	{
		boost::property_tree::ptree response_l;
		response_l.put ("work", rai::to_string_hex (work_a.value ()));
		rpc_l->response (response_l);
	}
	else
	{
		error_response (rpc_l->response, "Cancelled");
	}
}
```

And then having the function passed to the generate functions:

```
if (!use_peers) {
	node.work.generate (hash, callback);
} else {
	node.generate_work (hash, callback);
}
```

No idea how to do that in cpp though.

cryptocode: Yeah, that'll work, just put `auto` in front of the callback definition and you should be good to go (do test it though).

rai/node/rpc.cpp
```diff
@@ -4228,23 +4228,40 @@ void rai::rpc_handler::work_generate ()
 	if (rpc.config.enable_control)
 	{
 		std::string hash_text (request.get<std::string> ("hash"));
+	    bool use_peers (request.get_optional<bool> ("use_peers") == true);
```
PlasmaPower: I think the preferred way to do this is set the default value to false.

NiFNi: You are maybe right. It's always done like this with optional bools though. So convention seems like this.

rai/node/rpc.cpp
NiFNi: Alright worked exactly like that. Feel free to check it out again.

PR #864, opened 2018-05-12
Statistics pane under Advanced in Qt wallet

cryptocode:
Since we have stats we might as well show them (not everyone runs with RPC on). 

meltingice on Discord and possibly others are looking into graphing these stats on their web frontends, so I won't bother doing any Qt graphs for this.

stats-qt

### DIFF
--- a/rai/qt/qt.cpp
+++ b/rai/qt/qt.cpp
@@ -1,7 +1,10 @@
 #include <rai/qt/qt.hpp>
 
+#include <boost/foreach.hpp>
 #include <boost/property_tree/json_parser.hpp>
-
+#include <boost/property_tree/ptree.hpp>
+#include <cmath>
+#include <iomanip>
 #include <sstream>
 
 namespace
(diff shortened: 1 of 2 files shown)


rai/qt/qt.cpp
```diff
@@ -717,6 +720,91 @@ wallet (wallet_a)
...
+			}
+
+			if (type == "traffic")
+			{
+				std::stringstream numstream;
+
+				const std::vector<std::string> units = { " bytes", " KB", " MB", " GB", " TB", " PB" };
+				double bytes = std::stod (value);
+				auto index = static_cast<size_t> (std::floor (std::log2 (bytes) / 10));
+				if (index < units.size ())
```
PlasmaPower: Shouldn't matter, but why not `std::max (index, units.size() - 1)`?

rai/qt/qt.cpp
```diff
@@ -908,7 +997,7 @@ active_status (*this)
 	client_layout->setSpacing (0);
 	client_layout->setContentsMargins (0, 0, 0, 0);
 	client_window->setLayout (client_layout);
-	client_window->resize (320, 480);
+	client_window->resize (500, 620);
```
PlasmaPower: Will this affect the size of the whole wallet?

cryptocode: Yeah, and I think that's a good thing, unless there's a good reason to keep it tiny.

rai/qt/qt.cpp
cryptocode: Because the result would be a lie so I was thinking just show it raw. But yeah, don't think it matters, won't get into that range unless there's a bug.

PlasmaPower: I think the code there will still work even if it's bigger than 1024 PB.

cryptocode: Sure, I just mean it should flip over to EB at that point to be consistent. But hey, nobody will ever see it - I'll change it ;)

cryptocode: Updated

PlasmaPower:
LGTM

PR #865, opened 2018-05-12
Add DNS resolution support for work peers

PlasmaPower:


rai/node/node.cpp
```diff
@@ -992,12 +992,18 @@ bool rai::node_config::deserialize_json (bool & upgraded_a, boost::property_tree
...
-			uint16_t port;
-			result |= rai::parse_address_port (work_peer, address, port);
+			auto port_position (work_peer.rfind (':'));
+			result |= port_position == -1;
 			if (!result)
 			{
-				work_peers.push_back (std::make_pair (address, port));
+				auto port_str (work_peer.substr (port_position + 1));
+				uint16_t port;
+				result |= parse_port (port_str, port);
```
cryptocode: Minor issue: `parse_port` may throw (due to stoul) and `parse_address_port` handled that. 

I think the fix is to catch inside `parse_port` since it's a bool-error-returning function, that'll fix a similar issue later in this PR.

PlasmaPower: I was wondering about that, but we do call parse_port without a try catch elsewhere in deserialize_json. I think it gets caught by a try catch up the line.

We may need to fix it for the RPC function though. Either way, we should fix it to match the error return code.

cryptocode: Yeah, config deserialization errors tend to fail with information as to what is wrong. The error_code PR will be helpful in fixing that ahem ;)

PlasmaPower:
Updated

rai/node/node.cpp
PlasmaPower: I put a try-catch inside `parse_port`. Now looking through your error code PR.

argakiig:
lgtm, would fix #53 as well

PR #866, opened 2018-05-14
Simplify and improve performance of search_pending.

clemahieu:
Invert loop iteration since the number of accounts in the wallet is likely smaller than the number of pending entries.
Automatic block confirmation already checks if the destination is in a local wallet so we don't need to re-implement this.
Re-searching the pending table for source blocks for the confirmed account can be better implemented in the future with an account dirty flag or queued block insertion.
Since we don't need a callback after confirmation we don't need the search code in its own class.

### DIFF
--- a/rai/node/wallet.cpp
+++ b/rai/node/wallet.cpp
@@ -1186,116 +1186,41 @@ void rai::wallet::work_ensure (MDB_txn * transaction_a, rai::account const & acc
 	}
 }
 
-namespace
-{
-class search_action : public std::enable_shared_from_this<search_action>
-{
-public:
-	search_action (std::shared_ptr<rai::wallet> const & wallet_a, MDB_txn * transaction_a) :
-	wallet (wallet_a)
-	{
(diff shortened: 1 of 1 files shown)


rai/node/wallet.cpp
```diff
@@ -1186,116 +1186,41 @@ void rai::wallet::work_ensure (MDB_txn * transaction_a, rai::account const & acc
...
-					}
-					else
-					{
-						BOOST_LOG (wallet->node.log) << boost::str (boost::format ("Unable to fetch key for: %1%, stopping pending search") % key.account.to_account ());
+						rai::account_info info;
+						auto error (node.store.account_get (transaction, pending.source, info));
+						assert (!error);
+						std::shared_ptr<rai::block> block_l (node.store.block_get (transaction, info.head));
+						node.active.start (block_l);
+						node.network.broadcast_confirm_req (block_l);
```
PlasmaPower: I think it's better to call `node.block_confirm (block_l)` in replacement of active.start + broadcast_confirm_req.

PlasmaPower:
Other than that, LGTM

PR #870, opened 2018-05-14
Change frontier_count to count accounts table vs frontiers table

meltingice:
When calling the frontier_count RPC, it only counts accounts that have been opened via legacy blocks. This is because the frontiers table doesn't include state blocks (via @PlasmaPower). Changing this RPC call to count the accounts table seems to fix the issue.

This bug was originally discovered via Banano. There the RPC returns "1" because only 1 account (genesis) was created via legacy blocks, which is what originally tipped me off about this issue.

I'm not sure what the future of the frontiers table is like, so I leave that up to the core devs here. This simply fixes the count issue.


meltingice:
Sorry - accidentally included other commits. Resubmitting.

PR #871, opened 2018-05-14
Fix frontier_count RPC to include state block opened accounts

### LINKED ISSUE #869: frontier_count doesn't include accounts opened via state blocks
When calling the `frontier_count` RPC, it only counts accounts that have been opened via legacy blocks. This is because the frontiers table doesn't include state blocks (via @PlasmaPower). Changing this RPC call to count the accounts table would fix the issue.

This bug was originally discovered via Banano. There the RPC returns "1" because only 1 account (genesis) was created via legacy blocks, which is what originally tipped me off about this issue.

meltingice:
When calling the frontier_count RPC, it only counts accounts that have been opened via legacy blocks. This is because the frontiers table doesn't include state blocks (via @PlasmaPower). Changing this RPC call to count the accounts table seems to fix the issue.

This bug was originally discovered via Banano. There the RPC returns "1" because only 1 account (genesis) was created via legacy blocks, which is what originally tipped me off about this issue.

I'm not sure what the future of the frontiers table is like, so I leave that up to the core devs here. This simply fixes the count issue.

**frontier_count before**
```json
{
    "count": "502528"
}
```

**frontier_count after**
```json
{
    "count": "502838"
}
```

This fixes #869.

### DIFF
--- a/rai/blockstore.cpp
+++ b/rai/blockstore.cpp
@@ -893,7 +893,7 @@ void rai::block_store::frontier_del (MDB_txn * transaction_a, rai::block_hash co
 	assert (status == 0);
 }
 
-size_t rai::block_store::frontier_count (MDB_txn * transaction_a)
+size_t rai::block_store::account_count (MDB_txn * transaction_a)
 {
 	MDB_stat frontier_stats;
 	auto status (mdb_stat (transaction_a, accounts, &frontier_stats));
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ install:
 - cmd: >-
     SET GIT_COMMIT=%APPVEYOR_REPO_COMMIT:~0,3%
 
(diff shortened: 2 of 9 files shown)


PlasmaPower:
FYI the frontier table doesn't include state blocks both because it's not needed and it prevents old blocks from being built on top of state blocks.

LGTM

rai/blockstore.cpp
```diff
@@ -896,7 +896,7 @@ void rai::block_store::frontier_del (MDB_txn * transaction_a, rai::block_hash co
 size_t rai::block_store::frontier_count (MDB_txn * transaction_a)
```
cryptocode: Since state-block opened accounts are not included in the frontier table, the function should probably be renamed `account_count`.

`TEST (block_store, frontier_count)` => `TEST (block_store, account_count)` as well. 

But leave the RPC action name intact.

PlasmaPower: I'd prefer to keep the function name consistent with the RPC action.

cryptocode: Hm, not sure why the block store should care what RPC actions are (potentially wrongly) called, but I guess in this case there are no other users of the function beyond the test.

PlasmaPower: Oops, I'm on mobile and I thought this function was the rpc service function :P

Yeah renaming it's a good idea.

PlasmaPower:
Merged

PR #892, opened 2018-05-26
Option to force fork resolution in RPC process

SergiySW:


PlasmaPower:
- This should be a parameter for the `process` action.
- This should end any election with the same root.

SergiySW:
Now it ending election with same root

PlasmaPower:
LGTM

PR #894, opened 2018-05-27
Remove unneeded peers.insert after peers.contacted

orhanhenrik:
`insert` is already called from within `peers.contacted`, so there should be no need to call it again.

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -337,7 +337,6 @@ public:
 		}
 		node.stats.inc (rai::stat::type::message, rai::stat::detail::publish, rai::stat::dir::in);
 		node.peers.contacted (sender, message_a.version_using);
-		node.peers.insert (sender, message_a.version_using);
 		node.process_active (message_a.block);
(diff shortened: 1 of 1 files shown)


cryptocode:
@orhanhenrik A subtle point is that `contacted` inserts the IPv4 mapped IPv6 endpoint, while the second `insert` works on the original endpoint. With your patch, you'll only end up inserting the ipv6 endpoint.

This may be ok since we're supposed to be ipv6 only, but it's a change. @PlasmaPower ?

PlasmaPower:
I think it's fine since an IPv6 socket will receive IPv6 mapped socket addresses, at least on POSIX systems. We also assert IPv6 in some places.

PlasmaPower:
Also duplicate of #498

orhanhenrik:
Ouch, I didn't check other PRs before submitting this, sorry!

PlasmaPower:
No problem! I think #498 needed reconsideration anyways.

PR #898, opened 2018-05-30
Disable compute_rep_votes in config

SergiySW:
if enable_voting set to false, wallet shouldn't compute its account votes


PlasmaPower:
We call foreach_representative in other places right? I thought it would not call the callback if enable_voting was false.

SergiySW:
Only in 2 places, also here
https://github.com/nanocurrency/raiblocks/blob/0e46057d0ba5db7c0aef4cb3c914176926ad7ee0/rai/node/node.cpp#L145

PlasmaPower:
LGTM then

PR #904, opened 2018-06-04
RPC confirmation_active, confirmation_info

SergiySW:
confirmation_active: roots of active elections
confirmation_info: election info by "root" hash. Optional "representatives": representatives list with weight


SergiySW:
RPCs names are drafts, may be there are better options

rkeene:
What do you think about naming these "_confirmation_active" and "_confirmation_info" respectively, to indicate the level of guarantee of API stability we offer over these ?

rkeene:
We'll still try to get this in V16.0 as it is low risk, high reward.

PR #915, opened 2018-06-11
CLI unchecked_clear

SergiySW:
CLI "unchecked_clear" and optional unchecked_clear for "vacuum" & "snapshot"


argakiig:
LGTM

PlasmaPower:
This only works with `--vacuum` or `--snapshot` right? Why?

PlasmaPower:
Also, if we're going to keep it that way, we should warn/error if it's used without those options.

SergiySW:
No, it also works separate

PlasmaPower:
Oh, yeah I missed that.

PR #920, opened 2018-06-15
CLI validate blocks

SergiySW:


rkeene:
Does this support epoch blocks and the merged iterators ?

PR #924, opened 2018-06-18
Support lower end devices

schenkty:
128GB mapsize is sufficient to maintain node performance and allow lower end devices to operate. The existing 1TB mapsize requirement that is present in the current release prevents widespread node adoption.

### DIFF
--- a/rai/node/utility.cpp
+++ b/rai/node/utility.cpp
@@ -68,7 +68,7 @@ rai::mdb_env::mdb_env (bool & error_a, boost::filesystem::path const & path_a, i
 			assert (status1 == 0);
 			auto status2 (mdb_env_set_maxdbs (environment, max_dbs));
 			assert (status2 == 0);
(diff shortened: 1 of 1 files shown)


PlasmaPower:
The one potential problem is that I think some people's DBs are larger than 128GB due to bloating issues. I guess we'll just have to require they vacuum the DB before upgrading.

PlasmaPower:
@schenkty is now testing 256GB

argakiig:
I had clamped it down in my initial wallet ledger separation as well. When Colin redid the branch he set it to 64gb for ledger and 1gb for wallet. which from my testing worked fine

schenkty:
I just test 256GB and it works great. I did test 64GB and found it to crash but I have not found any reason for it to crash so I am going to attempt it again with a fresh build. I recommend that the lower the mapsize the better but it is understandable for the reasons that @PlasmaPower pointed out that it not be too low and prevent growth in the future. I believe 64GB would work for the most

schenkty:
Just tested 64GB for a bit and it works fine

PlasmaPower:
Ok I think 128GB is a good compromise then. We can tune it further with the wallet separation.

PR #929, opened 2018-06-23
Set permissions on backup files 

cryptocode:
So only owner can read and update the backup. https://github.com/nanocurrency/raiblocks/issues/928.


alexbakker:
The permissions should be set before anything is written to the file. Otherwise someone could theoretically read the file right after it's closed.

cryptocode:
@alexbakker Agreed, I'll push an update.

PR #946, opened 2018-07-04
node_id for RPC & QT wallet peers list

SergiySW:


PlasmaPower:
This is a breaking change for the peers RPC endpoint, right?

SergiySW:
If anyone is using version field there

SergiySW:
Now also with optinal "deprecated" field to return old style answer if anyone need this

SergiySW:
About +140% for size & about +100% for time (but still like 0.003-0.005 sec)

PR #947, opened 2018-07-05
IPC server and protobuf based API framework

cryptocode:
I've started documenting this over at https://nanoapi.github.io/. There's a chapter regarding the node implementation which should be useful when reviewing this.

The PR is mostly about the IPC/API *framework* - only a handful of actual APIs are added, just to get started. Going forward, people can submit PR's to extend the API, hopefully generating good discussions on how to best create focused and composable messages.

There are currently clients for C, C++, Go and Python, as well as a REST server. Rust, Node and JVM support are in the pipeline:
https://github.com/nanoapi/

If this PR is accepted, I can hand over the nanoapi org to core, or the repositories can be moved into the official repo (though a separate repo may make it easier to have contributors for client API stuff?)

Another git submodule is added for protobuf, as this seems to be the only robust way to integrate it with CMake on all OS's, so be sure to run `git submodule update --init --recursive` before building.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,3 +14,6 @@
 [submodule "phc-winner-argon2"]
 	path = phc-winner-argon2
 	url = https://github.com/clemahieu/phc-winner-argon2.git
+[submodule "protobuf-lib"]
+	path = protobuf-lib
+	url = https://github.com/google/protobuf.git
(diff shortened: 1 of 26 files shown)


cryptocode:
https://github.com/nanocurrency/raiblocks/pull/1434 is an alternative. Protobuf can be added on top of that one later if there's still demand.

cryptocode:
Closing in favor of https://github.com/nanocurrency/raiblocks/pull/1434 (and the external protobuf project mentioned there)

PR #948, opened 2018-07-05
Node ID RPC

SergiySW:
RPC "node_id"
RPC "node_id_delete"


rai/node/rpc.cpp
```diff
@@ -1886,6 +1886,32 @@ void rai::rpc_handler::mrai_to_raw (rai::uint128_t ratio)
...
+		rai::transaction transaction (node.store.environment, nullptr, false);
+		rai::keypair node_id (node.store.get_node_id (transaction));
+		response_l.put ("private", node_id.prv.data.to_string ());
+		response_l.put ("public", node_id.pub.to_string ());
+		response_l.put ("as_account", node_id.pub.to_account ());
+	}
+	response_errors ();
+}
+
+void rai::rpc_handler::node_id_delete ()
```
rkeene: Does deleting this cause any problems at runtime ?  If we add this as an interface that we guarantee to support, are we going to have a bad time ?

rai/node/rpc.cpp
```diff
@@ -1895,6 +1895,30 @@ void rai::rpc_handler::mrai_to_raw (rai::uint128_t ratio)
...
+{
+	rpc_control_impl ();
+	if (!ec)
+	{
+		rai::transaction transaction (node.store.environment, nullptr, true);
```
rkeene: The transaction constructor changed

PR #949, opened 2018-07-05
Allow unsafe RPC commands for TLS connections

SergiySW:
RPC "wallet_list", "wallet_seed"


PlasmaPower:
Is TLS client certificate pinning implemented and required for TLS in general?

cryptocode:
Feels like enabling these dangerous commands is orthogonal to TLS. Maybe you're running everything local, or maybe you use plain http over a secure tunnel. So perhaps a new config option, "enable_wallet_control" or something like that, is what's needed, where the node operator accepts the risks involved?

PlasmaPower:
It's still dangerous IMO. People will just Google how to use the RPC to list wallets and just enable that config option, without taking steps to secure the RPC.

cryptocode:
How about requiring these commands to be signed? The node operator would have to use the CLI to generate the keypair for this purpose.

PlasmaPower:
How would a key exchange provide authentication though? I think we should use asymetric encryption. The node's config has the public key, and the
RPC caller must have the private key to decrypt the response. We could definitely provide a command line option to generate these, or just have an openssl command to do so in the documentation.

SergiySW:
@PlasmaPower have to agree current PR is not secure yet

@cryptocode what about? should be safe with check (config.secure.client_certs_path.size () > 0)
https://github.com/nanocurrency/raiblocks/blob/8840c9dbefb72aef255744fd5157695201511195/rai/node/rpc_secure.cpp#L83

PlasmaPower:
That would work. A badly written client would still be vulnerable to a MITM, but it's a lot better than just exposing the wallet list.

cryptocode:
@PlasmaPower I meant the request would be signed for auth + key exchange if we wanted symmetric encryption (there's an extension that allows the same keypair to be used for both, but we probably don't want a fork of ed25519-donna)

cryptocode:
@SergiySW that's already in place, right? That's not what I think about as pinning, maybe I misunderstood @PlasmaPower

SergiySW:
@cryptocode yes it's there

Enforced

cryptocode:
I think this needs to be coordinated with the effort of taking RPC out of process, which would also remove the TLS stuff (in that case, maybe kick this to v19?). If these requests were encrypted or signed, it could work over any ipc mechanism.

devinus:
My thoughts: I feel it `enable_control` is enabled, then at least `wallet_seed` should be exposed regardless of whether TLS is enabled or not.

What can you do with the seed that you couldn't do anyway if the wallet is unlocked?

renesq:
@devinus I think we indeed need to distinct those cases: If a man in the middle logs the wallet_id but RPC is only listening to one specific remote host, he can't do any harm, whereas snatching a seed off a HTTP connection would enable him to raid the accounts.

cryptocode:
Related: https://github.com/nanocurrency/nano-node/pull/1851 enabled wallet_seed over ipc

cryptocode:
Closing as this is already support via IPC, and if there's a need we can add the same feature to RPC2 with access control.

PR #950, opened 2018-07-06
 Allow OpenSSL 1.1

SergiySW:
Tested with Debian Stretch OpenSSL 1.1.0f


PlasmaPower:
Does this also work with OpenSSL 1.0?

cryptocode:
The EXACT flag makes sure the LTS (1.0.2x) version of openssl is pulled in if multiple versions are installed, but that may be too conservative if this is a backwards compatible change.

SergiySW:
@PlasmaPower at least working with OpenSSL 1.0.2g (Ubuntu 16.04 LTS)

SergiySW:
1.1.1 also should become LTS
https://www.openssl.org/blog/blog/2018/05/18/new-lts/

PR #955, opened 2018-07-07
Add epoch blocks to disable old type blocks

PlasmaPower:
This PR adds epoch blocks, which are a variant of state blocks. They will be used to prevent future old type (non-state) blocks. This code may also be used for future upgrades.

How to detect if a state block is a epoch block:
- Its link field is epoch_link (currently `"epoch v1 block"` then a bunch of null bytes)
- Its balance is the same as the previous block for the account (or 0 if previous is 0)

Additional requirements of an epoch block (for more details see `rai::ledger_processor::epoch_block_impl`):
- Its account must not be the burn account
- It is signed by epoch_signer (currently the genesis account) instead of its account
- Its representative is the same as the previous block for the account (or the burn account if previous is 0)
- Its account's version must be 0

An epoch block will change the account version from 0 to 1, which means:
- It cannot be followed by old type blocks (like how state blocks already work)
- Any future sends will have a min_version of 1
  - These cannot be received by old type blocks
  - When received by a state block, this automatically upgrades the account to version 1

We'll make a script to generate epoch blocks for any currently open accounts and any unopened account with pending sends (except the burn account).

### DIFF
--- a/rai/common.hpp
+++ b/rai/common.hpp
@@ -110,13 +110,14 @@ std::unique_ptr<rai::block> deserialize_block (MDB_val const &);
 /**
  * Latest information about an account
  */
+#pragma pack(push, 1)
 class account_info
 {
 public:
 	account_info ();
 	account_info (MDB_val const &);
 	account_info (rai::account_info const &) = default;
-	account_info (rai::block_hash const &, rai::block_hash const &, rai::block_hash const &, rai::amount const &, uint64_t, uint64_t);
+	account_info (rai::block_hash const &, rai::block_hash const &, rai::block_hash const &, rai::amount const &, uint64_t, uint64_t, uint8_t);
 	void serialize (rai::stream &) const;
 	bool deserialize (rai::stream &);
 	bool operator== (rai::account_info const &) const;
@@ -129,24 +130,29 @@ public:
 	/** Seconds since posix epoch */
 	uint64_t modified;
 	uint64_t block_count;
+	uint8_t version;
 };
+#pragma pack(pop)
 
 /**
  * Information on an uncollected send
  */
+#pragma pack(push, 1)
 class pending_info
 {
 public:
 	pending_info ();
 	pending_info (MDB_val const &);
-	pending_info (rai::account const &, rai::amount const &);
+	pending_info (rai::account const &, rai::amount const &, uint8_t);
 	void serialize (rai::stream &) const;
 	bool deserialize (rai::stream &);
 	bool operator== (rai::pending_info const &) const;
 	rai::mdb_val val () const;
 	rai::account source;
 	rai::amount amount;
+	uint8_t min_version;
 };
+#pragma pack(pop)
 class pending_key
 {
 public:
--- a/rai/common.hpp
+++ b/rai/common.hpp
@@ -228,7 +228,7 @@ enum class process_result
 	old, // Already seen and was valid
 	negative_spend, // Malicious attempt to spend a negative amount
 	fork, // Malicious fork based on previous
-	unreceivable, // Source block doesn't exist or has already been received
+	unreceivable, // Source block doesn't exist, has already been received, or requires an account upgrade (epoch blocks)
 	gap_previous, // Block marked as previous is unknown
 	gap_source, // Block marked as source is unknown
 	opened_burn_account, // The impossible happened, someone found the private key associated with the public key '0'.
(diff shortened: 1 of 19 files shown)


rai/common.hpp
```diff
@@ -110,13 +110,14 @@ std::unique_ptr<rai::block> deserialize_block (MDB_val const &);
 /**
  * Latest information about an account
  */
+#pragma pack(push, 1)
```
cryptocode: It may not be a problem in this case, but this does mean we need to worry about unaligned access as we maintain these classes (i.e. no atomics in the future in these structs, no assignment from ptr-to-member, and for ARM, slow code for unaligned access may be generated). 

An alternative is maybe to pack manually instead of std::copy'ing the struct pointer.

PlasmaPower:
I think we can just use active graph for future epochs (so we wouldn't need epoch blocks at all).

rai/ledger.cpp
```diff
@@ -284,6 +301,69 @@ void ledger_processor::state_block_impl (rai::state_block const & block_a)
...
+								auto last_rep_block (ledger.store.block_get (transaction, info.rep_block));
+								assert (last_rep_block != nullptr);
+								result.code = block_a.hashables.representative == last_rep_block->representative () ? rai::process_result::progress : rai::process_result::representative_mismatch;
+							}
+						}
+					}
+				}
+				else
+				{
+					result.code = block_a.hashables.representative.is_zero () ? rai::process_result::progress : rai::process_result::representative_mismatch;
```
SergiySW: I guess here should be comparision with representative of previous block?

gr0vity:
With the merging of this PR, the genesis account now is able to sign blocks on my chain.
In my eyes this is a major change that needs a lot of thought.

Prior to this PullRequest, **my private key is the only key that can modify my own chain**.
This is a sound concept. I am the only owner of my chain and the only one who can change my chain.
**With epoch blocks this changes!**

After this PullRequest gets merged, **the genesis account now is able to sign (epoch) blocks on my behalf.**
Doesn't this go against decentralisation ? Can't this be exploitet as an additional attack vector ? What if the genesis private key gets lost or stolen ? 
Who tells me that in the future epoch blocks will not be able to modify my representative, my balance, ... ?

**Please explain a little more, why this change is not dangerous now and in the future.**
Thanks

clemahieu:
As a starting point, in the description it says the exactly one condition this block can be inserted.  Most specifically it cannot involve a balance transfer in any form; the only account state it changes is the account’s version. 

The reason we’re looking at these markers is as a way to upgrade the ledger, since all accounts are asynchronous we need an asynchronous form of chain upgrades.  We’re not able to say “upgrade at block X as in bitcoin”

If you have a way to mark cold-wallet or lost-key accounts I’m open to suggestions. There isn’t a way to determine the difference between these so this seemed like a good solution.

gr0vity:
I am okay with upgrading the ledger.
I am okay with accepting the epoch block myself.

What I dislike if the fact that a different account (genesis account) can alter my chain.
Right now the conditions for epoch blocks to be signed are _balance and rep must be the same_ .
**But you made these rules**.. and you could change these rules. 
Who prevents these rules from changing? Now that the genesis can sign blocks on my behalf... If some day these rules change, well I see a problem. I don't say they will. But they could... **Or can't they change ?**

So can we say that epoch blocks can never ever be used to alter my account balance and representative ?
And can we say that there is no bug in the implementation that could lead to a balance or rep change ?
If so, it's probably the best solution. Just saying.
Meybe it's only me, but **I see this is a major change of concept** which needs careful thought and review.

PlasmaPower:
If the majority of reps don't move, the new chain won't work, because nothing will reach quorum.

clemahieu:
At the beginning of the year we already demonstrated we won't alter account balances or representatives, this wasn't just words it was actions.

Reps already need to be upgraded because, as I already said, epoch blocks need to be explicitly programmed in to the node beforehand otherwise an old node will reject them as invalid.

Every code commit for every piece of software in the world is a possible rule change, that's how code works. That's also why this project is open source so everyone can verify what's being changed and can make sure it's in line with the spirit of the project.

This upgrade needs to be done and I'm not interested in dragging it out to pontificate on future scenarios that aren't going to happen.  Projects that can't evolve will die and we're going to evolve the Nano protocol to survive and serve its purpose as a store and transfer of value.

If you see a problem with this actual code in this commit let me know.

PR #956, opened 2018-07-07
Batch state blocks signatures verification

SergiySW:


clemahieu:
Very nice!  I'll take a look in a bit, if anyone else can too that'd be good.

rai/lib/numbers.cpp
```diff
@@ -428,6 +428,23 @@ bool rai::validate_message (rai::public_key const & public_key, rai::uint256_uni
...
+	size_t messages_lengths[batch_count];
+	const unsigned char * messages_pointers[batch_count];
+	const unsigned char * public_keys_pointers[batch_count];
+	const unsigned char * signatures_pointers[batch_count];
+	for (auto i (0); i < batch_count; i++)
+	{
+		messages_pointers[i] = messages[i].bytes.data ();
+		messages_lengths[i] = sizeof (messages[i].bytes);
+		public_keys_pointers[i] = public_keys[i].bytes.data ();
+		signatures_pointers[i] = signatures[i].bytes.data ();
```
PlasmaPower: I'd prefer to prevent an additional copy here. Could we change the function signature to require C style arrays of arrays of bytes?

SergiySW: Sure we can

PlasmaPower:
This PR is exciting! I'm worried about messing with signature verification, as that's a very critical piece of code, but this looks good to me.

Does this speed up bootstrapping?

rai/lib/numbers.cpp
PlasmaPower: While you're at it, where it's called from, instead of copying the e.g. signature bytes we could just make an array of pointers to them.

SergiySW:
While it's increasing verification speed up to 40-50% for large amounts of state blocks (I forget to add old test --debug_verify_profile_batch, thx for reminder), currently only 14.4% of blocks are state. So it won't increase current bootstrap speed significantly

clemahieu:
What about this process:
- In addition to block_processor::blocks there is block_processor::state_blocks
- block_processor::add puts state blocks in to the state block queue
- block_processor::process_blocks first runs signature checks on everything in the state_blocks deque.  Anything that fails a signature check is dropped.  Anything that succeeds is put in to the blocks deque.  The blocks deque is then processed as normal
- Signature checking in removed for state blocks in the ledger_processor
- A test is added to prove that incorrectly signed state blocks are rejected.

SergiySW:
Seems valid process

SergiySW:
Still I think that signature verification shouldn't be removed from ledger processor by default because other functions can use it except block_processor::process_receive_many

argakiig:
MSVC conforms to C90 standards and does not support variable length arrays (C99) which is what is causing the appveyor failures

SergiySW:
Have to disable batch signatures for MSVC builds

rai/node/node.cpp
```diff
@@ -1479,17 +1489,61 @@ bool rai::block_processor::should_log ()
...
+	return !blocks.empty () || !forced.empty () || !state_blocks.empty ();
+}
+
+void rai::block_processor::verify_state_blocks (std::unique_lock<std::mutex> & lock_a)
+{
+	lock_a.lock ();
+	std::deque<std::pair<std::shared_ptr<rai::block>, std::chrono::steady_clock::time_point>> items;
+	items.swap (state_blocks);
+	lock_a.unlock ();
+	auto size (items.size ());
```
cryptocode: `items.size ()` needs to be behind the lock I think, otherwise it's undefined behavior.

clemahieu: Items is the local variable that's been swapped for the shared container, it should be safe.

cryptocode: Yep, read it as the member variable :(

rai/node/node.cpp
```diff
@@ -1293,15 +1293,21 @@ bool rai::block_processor::full ()
...
+void rai::block_processor::add (std::shared_ptr<rai::block> block_a, std::chrono::steady_clock::time_point origination, bool unchecked)
 {
 	if (!rai::work_validate (block_a->root (), block_a->block_work ()))
 	{
 		std::lock_guard<std::mutex> lock (mutex);
 		if (blocks_hashes.find (block_a->hash ()) == blocks_hashes.end ())
 		{
-			blocks.push_back (std::make_pair (block_a, origination));
-			blocks_hashes.insert (block_a->hash ());
+			if (!unchecked && block_a->type () == rai::block_type::state && block_a->link () != node.ledger.epoch_link)
```
PlasmaPower: I believe the `unchecked` parameter is named incorrectly. Shouldn't it be called `checked`? If it's true, it bypasses signature validation.

PlasmaPower:
Other than that LGTM

rai/node/node.cpp
clemahieu: Can we construct a test on that?

SergiySW: Well, probably require renaming. Parameter for state blocks from unchecked table. As all state blocks in unchecked already was validated

PlasmaPower: I see. Is the performance improvement worth the risk?

SergiySW: I guess yes. Rare risks are signature corruption on disk in unchecked table or bit flips. Same risk for state blocks table & usual signature check.
But this can be moved to separate PR

PlasmaPower: In terms of risk I was thinking we might accidentally put unvalidated blocks in the unchecked table.

PlasmaPower: Though I think I've brought this up before. Either way should be fine.

clemahieu: Let’s go with the safer option. Can we make a test that has a block with a bad signature in unchecked and make sure it isn’t processed?

SergiySW: Unchecked thing removed. I guess we can return to this after 17.0, now better polish what we have for new release

rai/lib/blocks.cpp
```diff
@@ -1440,6 +1460,11 @@ rai::block_hash rai::receive_block::root () const
 	return hashables.previous;
 }
 
+rai::block_hash rai::receive_block::link () const
+{
+	return 0;
```
rkeene: Should we `assert(false)` here (or just not implement this virtual function for non-state blocks) ?  The receive block probably has the equivalent of a link

SergiySW: That's similar to source (), previous () & some other functions returning 0 for some block types. We can just not implement such functions when they return 0 I guess to save some lines in code

clemahieu: Is this a case of using release_assert since we shouldn’t be looking for the link of non-state blocks?

rai/node/node.cpp
```diff
@@ -1010,16 +1016,61 @@ bool rai::block_processor::should_log ()
...
+	{
+		auto & block (static_cast<rai::state_block &> (*items[i].first));
+		hashes.push_back (block.hash ());
+		messages.push_back (hashes.back ().bytes.data ());
+		lengths.push_back (sizeof (decltype (hashes)::value_type));
+		pub_keys.push_back (block.hashables.account.bytes.data ());
+		signatures.push_back (block.signature.bytes.data ());
+	}
+	auto code (rai::validate_message_batch (messages.data (), lengths.data (), pub_keys.data (), signatures.data (), size, verifications.data ()));
+	(void)code;
```
rkeene: Let's add some comments in here about how it populates `verifications` and why `code`'s value is ignored

rai/node/node.cpp
```diff
@@ -1051,7 +1102,7 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> &
 					node.ledger.rollback (transaction, successor->hash ());
 				}
 			}
-			auto process_result (process_receive_one (transaction, block.first, block.second));
+			auto process_result (process_receive_one (transaction, block.first, block.second, !force));
```
rkeene: Using `!force` here is not clear that it's directly related to state blocks... can we do something more clear ?

SergiySW: Forced state blocks are not validated in verify_state_blocks () function
Because of that we should set set validated_state_block as "false" for forced blocks (!force)

rai/node/node.cpp
SergiySW: Verifications is vector if signatures check results
validate_message_batch returing "true" if there are at least 1 invalid signature (not very useful here)

rai/node/node.cpp
rkeene: But `!force` will evaluate to true when `force` is false which can happen for things other than state blocks (or verified blocks ?)

rai/lib/blocks.cpp
rkeene: Either release_assert() or not declaring the virtual methods for these block types which should generate a fatal exception

rai/node/node.cpp
SergiySW: Yes, but only state blocks vefirication can be skipped by flag in ledger.processor

rkeene:
Awesome work !

PR #957, opened 2018-07-07
Speed up travis builds: enable ccache, cache cargo and reduce test timeouts

arugulaBowl:
These changes make the compilation on travis faster, with ~36 minutes as reference, it takes about ~20 minutes with cold caches and ~10 minutes with warm caches.

Summary of changes:
 * Enable build caches: ccache and cargo, this requires two slightly different approaches because the linux build runs inside a docker container. The solution on osx is just enabling the caches as explained in the travis documentation, the solution for linux maps a few folders from the docker container to the host to be able to store the cache. ccache is enabled through a cmake 3.4 feature, thus the change in the required version.
 * Enable homebrew cache: it does not avoid the compilation but speeds up the download.
 * Reduce tests timeout from 420 to 120. One of the tests hangs and was slowing down the build process.
 * A minor improvement to avoid docker waiting for password if the travis project is not configured with the docker password.

This is my first contribution so please don't hesitate to comment if I need to do something differently. I am planning to contribute a general improvement on the cmake code but it is taking longer than expected, soon...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,16 +17,23 @@ matrix:
 compiler:
   - clang
 
+cache:
+  - ccache: true
+
 before_install:
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo mkdir -p /etc/docker && echo '{"ipv6":true,"fixed-cidr-v6":"2001:db8:1::/64"}' | sudo tee /etc/docker/daemon.json && sudo service docker restart; fi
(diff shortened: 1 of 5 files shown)


arugulaBowl:
Some timings:

master:


this branch with cold caches:

this branch with warm caches:

arugulaBowl:
screenshots with timings:

master:
master

cold cache:
cold


warm cache:
warm

PlasmaPower:
Thanks!

PR #958, opened 2018-07-07
Update boost finder to version 1.67

arugulaBowl:
* Removed BOOST_CUSTOM option in favor of BOOST_ROOT.
* Only append to CMAKE_MODULE_PATH if BOOST_ROOT is defined.
* The change in appveyor.yml is needed because otherwise the automatic compiler detection thinks that the compiler is vc71.


arugulaBowl:
Already merged with #965

PR #964, opened 2018-07-10
Fix max vote sequence comparison

### LINKED ISSUE #962: [bug]is the judgement true:max_vote->sequence > vote_a->sequence? then it may be a replay, not a vote
https://github.com/nanocurrency/raiblocks/blob/ef86a5e4c5d5d8225c648934ef1abc6033c418f2/rai/node/node.cpp#L1224

PlasmaPower:
Fixes #962. I'm not 100% sure about this so I'll have to think about it a bit. This shouldn't affect much.


SergiySW:
Both options invalid.
max_vote->sequesnce should be == vote_a->sequence, then it's new vote
because of
https://github.com/nanocurrency/raiblocks/blob/master/rai/secure/blockstore.cpp#L1345

PR #965, opened 2018-07-10
Modernize and clean up cmake code

arugulaBowl:
This PR is modernizing the cmake code, reducing duplication and improving separation between libraries and maintainability.

These changes are aligned with the advice in guides like https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1.

There are a few changes that require discussion:
 * A few headers and cpp files were stored in the root folder, at first I moved them to the secure lib, and moved them to better places as I understood the code better.
 * rai_lib is not compiled twice, I thing this is fine but I may have misunderstood the original intention
 * A change was needed in lmdb to fix compilation on windows. For the moment this branch is pointing to a fork I made with the fix. I can create a PR to @clemahieu's fork of the library too
 * This branch requires the changes in https://github.com/nanocurrency/raiblocks/pull/958 because it needs the boost targets with the correct library dependencies and include diretories.

There is more work to be done but this PR is already too big and they could be done later, for example:
 * Continue moving library declarations out of the main cmakelist
 * Some defines are still declared in too many places, like -DQT_NO_KEYWORDS and -DBOOST_ASIO_HAS_STD_ARRAY=1
 * Review or limit the scope of the definition of WIN32_LEAN_AND_MEAN, it creates warnings in some places
 * Review if POSITION_INDEPENDENT_CODE is needed, nothing is compiled as a shared library so it is maybe not needed

Please test the branch, check what makes sense and what not or what may conflict with other planned changes. I have limited time but I will try to keep the PR updated and respond to the review.


arugulaBowl:
There is an error running the tests in travis: "Error: failed to spawn rai_node". I will investigate tomorrow.

PlasmaPower:
So should we merge #958 first or does this include its commits?

cryptocode:
I wonder in what ways this breaks my IPC/API PR, which has new CMake stuff...

clemahieu:
I’ll take a look at the separate commits just to make sure each one is a small discrete patch. We don’t want to squash these when committing so we can preserve the more understandable history instead of one big diff

arugulaBowl:
@PlasmaPower This branch includes #958 in the first commit, #958 can be merged separately and this branch rebased or merge only this one (if approved), as you prefer.

@cryptocode This PR should tidy things up and make the cmake code easier to maintain, I had a look at the IPC/API branch and I don't see any blocker. For sure it will take some work to rebase although it should not be too much. I can help too. Looking at that branch I see that the API is in rai_lib, is that why rai_lib was compile twice as shared and static? Are there clients of it already? If that is the case it it should be added back in this branch.

@clemahieu I tried to keep the history meaningful but in some places I just had to remove (or comment out) a lot of code to add it back as needed to fix compilation.

cryptocode:
@arugulaBowl Thanks for taking a look. The new API stuff is in `rai_protobuf` (new library) and `node`, and shouldn't be affected by any changes you've made to rai_lib. There may be existing clients to rai_lib, although those might have moved to the Rust version @PlasmaPower made (?)

CMakeLists.txt
```diff
@@ -5,11 +5,13 @@ set (CPACK_PACKAGE_VERSION_MAJOR "15")
...
 set (CPACK_PACKAGE_VERSION_PATCH "0")
 if (DEFINED GIT_COMMIT)
-set (CPACK_PACKAGE_VERSION_PATCH "GIT-${GIT_COMMIT}")
+	set (CPACK_PACKAGE_VERSION_PATCH "GIT-${GIT_COMMIT}")
 endif (DEFINED GIT_COMMIT)
 set (CPACK_PACKAGE_VENDOR "Nano Currency")
 
 set(CMAKE_INSTALL_RPATH "@executable_path/../Frameworks")
+# Create all libraries and executables in the root binary dir
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
```
cryptocode: Isn't having output directories for Debug/Release better than polluting the root dir ? An output directory makes it a tad easier/safer to wipe/move/copy build artifacts IMO. Is there an advantage?

arugulaBowl: This was needed to fix the tests, which expect rai_node to be in CMAKE_BINARY_DIR. I thought of building the install target and running the tests from the install/bin folder but the install target does not include the tests, so half of the executables come from there and half from the build folder. Instead of adding the tests to the install target I found this to be simpler way to keep the path of the executables as in master.

For any real distribution the install target is more appropriate but for the tests I am not sure.

arugulaBowl:
There is an issue with the formatting of plat/osx/working.mm, any idea?

>   Configuration file(s) do(es) not support Objective-C: /home/travis/build/nanocurrency/raiblocks/.clang-format
error: `clang-format -lines=1:15 rai/secure/plat/osx/working.mm` failed
Configuration file(s) do(es) not support Objective-C: /home/travis/build/nanocurrency/raiblocks/.clang-format
error: `clang-format -lines=1:15 rai/secure/plat/osx/working.mm` failed
Code formatting differs from expected - please run ci/clang-format-all.sh
The command "if [ -n "$ONE_TIME_TESTS" ]; then ci/check-commit-format.sh; fi" exited with 1.

argakiig:
There were a couple reasons due to type conversions that we stayed at 1.66

.gitmodules
```diff
@@ -9,7 +9,7 @@
 	url = https://github.com/weidai11/cryptopp.git
 [submodule "lmdb"]
 	path = lmdb
-	url = https://github.com/clemahieu/lmdb.git
+	url = https://github.com/arugulaBowl/lmdb.git
```
SergiySW: Seems like same LMDB (?)

PlasmaPower: Are you looking at the right branch?

SergiySW: clemahieu fork replaced with arugulaBowl/lmdb

PlasmaPower: Nevermind... yeah I don't see a difference.

.gitmodules
```diff
@@ -9,7 +9,7 @@
...
 	branch = mdb.RE/0.9
```
argakiig: should be `branch = lmdb_0_9_21` correct @arugulaBowl

argakiig: should be branch = lmdb_0_9_21 correct @arugulaBowl

.gitmodules
cryptocode: The only change to lmdb here is the definition of NTSTATUS which shouldn't be required unless targeting old MinGW versions - is that the case @arugulaBowl ?

arugulaBowl: Ah yes, I didn't explain that well. This is the same as clemahieu/lmdb_0_9_21 with this extra commit on top: https://github.com/arugulaBowl/lmdb/commit/58d12fb3abd8d50b7a81d435c6409bb1bc1af9f5.

I don't understand why this was needed now though.

arugulaBowl:
It's all green in the CI now.

In the two last commits I have:
 * Disabled clang-format for files other than *.{h,hpp,cpp} in ci/check-commit-format.sh
 * Applied clang-format-all.sh

It would be worth checking if the Windows executables work or require any other dll for running. I don't have a windows system at hand, could someone check?

argakiig:
Will build and test windows tonight, and report back. Any reason we can't fix boost to be exact 1.66, also msvc 14.1 builds fine as well as 14.0 as long as libs are built for it.Will update with these results

argakiig:
Rebase please to bring in line with master, otherwise LGTM

arugulaBowl:
@argakiig Hopefully I will have some time to rebase tonight (that is in more than 12h).

Regarding the boost version, the branch includes the finder from the latest cmake master, which is compatible with all previous versions. I was not aware that 1.67 caused problems, but even in that case I don't see the downside of having a newer finder and still use 1.66 for compilation.

arugulaBowl:
It is already rebased, I gave it a shot and it rebased without conflicts.

PR #968, opened 2018-07-11
Mark epoch blocks in history

SergiySW:


rai/node/rpc.cpp
```diff
@@ -1790,6 +1790,11 @@ class history_visitor : public rai::block_visitor
 					tree.put ("subtype", "change");
 				}
 			}
+			else if (balance == previous_balance && !handler.node.ledger.epoch_link.is_zero () && block_a.hashables.link == handler.node.ledger.epoch_link)
+			{
+				tree.put ("type", "epoch");
+				tree.put ("account", handler.node.ledger.epoch_signer.to_account ());
```
PlasmaPower: I'd put an additional `if (raw)` here.

rai/node/rpc.cpp
```diff
@@ -1790,6 +1790,18 @@ class history_visitor : public rai::block_visitor
...
 			}
+			else if (balance == previous_balance && !handler.node.ledger.epoch_link.is_zero () && block_a.hashables.link == handler.node.ledger.epoch_link)
+			{
+				if (raw)
+				{
+					tree.put ("subtype", "epoch");
+				}
+				else
+				{
+					tree.put ("type", "epoch");
```
PlasmaPower: I don't think we should include them in the history at all otherwise, just like with change blocks.

rai/node/rpc.cpp
```diff
@@ -1790,6 +1790,11 @@ class history_visitor : public rai::block_visitor
 					tree.put ("subtype", "change");
 				}
 			}
+			else if (raw && balance == previous_balance && !handler.node.ledger.epoch_link.is_zero () && block_a.hashables.link == handler.node.ledger.epoch_link)
```
PlasmaPower: The `raw` condition needs to be in a separate if statement, like with the change condition above, otherwise it'll fall through to the receive code.

PR #969, opened 2018-07-12
Require core tests to pass in Travis

PlasmaPower:


cryptocode:
`node.fork_stale` hangs a long time sometimes for me, as does `system.generate_send_existing`

but never when run separately (that's true for `fork_pre_confirm` as well)

Some kind of interaction between tests going on?

PlasmaPower:
I think that's a coincidence, or maybe due to CPU or HDD load.

cryptocode:
Maybe, though fork_pre_confirm is never slow when executed separately, `repeat 1000 GTEST_FILTER=fork_pre_confirm core_test` runs without a hickup.

PlasmaPower:
I was able to get it to fail multiple times alone with `until ! ./core_test --gtest_filter=node.fork_pre_confirm; do; done`

cryptocode:
ok, system dependent then

PR #970, opened 2018-07-13
Rewriting RPC with error codes

SergiySW:
Also:
- Merge RPCs chain & RPC successors
- Merge RPCs mrai_from_raw, krai_from_raw & rai_from_raw
- Merge RPCs mrai_to_raw, krai_to_raw & rai_to_raw
- Add functions to get wallet, account, hash, amount from request to reduce code size


SergiySW:
Some useful functions were separated to reduce code size (get wallet, account, hash, amount etc. from request)

PR #973, opened 2018-07-18
Allow the "bulk_pull" bootstrapping message to accept a block hash as…

rkeene:
… its first argument in replace of an account hash;

This is useful to enable bootstrapping to follow chains it finds as references in a link in a receiving block and wants to follow that chain downward as well.

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -1575,21 +1575,14 @@ void rai::bulk_pull_server::set_current_end ()
 		}
 		else
 		{
+			current = info.head;
 			if (!request->end.is_zero ())
 			{
 				auto account (connection->node->ledger.account (transaction, request->end));
-				if (account == request->start)
-				{
-					current = info.head;
-				}
-				else
+				if (account != request->start)
 				{
 					current = request->end;
 				}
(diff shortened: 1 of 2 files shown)


rkeene:
Another option, to make this more explicit is to use an extension bit to signify "alternate meaning" of a parameter, in this case the "account" parameter

rai/node/bootstrap.cpp
```diff
@@ -1547,33 +1551,43 @@ void rai::bulk_pull_server::set_current_end ()
...
+				}
+				else
+				{
+					current = request->end;
+				}
+			}
+			else
+			{
+				current = info.head;
+			}
```
cryptocode: Minor, but the last else could be simplified to

```
current = info.head;
if (!request->end.is_zero () 
    && request->start != connection->node->ledger.account (transaction, request->end))
{
	current = request->end;
}
```

PlasmaPower: or even
```cpp
if (!request->end.is_zero () 
    && request->start != connection->node->ledger.account (transaction, request->end))
{
	current = request->end;
}
else
{
	current = info.head;
}
```

PR #977, opened 2018-07-20
Fix possible wallet locked errors in RPC

SergiySW:
send, receive, change


renesq:
Does this solve the 'unintended self-locking' bug, or just make the error message more precise (previously: "Account not found" if the wallet was locked with an empty password while trying to send funds)?

SergiySW:
Better messages Before it can be error generating blocks or 0 hash which is not correct

PR #981, opened 2018-07-22
Vote/election related docs and minor refactorings

cryptocode:
Vote/election related doxygen docs, encapsulation, more consistent naming, typed durations/chrono literals, and some minor simplifications.


cryptocode:
Going to break this up to avoid constant rebasing

PR #994, opened 2018-07-25
Add wallet_history with timestamps

SergiySW:
Allowing easily track local wallet history actions with RPC


SergiySW:
To be rebased on top of https://github.com/nanocurrency/raiblocks/tree/block_sideband

nano/node/rpc.cpp
```diff
@@ -3347,6 +3347,60 @@ void nano::rpc_handler::wallet_frontiers ()
 	response_errors ();
 }
 
+void nano::rpc_handler::wallet_history ()
+{
+	uint64_t modified_since (1);
+	boost::optional<std::string> modified_since_text (request.get_optional<std::string> ("modified_since"));
+	if (modified_since_text.is_initialized ())
+	{
+		modified_since = strtoul (modified_since_text.get ().c_str (), NULL, 10);
```
cryptocode: Not sure if it matters much here, but `stroul` returns 0 on erroneous input.

SergiySW: Thx, replaced with decode_unsigned () !

PR #996, opened 2018-07-25
Reduce announcement_minimum 

### LINKED ISSUE #992: Reduce Bandwidth Utilization
This ticket is to track work associated with reducing bandwidth by reducing voting re-announcments.  Currently it is repeating votes 4x (reportedly), but this is no longer needed and only 1x is needed due to changes required for quorum.

argakiig:
as quorum must be met for block confirmation now and announcements will continue until minimum quorum is met fixes #992

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -3637,7 +3637,7 @@ void rai::active_transactions::announce_votes ()
 				unconfirmed_announcements += i->announcements;
 			}
 			election_l->broadcast_winner (transaction);
-			if (i->announcements % announcement_min == 2)
+			if (i->announcements % 4 == 1)
 			{
 				auto reps (std::make_shared<std::vector<rai::peer_information>> (node.peers.representatives (std::numeric_limits<size_t>::max ())));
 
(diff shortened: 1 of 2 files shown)


rai/node/node.cpp
```diff
@@ -3559,7 +3559,7 @@ void rai::active_transactions::announce_votes ()
 				unconfirmed_announcements += i->announcements;
 			}
 			node.background ([election_l]() { election_l->broadcast_winner (); });
-			if (i->announcements % announcement_min == 2)
+			if (i->announcements % announcement_min == 1)
```
SergiySW: May be increase to
`if (i->announcements % (announcement_min * 2) == 1)`
?

argakiig: what about `if (i->announcements % 4 == 0)` that way the loop is entered every 4th run

SergiySW: I guess better have == 1 or == 2 so new block can propagate in network before confirmation request. Not sure

argakiig:
cleaned up and rebased for master

PR #998, opened 2018-07-26
If there were more than <announcement_long> announcements, broadcast …

SergiySW:
…confirm request to all known peers

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -3557,6 +3557,8 @@ void rai::active_transactions::announce_votes ()
 			{
 				++unconfirmed_count;
 				unconfirmed_announcements += i->announcements;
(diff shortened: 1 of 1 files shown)


SergiySW:
Since often issue is not enough observed representatives

PlasmaPower:
This seems like a really bad idea. How about just speeding up rep crawler?

SergiySW:
If it can help. There are only about 350 peers now, but sometimes rep_crawler cannot find enough representatives for thousands of seconds while it should check about 2.5 peers every second

SergiySW:
https://github.com/nanocurrency/raiblocks/pull/999 should be better solution

rkeene:
Can I close this PR in preference to #999 ?

PR #999, opened 2018-07-26
Send confirmation request to all peers if representatives list is empty…

SergiySW:
… or observed representatives weight is less than online_weight_minimum

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -3535,6 +3535,7 @@ void rai::active_transactions::announce_votes ()
 	std::lock_guard<std::mutex> lock (mutex);
 	unsigned unconfirmed_count (0);
 	unsigned unconfirmed_announcements (0);
+	unsigned mass_request_count (0);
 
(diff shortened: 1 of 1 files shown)


rai/node/node.cpp
```diff
@@ -3582,7 +3583,7 @@ void rai::active_transactions::announce_votes ()
 						}
 					}
 				}
-				if (!reps->empty () && node.online_reps.online_stake () != node.config.online_weight_minimum.number ())
+				if (!reps->empty () && node.config.online_weight_minimum.number () < total_weight)
```
rkeene: Is this test correct ?

SergiySW: If total_weight > online_weight_minimum from config

rkeene: Okay, that's easier to parse -- seems good !

SergiySW: Yeah, perhaps I should replace it. Opposite comparison seems counterintuitive

PR #1005, opened 2018-07-27
Remove transaction write lock from bootstrap

SergiySW:


PlasmaPower:
I think we should additionally consolidate the frontier_response transactions.

PR #1008, opened 2018-07-28
Update README.md

argakiig:
Testing a change to appveyor will delete after test


argakiig:
openning and trying again

PR #1015, opened 2018-07-31
Fix duplicate send ID insufficient balance

PlasmaPower:


rkeene:
Moving this to rebase on the next patch release

PlasmaPower:
Why?

rkeene:
It needed to be rebased before it could be merged and this was not done for V15.1, and then it got overtaken by other events.  It was a lower priority than getting a V15 release out based on a tested integration.

PR #1025, opened 2018-08-04
Add hash only votes, which can be batched

### LINKED ISSUE #987: Hash-only Voting Support
Implement support for accepting a voting block which does not include the contents of the block, but rather the hash (or up-to 12).

This does not include generating the blocks, but accepting them in place of the current voting block so that in the future the current voting blocks may be deprecated.

PlasmaPower:
Hash-only voting support, closes #987

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -1990,7 +1990,7 @@ void rai::gap_cache::vote (std::shared_ptr<rai::vote> vote_a)
 		if (existing != blocks.get<1> ().end ())
 		{
 			auto is_new (false);
-			blocks.get<1> ().modify(existing, [&](rai::gap_information & info) { is_new = info.voters.insert(vote_a->account).second; });
+			blocks.get<1> ().modify (existing, [&](rai::gap_information & info) { is_new = info.voters.insert (vote_a->account).second; });
 			if (is_new)
 			{
 				uint128_t tally;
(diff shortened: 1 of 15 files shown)


rai/node/node.cpp
```diff
@@ -1876,27 +1891,37 @@ void rai::gap_cache::vote (std::shared_ptr<rai::vote> vote_a)
...
-			auto now (std::chrono::steady_clock::now ());
-			node.alarm.add (rai::rai_network == rai::rai_networks::rai_test_network ? now + std::chrono::milliseconds (5) : now + std::chrono::seconds (5), [node_l, hash]() {
-				rai::transaction transaction (node_l->store.environment, nullptr, false);
-				if (!node_l->store.block_exists (transaction, hash))
+			auto is_new (false);
+			blocks.get<1> ().modify(existing, [&](rai::gap_information & info) { is_new = info.voters.insert(vote_a->account).second; });
+			if (is_new)
+			{
+				uint128_t tally;
+				for (auto voter : existing->voters)
```
cryptocode: maybe `auto&` to avoid copying the uint256_union's, maybe even `const` it

rai/secure/common.cpp
```diff
@@ -746,20 +779,49 @@ rai::uint256_union rai::vote::hash () const
...
+	}
 }
 
 void rai::vote::serialize (rai::stream & stream_a)
 {
 	write (stream_a, account);
 	write (stream_a, signature);
 	write (stream_a, sequence);
-	rai::serialize_block (stream_a, *block);
+	for (auto block : blocks)
```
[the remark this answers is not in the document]
PlasmaPower: It will eventually be bounded in `rai::active_transactions::announce_votes`, though currently it just  produces normal 1 full block votes.

rai/node/node.cpp
```diff
@@ -3795,21 +3901,40 @@ bool rai::active_transactions::start (std::pair<std::shared_ptr<rai::block>, std
...
+	bool processed (false);
 	{
 		std::lock_guard<std::mutex> lock (mutex);
-		auto root (vote_a->block->root ());
-		auto existing (roots.find (root));
-		if (existing != roots.end ())
+		for (auto vote_block : vote_a->blocks)
 		{
-			election = existing->election;
+			rai::election_vote_result result { false, false };
```
cryptocode: I think this initialization should happen in `election_vote_result` so we don't get the problem of uninitialized members if we use the class elsewhere down the line.

PR #1026, opened 2018-08-05
Deadlines in tests

cryptocode:
Summary:

* Uses deadlines (durations in any time unit) instead of iterations
* gtest extended with ASSERT_NO_ERROR which prints std::error_code messages on error. 
* system::poll returns std::error_code. ASSERT_NO_ERROR will have more uses in the future as more code migrates to error_code.
* Deadline scale factor support

The scale factor is an environment variable which makes it easier to experiment with deadlines (instead of recompiling).

It can also be set when running tests on slow machines.

For instance, to see if 10% of the deadline (1 second in this case) is enough to pass the `node_receive_quorum` test:

`DEADLINE_SCALE_FACTOR=0.1 GTEST_FILTER=node.node_receive_quorum Debug/core_test`

To allow 20% more time on all tests on a slow machine, a Travis config or whatever can export:

`DEADLINE_SCALE_FACTOR=1.2`

Sample error when using ASSERT_NO_ERROR:

```
node.cpp:230: Failure
Value of: system.poll ()
  Actual: Deadline expired
Expected:
```

### DIFF
--- a/rai/core_test/CMakeLists.txt
+++ b/rai/core_test/CMakeLists.txt
@@ -1,4 +1,5 @@
 add_executable (core_test
+	testutil.hpp
 	block.cpp
 	block_store.cpp
 	interface.cpp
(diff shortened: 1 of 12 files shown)


cryptocode:
`bulk.offline_send` seems to have been failing for a while, on master as well

PlasmaPower:
I'm thinking this should merge before it causes more conflicts.

PR #1028, opened 2018-08-05
Remove active blocks from *_pending RPC calls

PlasmaPower:


SergiySW:
There should be option to show all pending including active

PlasmaPower:
Added that

PR #1031, opened 2018-08-06
Log votes for very long unconfirmed elections

SergiySW:


PlasmaPower:
Have you tested this on the live net? Is it a reasonable amount of logging?

SergiySW:
This is arbitrary value thats hsould be enough to discover possible issues.Most nodes never have such issues so it's hard to test in live network.

PR #1039, opened 2018-08-08
Bulk pull account

### LINKED ISSUE #988: Bulk Pull by Account Support
Add support for a new bootstrapping command called "bulk_pull_account" which accepts as arguments:
1. Account (256-bits)
2. Minimum Pending Amount (128 bits)
3. Flags (8-bits)

The _Account_ argument is the public key of an account which we are interested in.  The "bulk_pull_account" command will request that the remote node return all blocks related to this account.  This includes the frontier for the account and information related to transactions destined for that account that have not yet been received with receiving blocks (pending).

The _Minimum Pending Amount_ argument is the minimum amount that must be pending in a single transaction for an account or block hash to be included in the _pending_ result section.

The _Flags_ argument is a bitfield which is a mask of the following options OR'd together:
- 0x01: PENDING_ADDRESS_ONLY: The presence of this flag indicates that the result should only include the account public keys in the _pending_ section;  The absence of this flag indicates that the _pending_ section should be comprised of tuples containing the block hash and amount for all pending receives

The result will be in the format of
```
struct frontierInformation {
        blockHash_t frontierHash
        rawAmount_t frontierBalance
}
union pending {
     struct addressOnly[] {
          address_t sendingAccount
     }
     struct hashBalanceTuple[] {
          blockHash_t sendingBlockHash
          rawAmount_t sendingBlockAmount
     }
}
```
Where blockHash_t and address_t are a 256-bit bitfield, and rawAmount_t is a 128-bit bitfield.

The pending section will terminate with a final entry where all entries have a value of 0.  That is, for the case where the PENDING_ADDRESS_ONLY flag is set the result will terminate with a value of a 256-bits 0, and otherwise it will terminate with a value of 384 (256+128) bits of 0.

rkeene:
This will close #988 

Still to do:

- [x] Rebase from master
- [x] Look into using an ephemeral DB transaction to avoid locking pending records for a huge amount of time

Tests will be added as a separate change set.

### DIFF
--- a/rai/core_test/message_parser.cpp
+++ b/rai/core_test/message_parser.cpp
@@ -12,6 +12,7 @@ public:
 	confirm_req_count (0),
 	confirm_ack_count (0),
 	bulk_pull_count (0),
+	bulk_pull_account_count (0),
 	bulk_pull_blocks_count (0),
 	bulk_push_count (0),
 	frontier_req_count (0)
@@ -37,6 +38,10 @@ public:
 	{
 		++bulk_pull_count;
 	}
+	void bulk_pull_account (rai::bulk_pull_account const &) override
+	{
+		++bulk_pull_account_count;
+	}
 	void bulk_pull_blocks (rai::bulk_pull_blocks const &) override
 	{
 		++bulk_pull_blocks_count;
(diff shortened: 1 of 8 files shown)


clemahieu:
The reason I don’t like continue and early return statements is because anything that contains them can’t be trivially factored out in to a function.

I know the nesting hurts readability and makes diffs annoying if you add or remove a nested block so I think it’s a compromise.

PR #1041, opened 2018-08-09
Remove dead local in have_quorum and clarify

cryptocode:
`first` was never used + rename vars to clarify what's going


cryptocode:
The function at hand has been rewritten

PR #1046, opened 2018-08-10
 confirm_req_hash to reduce bandwidth usage

SergiySW:
Initially support for representative side.
Once network is upgraded, replacing send_confirm_req with send_confirm_req_hash in next version


PlasmaPower:
Why include the block hash in the message? Just the root should do.

SergiySW:
Actually I thought about answering with full block if fork & just hash vote if same block requested

SergiySW:
Except that hash can be removed from request (& renamed to confirm_req_root)

PlasmaPower:
If `root` is null, we should just check if we have the block. That way, this can still be used to confirm frontiers.

We should probably also support batching here.

Inkeliz:
It's very strange. We have one single `confirm_ack` that supports votes by hash and vote by block, simply using the `Extensions`. However, the `confirm_req` get's a totally new type, the `confirm_req_hash`.

Why not follow the same way of `confirm_ack`, using`Extensions` header. 👎

SergiySW:
It should be possible as well

cryptocode:
I think it's more robust to write out the hash count instead of relying on EOS when deserializing (premature EOS would be an error as in other message types; a point security wise too - we check against the udp size, but there's a potential for mistakes here). Length prefixing makes testing/tooling/other transports easier too as you can put messages back to back.

nano/node/node.cpp
```diff
@@ -510,19 +605,52 @@ class network_message_visitor : public nano::message_visitor
...
+				if (successor != nullptr)
+				{
+					auto same_block (successor->hash () == message_a.block->hash ());
+					confirm_block (transaction, node, sender, std::move (successor), !same_block);
+				}
+			}
+			else if (!message_a.roots_hashes.empty ())
+			{
+				std::vector<nano::block_hash> blocks_bundle;
+				for (auto root_hash : message_a.roots_hashes)
```
cryptocode: Maybe `auto &` to save an array copy, possibly same in other auto range-based loops.

nano/core_test/block.cpp
```diff
@@ -332,6 +332,58 @@ TEST (block, confirm_req_serialization)
...
+	ASSERT_EQ (req, req2);
+	ASSERT_EQ (req.roots_hashes, req2.roots_hashes);
+}
+
+TEST (block, confirm_req_hash_batch_serialization)
+{
+	nano::keypair key;
+	nano::keypair representative;
+	std::vector<std::pair<nano::block_hash, nano::uint512_union>> roots_hashes;
+	auto open (std::unique_ptr<nano::state_block> (new nano::state_block (key.pub, 0, representative.pub, 2, 4, key.prv, key.pub, 5)));
```
wezrule: I think this can be a stack variable rather than a smart pointer.

nano/core_test/block.cpp
```diff
@@ -332,6 +332,58 @@ TEST (block, confirm_req_serialization)
...
+	roots_hashes.push_back (std::make_pair (open->hash (), nano::uint512_union (open->previous (), open->root ())));
+	for (auto i (roots_hashes.size ()); i < 7; i++)
+	{
+		nano::keypair key1;
+		nano::keypair previous;
+		auto block (std::unique_ptr<nano::state_block> (new nano::state_block (key1.pub, previous.pub, representative.pub, 2, 4, key1.prv, key1.pub, 5)));
```
wezrule: Use stack variable

nano/core_test/message_parser.cpp
```diff
@@ -130,6 +130,38 @@ TEST (message_parser, exact_confirm_req_size)
...
 }
 
+TEST (message_parser, exact_confirm_req_hash_size)
+{
+	nano::system system (24000, 1);
+	test_visitor visitor;
+	nano::block_uniquer block_uniquer;
+	nano::vote_uniquer vote_uniquer (block_uniquer);
+	nano::message_parser parser (block_uniquer, vote_uniquer, visitor, system.work);
+	auto block (std::unique_ptr<nano::send_block> (new nano::send_block (1, 1, 2, nano::keypair ().prv, 4, system.work.generate (1))));
```
wezrule: Use stack variable

nano/node/common.cpp
```diff
@@ -500,14 +547,63 @@ void nano::confirm_req::visit (nano::message_visitor & visitor_a) const
...
-	assert (block != nullptr);
 	header.serialize (stream_a);
-	block->serialize (stream_a);
+	if (header.block_type () == nano::block_type::not_a_block)
+	{
+		assert (!roots_hashes.empty ());
+		// Calculate size
+		uint8_t count (roots_hashes.size ());
+		write (stream_a, count);
+		assert (count <= 32);
```
wezrule: If roots_hashes.size () is 257 then count will be 1 and this assert won't be firing correctly. Should this instead be on roots_hashes.size ()?

nano/node/node.cpp
```diff
@@ -305,6 +322,22 @@ void nano::network::republish_block (std::shared_ptr<nano::block> block)
 	}
 }
 
+void nano::network::republish_block (std::shared_ptr<nano::block> block, nano::endpoint const & peer_a)
+{
+	auto hash (block->hash ());
+	nano::publish message (block);
+	std::shared_ptr<std::vector<uint8_t>> bytes (new std::vector<uint8_t>);
```
wezrule: This can be a stack variable

nano/node/node.cpp
```diff
@@ -3413,15 +3542,60 @@ void nano::active_transactions::request_confirm (std::unique_lock<std::mutex> &
...
+							auto root_hash (std::make_pair (block->hash (), nano::uint512_union (block->previous (), block->root ())));
+							if (rep_request == requests_bundle.end ())
+							{
+								std::vector<std::pair<nano::block_hash, nano::uint512_union>> insert_vector (1, root_hash);
+								if (requests_bundle.size () < max_broadcast_queue * 2)
+								{
+									requests_bundle.insert (std::make_pair (j->endpoint, insert_vector));
+								}
+							}
+							else if (rep_request->second.size () < max_broadcast_queue * 6)
```
wezrule: How is this magic number chosen? (and the one above it). If it's dependent on something can it be derived from that? or use a variable describing it or a comment.

nano/node/node.cpp
```diff
@@ -3413,15 +3542,60 @@ void nano::active_transactions::request_confirm (std::unique_lock<std::mutex> &
...
+					else
+					{
+						for (auto j (reps->begin ()), m (reps->end ()); j != m; j++)
+						{
+							auto rep_request (requests_bundle.find (j->endpoint));
+							auto block (i->election->status.winner);
+							auto root_hash (std::make_pair (block->hash (), nano::uint512_union (block->previous (), block->root ())));
+							if (rep_request == requests_bundle.end ())
+							{
+								std::vector<std::pair<nano::block_hash, nano::uint512_union>> insert_vector (1, root_hash);
```
wezrule: std::vector has a few confusing argument constructors, I find it easier to avoid them entirely. As this is only inserting 1 element it might look less confusing with the initializer syntax `= { root_hash };` (although not required)

nano/node/node.cpp
SergiySW: Probably doesn't make much sense here because nano::network::republish accept std::shared_ptr<std::vector<uint8_t>> buffer_a

nano/node/common.cpp
SergiySW: Good point

nano/node/node.cpp
SergiySW: 6 is max hashes in confirm_req (now const in network), while above should be just max_broadcast_queue

nano/node/node.cpp
wezrule: Yeh fair enough, I didn't see that

PR #1050, opened 2018-08-11
In debug mode, rethrow caught exceptions

rkeene:
In debug mode, rethrow caught exceptions for service procedures

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -3718,7 +3718,16 @@ rai::thread_runner::thread_runner (boost::asio::io_service & service_a, unsigned
 			}
 			catch (...)
 			{
+#ifndef NDEBUG
+				/*
+				 * In a release build, catch and swallow the
+				 * service exception, in debug mode pass it
+				 * on
+				 */
+				throw;
+
 				assert (false && "Unhandled service exception");
+#endif
 			}
 		}));
 	}


rai/node/node.cpp
```diff
@@ -3718,7 +3718,16 @@ rai::thread_runner::thread_runner (boost::asio::io_service & service_a, unsigned
...
 			{
+#ifndef NDEBUG
+				/*
+				 * In a release build, catch and swallow the
+				 * service exception, in debug mode pass it
+				 * on
+				 */
+				throw;
+
 				assert (false && "Unhandled service exception");
```
PlasmaPower: Shouldn't we get rid of this assert?

PlasmaPower: Also, in release mode, should we start a replacement thread?

rkeene: Maybe, but that will have to be a different changeset -- we shouldn't generally have unhandled
exceptions reach this point, so if they are thrown from here in debug mode we can find them

cryptocode: fwiw, the exceptions I've seen reaching this point in the past were boost log exceptions (not enough arguments for the format string), but all those should've been fixed. Could be others lurking of course.

nallar:
Now in release mode, an exception here is completely ignored? That doesn't seem right.

PlasmaPower:
Asserts are ignored in debug mode anyway. I don't think we can access the logging system if we get an exception there. Maybe we could put it into a global queue, or log to cerr?

PR #1061, opened 2018-08-14
Fix rai::store_merge_iterator::cursor_current

PlasmaPower:


PlasmaPower:
We might not need the size checks with the new comparison function.

PR #1105, opened 2018-08-26
Allow local preconfigured peers

PlasmaPower:
Untested


rai/node/node.cpp
```diff
@@ -95,15 +95,20 @@ void rai::network::send_keepalive (rai::endpoint const & endpoint_a)
 	});
 }
 
-void rai::node::keepalive (std::string const & address_a, uint16_t port_a)
+void rai::node::keepalive (std::string const & address_a, uint16_t port_a, bool preconfigured_peer_a)
 {
 	auto node_l (shared_from_this ());
-	network.resolver.async_resolve (boost::asio::ip::udp::resolver::query (address_a, std::to_string (port_a)), [node_l, address_a, port_a](boost::system::error_code const & ec, boost::asio::ip::udp::resolver::iterator i_a) {
+	network.resolver.async_resolve (boost::asio::ip::udp::resolver::query (address_a, std::to_string (port_a)), [node_l, preconfigured_peer_a, address_a, port_a](boost::system::error_code const & ec, boost::asio::ip::udp::resolver::iterator i_a) {
```
cryptocode: Not specific to this PR, but this particular async_resolve overload is deprecated and will be removed (https://www.boost.org/doc/libs/1_67_0/doc/html/boost_asio/reference/ip__basic_resolver/async_resolve.html)

PlasmaPower: What should we use instead?

cryptocode: One of the string_view overloads I think

SergiySW:
Related to https://github.com/nanocurrency/nano-node/pull/1370/files

PR #1106, opened 2018-08-26
Fix duplicate send ID insufficient balance again

PlasmaPower:
The non-test part of edcdb15 was reverted in 2a6030c, which broke the test

### DIFF
--- a/rai/core_test/rpc.cpp
+++ b/rai/core_test/rpc.cpp
@@ -375,7 +375,8 @@ TEST (rpc, send_idempotent)
 		system.poll ();
 	}
 	ASSERT_EQ (200, response2.status);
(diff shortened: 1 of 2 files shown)


rai/core_test/rpc.cpp
```diff
@@ -375,7 +375,8 @@ TEST (rpc, send_idempotent)
 		system.poll ();
 	}
 	ASSERT_EQ (200, response2.status);
-	ASSERT_EQ (response2.json.get<std::string> ("block"), block_text);
+	ASSERT_EQ ("", response2.json.get<std::string> ("error", ""));
+	ASSERT_EQ (block_text, response2.json.get<std::string> ("block"));
```
PlasmaPower: This makes the error message clearer when it errors.

PR #1114, opened 2018-08-28
Added start of "release_assert()"

### LINKED ISSUE #1098: Create a release-assert mechanism
Currently we use `assert(3)` for some checks that should be done in release builds, but `assert(3)` is a no-op in release builds.

Let's create an assert-alike that aborts if those checks fail in release builds.

Possible names:
 - `always_assert()`
 - `assert()` (and rename `assert(3)` to `debug_assert()`, but this is non-standard and confusing to developers)
 - `release_assert()`

rkeene:
Closes #1098

### DIFF
--- a/rai/lib/utility.cpp
+++ b/rai/lib/utility.cpp
@@ -1,0 +1,13 @@
+#include <iostream>
+#include <rai/lib/utility.hpp>
+
+void release_assert_internal (bool check, const char * check_expr, const char * file, unsigned int line)
+{
+	if (check)
+	{
+		return;
+	}
+
+	std::cerr << "Assertion (" << check_expr << ") failed " << file << ":" << line << std::endl;
+	abort();
+}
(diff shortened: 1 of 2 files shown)


rai/lib/utility.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+void release_assert_internal (bool check, const char * check_expr, const char * file, unsigned int line)
+{
+	if (check)
+	{
+		return;
+	}
+
+	std::cerr << "Assertion (" << check_expr << ") failed " << file << ":" << line << std::endl;
+	abort();
```
PlasmaPower: Will calling `abort()` cause problems for our test suite?

PlasmaPower: As in, will it abort the test suite instead of just failing the test

rkeene: This is what `assert(3)` does, so our test suite should handle it.  I did test it and this is the result:

```
[ RUN      ] block.send_serialize_json
Assertion (false) failed /home/rkeene/devel/raiblocks/rai/lib/numbers.cpp:299
Aborted
```

PlasmaPower: Ok, if its behavior matches assert, then LGTM.

PR #1117, opened 2018-08-29
Add support for failing tests when a release_assert fails

rkeene:
Rather than aborting the test suite when a "release_assert" fails, simply abort the current test.

This is only applied when compiling for the test network.

### DIFF
--- a/nano/core_test/CMakeLists.txt
+++ b/nano/core_test/CMakeLists.txt
@@ -20,7 +20,8 @@ add_executable (core_test
 	versioning.cpp
 	wallet.cpp
 	wallets.cpp
-	work_pool.cpp)
+	work_pool.cpp
+	testutil.cpp)
 
 target_compile_definitions(core_test
 		PRIVATE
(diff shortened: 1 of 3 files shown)


rai/lib/utility.cpp
```diff
@@ -1,13 +1,21 @@
 #include <iostream>
 #include <rai/lib/utility.hpp>
 
+#if ACTIVE_NETWORK == rai_test_network
+/*
+ * If we are running under the test harness, create this as a
+ * weak symbol so the one we provide as part of gtest can
+ * override it.
+ */
+#  pragma weak release_assert_internal
```
cryptocode: Does msvc support this?

PlasmaPower:
We should probably also override the regular `assert`.

Also, does this still support building rai_node for the test net?

rkeene:
You can still build rai_node for the test net, it will use the weak symbol for "assert_release" and not the one from gtest.

rai/lib/utility.cpp
rkeene: I don't think MSVC supports it, but my understanding is that it will ignore the unknown pragma and during linking it will pick one of the symbols.  If it picks the less-than-optimal one on MSVC then there will slightly less optimal tests, but overall there will be no change.

rkeene:
Doing this for regular assert is more difficult because we can't easily supply a new assert without being careful.

cryptocode:
This won't work with network_params. Replaced by https://github.com/nanocurrency/nano-node/pull/1856

PR #1118, opened 2018-08-29
Add warning tag to constructors using release_assert

cryptocode:


rai/lib/utility.cpp
```diff
@@ -9,5 +9,5 @@ void release_assert_internal (bool check, const char * check_expr, const char *
 	}
 
 	std::cerr << "Assertion (" << check_expr << ") failed " << file << ":" << line << std::endl;
-	abort();
+	abort ();
```
rkeene: Oops, I missed this initially.

PR #1124, opened 2018-08-30
Check genesis on startup

cryptocode:
There seems to be nodes on betanet running with livenet databases (blocks are from livenet). Since it's easy to mix up databases when using `--data_path`, this PR proposes to check that the db contains the correct genesis block ~~write the network id into the `meta` db and abort the process if it doesn't match the process' network id on startup.~~

### DIFF
--- a/rai/lib/utility.cpp
+++ b/rai/lib/utility.cpp
@@ -9,5 +9,5 @@ void release_assert_internal (bool check, const char * check_expr, const char * 
 	}
 
 	std::cerr << "Assertion (" << check_expr << ") failed " << file << ":" << line << std::endl;
-	abort();
+	abort ();
 }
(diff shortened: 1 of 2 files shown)


clemahieu:
Can this be checked by checking of the genesis block exists in the ledger?

cryptocode:
@clemahieu haven't the genesis changed on beta a few times? If someone pulls out an old db.

cryptocode:
@clemahieu better?

rai/lib/utility.cpp
```diff
@@ -9,5 +9,5 @@ void release_assert_internal (bool check, const char * check_expr, const char *
 	}
 
 	std::cerr << "Assertion (" << check_expr << ") failed " << file << ":" << line << std::endl;
-	abort();
+	abort ();
```
rkeene: This is already changed in master, I think

cryptocode: I rebased but this still shows up after running ci/clang-format-all.sh. Was master force-updated maybe? I don't think there'll be a merge conflict.

PR #1126, opened 2018-08-30
Event recorder

cryptocode:
https://github.com/cryptocode/raiblocks/wiki/Event-Recorder

This is an extension to the statistics api to get more detailed diagnostics when debugging issues. Instrumentation points are typically one-liners, and shouldn't impact regular code much. 

Pushing the PR to get discussions on whether or not this is something we want to add in this form. This can be extended in all sorts of directions - there's a few suggestions at the end of the wiki page.

Update: opt-in stacktrace support (used by transaction logging has been added). Currently, this feature is only supported on macOS.


rkeene:
I want to use the event recorder functionality from `rai::transaction::transaction()` and `rai::transaction::~transaction()` for Issue #1134, but I have no `rai::node` object -- is there a way to make the recording independent of the `rai::node` object in practice ?

One possibility is a singleton that buffers all events until it is given a `rai::node` object to attach to then follows its policy on queued and future events.

rai/node/eventrecorder.hpp
```diff
@@ -0,0 +1,552 @@
...
+		/**
+		 * Iterates all hash entries in all event tables, invoking the callback for each entry
+		 */
+		std::error_code iterate_hash (rai::block_hash hash, std::function<void(db_info * db_info, std::unique_ptr<nano::events::event>)> callback);
+
+		/** 
+		 * On startup, this is set to the number of event entries (this works because we never delete events).
+		 * Used to set the event ordinal.
+		 */
+		std::atomic<uint32_t> counter{ 0 };
```
rkeene: `std::atomic<>` requires the header <atomic>

rkeene:
`rai::node::recorder_get()` is fine with me as long as it works in some way before any node objects are created (since it may get called during the constructor initialization)

clemahieu:
I’d prefer if we didn’t use singletons. Static initialization has a lot of nuances. Can we make it a part of the logging class?

cryptocode:
@clemahieu Yeah, singletons are indeed problematic, but not sure how to call the event recorder from rai::transaction without it, unless we inject it. Putting the singleton in the recorder class instead of the node is no problem, if that's what you meant.

cryptocode:
Colin noted that https://github.com/nanocurrency/raiblocks/pull/1127 will introduce `store::tx_begin()`, which should make it possible to reference the recorder without a singleton. The recorder will probably have to be injected into the store constructor.

cryptocode:
Added support for tx recording, rebased on top of the db abstraction change

cryptocode:
This was originally intended to track what happened to hashes throughout the node, but seems like we really want a general-purpose event reporting system. 

I'm not sure manual serialization is best for that. One idea is to rework this to use sqlite (e.g. just add the amalgated source files), which would make adding new event types trivial (just new tables) and it would give powerful event query capabilities.

cryptocode:
The original block monitoring idea is now mostly covered by packet capture tools. For general purpose event monitoring, we might want something like this PR, but based on sqlite or something like that for easy querying. Thus closing this one.

PR #1128, opened 2018-08-30
Check for invalid private keys in wallet_decrypt_unsafe

SergiySW:
I.e. corrupted database


PlasmaPower:
We already check for this in `rai::wallet_store::fetch`, right?

SergiySW:
Yes, and return general bool error that can mean several different things. Also not showed to user, only assert

PlasmaPower:
Ok, then this change seems good. :+1:

rai/node/cli.cpp
```diff
@@ -433,7 +433,14 @@ std::error_code rai::handle_node_options (boost::program_options::variables_map
...
 							auto error (existing->second->store.fetch (transaction, account, key));
 							assert (!error);
-							std::cout << boost::str (boost::format ("Pub: %1% Prv: %2%\n") % account.to_account () % key.data.to_string ());
+							if (rai::pub_key (key.data) == account)
+							{
+								std::cout << boost::str (boost::format ("Pub: %1% Prv: %2%\n") % account.to_account () % key.data.to_string ());
+							}
+							else
+							{
+								std::cout << boost::str (boost::format ("Pub: %1% Prv: %2% Invalid private key!\n") % account.to_account () % key.data.to_string ());
```
rkeene: Should this be stderr, rather than stdout ?

SergiySW: If add "Invalid private key" to different line

PR #1130, opened 2018-08-31
Make build from build folder instead of root source folder

tiendq:
Build from `build` folder to not pollute root source folder with CMake generated files, and easy cleaning up by just `rm -rf build`.

Configure the project build with the following command to create the `build` directory with the configuration.

```shell
mkdir build
cd build               # Create a build directory
cmake ..              # Configure the project
cmake --build .  # Build all default targets
```

To build `rai_node`: `cmake --build .  --target rai_node`
To build `wallet`: `cmake --build .  --target nano_wallet`

Target built files are in `<source>/build/bin`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,9 @@ CTestTestfile.cmake
 install_manifest.txt
 Makefile
 Testing
+/build
+
+# Build generated files
+bootstrap_weights.cpp
+counters.stat
+samples.stat
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,9 @@ endif ()
 set (CPACK_PACKAGE_VENDOR "Nano Currency")
 
 set(CMAKE_INSTALL_RPATH "@executable_path/../Frameworks")
+
 # Create all libraries and executables in the root binary dir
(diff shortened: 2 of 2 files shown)


PlasmaPower:
I already build from subdirs, and it works fine for me. What's the problem you're experiencing without this PR?

rkeene:
@tiendq Thanks for this contribution !  I'll look into it further !

clemahieu:
The way to do this is to run cmake from the directory you want to build in and it won't pollute the source.  I build out-of-tree all the time.

rkeene:
I think I understand the request now.  @tiendq can you confirm that this already works (without this changeset) if you create a directory and run `cmake` from there, basically as described in your initial description ?

tiendq:
@rkeene I followed building instructions from root `README.md` yesterday, and it told me to build in the root source folder e.g. `raiblocks` where I cloned the repo. I then tried to build from `build` folder, and just update `CMakeLists.txt` to put targets into `/build/bin`. Actually I wanted to update building instructions too but it wasn't a `.md` file so I didn't know how to do.

PR #1137, opened 2018-09-02
Added incomplete transaction tracing

### LINKED ISSUE #1134: DB transaction tracing
Production-grade tool to keep track of DB transactions;  Possibly integrate with @cryptocode's event tracing framework

rkeene:
Closes #1134

### DIFF
--- a/rai/node/lmdb.cpp
+++ b/rai/node/lmdb.cpp
@@ -264,14 +264,29 @@ rai::mdb_val::operator MDB_val const & () const
 	return value;
 }
 
+std::atomic<uint64_t> rai::transaction::id_counter;
+
 rai::transaction::transaction (rai::mdb_env & environment_a, bool write, MDB_txn * parent_a)
 {
+	/*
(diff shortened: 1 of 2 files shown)


cryptocode:
This has been incorporated into https://github.com/nanocurrency/raiblocks/pull/1126 with the following changes/notes:

* transaction id is initialized with current timestamp to avoid id clashes between runs
* rebased on top of the db abstraction merge
* the recorder currently exposes a singleton, so it can be accessed from the transaction - this should be refactored later so we can use the recorder in multi-node tests
* the tx tracing changes are (partially) documented in https://github.com/cryptocode/raiblocks/wiki/Event-Recorder
    * tx tracing and backtrace logging are both config options, default off
    * backtrace logging adds measurable time overhead, only turn on when backtraces are needed (at least until observers are fixed, as they produce large amounts of unnecessary transactions)

rkeene:
@cryptocode Thanks !

PR #1138, opened 2018-09-02
Trace vote paths through extensions field

PlasmaPower:
Untested


PlasmaPower:
This seems to be working from my testing. I'd like to get this on the beta network.

zhyatt:
@cryptocode Can you verify Colin's comment and close out if we are good?

PR #1142, opened 2018-09-04
Improve rai::election::publish

PlasmaPower:


rai/node/node.cpp
```diff
@@ -3696,7 +3696,31 @@ bool rai::election::publish (std::shared_ptr<rai::block> block_a)
...
+			if (node.store.block_exists (transaction, block_a->previous ()))
+			{
+				account = node.ledger.account (transaction, block_a->previous ());
+			}
+		}
+		else
+		{
+			account = block_a->root ();
+		}
+		if (account.is_zero () || rai::validate_message (account, block_a->hash (), block_a->block_signature ()))
```
SergiySW: Support for epoch blocks?

PlasmaPower: Fixed and added a test for it

rai/node/node.cpp
```diff
@@ -3697,7 +3697,58 @@ bool rai::election::publish (std::shared_ptr<rai::block> block_a)
...
+			}
+			if (!result)
+			{
+				if (block_l->hashables.balance == prev_balance && !node.ledger.epoch_link.is_zero () && block_l->hashables.link == node.ledger.epoch_link)
+				{
+					account = node.ledger.epoch_signer;
+				}
+			}
+		}
+		if (!result && (account.is_zero () || rai::validate_message (account, block_a->hash (), block_a->block_signature ())))
```
rkeene: I think it would be better if we moved all the work to validate a block and move it into a different function, probably a method of the block.

PR #1145, opened 2018-09-04
Added start of debugging mechanism

rkeene:
This changeset introduces a unified way to add debugging logs of various levels to subsystems within Nano.


cryptocode:
Closing as it's incomplete and needs more discussion. Can probably be replaced by boost logging levels, which we likely want anyway to differentiate on severity.

PR #1157, opened 2018-09-07
Require 2 conditions to confirm election

SergiySW:
Both sum >= node.config.online_weight_minimum & have_quorum () required
To prevent possible situations with confirmed election without calling block_processor.force or block_processor.force without confirmed election

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -3600,14 +3600,14 @@ void rai::election::confirm_if_quorum (rai::transaction const & transaction_a)
 	{
 		sum += i.first;
 	}
-	if (sum >= node.config.online_weight_minimum.number () && !(*block_l == *status.winner))
+	if (sum >= node.config.online_weight_minimum.number () && have_quorum (tally_l))
 	{
-		auto node_l (node.shared ());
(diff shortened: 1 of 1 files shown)


clemahieu:
Blocks need to be able to be forced out before quorum is reached otherwise consensus could deadlock.

SergiySW:
For forcing blocks
Confirmation still should require sum >= online_weight_minimum

rai/node/node.cpp
```diff
@@ -3601,13 +3601,13 @@ void rai::election::confirm_if_quorum (rai::transaction const & transaction_a)
 	{
 		sum += i.first;
 	}
-	if (sum >= node.config.online_weight_minimum.number () && !(*block_l == *status.winner))
+	if (sum >= node.config.online_weight_minimum.number () && block_l->hash () != status.winner->hash ())
```
rkeene: Is this already done as a part of PR 916 ?

https://github.com/nanocurrency/raiblocks/pull/916/files#diff-ced934b3fddcffa6f71a6c8fc44ff563L501

Either way it's more clearly written like this to me, but I want to make sure that if we need further changes we identify them.

PR #1158, opened 2018-09-07
List of rolled back blocks in block_processor

SergiySW:


clemahieu:
We can't filter out blocks that have been forced out because the winner can oscillate before quorum.

SergiySW:
If winner will be changed before quorum it will be placed in forced deque, not regular blocks. So ongoing election won't be affected, only fresh bootstrap blocks cannot start same election again & again

clemahieu:
That makes sense.  So this is to prevent the same block from starting an election if it's already failed once. Should add every hash that's not the winner at the end of an election in to this set? We need to consider how big we should allow this container to get and when we want to prune it.

SergiySW:
No significant bootstrap speed improvements, proposed to close

rkeene:
We addressed this in other ways, closing this out.

PR #1165, opened 2018-09-09
Compare value in merge itr only when key is equal

cryptocode:
Fixes https://github.com/nanocurrency/raiblocks/issues/1164


cryptocode:
Added test

PR #1169, opened 2018-09-10
Exclude active blocks from the pending RPC

### LINKED ISSUE #1122: RPC call "pending" returning active blocks
During investigation of Issue #1107 it was discovered that the RPC call "pending" may return active blocks.  As part of pull request #1028 active blocks were removed from "*_pending" RPC calls but may not have been removed from "pending".

Investigate if "pending" does or does not filter out active blocks, then update it to filter out active blocks unless the "include_active" object is specified with a value of true, just like "accounts_pending" and "wallet_pending"

PlasmaPower:
Fixes #1122


PlasmaPower:
Not quite, I think I fixed existing tests

PlasmaPower:
I'd like to work on the tests before merging this

PlasmaPower:
Tests fixed, and also applied this change to pending_exists

PR #1170, opened 2018-09-10
 Remove generate_hash_votes_at from config

SergiySW:
As it's no longer required


rai/node/node.cpp
```diff
@@ -3819,7 +3806,6 @@ void rai::active_transactions::announce_votes ()
 					else
 					{
 						election_l->compute_rep_votes (transaction);
```
PlasmaPower: I think this else branch should be removed entirely.

rai/node/node.hpp
```diff
@@ -518,6 +520,11 @@ class node_config
 	rai::stat_config stat_config;
 	rai::uint256_union epoch_block_link;
 	rai::account epoch_block_signer;
+<<<<<<< HEAD
+=======
+	std::chrono::system_clock::time_point generate_hash_votes_at;
+	std::chrono::milliseconds block_processor_batch_max_time;
+>>>>>>> upstream/master
```
PlasmaPower: Looks like you left in a merge conflict by mistake

PlasmaPower:
It looks like the executable permission of some scripts under utils were removed.

SergiySW:
Weird Windows. Fixed

SergiySW:
It seems common.sh also should be executable

PlasmaPower:
Actually, I believe common.sh was meant to be sourced (and therefore it shouldn't be executable)

PlasmaPower:
LGTM other than that

SergiySW:
Good

Issue #1123, opened 2018-08-29
Add representative weights for the Beta network


rkeene:
Add representative weights for the Beta network, similar to the Live/Main network, to the node.

rkeene:
Resolved by #1179

PR #1179, opened 2018-09-12
Initial bootstrap weight for betanet

cryptocode:
This adds support for initial repweights on betanet. There are now `rep_weights_beta.bin` and  `rep_weights_live.bin`, the CMake file picks the right one.

Also updated the Python script to dump the accounts and weights included in the resulting bin file.


rkeene:
Closes #1123

PR #1182, opened 2018-09-13
Enable bootstrap weights on betanet

cryptocode:
@rkeene oops, this change wasn't added to the commit in https://github.com/nanocurrency/raiblocks/pull/1179


cryptocode:
@SergiySW I opted for explicit tests in case https://github.com/nanocurrency/raiblocks/issues/1133 adds a new network type, which is likely to not have repweights (?)

PR #1184, opened 2018-09-13
Minimum bootstrap node version

cryptocode:
About 18% of the bootstrap targets are nodes that can never be in sync - this sets the minimum version to ensure the node groks VBH (please verify version). An alternative is to depeer them outright, or we can use this as a first step.


cryptocode:
Not ready for merge yet.

Seems to run well on livenet, but a bunch of tests dislike the version change.

cryptocode:
Tests pass locally, not sure what's going on with CI.

PR #1193, opened 2018-09-16
Add "latest-rc" tag for Docker

BitDesert:
So we can pull the "latest-rc" tag that points to the latest release candidate.

### DIFF
--- a/ci/deploy-docker.sh
+++ b/ci/deploy-docker.sh
@@ -12,7 +12,7 @@ fi
 
 tags=()
 if [[ "${TRAVIS_TAG}" =~ 'RC' ]]; then
-    tags+=("$TRAVIS_TAG")
+    tags+=("$TRAVIS_TAG" latest-rc)
 elif [ -n "$TRAVIS_TAG" ]; then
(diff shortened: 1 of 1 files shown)


rkeene:
The disadvantage of this is that it will sometimes be the case that "latest-rc" is older than "latest" and sometimes be the case that "latest" is older than "latest-rc".

I think a better solution would be to create a new tag that either points to "latest" or the proposed "latest-rc", depending on which is newer -- something like "unstable".

severalif:
@rkeene The issue right now is that the nodes (the official etc) on the beta network run v16rc2, which is older than the latest release (v16), and this in combination with that nodes on the beta network de-peer when the protocol version differs. Thus, if you use the latest tag for beta network, then you will run v16 and get no peers. If you add peers manually, then you get issues with block confirmation since there are not enough reps using v16.

It's not really clear what version we should run on the beta network. One response we got from core in the Discord channel beta-net was that we should only run release candidates (not releases if they are newer). That's why Desert created this PR.

For me it makes more sense to run the latest release candidate or latest release depending on which one is newer (thus, an unstable tag).

Issue #1196, opened 2018-09-18
Conflicting bit positions in message_header


catenocrypt:
The following message header extensions bit positions are defined:
(file rai/node/common.hpp)
class message_header:
  static size_t constexpr ipv4_only_position = 1;
  static size_t constexpr bootstrap_server_position = 2;
class node_id_handshake:
  static size_t constexpr query_flag = 0;
  static size_t constexpr response_flag = 1;

Both ipv4_only and response uses bit 1, which may lead to conflict.
It looks like the ipv4_only bit is in fact not used, nonetheless this is confusing.

PlasmaPower:
Neither ipv4_only nor bootstrap_server are used. We should probably just get rid of the constants.

catenocrypt:
Correct.  The bitfield constants should be moved to class message_header (from class node_id_handshake), so that they are in one place, and conflicts are less likely when new flags are introduced in the future.

catenocrypt:
Solution is available here:

https://github.com/nanocurrency/raiblocks/compare/master...catenocrypt:issue_1196

catenocrypt:
Merged to master (nov 19)

cryptocode:
Fixed by #1199

PR #1199, opened 2018-09-19
Message header extensions: remove unused flags

catenocrypt:
Issue #1196


catenocrypt:
BI build failed, but there is only one timeout error (node.vote_republish)

PlasmaPower:
I'd prefer to keep query/response local to node ID messages

catenocrypt:
I understand that the query/response flags are used only by the handshake message type, but I suggest to put the flag definition into the central message part, because if a new flag is defined in the general part for all messages, it can conflict with message-type-specific flags.  That's why I think it's better to have all flag definitions in one place  (OK, I know that there are not too many flags definitions).

PlasmaPower:
However, there can be conflicts between different types of messages, and that's fine. E.g. there already are some between node id handshake and bulk pull account.

catenocrypt:
As suggested, I have left query/response flag definitions where they were.

PR #1201, opened 2018-09-19
Added support for PENDING_HASH_AMOUNT_AND_ADDRESS to bulk_pull_account

### LINKED ISSUE #1159: Support a flag for `bulk_pull_account` for including the account
@PlasmaPower mentioned it would be helpful for `bulk_pull_account` (added per #988 by #1039) to optionally include the sender's account in addition the the hash and amount, for UI purposes most likely

rkeene:
Closes #1159

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -1796,13 +1796,23 @@ void rai::bulk_pull_account_server::set_params ()
 	 * Parse the flags
 	 */
 	invalid_request = false;
+	pending_include_address = false;
+	pending_address_only = false;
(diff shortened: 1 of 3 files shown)


PlasmaPower:
I was thinking we'd just check if both flags were set, but other than that, LGTM

rkeene:
Treating it as a flag doesn't make sense in this combination, so I changed it to be an Option/Enumeration

00b (0x00) == Default; Amount and Hash
01b (0x01) == Address-only
10b (0x02) == Default + Address
11b (0x03) == Gibberish (Address-only+Address?)

PlasmaPower:
Oh, I see. I thought those were bit positions instead of actual flags. LGTM then.

PR #1210, opened 2018-09-21
Service UDP receive socket more frequently

clemahieu:
Previously the UDP servicing routine would do potentially slow things like database reads, memory allocations etc.
This change adds the udp_buffer class which is designed to service the UDP socket as fast as possible.
udp_buffer:
- Will internally drop network data instead of relying on the operating system to do it.
- Does not allocate memory except on node startup
- Only queues buffers within the IO thread routines
- Does not copy any data while queueing

Results from this patch and running a bootstrap went from tens of thousands of OS dropped UDP packets to around 100 and no internal packet drops.

### DIFF
--- a/rai/core_test/block_store.cpp
+++ b/rai/core_test/block_store.cpp
@@ -285,7 +285,6 @@ TEST (unchecked, multiple_get)
 	auto block2 (std::make_shared<rai::send_block> (3, 1, 2, rai::keypair ().prv, 4, 5));
 	auto block3 (std::make_shared<rai::send_block> (5, 1, 2, rai::keypair ().prv, 4, 5));
 	{
-		
 		auto transaction (store.tx_begin (true));
 		store.unchecked_put (transaction, block1->previous (), block1); // unchecked1
 		store.unchecked_put (transaction, block1->hash (), block1); // unchecked2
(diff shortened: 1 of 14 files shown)


VenKamikaze:
I just built and ran this branch on betanet as I was wondering if it would help with the v16 sync issues. 

It does not seem to be able to bootstrap from other peers and I'm getting the following in the logs:
[2018-09-21 15:09:19.511590]: Starting bootstrap attempt
[2018-09-21 15:09:19.511922]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512045]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512146]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512230]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512318]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512399]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512480]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512561]: Bootstrap stopped because there are no peers
[2018-09-21 15:09:19.512644]: Exiting bulk push client
[2018-09-21 15:09:19.512749]: Exiting bootstrap attempt
[2018-09-21 15:09:19.743774]: Could not deserialize buffer
[2018-09-21 15:09:19.762357]: Could not deserialize buffer
[2018-09-21 15:09:19.762815]: Could not deserialize buffer
[2018-09-21 15:09:19.763743]: Could not deserialize buffer
... etc.

The 'could not deserialize buffer' only occurs in 1 minute intervals, whereas the bootstrap attempts try every 5 seconds (and fail).

cryptocode:
@VenKamikaze the betanet nodes only peers with other nodes of the exact same protocol version. You will have to change the protocol version check in the source or wait for the next RC.

PR #1218, opened 2018-09-22
Retry failed core test a few times for 90 days until we can find and …

rkeene:
…fix flakey tests


PlasmaPower:
Travis runs are already close to the time limit of 50 minutes. I'm afraid that this may push it over that limit.

PlasmaPower:
Per platform

PlasmaPower:
I guess some previous PRs have shortened the builds, so this should work. :+1:

rkeene:
Rebased against master after pull request #1222

PR #1219, opened 2018-09-22
Start elections for processed blocks after transaction commit

SergiySW:


rai/node/node.cpp
```diff
@@ -3882,7 +3889,7 @@ void rai::active_transactions::announce_votes ()
 						node.network.republish_block (transaction, election_l->status.winner);
 					}
 				}
-				else if (i->announcements > 3)
+				else
```
SergiySW: Delay was introduced to wait for transaction commit. Now all blocks in elections should exist in ledger

PR #1223, opened 2018-09-24
Zero-terminate all things we refer to as strings

### LINKED ISSUE #290: xrb_uint256_to_string  does not null-terminate the address string
https://github.com/clemahieu/raiblocks/blob/0f06b03a1c42655f3217b1a8f2936320a7b0cf67/rai/lib/interface.cpp#L21

rkeene:
Fixes #290

### DIFF
--- a/nano/core_test/interface.cpp
+++ b/nano/core_test/interface.cpp
@@ -26,7 +26,7 @@ TEST (interface, xrb_uint256_to_string)
 TEST (interface, xrb_uint256_to_address)
 {
 	nano::uint256_union zero (0);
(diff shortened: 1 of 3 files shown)


PlasmaPower:
I don't love making a very breaking change to rai_lib. Maybe add a version suffix?

But yes rust-rai-lib should be a drop-in replacement for rai_lib.

rkeene:
You mean version the symbol (does Windows support this?;  do we even create a shared object ?) or create a new symbol with a different name and different semantics ?

PR #1228, opened 2018-09-24
For Docker, log output to a file (unrotated !)

rkeene:
Within Docker, we don't capture the output so users may have a hard-time finding information about failed `release_assert ()` s  which #1226 added even more of


PlasmaPower:
Couldn't we just enable log_to_cerr in the default config?

rkeene:
Maybe we should rotate this log file before merging this.

PR #1229, opened 2018-09-25
Basic confirm_req improvements

rkeene:
Improves logging to indicate how many representatives it is broadcasting to in each batch, and delays the second group by 50ms from the first group (instead of 0ms).


SergiySW:
Not sure if we want to log each cycle again

rkeene:
Why did we stop logging each batch ?  I think it's helpful to know when those kick off

rkeene:
So that we can correlate the network activity to other things in the logs -- along with #1230 it should be much fewer than 20 lines in most cases

PR #1231, opened 2018-09-25
Ensure the async_write lambda is a closure around the buffer

rkeene:


rkeene:
I can't approve my own pull request !

rkeene:
@clemahieu 👍   That's the first part of the changeset, it just looks like the others so it's easy to miss

PR #1233, opened 2018-09-25
Fix vote check

rkeene:
Fix supplied in #964 was incorrect, updated

### DIFF
--- a/rai/node/lmdb.cpp
+++ b/rai/node/lmdb.cpp
@@ -1891,7 +1891,6 @@ std::shared_ptr<rai::vote> rai::mdb_store::vote_max (rai::transaction const & tr
 	auto current (vote_current (transaction_a, vote_a->account));
 	auto result (vote_a);
 
-
 	if (current != nullptr && current->sequence > result->sequence)
 	{
 		vote_result = rai::vote_code::replay;
--- a/rai/node/lmdb.cpp
+++ b/rai/node/lmdb.cpp
@@ -1896,7 +1896,7 @@ std::shared_ptr<rai::vote> rai::mdb_store::vote_max (rai::transaction const & tr
 	 * database or vote_cache, and "result" represents the
 	 * vote under question.
 	 */
-	if (current != nullptr && current->sequence > result->sequence)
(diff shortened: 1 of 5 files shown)


PlasmaPower:
I think we should get rid of that part entirely. Why's it necessary?

rkeene:
I think we need to update the `vote_max()` interface to return a `vote_code` so we know if it came from the DB/cache or the current vote was inserted (vote) or not (replay)

rai/node/lmdb.cpp
```diff
@@ -1884,15 +1885,24 @@ std::shared_ptr<rai::vote> rai::mdb_store::vote_generate (rai::transaction const
...
 
-std::shared_ptr<rai::vote> rai::mdb_store::vote_max (rai::transaction const & transaction_a, std::shared_ptr<rai::vote> vote_a)
+std::shared_ptr<rai::vote> rai::mdb_store::vote_max (rai::transaction const & transaction_a, std::shared_ptr<rai::vote> vote_a, rai::vote_code & vote_result)
 {
 	std::lock_guard<std::mutex> lock (cache_mutex);
 	auto current (vote_current (transaction_a, vote_a->account));
 	auto result (vote_a);
+
+
 	if (current != nullptr && current->sequence > result->sequence)
```
SergiySW: I guess this can be  
`current->sequence >= result->sequence`  
to prevent vote_code::vote result for current vote == new vote

rkeene: If current (DB/cache) is equal to the vote-in-question (result), we should yield "replay", right ?

SergiySW: I think so. Otherwise old solution with == was enough

rkeene:
After looking into `node.active.vote()` (`rai::active_transactions::vote()`) more, I think just removing the consideration altogether is good.  It will return true if a replay *OR* the vote is being ignored because it is being too spammy (from `existing->election->vote()`, aka `rai::election::vote()`).

rkeene:
The reason this code was correct before #964 was because it would always evaluate the second par of the `if (... || ...)` expression as false and thus irrelevant

PR #1237, opened 2018-09-26
 Escalation for long unconfirmed elections

SergiySW:
Start new elections for previous block & source if there are less than 100 active elections

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -3888,7 +3888,7 @@ void rai::active_transactions::announce_votes ()
 							auto previous (node.store.block_get (transaction, previous_hash));
 							if (previous != nullptr)
 							{
-								start (previous);
+								start (std::move (previous));
 							}
 						}
 						auto source_hash (node.ledger.block_source (transaction, *election_l->status.winner));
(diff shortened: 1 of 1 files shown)


rai/node/node.cpp
```diff
@@ -3878,6 +3878,30 @@ void rai::active_transactions::announce_votes ()
 					auto tally_l (election_l->tally (transaction));
 					election_l->log_votes (tally_l);
 				}
+				/* Escalation for long unconfirmed elections
+				Start new elections for previous block & source
+				if there are less than 100 active elections */
+				if (i->announcements % announcement_long == 1 && roots.size () < 100)
```
rkeene: Is this supposed to be:
`if (i->announcements / announcement_long > 0 && i->announcements % announcement_long == 1 && roots.size () < 100)`

So that it doesn't fire when i->announcements is 1 ?

SergiySW: It's inside of
`if (i->announcements > announcement_long)`

rkeene: Oops !

rkeene:
After testing, this creates a deadlock.  I'm going to try to revert it and see what happens.

rkeene:
@SergiySW Can you try adding another commit to this branch (SergiySW:election_escalation) to see if we can re-open this same pull request ?

rkeene:
D'oh, can you submit a new PR then ?

Issue #1235, opened 2018-09-25
Mutex deadlock


rkeene:
Thread #13:
        `rai::block_processor::process_blocks()` ->
        `rai::block_processor::process_receive_many()` locks 0x62400000bcd0 ->
        `rai::active_transactions::start()` ->
        `rai::active_transactions::start()` -> wait to lock 0x62400000a5b8

Thread #12:
        `rai::vote_processor::process_loop()` ->
        `rai::vote_processor::vote_blocking()` ->
        `rai::active_transactions::vote()` locks 0x62400000a5b8 ->
        `rai::election::vote()` ->
        `rai::election::confirm_if_quorum()` ->
        `rai::block_processor::force()` wait to lock 0x62400000bcd0

rkeene:
We now have 3 primary options for solving this (see above referencing pull requests).

PR #1238, opened 2018-09-26
Create a new mutex just to protect "processed_active"

rkeene:
Option 1 for fixing #1235


rkeene:
We are going with option 2 (#1239).

PR #1240, opened 2018-09-26
Enqueue forcing of blocks from confirm_if_quorum

rkeene:
This helps ensure a deadlock is not reached

Option 3 for issue #1235


rkeene:
We are going with option 2 (#1239)

PR #1241, opened 2018-09-27
RPC "sign" (hash, block)

SergiySW:
Allowing to sign hash or block with private key


cryptocode:
Instead of two separate actions, could this be a single general action (kind of like https://github.com/nanocurrency/raiblocks/pull/647) so it can be used to sign anything? The `message` can be a block or hash, but also whatever the callee wants.

SergiySW:
It was initially (hash & block in one RPC @sign", later may be messages. But @rkeene don't like it

rkeene:
Let's move this to V18.0 for now so we can figure out what semantics are best !

rkeene:
Gives some UBSan errors `/home/rkeene/devel/nano-node/nano/lib/jsonconfig.hpp:241:17: runtime error: load of value 110, which is not a valid value for type 'bool'`

SergiySW:
@rkeene hopefully fixed

nano/node/rpc.cpp
```diff
@@ -2954,6 +2959,92 @@ void nano::rpc_handler::send ()
...
+	if (hash_text.is_initialized ())
+	{
+		hash = hash_impl ();
+	}
+	// Retrieving block
+	std::shared_ptr<nano::block> block;
+	boost::optional<std::string> block_text (request.get_optional<std::string> ("block"));
+	if (!ec && block_text.is_initialized ())
+	{
+		block = block_impl ();
```
rkeene: Right now you have to pass in a value for `signature` for the block, even though it will overwrite it.  Maybe better to supply our own.
```suggestion
		block = block_impl (true);
```

nano/node/rpc.cpp
```diff
@@ -244,6 +278,29 @@ nano::amount nano::rpc_handler::amount_impl ()
...
+	if (!ec)
+	{
+		std::string block_text (request.get<std::string> ("block"));
+		boost::property_tree::ptree block_l;
+		std::stringstream block_stream (block_text);
+		boost::property_tree::read_json (block_stream, block_l);
+		if (!signature_work_required)
+		{
+			block_l.put ("signature", "0");
+			block_l.put ("work", "0");
```
rkeene: For atleast `sign()` it might be helpful not overwrite `work` if it was supplied.

rkeene: Nevermind, this only succeeds if the attribute doesn't already exist it seems.

nano/node/rpc.cpp
SergiySW: Commited

PR #1242, opened 2018-09-27
Lock ordering potential deadlock detection

PlasmaPower:
This will identify lock ordering issues and print out helpful traces when such issues are identified. For the traces to be helpful, you'll need to compile with `CFLAGS=-rdynamic CXXFLAGS=-rdynamic LDFLAGS=-rdynamic`, and you may need to clear your cmake build directory first.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,7 +120,7 @@ if (BOOST_ROOT)
     list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules")
 endif ()
 
-find_package (Boost 1.66.0 REQUIRED COMPONENTS filesystem log thread program_options stacktrace_basic)
+find_package (Boost 1.66.0 REQUIRED COMPONENTS filesystem log thread program_options)
 
 add_subdirectory(ed25519-donna)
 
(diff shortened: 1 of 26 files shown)


clemahieu:
Doesn’t TSAN detect potential deadlocks?

rkeene:
Yes, though it's really slow and won't consider database write transactions

clemahieu:
Why won’t it detect them?  Is lmdb being compiled with tsan on?

clemahieu:
Can we turn off other tsan options and turn on just deadlock detection to get the speed we need? https://github.com/google/sanitizers/wiki/ThreadSanitizerFlags
https://github.com/google/sanitizers/wiki/ThreadSanitizerDeadlockDetector
Reimplementing our own trading tools that senitizers already do is almost never a good option imo.

rkeene:
LMDB's locking mechanism isn't always based on mutexes -- it could be doing something else (and does on Windows, but maybe we don't care since we don't use Windows for most development)

PlasmaPower:
If we can get TSAN deadlock detection enabled by default in debug builds, that'd also work. I just think we need some detection for deadlocks by default.

rkeene:
Yes, the next step is to do a head-to-head of TSAN with deadlock detection-only and this dead-lock detection against known deadlocks.

Then pros and cons, con for TSAN: has to have a special build, may be slow; con for Custom: maintenance surface area, big changeset, new tool for everyone to learn

Etc.

PlasmaPower:
This has way too many merge conflicts, and TSAN is being used instead.

PR #1246, opened 2018-10-01
 Remove processing blocks from confirm_req

SergiySW:
To reduce unnecessary average nodes load.

- addditional disable successor confirmation for nodes with disabled voting in config

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -515,8 +515,6 @@ public:
 		}
 		node.stats.inc (rai::stat::type::message, rai::stat::detail::confirm_req, rai::stat::dir::in);
 		node.peers.contacted (sender, message_a.header.version_using);
-		node.process_active (message_a.block);
-		node.active.publish (message_a.block);
 		auto transaction_a (node.store.tx_begin_read ());
 		auto successor (node.ledger.successor (transaction_a, message_a.block->root ()));
 		if (successor != nullptr)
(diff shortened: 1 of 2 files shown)


clemahieu:
Can we split this in to 2 commits.  1 to remove process_active and the other to put confirm_block behind the check for vote enabling?

This way if we need to bisect commits to find which created the improvements or which caused an issue they're not together.

SergiySW:
Ok

rai/node/node.cpp
```diff
@@ -515,13 +515,15 @@ class network_message_visitor : public rai::message_visitor
 		}
 		node.stats.inc (rai::stat::type::message, rai::stat::detail::confirm_req, rai::stat::dir::in);
 		node.peers.contacted (sender, message_a.header.version_using);
-		node.process_active (message_a.block);
-		node.active.publish (message_a.block);
```
rkeene: This was added by vote-by-hash (#1025), probably to republish a block that's being voted on without the contents of the block.  I agree it seems un-needed here, atleast implicitly.  I will double-check with @PlasmaPower to find out of there's some other intention.

PlasmaPower: Yeah I agree that if we're not processing the block, we don't need to call .publish

rai/node/node.cpp
```diff
@@ -515,6 +515,7 @@ class network_message_visitor : public rai::message_visitor
 		}
 		node.stats.inc (rai::stat::type::message, rai::stat::detail::confirm_req, rai::stat::dir::in);
 		node.peers.contacted (sender, message_a.header.version_using);
+		node.process_active (message_a.block);
```
cryptocode: Both process_active and publish were removed, should both be put back ?

rkeene: I think it's safe to leave it out here from the previous discussion about it above.  Plus, then we'd have 2 more pull requests to change things around !

PR #1247, opened 2018-10-01
Change default desktop config.json

SergiySW:
set "enable_voting": "false", "bootstrap_connections_max": "4"
to reduce load on slow PCs


pedfx:
I'm having trouble when I change anything on config.json, it says `error deserializing config`
No matter what I change I can only open the node `.app` if I delete the config.json file and lost all the changes.

I recently installed a node on a Mac but I'm only used to how Linux configuration works d

PR #1254, opened 2018-10-03
Fix deadlock for elections escalation

SergiySW:
Replaces pull request #1237 

Start new elections for previous block & source if there are less than 100 active elections


SergiySW:
https://github.com/nanocurrency/raiblocks/pull/1253

PR #1255, opened 2018-10-03
Check epoch_link before checking balance of state blocks

PlasmaPower:
Checking the balance of the previous block may be slow if the previous block is not a state block.

Therefore, re-order the epoch test to check the link codes first before the more expensive test of checking to see if the delta_balance is zero.


SergiySW:
Then we can remove gap_previous check from epoch_block_impl

PR #1256, opened 2018-10-04
Add a max time to batch processes

PlasmaPower:


rai/node/node.cpp
```diff
@@ -1601,9 +1610,9 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> &
 {
 	{
 		auto transaction (node.store.tx_begin_write ());
+		auto start_time (std::chrono::steady_clock::now ());
 		lock_a.lock ();
-		auto count (0);
-		while (have_blocks () && count < 16384)
+		while (have_blocks () && std::chrono::steady_clock::now () - start_time < node.config.block_processor_batch_max_time)
```
rkeene: Maybe and also keep the count below 16384 ?

rkeene: Then again, probably not

PlasmaPower: Yeah I don't see a reason for a count limit. I think time is what we care about.

PR #1257, opened 2018-10-04
Iterative amount/balance visitor

cryptocode:
The amount and balance visitors are merged into a `summation_visitor` with functions for computing amount and balance. A loop emulates the mutual recursion with an emplace based helper stack. Other than that, I kept the visitor and computation loop logic as close to the original as I could.


PlasmaPower:
Do we have tests for this?

cryptocode:
This is a drop-in replacement that gets tested the same way as the old visitors (several tests employs the  ledger.balance and ledger.amount calls, but I'm not sure if they cover all eventualities.)

PlasmaPower:
> several tests employs the ledger.balance and ledger.amount calls

Ok, then this sounds good to me. :+1:

PR #1258, opened 2018-10-04
Name threads by role

rkeene:
In order to help correlate activities to the current running thread, name the threads when they are spawned

TODO

- [x] Add all thread roles
- [x] Name the threads (C++ thread equivalent of `pthread_setname_np()`)


PlasmaPower:
Why use an enum instead of just passing around a `const char *`?

rkeene:
So that it's easier to do things with, such as

https://github.com/nanocurrency/raiblocks/pull/1258/files#diff-125b1e30a208d9bab5533b3660857fc1R289

PlasmaPower:
Hmm, that does make sense. We could consider https://github.com/aantron/better-enums , simply comparing strings, or having dedicated boolean flags in TLS for stuff like IO threads.

rkeene:
What would be the benefit of separating the flag out from the tag  ?

Same for string comparison, we'd still have the same issue of coordinating the values among different parts of the code, but with no authoritative source.

rai/lib/utility.cpp
```diff
@@ -1,6 +1,72 @@
...
+				break;
+			case rai::thread_role::name::bootstrap_initiator:
+				thread_role_name_string = "Bootstrap init";
+				break;
+			case rai::thread_role::name::boost_io_context:
+				thread_role_name_string = "Boost io ctx";
+				break;
+		}
+
+		rai::thread_role::set_name (thread_role_name_string);
```
cryptocode: Maybe an `assert(thread_role_name_string.size() < 16)` to make it clear why the names are shortened (and to catch future "typo fixes")

rkeene: Good idea

PlasmaPower: It's not a huge deal though since they'll just be truncated on Linux

PR #1262, opened 2018-10-05
Factor out logging

cryptocode:
Part of https://github.com/nanocurrency/raiblocks/issues/1112

This moves the logging class into a separate file. This change showed that `boost::shared_ptr` was imported indirectly from the boost logging includes. No need to keep this dependency - moved to `std::shared_ptr`.


rai/node/node.cpp
```diff
@@ -4281,7 +4067,7 @@ void rai::thread_runner::join ()
 
 rai::inactive_node::inactive_node (boost::filesystem::path const & path) :
 path (path),
-service (boost::make_shared<boost::asio::io_service> ()),
+service (std::make_shared<boost::asio::io_service> ()),
```
rkeene: Is this change part of the same logical changeset ?

cryptocode: The removal of the boost logging includes prevents this from compiling. Instead of adding a new boost header dep, it seemed like a good opportunity to move this to std::shared_ptr (this is the only place where boost's shared_ptr is used, probably by mistake.)

rkeene: Thanks !

PR #1264, opened 2018-10-05
No write transactions on I/O threads

rkeene:
To ensure I/O threads are not blocked, add a debugging assert if a write DB transaction is created on one of those threads

Wesley: During normal node execution the common place write transactions are created are in the `ongoing_*` node functions executed from an `alarm`; the other place area is RPC calls. I've created a simple worker thread which gets forwarded the calls from the io threads. As these calls do not happen very regularly I didn't bother making a thread pool for it.

To make sure the RPC calls are working correctly all the tests now force the main thread to have an io-thread name which should catch the asserts for this,


cryptocode:
An alternative would be to use Boost's thread pool and submit work to that, like we do in the signature checker. That way we could have a configurable number of background workers instead of a single thread. But a single non-io thread is still an improvement so maybe the pool is for a future PR.

PR #1265, opened 2018-10-05
Remove process_active from confirm_req

rkeene:
Second half of pull request #1246


rkeene:
Merging this as a part of previously approved #1246, we just wanted to split it out for bisect'ability reasons.

PR #1266, opened 2018-10-05
Add support for lazily bootstrapping

### LINKED ISSUE #995: Implement a Lazy Bootstrapping Client
Currently the bootstrapping process is active and tries to import the entire ledger to every node, account by account.

In order to ensure that the blocks are valid in the face of forks, many blocks have to be voted on during bootstrapping, which is slow, bandwidth intensive, and I/O intensive.

However, a more lazy approach can be taken.  On the real-time peer-to-peer network, new blocks are constantly being voted on and, ultimately, confirmed by the network.  Each block is a node, and likely a head, of a Merkle tree where each node has 1 or 2 preceding nodes (the "previous" block, unless its the first block in a chain, and also the "link" or "source" block, if it is a receiving block).

Therefore, the confirmation of a single block also confirms MANY other blocks -- on a highly-connected (on the Merkle tree) account, it can confirm a significant amount of blocks without additional voting.

This ticket is to implement the client portion of this bootstrapping into the Nano node.

When a new block is confirmed on the network, the bootstrapping client should:

1. Verify it does not already have the block
  a. If so: Do nothing
2. Verify the block does not replace an existing block
  a. If so: Rollback
3. Determine the account the block belongs to
4. Determine the frontier for that account
5. Determine if there is a "gap" between the block and the frontier
  a. If not: Insert the block and continue
  b. If so, issue a "bulk_pull" on the bootstrapping network from the block just verified down to the current local frontier;  Process all blocks as below as well
6. If the block is a receiving block, the sending block must have already been confirmed, perform this entire procedure from that block down
7. Insert the confirmed blocks into the ledger database as confirmed blocks, transactionally

rkeene:
Closes #995 

Issue #995 goes into how it should work, but basically it will passively listen for votes on the real-time network, as soon as a block hits an irreversible threshold, the entire merkle tree can be confirmed and we can recursively (iteratively) add blocks to a queue to pull down using `bulk_pull` which supports the "start" parameter as a block.

For now the existing bootstrap client will still run to fill in the rest, but much less frequently.

For accounts in local wallets, we will support pulling those chains pro-actively (optionally without confirmation -- do you trust yourself to use this seed on only one node?) using `bulk_pull_account`.

### DIFF
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -2382,3 +2382,21 @@ void rai::frontier_req_server::next ()
 		current.clear ();
 	}
 }
+
+/*
+ * Lazy Bootstrapping Client
+ */
+rai::bootstrap_lazy::bootstrap_lazy ()
+{
+	return;
+}
+
+rai::bootstrap_lazy::~bootstrap_lazy ()
+{
+	return;
+}
+
+void rai::bootstrap_lazy::add_confirmed_block (std::shared_ptr<rai::block> block_a, rai::account const & account_a, rai::amount const & amount_a)
+{
+	return;
+}
--- a/rai/node/bootstrap.cpp
+++ b/rai/node/bootstrap.cpp
@@ -2386,8 +2386,13 @@ void rai::frontier_req_server::next ()
 /*
  * Lazy Bootstrapping Client
  */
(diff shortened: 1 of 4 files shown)


rkeene:
Closing this PR in favor of Serg's in #1332

PR #1269, opened 2018-10-06
Fix node.password_fanout test with running beta node

SergiySW:
Conflicting port 54000 default for beta & test networks


rkeene:
After looking at it some more, this is a bug since everything else on the test side uses 24000

PR #1273, opened 2018-10-08
Don't allocate callbacks for rai::socket on the heap

PlasmaPower:


cryptocode:
Closing as this will have to be redone with the tcp changes being done.

Issue #1275, opened 2018-10-09
Support Mojave dark mode


cryptocode:
The Qt wallet is barely usable under Mojave dark mode, a common problem with Qt applications apparently.

I think fixing this will require an upgrade to Qt 5.12 (planned release is late November) as no backporting will be done (https://bugreports.qt.io/browse/QTBUG-68891)

PR #1279, opened 2018-10-09
Force macOS wallet to use aqua light mode whilst QT adds support for …

elliottminns:
…Mojave Dark Mode

Dark mode for macOS won't be supported in QT until version 5.12 which is expected to be released in November. Until that time, this PR adds a patch to force the macOS client to use light mode and ignore the system setting of dark mode. References #1275


cryptocode:
Cool. Might solve https://github.com/nanocurrency/raiblocks/issues/1275

elliottminns:
Definitely a temporary fix until it can be supported fully.

PR #1283, opened 2018-10-09
 Add config option for minimum local representative weight

SergiySW:
Default 1000 Nano
Large services can set higher values to limit votes CPU load, while singe users willing to have small working representative can set it lower


clemahieu:
The purpose of this patch is so node operators can choose a weight amount which much be reached for an account before the node will spend effort sending out votes.

LGTM

rkeene:
Tests fail:
```
[  FAILED  ] node.fork_publish
[  FAILED  ] node.rep_self_vote
```

SergiySW:
@rkeene fixed in https://github.com/nanocurrency/raiblocks/pull/1281

rkeene:
Oh I see, I thought this was targeting master

PR #1285, opened 2018-10-10
 block_balance DB table (draft)

SergiySW:
to improve I/O of balance/amount visitors

data.ldb size
Before: 5302931456
After:    5699977216


clemahieu:
Can we put the balance in the value of the block table? hash -> [block_data, balance]

Then we don’t need a separate table and it should pack more efficiently.

SergiySW:
make receive_v1, open_v1, change_v1 tables?

clemahieu:
We could detect the length. If len = [block_data] it needs update. If len = [block_data, balance] no update needed.

rkeene:
This will tie into the change to store block confirmation height (#1303), which has been moved to V18.0

SergiySW:
Tied with DB changes in https://github.com/nanocurrency/raiblocks/pull/1303

SergiySW:
@zhyatt

SergiySW:
Closing in favor of https://github.com/nanocurrency/raiblocks/tree/block_sideband

PR #1287, opened 2018-10-10
Better changelog generation

rkeene:
Improve the changelog generation script to produce higher quality output


rkeene:
This doesn't change anything in the build pipeline so there's no point in waiting for CI changes to finish up

PR #1291, opened 2018-10-11
Add mutex around peers.get<T>()

rkeene:
There seems to be a missing mutex around this call to a huge boost multi-index item


rkeene:
Originally broken in [ae55a19172033e5de61a64ac2f46c66bced91af0]

PR #1294, opened 2018-10-11
Republish & request confirmation for new elections immediately

SergiySW:


PlasmaPower:
I believe this was implemented as a separate PR

PR #1300, opened 2018-10-13
Add weight option and account filter to representative_online rpc cal…

### LINKED ISSUE #1192: 2 feature requests for the RPC call 'representatives_online'
**Description of bug:**

Make a representatives_online RPC call

    "representatives": {
        "xrb_14nfwekkenidj3nczefx5snpe6do6d49e4f33cpaeqdq59hkc6tc6wxwkem5": "",
        "xrb_1cmtk7x7eeicbe434tqcjzhdiwxkaqdntbfj1tgxmufyidcdjw3wuk5p67d1": "",
        "xrb_1beta4nkzb3g6b1a1qhae89earmz3gk3kfrp3f8hyztm8qyjkeyz9kajfutq": "",
        "xrb_1beta5zm6rq4ghp6c9n99j8fz3am8e44dewae5ea7addsjtfo8umsafmr1r7": ""
    }

1) Include voting weight in the response. Or even better, add an option argument to the call, 'weight' with which the client can specify whether they want that information.
2) Provide an optional argument, 'account(s)' or 'representative(s)' with which the client can query the status of a specific representative(s).

CathalT:
…l. Fixes #1192

### DIFF
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -2528,11 +2528,48 @@ void rai::rpc_handler::representatives ()
 
 void rai::rpc_handler::representatives_online ()
 {
+	const auto accounts_node = request.get_child_optional ("accounts");
+	const bool weight = request.get<bool> ("weight", false);
+	std::vector<rai::public_key> accounts_to_filter;
+	if (accounts_node.is_initialized ())
+	{
+		for (auto & a : (*accounts_node))
+		{
+			rai::public_key account;
+			auto error (account.decode_account (a.second.get<std::string> ("")));
+			if (!error)
+			{
+				accounts_to_filter.push_back (account);
+			}
+		}
+	}
 	boost::property_tree::ptree representatives;
 	auto reps (node.online_reps.list ());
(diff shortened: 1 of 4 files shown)


CathalT:
Hmm , will look into that failure, it was passing locally...

CathalT:
Ah looks like a timing issue, I see there are deadline / poll functions.

rai/node/rpc.cpp
```diff
@@ -2528,11 +2528,48 @@ void rai::rpc_handler::representatives ()
...
+	if (accounts_node.is_initialized ())
+	{
+		for (auto & a : (*accounts_node))
+		{
+			rai::public_key account;
+			auto error (account.decode_account (a.second.get<std::string> ("")));
+			if (!error)
+			{
+				accounts_to_filter.push_back (account);
+			}
```
PlasmaPower: If there's an error here, we should set `ec` to `nano::error_common::bad_account_number`, `break`, and not populate the response below (check `if (!ec)` below).

PR #1301, opened 2018-10-13
New unit labels from legacy XRB to NANO style

elliottminns:
Updates the GUI labelling to use `NANO`, `Mnano`, `knano` and `nano` instead of `XRB`, `Mxrb`, `kxrb` and `xrb` respectively.

### DIFF
--- a/rai/qt/qt.cpp
+++ b/rai/qt/qt.cpp
@@ -732,7 +732,7 @@ wallet (wallet_a)
 			show_line_ok (*account_line);
 			this->history.refresh ();
 			auto balance (this->wallet.node.balance_pending (account));
-			auto final_text (std::string ("Balance (XRB): ") + wallet.format_balance (balance.first));
+			auto final_text (std::string ("Balance (NANO): ") + wallet.format_balance (balance.first));
 			if (!balance.second.is_zero ())
 			{
 				final_text += "\nPending: " + wallet.format_balance (balance.second);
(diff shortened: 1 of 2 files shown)


renesq:
To me, it seemed like those units are *not* pursued anymore, since both the NANO website *and* the NANO developer guide has removed any reference to those kind of units. Because they obviously did more harm than good, as they were incredibly confusing and weren't eliminated after each of the unit changes (rai->xrb->nano). This has lead the community to even introduce their own units like "naneroo" instead of small-caps-nano. 

A better choice in my opinion would be to use NANO, µNANO and mNANO instead of NANO/Mnano, knano and nano. Or officially introduce new names for fractions of NANO. Or make the Qt wallet compatible with decimals and/or raw 🤔

cryptocode:
Units are also updated in https://github.com/nanocurrency/raiblocks/wiki/Distribution,-Mining-and-Units#divider. Decimals are discussed in https://github.com/nanocurrency/raiblocks/issues/247.

rotilho:
I think the nano Nano is really confusing, I recently mixed both when developing [jnano-client](https://github.com/rotilho/jnano-commons/blob/master/src/main/java/com/rotilho/jnano/commons/NanoUnit.java#L14).

I decided to call nano "small nano" but it would be really nice have some kind of prefix for it like snano and keep MNano as just Nano

pedfx:
I agree it should call the `nano = 1000000000000000000000000, 10^24`   snano (or another letter before nano)

So we don't have any unit being called just Nano or nano or NANO, that is the most confusing thing possible.

It would be called Nano/NANO/nano just for the currency name and not used on any unit alone.

cryptocode:
The Qt wallet doesn't support decimals and this PR correctly uses the units defined in https://github.com/nanocurrency/raiblocks/wiki/Distribution,-Mining-and-Units#divider. A new PR can be made if we decide on new units or add decimal input support.

PR #1302, opened 2018-10-14
Separate vote generation process

clemahieu:
This patch decouples vote generation from quorum solicitation. Previously both were done in one large announce loop which on average would delay vote generation by 1/2 the announcement interval.

Now we announce votes as soon as a local decision is made i.e. a block is inserted in to the ledger.

### DIFF
--- a/rai/node/voting.cpp
+++ b/rai/node/voting.cpp
@@ -45,11 +45,13 @@ void rai::vote_generator::send (std::unique_lock<std::mutex> & lock_a)
 		hashes.pop_front ();
 	}
 	lock_a.unlock ();
-	auto transaction (node.store.tx_begin_write ());
-	node.wallets.foreach_representative (transaction, [this, &hashes_l, &transaction](rai::public_key const & pub_a, rai::raw_key const & prv_a) {
-		auto vote (this->node.store.vote_generate (transaction, pub_a, prv_a, hashes_l));
-		this->node.vote_processor.vote (vote, this->node.network.endpoint ());
-	});
+	{
+		auto transaction (node.store.tx_begin_read ());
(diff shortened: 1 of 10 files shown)


PlasmaPower:
I wouldn't remove the announcements in `rai::active_transactions::announce_votes`.

clemahieu:
Why?

PlasmaPower:
We could ignore them for the first interval, but I think they're important for ensuring confirmation.

clemahieu:
Not having received quorum doesn’t imply our created vote was missed. If our vote was missed someone would solicit it with confirm_req, receive the reply, then relay it out to the network.

PlasmaPower:
If a lot of other peoples' votes are missing, most likely it's due to network congestion, which means that our votes are likely missing too.

rai/node/voting.cpp
```diff
@@ -0,0 +1,94 @@
+#include <rai/node/voting.hpp>
+
+#include <rai/node/node.hpp>
+
+rai::vote_generator::vote_generator (rai::node & node_a, std::chrono::milliseconds wait_a) :
+node (node_a),
+wait (wait_a),
+stopped (false),
+started (false),
+thread ([this]() { run (); })
```
rkeene: We need to also name this thread

rai/node/voting.hpp
```diff
@@ -0,0 +1,32 @@
...
+	void run ();
+	void send (std::unique_lock<std::mutex> &);
+	rai::node & node;
+	std::mutex mutex;
+	std::condition_variable condition;
+	std::deque<rai::block_hash> hashes;
+	std::chrono::milliseconds wait;
+	bool stopped;
+	bool started;
+	std::thread thread;
```
rkeene: We should move this to `boost::thread` as well as we did for our other threads so that we can initialize them with greater stack sizes.

PR #1303, opened 2018-10-14
Store confirmation height in database to allow synchronously querying confirmation

PlasmaPower:
As a byproduct, this results in all confirmed blocks being immediately cemented. Attempting to replace a confirmed block with a different confirmed block results in a crash (from `release_assert`).

New terminology:
- Block account height, aka block height or a block's account height: given a block, the block height of its account if the block was the frontier. For an open block, this is always 1.
- Block sideband: this is all the information stored alongside a block in the database. Previously, this was just the block's successor hash, but this PR also stores the block's account height.

New information in the database:
- `rai::account_info::confirmation_height`: the maximum account height of any confirmed block in the account. Any blocks in the account with an account height less than the confirmation height have been confirmed.
- `rai::pending_info::source_account_height`: the source block's account height. Used to check if a pending block has been confirmed.
- `rai::block_sideband::account_height`: the account height of the block. This is needed when we roll back a receive block and have to recreate the pending entry.

New node logic:
- When searching for pending incoming sends, we now always confirm_req the source's account frontier instead of the specific send block.
- `process_confirmed` is now recursively called for the confirmed block's previously unconfirmed dependencies, and this process also updates the confirmation_height of the account (note: this is not actually implemented with recursion to avoid running into a stack overflow).

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -2321,6 +2321,7 @@ rai::uint128_t rai::node::account_pending (rai::transaction const & transaction_
 {
 	rai::uint128_t result (0);
 	rai::account end (account_a.number () + 1);
+	std::unordered_set<rai::account> accounts_to_confirm;
 	auto add_pending ([&, this](rai::pending_key key, rai::pending_info info) {
 		if (include_unconfirmed || ledger.block_confirmed (transaction_a, key.hash))
 		{
@@ -2328,7 +2329,7 @@ rai::uint128_t rai::node::account_pending (rai::transaction const & transaction_
 		}
 		else
 		{
-			this->block_confirm (this->ledger.store.block_get (transaction_a, key.hash));
+			accounts_to_confirm.insert (info.source);
 		}
 	});
 	for (auto i (store.pending_v0_begin (transaction_a, rai::pending_key (account_a, 0))), n (store.pending_v0_begin (transaction_a, rai::pending_key (end, 0))); i != n; ++i)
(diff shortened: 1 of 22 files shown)


rai/node/rpc.cpp
```diff
@@ -1578,6 +1579,25 @@ void rai::rpc_handler::frontiers ()
 	response_errors ();
 }
 
+void rai::rpc_handler::is_confirmed ()
```
SergiySW: I suggest something more consistent with other RPCs, such as
"block_confirmed"

PlasmaPower: Should I also rename the rai::node and rai::ledger methods?

rai/node/node.cpp
```diff
@@ -2337,12 +2369,11 @@ class confirmed_visitor : public rai::block_visitor
 void rai::node::process_confirmed (std::shared_ptr<rai::block> block_a)
 {
 	auto hash (block_a->hash ());
-	bool exists (ledger.block_exists (hash));
-	// Attempt to process confirmed block if it's not in ledger yet
-	if (!exists)
+	bool exists;
 	{
+		// We always re-process the block to guarantee it's marked as confirmed
```
SergiySW: Wouldn't this result in terrible performance? Single write transaction for each confirmed block

PlasmaPower: Good point, though it could cause a lot of problems if block_confirmed sometimes returned false for blocks from the callback.

PlasmaPower: Maybe I could move the block observer call into process_return_one after processing a confirmed block. That'd actually make a lot of sense.

rkeene:
We're moving this to V18.0 as it needs more testing before we're willing to accept it.

PlasmaPower:
An separate implementation of this is making progress. This PR has way too many merge conflicts to be merged, so I'm closing it.

PR #1304, opened 2018-10-14
Add http keep alive for RPC callback. Fixes #1125

### LINKED ISSUE #1125: Update HTTP Callback to Support HTTP keep-alive
Currently the HTTP callback mechanism does not use HTTP "Connection: keep-alive"

After moving it to a plugin (#1096), look at adding support for keep-alive to improve performance.

CathalT:
Assuming this is what https://github.com/nanocurrency/raiblocks/issues/1125 is referring to?

Not sure on the actual timeout value itself, should I make it configurable?

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -1286,6 +1286,8 @@ stats (config.stat_config)
 							for (auto i (i_a), n (boost::asio::ip::tcp::resolver::iterator{}); i != n; ++i)
 							{
 								auto sock (std::make_shared<boost::asio::ip::tcp::socket> (node_l->service));
+								boost::asio::socket_base::keep_alive option (true);
+								sock->set_option (option);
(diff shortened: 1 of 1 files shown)


PlasmaPower:
Not quite, the keepalive needed is at the HTTP level, not the TCP level as this PR sets.

CathalT:
2nd part is the http keep alive or am I in the wrong place altogether? 

```
req->insert (boost::beast::http::field::connection, "keep-alive");
req->insert (boost::beast::http::field::keep_alive, "timeout=120");
```

cryptocode:
The callback isn't a persistent connection, it makes a new connection on every observed block.

PlasmaPower:
That tells the server that we support HTTP keepalive, except we don't, which would cause problems with servers that actually support HTTP keepalive.

CathalT:
Ahhh I get you, was looking for a quick issue from the list to get my feet wet, I guess this is not one :smiley:

PR #1306, opened 2018-10-15
Using decode_account instead of decode_hex in account_move RPC handler

### LINKED ISSUE #666: account_move is broken
I get:
`
{
"moved": "0"
}
`
The account shows up in the accounts_balance for the source wallet. 
The dest wallet exists. I am able to get the balances of the accounts in that wallet.
Nothing in the logs except the json being sent.

elliottminns:
This patch resolves an issue with the RPC handler for `account_move` incorrectly decoding accounts by hex. According to https://github.com/nanocurrency/raiblocks/wiki/RPC-protocol#account-move this method should accept addresses rather than public keys and therefore needed `decode_account` instead.  

Closes #666

### DIFF
--- a/rai/core_test/rpc.cpp
+++ b/rai/core_test/rpc.cpp
@@ -683,7 +683,7 @@ TEST (rpc, account_move)
 	request.put ("source", source_id.pub.to_string ());
 	boost::property_tree::ptree keys;
 	boost::property_tree::ptree entry;
-	entry.put ("", key.pub.to_string ());
+	entry.put ("", key.pub.to_account ());
 	keys.push_back (std::make_pair ("", entry));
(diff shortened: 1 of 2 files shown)


PlasmaPower:
This is technically a breaking change, though most likely nobody is using this RPC.

PlasmaPower:
We could do `error = account.decode_account (str) && account.decode_hex (str);` (we also should have error handling here).

clemahieu:
I’m fine with breaking this, accounts should never be moved around by hex because it discards the checksum.

The test RPC.account_move needs to be fixed.

elliottminns:
Test RPC.account_move updated to match the new behaviour.

Issue #1316, opened 2018-10-20
MSVC linker error rai_bootstrap_weights


CathalT:
This issue has been around for a while, I see that its been mentioned before:
https://github.com/nanocurrency/raiblocks/issues/747

Current workaround is a script to replace the variables in node.cpp.

I've raised a bug on visual studio and its currently under investigation:
https://developercommunity.visualstudio.com/content/problem/359987/declaring-local-extern-variable-causes-linker-erro.html

Changing the variables to globals under the "rai" namespace fixes the issue.
I think this might be preferable to the script.

e.g.

```
namespace rai
{
extern unsigned char rai_bootstrap_weights[];
extern size_t rai_bootstrap_weights_size;
}
```

I'll open a PR and see what you think, if there's some reason not to make the variables global.

PR #1317, opened 2018-10-20
Fix MSVC linker error using rai_bootstrap_weights

CathalT:
see #1316


CathalT:
No problem! Glad to help , I'm mainly a Windows dev myself so it it was definitely self beneficial haha

PR #1318, opened 2018-10-21
Insert race workaround

clemahieu:


PlasmaPower:
IMO either stopped should be set when confirmed is set, or stopped should be a method that returns confirmed || aborted.

clemahieu:
Let's do that as a followup patch.  This is just a rename and workaround for the tx commit race condition.

clemahieu:
Let's do that as a followup patch.  This is just a rename and workaround for the tx commit race condition.

PR #1321, opened 2018-10-21
Remove default argument to constructor.

clemahieu:


PlasmaPower:
IIRC this has problems, e.g. `outstanding`. I originally tried to implement it this way, but failed.

clemahieu:
Hmm interesting, I assumed since 'last' is set atomically when the last connection is destroyed it'd ensure it's the tail of the recursion cycle.

Any recollection of the behavior that happened when it didn't work right?

PlasmaPower:
A simple case is that after the work peers fail, they'll be removed from `outstanding` (and `need_resolve` has already been emptied), which means that the retry won't see any work peers to use.

I tried to fix this, but ran into other side effects which I don't remember.

clemahieu:
Hmm, this patch doesn't make sense anyway because we'd be passing the only reference to a weak pointer.

PR #1323, opened 2018-10-21
Qt fixes

clemahieu:


clemahieu:
The one failure looks to be build-time related.

PR #1334, opened 2018-10-28
Increase minimum protocol version to when epoch blocks were introduced (V15.2)

PlasmaPower:


PlasmaPower:
I guess this removes legacy peers, so we can get rid of that code.

PlasmaPower:
Done

rkeene:
We'll still have legacy peers in the future, just a different legacy -- some of those tests should probably stay but with a variable version number (probably the minimum we'll peer with ?  I think legacy has a more specific check)

PlasmaPower:
The term legacy in peer_container refers to peers which do not support node id.

PR #1335, opened 2018-10-28
Reply to confirm_reqs with a vote by hash

PlasmaPower:


SergiySW:
Do we have a test to check efficency of sending vote for hash (fork) & then fork block? Wouldn't vote be just ignored?

PlasmaPower:
Nope, votes are stored by their hash, and later the hash->block table is used when votes are tallied.

PR #1344, opened 2018-11-01
Log vote processing time

rkeene:
Log vote processing time so we can tell how many votes are coming in without the spam of logging every vote

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -805,7 +805,7 @@ void rai::vote_processor::process_loop ()
 				end_time = std::chrono::steady_clock::now ();
 				elapsed_time = end_time - start_time;
 				elapsed_time_ms = std::chrono::duration_cast<std::chrono::milliseconds> (elapsed_time);
-				BOOST_LOG (node.log) << boost::str (boost::format ("Processed %1% votes in %2% milliseconds (rate of %3% votes per second)") % votes_l.size () % elapsed_time_ms.count () % ((votes_l.size () * 1000ULL) / elapsed_time_ms.count ()) );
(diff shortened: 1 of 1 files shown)


rkeene:
We can ignore the AppVeyor failure above, it built successfully and there was some transient issue uploading the artifact

rai/node/node.cpp
```diff
@@ -791,6 +811,36 @@ void rai::vote_processor::process_loop ()
...
+					 * If the time spent was less than 100ms then
+					 * the results are probably not useful as well,
+					 * so don't spam the logs.
+					 */
+					log_this_iteration = false;
+				}
+			}
+			if (log_this_iteration)
+			{
+				if (elapsed_time_ms_int == 0)
```
cryptocode: Seems like this won't happen since we're only logging if > 100 ms.
As a result, BOOST_LOG can be moved to a flipped `if (elapsed_time_ms_int < 100)` to simplify the code.

rai/node/node.cpp
```diff
@@ -779,6 +785,20 @@ void rai::vote_processor::process_loop ()
...
+			{
+				/*
+				 * Only log the timing information for this iteration if
+				 * there are a sufficient number of items for it to be relevant
+				 */
+				log_this_iteration = true;
+			}
+			if (log_this_iteration)
+			{
+				start_time = std::chrono::steady_clock::now ();
```
cryptocode: Can move this to the above `if`?

rai/node/node.cpp
rkeene: I left it here in case we make the limits configurable

PR #1345, opened 2018-11-02
Reduce the amount of time spent holding the active_transactions lock.

clemahieu:


rai/node/node.cpp
```diff
@@ -3026,7 +3029,7 @@ bool rai::active_transactions::add (std::pair<std::shared_ptr<rai::block>, std::
 		if (existing == roots.end ())
 		{
 			auto election (std::make_shared<rai::election> (node, primary_block, confirmation_action_a));
-			roots.insert (rai::conflict_info{ root, election, 0, blocks_a });
+			roots.insert (rai::conflict_info{ root, election, static_cast<unsigned> (0 - 1), blocks_a });
```
cryptocode: I think we use `std::numeric_limits<unsigned>::max()` elsewhere

rkeene:
```suggestion
			roots.insert (rai::conflict_info{ root, election, std::numeric_limits<unsigned>::max(), blocks_a });
```

rkeene: Testing out the new "Suggested change"

PR #1331, opened 2018-10-26
Strict check for vote_code::vote

SergiySW:
highest sequence number required

PR #1346, opened 2018-11-02
Update to ask the database layer if the vote is a replay or not

rkeene:
Fix of #1233, which is a fix of #964, which addresses a bug in #962 (introduced in #850).

In this patch we ask the database to tell us if the vote is a replay or not, since it's the layer that knows this answer.


PlasmaPower:
This will do more harm than good, as it'll go back to sequence numbers being global, which means that out of order votes will result in permanently lost votes.

rkeene:
Also related is #1331

zhyatt:
Closing this out as invalid as we need to track account specific voting otherwise votes can be lost.

PR #1348, opened 2018-11-02
Make config upgrades more robust

cryptocode:
There's been a couple of instances where the version number haven't been changed in all places. This puts the json config version in a member so it's only one place to change it.

### DIFF
--- a/rai/node/logging.cpp
+++ b/rai/node/logging.cpp
@@ -80,7 +80,6 @@ bool rai::logging::upgrade_json (unsigned version_a, boost::property_tree::ptree
 			result = true;
 		case 3:
 			tree_a.put ("network_node_id_handshake", "false");
-			tree_a.put ("version", "4");
 			result = true;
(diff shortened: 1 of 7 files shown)


rai/node/logging.cpp
```diff
@@ -67,25 +67,23 @@ void rai::logging::serialize_json (boost::property_tree::ptree & tree_a) const
...
-			tree_a.put ("version", "2");
 			result = true;
 		case 2:
 			tree_a.put ("rotation_size", "4194304");
 			tree_a.put ("flush", "true");
-			tree_a.put ("version", "3");
 			result = true;
 		case 3:
 			tree_a.put ("network_node_id_handshake", "false");
 			tree_a.put ("version", "4");
```
rkeene: Was this missed ?

cryptocode: Yeah, good catch. Fixed.

PR #1353, opened 2018-11-02
Make signatures non-deterministic

PlasmaPower:
Advantage: resistance to side channel attacks
Disadvantage: signatures will be non-deterministic

### DIFF
--- a/ed25519-donna/ed25519.c
+++ b/ed25519-donna/ed25519.c
@@ -61,12 +61,15 @@ ED25519_FN(ed25519_sign) (const unsigned char *m, size_t mlen, const ed25519_sec
 	bignum256modm r, S, a;
 	ge25519 ALIGN(16) R;
 	hash_512bits extsk, hashr, hram;
+	unsigned char randr[32];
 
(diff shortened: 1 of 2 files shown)


renesq:
Are existing signature verification libraries compatible with this? 
Are existing signature creation procedures still valid for a long time to come?

PlasmaPower:
Yes and yes

PlasmaPower:
I should probably add that there's no way to tell if a signature was generated with this method, unless you either have the private key, or you see two different signatures (which means at least one was clearly generated non-deterministically).

SergiySW:
Deterministic signtures was one of the main features of Ed25119
http://blog.cr.yp.to/20140205-entropy.html
https://tools.ietf.org/html/rfc8032#section-8.2

Of course there are different opinions
https://cybermashup.files.wordpress.com/2017/10/practical-fault-attack-against-eddsa_fdtc-2017.pdf
https://eprint.iacr.org/2017/985.pdf

Could we make proper cryptography audit of this change?

PR #1354, opened 2018-11-05
Prevent rebroadcasting the result of a confirm_req

PlasmaPower:


rai/node/node.cpp
```diff
@@ -3079,7 +3082,7 @@ bool rai::active_transactions::vote (std::shared_ptr<rai::vote> vote_a)
 			processed = processed || result.processed;
 		}
 	}
-	if (processed)
+	if (processed && !vote_a->from_confirm_req ())
```
clemahieu: Can you move this in to republish_vote?

PlasmaPower: All the requirements of republishing are checked outside of the republish function itself, so I think this is more consistent.

PlasmaPower:
Replaced by #1382

PR #1355, opened 2018-11-06
RPC Account history filter

### LINKED ISSUE #1110: Enhancement Request: add optional account_filter option to account_history RPC
It would be helpful to easily filter the results of account_history to search for a specific account (or array of accounts) that have sent/received to the account.  At the moment the only filters are a limit/count and head block.  If an account_filter option was added it would be much easier to identify accounts that have sent to the account or received from the account.

One application would be to identify the net balance in/out from other accounts.

CathalT:
Fixes #1110


CathalT:
Need to have a look at those failing tests, probably timing again, seemed ok locally

rai/core_test/rpc.cpp
```diff
@@ -962,6 +962,28 @@ TEST (rpc, history)
...
+	boost::property_tree::ptree filtered_accounts;
+	filtered_accounts.push_back (std::make_pair ("", other_account));
+	request2.add_child ("account_filter", filtered_accounts);
+	request2.put ("count", 100);
+	test_response response2 (request2, rpc, system.service);
+	while (response2.status == 0)
+	{
+		system.poll ();
+	}
+	auto history_node2 (response2.json.get_child ("history"));
```
argakiig: currently this returns an error rather than a history node. I think you will find that using account_history rather than history as your RPC call will work better as history has been deprecated and does not function exactly the same as account_history would.

rai/core_test/rpc.cpp
```diff
@@ -962,6 +962,28 @@ TEST (rpc, history)
...
+	ASSERT_EQ (history_node2.size (), 2);
```
argakiig: This should be ASSERT_EQ (history_node2.size (), 1); if you are using account_history as there would be only one expected block relating to account2 on the genesis account chain

CathalT:
Must not have ran the test with those last two lines at all my bad. 
I went with creating a new test for "account_history". We should be ok now, unless you want me to refactor the tests a bit, I can spend a bit more time on it another day.

argakiig:
@CathalT can you rebase and check formatting?

CathalT:
Just gone on holidays for a week, I'll be back Sunday the 3rd of Feb

CathalT:
I'm having trouble building after updating to latest, cryptopp is failing to build some assembly files on windows.
Haven't got a chance to sort it out yet and there are a few more issues to sort for this PR to pass...

CathalT:
@argakiig those latest failures don't look related to my changes? I think we're looking good

```
[  FAILED  ] wallet.select_account
[  FAILED  ] wallet.seed_work_generation
```

wezrule:
@CathalT yeh those are intermittent failures don't worry. However could please run clang-format on your files (ci/clang-format-all.sh). You can see the diff near the top of the travis ci output complaining.

guilhermelawless:
@CathalT  there's a conflict

PR #1356, opened 2018-11-06
(Proposal) Fix multiple rpc callbacks for same host.

CathalT:
https://github.com/nanocurrency/raiblocks/issues/1073
Instead of looping through all resolved addresses, only iterate if there is a failure and try again.

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1823,8 +1823,6 @@ void nano::node::do_rpc_callback (boost::asio::ip::tcp::resolver::iterator i_a, 
 										BOOST_LOG (node_l->log) << boost::str (boost::format ("Callback to %1%:%2% failed with status: %3%") % address % port % resp->result ());
 									}
 									node_l->stats.inc (nano::stat::type::error, nano::stat::detail::http_callback, nano::stat::dir::out);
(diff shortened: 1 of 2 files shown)


cryptocode:
@CathalT any chance you could rebase this?

CathalT:
Will have a look this evening

CathalT:
Apologies lads, bit caught up over Christmas, will get around to it in the next day or two!

nano/node/node.cpp
```diff
@@ -1853,6 +1787,84 @@ nano::node::~node ()
...
+								}
+								else
+								{
+									if (node_l->config.logging.callback_logging ())
+									{
+										BOOST_LOG (node_l->log) << boost::str (boost::format ("Callback to %1%:%2% failed with status: %3%") % address % port % resp->result ());
+									}
+									node_l->stats.inc (nano::stat::type::error, nano::stat::detail::http_callback, nano::stat::dir::out);
+									++i_a;
+									node_l->do_rpc_callback (i_a, address, port, target, body, resolver);
```
cryptocode: Should this only be done in the `async_connect` callback? Seems like resolving has succeeded if we get to async_write/read.

PR #1362, opened 2018-11-08
Fix vote_cache

### LINKED ISSUE #1206: Fix `mdb_store::flush()`
we swap the vote cache before writing it to disk

PlasmaPower:
This adds a second cache layer, so that the vote store remains consistent during flushing.

Fixes #1206

### DIFF
--- a/rai/node/lmdb.cpp
+++ b/rai/node/lmdb.cpp
@@ -1829,12 +1829,12 @@ void rai::mdb_store::checksum_del (rai::transaction const & transaction_a, uint6
 
 void rai::mdb_store::flush (rai::transaction const & transaction_a)
 {
-	std::unordered_map<rai::account, std::shared_ptr<rai::vote>> sequence_cache_l;
 	{
 		std::lock_guard<std::mutex> lock (cache_mutex);
-		sequence_cache_l.swap (vote_cache);
+		vote_cache_l1.swap (vote_cache_l2);
+		vote_cache_l1.clear ();
 	}
-	for (auto i (sequence_cache_l.begin ()), n (sequence_cache_l.end ()); i != n; ++i)
+	for (auto i (vote_cache_l2.begin ()), n (vote_cache_l2.end ()); i != n; ++i)
 	{
 		std::vector<uint8_t> vector;
 		{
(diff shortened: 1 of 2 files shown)


rai/node/lmdb.cpp
```diff
@@ -1829,12 +1829,12 @@ void rai::mdb_store::checksum_del (rai::transaction const & transaction_a, uint6
 
 void rai::mdb_store::flush (rai::transaction const & transaction_a)
 {
-	std::unordered_map<rai::account, std::shared_ptr<rai::vote>> sequence_cache_l;
 	{
 		std::lock_guard<std::mutex> lock (cache_mutex);
-		sequence_cache_l.swap (vote_cache);
+		vote_cache_l2 = vote_cache_l1;
```
clemahieu: Since we're checking both l1 and l2, do we need to copy this, it seems like a swap would still work.
At the end of the for-look that checks l2 we could clear l2.

rai/node/lmdb.cpp
```diff
@@ -1829,12 +1829,12 @@ void rai::mdb_store::checksum_del (rai::transaction const & transaction_a, uint6
...
+		vote_cache_l1.clear ();
 	}
-	for (auto i (sequence_cache_l.begin ()), n (sequence_cache_l.end ()); i != n; ++i)
+	for (auto i (vote_cache_l2.begin ()), n (vote_cache_l2.end ()); i != n; ++i)
```
clemahieu: Actually accessing l2 here isn't safe because cache_mutex is no longer locked.

PlasmaPower: It's indeed safe because the function won't be called concurrently, and the function is the only place where it's written to.

rai/node/lmdb.cpp
PlasmaPower: > Since we're checking both l1 and l2, do we need to copy this, it seems like a swap would still work.
Yeah good point.

>  At the end of the for-look that checks l2 we could clear l2.
I believe that would result in worse performance. Also, we'd need to do it after the transaction expires, not at the end of the function.

rai/node/lmdb.cpp
clemahieu: Good point, looks right.

clemahieu: Good point, looks right.

PR #1365, opened 2018-11-08
Add vote_processor overflow mode

SergiySW:
If there are more than 256 * 1024 new unprocessed votes


rkeene:
I moved the follow-up work to a separate release so that we get the basic queue implemented sooner.

PR #1366, opened 2018-11-09
CLI --debug_profile_votes

SergiySW:
To check vote_processor speed


rkeene:
Can we split this into two different pull requests -- one with the --debug_profile_votes and the other with the changes to timing ?

SergiySW:
Yes. But probably CLI  will not return correct results with 100k ms delay

rkeene:
Seems reasonable

Missing a close parentheses, other than that we'll merge it:

```
/workspace/rai/node/node.cpp: In member function 'void rai::active_transactions::announce_loop()':
/workspace/rai/node/node.cpp:3090:91: error: expected ')' before ';' token
   condition.wait_for (lock, std::chrono::milliseconds (announce_interval_ms + extra_delay);
                                                                                           ^
```

PR #1368, opened 2018-11-09
Dedicate a full r hash block to attacker unknown bytes

PlasmaPower:
Follow up to #1353. I noticed that https://eprint.iacr.org/2017/985.pdf recommended dedicating a full hash block to the key and random bytes to prevent side channel attacks (I don't recall that from the paper I had previously read). This is also similar to how keys work in blake2b, and again there's really no reason not to do it.

### DIFF
--- a/ed25519-donna/ed25519.c
+++ b/ed25519-donna/ed25519.c
@@ -62,6 +62,7 @@ ED25519_FN(ed25519_sign) (const unsigned char *m, size_t mlen, const ed25519_sec
 	ge25519 ALIGN(16) R;
 	hash_512bits extsk, hashr, hram;
 	unsigned char randr[32];
+	static const unsigned char rzero[64] = {0};
 
 	ed25519_extsk(extsk, sk);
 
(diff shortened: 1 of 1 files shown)


PlasmaPower:
That paper was for SHA-512, though its strategies are likely applicable to blake2b (and the block size is the same for both).

ed25519-donna/ed25519.c
```diff
@@ -62,6 +62,7 @@ ED25519_FN(ed25519_sign) (const unsigned char *m, size_t mlen, const ed25519_sec
 	ge25519 ALIGN(16) R;
 	hash_512bits extsk, hashr, hram;
 	unsigned char randr[32];
+	static const unsigned char rzero[64] = {0};
```
clemahieu: Are these attacker unknown bytes?  This looks like 64 zeros.

PlasmaPower: Yeah it's just zeroes. The paper recommends padding it with zeroes or another constant to reach the 128 byte block size. I think the principle is that the next round will start with attacker unknown values.

rkeene: Should we assert mlen is 32 to ensure that padding is correct, and also include a comment to this effect
```suggestion
	/*
	 * Pad the message to 128 bytes in size; 32-byte signature; 32-byte random; 64-byte zero
	 */
	static const unsigned char rzero[64] = {0};
	assert (mlen == 32);
```

PlasmaPower: mlen is irrelevant here because that comes after the anti-attacker block

PR #1372, opened 2018-11-09
Cached local representatives list

SergiySW:
with vote_minimum for config.json to prevent accidental multiple account voting for large services

### DIFF
--- a/nano/node/wallet.cpp
+++ b/nano/node/wallet.cpp
@@ -1735,3 +1735,24 @@ MDB_txn * nano::wallet_store::tx (nano::transaction const & transaction_a) const
 	auto result (boost::polymorphic_downcast<nano::mdb_txn *> (transaction_a.impl.get ()));
 	return *result;
 }
+
+namespace nano
+{
+std::unique_ptr<seq_con_info_component> collect_seq_con_info (wallets & wallets, const std::string & name)
+{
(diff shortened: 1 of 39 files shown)


nano/node/wallet.cpp
```diff
@@ -1579,6 +1602,38 @@ void nano::wallets::clear_send_ids (nano::transaction const & transaction_a)
...
+		std::lock_guard<std::mutex> representatives_lock (wallet.representatives_mutex);
+		wallet.representatives.swap (representatives_l);
+	}
+}
+
+void nano::wallets::ongoing_compute_reps ()
+{
+	compute_reps ();
+	auto & node_l (node);
+	auto compute_delay (nano::nano_network == nano::nano_networks::nano_test_network ? std::chrono::milliseconds (10) : std::chrono::milliseconds (15 * 60 * 1000)); // Representation drifts quickly on the test network but very slowly on the live network
```
cryptocode: I wonder if `compute_reps` on a node with many wallets and a ton of accounts warrants a less frequent scan than 15 seconds (or a config option) - or is the scan fast enough?

SergiySW: 15 minutes

cryptocode: thanks 😆

PR #1373, opened 2018-11-10
Add account_version to RPC ledger, wallet_ledger

SergiySW:


rkeene:
Right now the account version is an internal artifact -- do we want to expose it and make that part of our public interface ?

SergiySW:
Yes, I guess it won't be required after epoch blocks propagation.

PR #1377, opened 2018-11-14
Add batch votes verification

SergiySW:


rai/node/node.cpp
```diff
@@ -823,11 +824,49 @@ void rai::vote_processor::vote (std::shared_ptr<rai::vote> vote_a, rai::endpoint
...
+	{
+		hashes.push_back (vote.first->hash ());
+		messages.push_back (hashes.back ().bytes.data ());
+		pub_keys.push_back (vote.first->account.bytes.data ());
+		signatures.push_back (vote.first->signature.bytes.data ());
+	}
+	/* Verifications is vector if signatures check results
+	validate_message_batch returing "true" if there are at least 1 invalid signature */
+	auto code (rai::validate_message_batch (messages.data (), lengths.data (), pub_keys.data (), signatures.data (), size, verifications.data ()));
+	std::deque<std::pair<std::shared_ptr<rai::vote>, rai::endpoint>> result;
```
rkeene: Maybe make this a bit more generic:

```suggestion
	decltype(vote_a) result;
```

rai/node/node.cpp
```diff
@@ -823,11 +824,49 @@ void rai::vote_processor::vote (std::shared_ptr<rai::vote> vote_a, rai::endpoint
...
+	for (auto & vote : votes_a)
+	{
+		hashes.push_back (vote.first->hash ());
+		messages.push_back (hashes.back ().bytes.data ());
+		pub_keys.push_back (vote.first->account.bytes.data ());
+		signatures.push_back (vote.first->signature.bytes.data ());
+	}
+	/* Verifications is vector if signatures check results
+	validate_message_batch returing "true" if there are at least 1 invalid signature */
+	auto code (rai::validate_message_batch (messages.data (), lengths.data (), pub_keys.data (), signatures.data (), size, verifications.data ()));
```
rkeene: We don't do anything with this, maybe we shouldn't bother storing it --  the comment above would need to be updated too, since it talks about the return value, which we don't use...

```suggestion
	rai::validate_message_batch (messages.data (), lengths.data (), pub_keys.data (), signatures.data (), size, verifications.data ());
```

rai/node/node.cpp
SergiySW: Makes sense

rai/node/node.cpp
SergiySW: Yes

rkeene:
Oops !

PR #1378, opened 2018-11-14
Ensure ".clang-format" works with current clang-format tool

rkeene:
"clang-format" changes the formatting occasionally and requires us to add new options to preserve our existing formatting, however the configuration file used do this adds new keys.  These new keys cannot be parsed by older versions of "clang-format".  To work around this, ignore keys which cannot be parsed by the local "clang-format" tool.


ci/check-commit-format.sh
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env bash
+
 REPO_ROOT=$(git rev-parse --show-toplevel)
+
+"${REPO_ROOT}/ci/update-clang-format"
```
cryptocode: Should `clang-format-all.sh` do the same? I run that before committing, so I'd get any errors (pre-commit hook calls check-commit-format.sh so I don't run that manually)

PR #1379, opened 2018-11-14
RPC confirmation_quorum: optionally include account/ip/rep weight

cryptocode:
If `peer_details` is true, add account/ip/rep weight to the `confirmation_quorum` RPC call.


cryptocode:
Added option to https://github.com/nanocurrency/raiblocks/wiki/RPC-protocol#confirmation-quorum

PR #1382, opened 2018-11-16
Replay votes in response to a confirm_req for an active block

PlasmaPower:
The base branch for this is releases/v16 because this is targeting a V16.3.


SergiySW:
We need some > 0.1% rep to test that

SergiySW:
@PlasmaPower is active_transactions mutex lock required here?

PlasmaPower:
Good catch, fixed

PR #1404, opened 2018-11-21
Fix secure rpc build

cryptocode:
The cmake restructuring removed the linker reference.


PlasmaPower:
Shouldn't the `add_library (${OPENSSL_LIBRARIES})` be in the `if (RAIBLOCKS_SECURE_RPC)` section above?

cryptocode:
`${OPENSSL_LIBRARIES}` is added to `target_link_libraries`- it's blank if building without secure rpc.

PR #1406, opened 2018-11-22
 Early detection for votes overflow

### LINKED ISSUE #1391: Implement queue management
Right now as a part of #1365 we drop votes when the voting queue gets full, we should implement a WRED-esque scheme to handle this more gracefully.

SergiySW:
Fixes #1391

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -873,17 +873,17 @@ void rai::vote_processor::vote (std::shared_ptr<rai::vote> vote_a, rai::endpoint
 		// Level 1
 		else if (votes.size () < 112 * 1024)
 		{
-			process = (representatves_1.find (vote_a->account) != representatves_1.end ());
+			process = (representatives_1.find (vote_a->account) != representatives_1.end ());
 		}
 		// Level 2
 		else if (votes.size () < 128 * 1024)
 		{
-			process = (representatves_2.find (vote_a->account) != representatves_2.end ());
+			process = (representatives_2.find (vote_a->account) != representatives_2.end ());
 		}
 		// Level 3
 		else if (votes.size () < 144 * 1024)
 		{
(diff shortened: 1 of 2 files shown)


rai/node/node.cpp
```diff
@@ -1899,6 +1953,20 @@ void rai::node::ongoing_rep_crawl ()
 	}
 }
 
+void rai::node::ongoing_rep_calculation ()
+{
+	auto now (std::chrono::steady_clock::now ());
+	vote_processor.calculate_weights ();
+	std::weak_ptr<rai::node> node_w (shared_from_this ());
+	// Repeat in 10 minutes
+	alarm.add (now + std::chrono::seconds (10 * 60), [node_w]() {
```
cryptocode: Minor: maybe `std::chrono::minutes` here, and the comment can be removed?

rai/node/node.cpp
```diff
@@ -863,7 +863,30 @@ void rai::vote_processor::vote (std::shared_ptr<rai::vote> vote_a, rai::endpoint
 	std::lock_guard<std::mutex> lock (mutex);
 	if (!stopped)
 	{
-		if (votes.size () < 256 * 1024)
+		bool process (rai::rai_network == rai::rai_networks::rai_test_network);
```
PlasmaPower: This initial value doesn't do anything as it'll be overridden later. I'd put this condition as the first if statement, and keep the rest as "else if"s.

PlasmaPower: I guess it does help if the rep is >= 5%. Was that intended?

SergiySW: Test network value required only for votes processor profiling, good catch

rai/node/node.cpp
SergiySW: Sure

PR #1409, opened 2018-11-27
[v17] Replay votes in response to a confirm_req for an active block

SergiySW:
Port of https://github.com/nanocurrency/raiblocks/pull/1382/


PlasmaPower:
IIRC compute_rep_votes is never called, so we can get rid of that function entirely.

I was hoping that V17 would be a good opportunity to do some kind of LRU cache instead of tying it to active_transactions, but this is still a lot better than nothing, and we can improve it later if we need to. :+1:

SergiySW:
If you have LRU cache in mind, we can merge this in 17.0 & not merge to master, wait for cache I guess
@rkeene ?

rkeene:
That's fine -- getting it fixed in V17.0 and doing something better in V18+ works

rai/node/voting.cpp
```diff
@@ -49,6 +49,16 @@ void rai::vote_generator::send (std::unique_lock<std::mutex> & lock_a)
 		auto transaction (node.store.tx_begin_read ());
 		node.wallets.foreach_representative (transaction, [this, &hashes_l, &transaction](rai::public_key const & pub_a, rai::raw_key const & prv_a) {
 			auto vote (this->node.store.vote_generate (transaction, pub_a, prv_a, hashes_l));
+			std::unique_lock<std::mutex> lock (node.active.mutex);
```
rkeene: MSVC is failing to compile this:

```
c:\projects\myproject\rai\node\voting.cpp(52): error C2143: syntax error: missing ')' before '.'
c:\projects\myproject\rai\node\voting.cpp(52): error C3484: syntax error: expected '->' before the return type
c:\projects\myproject\rai\node\voting.cpp(52): error C3613: missing return type after '->' ('int' assumed)
c:\projects\myproject\rai\node\voting.cpp(52): error C3646: 'active': unknown override specifier
c:\projects\myproject\rai\node\voting.cpp(52): error C3551: if a trailing return type is used then the leading return type shall be the single type-specifier 'auto' (not 'std::unique_lock<std::mutex>')
c:\projects\myproject\rai\node\voting.cpp(52): error C2059: syntax error: '.'
c:\projects\myproject\rai\node\voting.cpp(52): error C2059: syntax error: ')'
```

PlasmaPower: We've seen this type of problem before, it needs to be `this->node`. I wonder if we can enable a warning for this in GCC.

PR #1410, opened 2018-11-27
Ensure the block exists in account_history

rkeene:
When calling account_history with a block that does not exist, it currently fails to find the block resulting in dereferencing a NULL pointer.

This change makes it return an error.  Another option is to make it return an empty history for the hash.

### DIFF
--- a/rai/node/rpc.cpp
+++ b/rai/node/rpc.cpp
@@ -1751,7 +1751,7 @@ void rai::rpc_handler::account_history ()
 	{
 		if (!hash.decode_hex (*head_str))
 		{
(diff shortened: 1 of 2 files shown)


rai/node/rpc.cpp
```diff
@@ -1751,7 +1751,14 @@ void rai::rpc_handler::account_history ()
 	{
 		if (!hash.decode_hex (*head_str))
 		{
-			account = node.ledger.account (transaction, hash);
+			if (node.store.block_get (transaction, hash))
```
clemahieu: This can be blocks_exists to avoid a full deserialization

PR #1412, opened 2018-11-28
Add scalar support to wallets

PlasmaPower:
This is a big change, it has merge conflicts, and may not be good in its current state, but I think some of the patterns here will be useful for adding better scalar support to wallets.

With this change, keys generated by tools like https://github.com/PlasmaPower/nano-bip32 can be used with rai_node. That said, this PR doesn't contain an RPC for the insert_scalar function, though one shouldn't be hard to make (it's the same as insert_adhoc).

### DIFF
--- a/ed25519-donna/ed25519.c
+++ b/ed25519-donna/ed25519.c
@@ -23,8 +23,8 @@
 	Generates a (extsk[0..31]) and aExt (extsk[32..63])
 */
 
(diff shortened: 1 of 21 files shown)


PlasmaPower:
This has a lot of merge conflicts, so it'd probably be better to start a future implementation from scratch (though the same strategy could be used).

PR #1421, opened 2018-12-02
MDB improvements

argakiig:
MDB_NOSYNC could provide improvements with disk usage by not calling fsync after commit
MDB_NORDAHEAD improves memory usage by not trying to read ahead in the DB, This has no effect on windows but other OSes should benefit. Windows does not read ahead so it is not affected

### DIFF
--- a/nano/node/lmdb.cpp
+++ b/nano/node/lmdb.cpp
@@ -25,7 +25,8 @@ nano::mdb_env::mdb_env (bool & error_a, boost::filesystem::path const & path_a, 
 			release_assert (status3 == 0);
 			// It seems if there's ever more threads than mdb_env_set_maxreaders has read slots available, we get failures on transaction creation unless MDB_NOTLS is specified
 			// This can happen if something like 256 io_threads are specified in the node config
-			auto status4 (mdb_env_open (environment, path_a.string ().c_str (), MDB_NOSUBDIR | MDB_NOTLS, 00600));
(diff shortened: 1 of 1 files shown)


cryptocode:
Yeah I think this would need to be a flag since it can more easily cause corruption. A pattern I've seen is "fast flags", e.g. MDB_NOSYNC is fast and MDB_NOSYNC | MDB_WRITEMAP | MDB_MAPASYNC is fastest, but the default should be neither.

argakiig:
if MDB_NOSYNC is enabled without MDB_WRITEMAP you still maintain the ACI, but the cost of durability. the risk is you lose some final transactions and have to bootstrap them later. How is this different from the node being offline for extended periods.

cryptocode:
MDB_NOSYNC can corrupt the database on system crashes/outage.

argakiig:
ah, I see, let me see about getting a config flag option, up

cryptocode:
@argakiig seems like you're right about MDB_WRITEMAP still maintaining ACI according to docs, a config flag would be nice though.

clemahieu:
I think the first order of business is to figure out if there's any performance gain from these flags, if not we should document it and not make a change.

argakiig:
Will create a separate PR for testing MDB_NOSYNC/MDB_WRITEMAP with config options This is now just for MDB_NORDAHEAD which, when supported, will stop the DB from being loaded

PR #1434, opened 2018-12-08
Support external RPC servers via IPC

cryptocode:
Remaining
- [x] Add tests (write an IPC client class for this)
- [x] Document the framing format in the code
- [x] Add ipc logging option with upgrade path
- [x] Config upgrade path ~(merge https://github.com/nanocurrency/raiblocks/pull/1416 first)~
After reviewing/testing, will also add this:
- [x] Replace boost::bind with lambdas
- [x] Utilize the new nano::timer class
- [x] Introduce session id for improved logging

This adds support for external RPC servers that communicate with the node via IPC (currently domain sockets and tcp).

Using external web servers can benefit service scalability and security. IPC connections are persistent, which should reduce time_wait (file descriptor exhaustion) issues on busy servers.

The internal RPC server can thus be disabled in the config, yet have the functionality exposed via IPC. The JSON format is exactly the same as the internal server - clients only need to change URL. 

A proof-of-concept REST server in Go is available at https://github.com/nanocurrency/rpc-go.

The IPC gateway allows for other message payloads like protobuf (there's a PR for protobuf which would need some work to work with this PR)

Technically, the IPC gateway defines a simple framing format with an encoding flag. In this case, the encoding is 1 (json rpc) which expects a payload of 32-bit big-endian size + json, which is also the case for the response. The json payload is forwarded to the node's existing RPC action handler. More info will be added to the wiki.

#### Configuration
To test the rpc-go server, enable either tcp or local ipc in the node config. Do the corresponding change in rpc-go's config.json (which uses domain sockets by default)

The rpc_enable flag can still be active, to access the current RPC server.

### DIFF
--- a/nano/nano_node/daemon.cpp
+++ b/nano/nano_node/daemon.cpp
@@ -3,10 +3,9 @@
 #include <iostream>
 #include <nano/lib/jsonconfig.hpp>
 #include <nano/lib/utility.hpp>
-#include <nano/nano_node/daemon.hpp>b
+#include <nano/nano_node/daemon.hpp>
 #include <nano/node/ipc.hpp>
 #include <nano/node/working.hpp>
-#include <nano/nano_node/daemon.hpp>
 
 nano_daemon::daemon_config::daemon_config () :
 rpc_enable (false),
--- a/nano/nano_node/daemon.cpp
+++ b/nano/nano_node/daemon.cpp
@@ -145,7 +145,6 @@ void nano_daemon::daemon::run (boost::filesystem::path const & data_path, nano::
 					rpc->start (config.rpc_enable);
 				}
 				nano::ipc::ipc_server ipc (*node, *rpc);
-				config.node.disable_lazy_bootstrap = disable_lazy_bootstrap;
 				runner = std::make_unique<nano::thread_runner> (io_ctx, node->config.io_threads);
 				runner->join ();
 			}
--- a/nano/nano_node/daemon.cpp
+++ b/nano/nano_node/daemon.cpp
@@ -1,11 +1,12 @@
-#include <nano/lib/jsonconfig.hpp>
-#include <nano/lib/utility.hpp>
-#include <nano/nano_node/daemon.hpp>
-
 #include <boost/property_tree/json_parser.hpp>
 #include <fstream>
 #include <iostream>
+#include <nano/lib/jsonconfig.hpp>
+#include <nano/lib/utility.hpp>
+#include <nano/nano_node/daemon.hpp>b
+#include <nano/node/ipc.hpp>
 #include <nano/node/working.hpp>
+#include <nano/nano_node/daemon.hpp>
 
 nano_daemon::daemon_config::daemon_config () :
 rpc_enable (false),
@@ -139,10 +140,12 @@ void nano_daemon::daemon::run (boost::filesystem::path const & data_path, nano::
 				node->flags = flags;
 				node->start ();
 				std::unique_ptr<nano::rpc> rpc = get_rpc (io_ctx, *node, config.rpc);
-				if (rpc && config.rpc_enable)
+				if (rpc)
 				{
-					rpc->start ();
+					rpc->start (config.rpc_enable);
 				}
+				nano::ipc::ipc_server ipc (*node, *rpc);
+				config.node.disable_lazy_bootstrap = disable_lazy_bootstrap;
 				runner = std::make_unique<nano::thread_runner> (io_ctx, node->config.io_threads);
 				runner->join ();
 			}
--- a/nano/node/ipc.cpp
+++ b/nano/node/ipc.cpp
@@ -1,0 +1,425 @@
+#include <algorithm>
+#include <boost/array.hpp>
+#include <boost/asio.hpp>
+#include <boost/bind.hpp>
+#include <boost/endian/conversion.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+#include <boost/thread/thread_time.hpp>
+#include <cstdio>
+#include <fstream>
+#include <iostream>
+#include <nano/node/common.hpp>
+#include <nano/node/ipc.hpp>
+#include <nano/node/node.hpp>
+#include <nano/node/rpc.hpp>
+#include <thread>
+
+using namespace boost::log;
+
+std::string nano::error_ipc_messages::message (int ev) const
+{
+	switch (static_cast<nano::error_ipc> (ev))
+	{
+		case nano::error_ipc::generic:
+			return "Unknown error";
+		case nano::error_ipc::invalid_preamble:
+			return "Invalid preamble";
+	}
+}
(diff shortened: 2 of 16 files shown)


cryptocode:
@meltingice is working on an IPC lib and gRPC interface based on this PR

https://github.com/meltingice/nano-ipc-js
https://github.com/meltingice/nano-grpc

nano/nano_node/daemon.cpp
```diff
@@ -125,10 +125,12 @@ void nano_daemon::daemon::run (boost::filesystem::path const & data_path, nano::
...
 				node->start ();
 				std::unique_ptr<nano::rpc> rpc = get_rpc (io_ctx, *node, config.rpc);
-				if (rpc && config.rpc_enable)
+				if (rpc)
 				{
-					rpc->start ();
+					rpc->start (config.rpc_enable);
 				}
+				nano::ipc::ipc_server ipc (*node, *rpc);
+				config.node.disable_lazy_bootstrap = disable_lazy_bootstrap;
```
rkeene: Is this part of this changeset or is it from a merge error ?

cryptocode: It should look like

```
				nano::ipc::ipc_server ipc (*node, *rpc);
				runner = std::make_unique<rai::thread_runner> (service, node->config.io_threads);
				runner->join ();
```

meltingice:
FYI if anyone else runs into this issue... after updating to the latest commit on this PR, the node could no longer deserialize the config. It was missing the `node.logging.timing` config option, which didn't get added by a config upgrade. Works fine once added. Default is false.

cryptocode:
Yeah, the timing upgrade was recently added to master

meltingice:
I am getting an exception that's crashing the node when attempting to call `account_representative_set` via IPC on an unopened account. The crash does not happen when attempting the same call via the HTTP RPC.

The call I'm making is:

```
{
  "wallet": "523325FC5B61CB450035B32AAE11FE7F02BEB5D3F736EF59FC3C7B0FAD35C46C",
  "account": "xrb_118fd6yj4mzc17k3q9ugygdqwj8cxqia8hgrxijb1efahsmnwzk8qs3foxnt",
  "representative": "xrb_1x7biz69cem95oo7gxkrw6kzhfywq4x5dupw4z1bdzkb74dk9kpxwzjbdhhs"
}
```

And the crash message is:

```
libc++abi.dylib: terminating with uncaught exception of type std::__1::bad_weak_ptr: bad_weak_ptr
[1]    11184 abort      ./nano_node --daemon --data_path=/Users/ryanlefevre/RaiBlocks/
```

cryptocode:
@meltingice thanks for the report. This seems to be related to action handlers with asynchronous responses (wallet->change_async in this case). I'll look into it.

meltingice:
Just tested and appears to be fixed now. Thanks @cryptocode!

nano/node/ipc.cpp
```diff
@@ -0,0 +1,682 @@
...
+
+private:
+	nano::ipc::ipc_server & server;
+	nano::ipc::ipc_config_transport & config_transport;
+	std::unique_ptr<nano::thread_runner> runner;
+	std::unique_ptr<boost::asio::io_context> io_ctx;
+	std::unique_ptr<ACCEPTOR_TYPE> acceptor;
+};
+
+/** The domain socket file is attemped removed at both startup and shutdown. */
```
wezrule: > attemped 

typo

nano/node/ipc.cpp
```diff
@@ -0,0 +1,682 @@
...
+class nano::ipc::dsock_file_remover
+{
+public:
+	dsock_file_remover (std::string const & file_a) :
+	filename (file_a)
+	{
+		std::remove (filename.c_str ());
+	}
+	~dsock_file_remover ()
```
wezrule: Either make the class `final` (preventing subclassing) or make the destructor virtual

nano/node/ipc.hpp
```diff
@@ -0,0 +1,154 @@
...
+		ipc_config_domain_socket transport_domain;
+		ipc_config_tcp_socket transport_tcp;
+	};
+
+	/** The IPC server accepts connections on one or more configured transports */
+	class ipc_server
+	{
+	public:
+		ipc_server (nano::node & node, nano::rpc & rpc);
+		~ipc_server ();
```
wezrule: Same as earlier, make class `final` or make destructor virtual.

nano/node/ipc.hpp
```diff
@@ -0,0 +1,154 @@
...
+	public:
+		virtual ~ipc_client_impl () = default;
+	};
+
+	/** IPC client */
+	class ipc_client
+	{
+	public:
+		ipc_client (boost::asio::io_context & io_ctx_a);
+		~ipc_client ();
```
wezrule: Destructor should be virtual as it is being subclassed. It is currently empty so can be `= default`

nano/node/ipc.cpp
```diff
@@ -0,0 +1,682 @@
...
+};
+
+nano::ipc::ipc_client::ipc_client (boost::asio::io_context & io_ctx_a) :
+io_ctx (io_ctx_a)
+{
+}
+
+nano::ipc::ipc_client::~ipc_client ()
+{
+}
```
wezrule: Empty destructor can be removed and made = default in class definition

nano/node/ipc.hpp
```diff
@@ -0,0 +1,154 @@
...
+	/** IPC transport interface */
+	class transport
+	{
+	public:
+		virtual void stop () = 0;
+		virtual ~transport () = default;
+	};
+
+	/** Base class for transport configurations */
+	class ipc_config_transport
```
wezrule: This needs a virtual destructor as it is being subclassed

nano/node/ipc.cpp
```diff
@@ -0,0 +1,682 @@
...
+ * stay out of the header file.
+ */
+class client_impl : public nano::ipc::ipc_client_impl
+{
+public:
+	client_impl (boost::asio::io_context & io_ctx_a) :
+	io_ctx (io_ctx_a)
+	{
+	}
+	~client_impl ()
```
wezrule: Empty destructor can be removed

nano/node/ipc.cpp
```diff
@@ -0,0 +1,682 @@
...
+
+	void timer_expired ()
+	{
+		close ();
+	}
+
+	void timer_cancel ()
+	{
+		boost::system::error_code ec;
+		this->io_timer.cancel (ec);
```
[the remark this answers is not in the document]
cryptocode: Yeah, it's sometimes needed in lambdas and this was moved from one. Fixed.

nano/node/ipc.cpp
cryptocode: Seems like that may fail on some compilers, but let's try and see what CI thinks (https://stackoverflow.com/questions/8619708/must-provide-destructor-in-the-pimpl)

meltingice:
The JS client library has been updated to match the preamble changes. https://www.npmjs.com/package/nano-ipc

PR #1436, opened 2018-12-09
Minimum blocks in ledger to start effective lazy bootstrap

SergiySW:
15,000,000 for live network
4,475,000 for beta network

Preventing start of lazy bootstrap on fresh node without launch flags


rkeene:
I think with the work in #1437 to fallback to legacy bootstrapping if not enough peers are helping to meet the goals that this is no longer needed

PR #1438, opened 2018-12-11
Confirm dependent active elections

SergiySW:
If possible. Confirm elections for: previous block, source or link


rkeene:
Awesome !

SergiySW:
Of course with https://github.com/nanocurrency/raiblocks/pull/1303 (v18) we can confirm all dependencies, now it's limited only to active elections

zhyatt:
Depends on #1570

PR #1439, opened 2018-12-11
Added start of automatic memory stats

rkeene:
This will (when complete) give you a breakdown of all members of all items within the `rai::node` class, recursively from the `stats` RPC call when asked for `objects` type stats.

TODO:

- [x] Merge this into the `stats` RPC
- [ ] Automate collection of members
- [ ] Better mutex handling
- [ ] Integrate generation into build system


rai/node/rpc_memory_generated.hpp
```diff
@@ -0,0 +1,60 @@
+while (true)
+{
+	if (std::try_lock (node.active.mutex, node.alarm.mutex, node.block_processor.mutex, node.bootstrap.mutex, node.bootstrap_initiator.mutex, node.gap_cache.mutex, node.vote_processor.mutex) == -1)
```
clemahieu: How's this loop different than a lock_guard?

rkeene: It tries to lock all the mutexes, if any fail to be locked, it unlocks all that were locked and tries again -- to avoid lock ordering issues (by brute force)

clemahieu: Why don't we lock them individually and then release them?

rkeene: In the vein of `std::lock_guard` there is `std::scoped_lock` which we could use here

rkeene: We can do `std::lock` to get rid of the loop

rkeene: If we lock them individually we may lock them in the wrong order versus a competing regime of locks and cause a deadlock.  Since this is a diagnostic facility, we don't actually care if we are a bit slow in acquiring the locks to avoid deadlocks.

PR #1448, opened 2018-12-13
Count state blocks in rai::block_processor::full ()

SergiySW:


rai/node/node.cpp
```diff
@@ -1285,9 +1285,9 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> &
 	// Processing blocks
 	while ((!blocks.empty () || !forced.empty ()) && std::chrono::steady_clock::now () - start_time < node.config.block_processor_batch_max_time)
 	{
-		if (blocks.size () > 64 && should_log ())
+		if ((blocks.size () + state_blocks.size ()) > 64 && should_log ())
```
rkeene: Is this part correct ?  In this loop we're only dequeuing `blocks` (items from `state_blocks` were enqueued to `blocks` after being verified at the start of `process_receive_many()` with a call to `verify_state_blocks()`)

SergiySW: Yes, but I guess we want to see all blocks remaining to process (?)

rai/node/node.cpp
```diff
@@ -1285,9 +1285,9 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> &
...
 		{
-			BOOST_LOG (node.log) << boost::str (boost::format ("%1% blocks in processing queue") % blocks.size ());
+			BOOST_LOG (node.log) << boost::str (boost::format ("%1% blocks in processing queue") % (blocks.size () + state_blocks.size ()));
```
rkeene: Maybe for now, until we do merge in #1450:

```suggestion
			BOOST_LOG (node.log) << boost::str (boost::format ("%1% (+ %2%) blocks in processing queue") % blocks.size () % state_blocks.size ());
```

PR #1450, opened 2018-12-13
Verify more state blocks if blocks deque is empty

SergiySW:
This may be temporary solution as initially all unchecked state blocks should be placed in blocks deque (second verification is not required)

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -1233,7 +1233,6 @@ bool rai::block_processor::have_blocks ()
 
 void rai::block_processor::verify_state_blocks (std::unique_lock<std::mutex> & lock_a, size_t max_count)
 {
-	lock_a.lock ();
 	std::deque<std::pair<std::shared_ptr<rai::block>, std::chrono::steady_clock::time_point>> items;
 	if (max_count == std::numeric_limits<size_t>::max () || max_count >= state_blocks.size ())
 	{
@@ -1284,12 +1283,16 @@ void rai::block_processor::verify_state_blocks (std::unique_lock<std::mutex> & l
 		}
 		items.pop_front ();
 	}
-	lock_a.unlock ();
 }
 
 void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> & lock_a)
 {
-	verify_state_blocks (lock_a);
+	lock_a.lock ();
+	if (!state_blocks.empty ())
+	{
+		verify_state_blocks (lock_a);
+	}
+	lock_a.unlock ();
 	auto transaction (node.store.tx_begin_write ());
 	auto start_time (std::chrono::steady_clock::now ());
 	lock_a.lock ();
@@ -1314,6 +1317,7 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> & 
 			forced.pop_front ();
 			force = true;
 		}
+		lock_a.unlock ();
 		auto hash (block.first->hash ());
 		if (force)
 		{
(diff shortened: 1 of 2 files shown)


rai/node/node.cpp
```diff
@@ -1303,7 +1303,14 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> &
 			forced.pop_front ();
 			force = true;
 		}
+		/* Verify more state blocks if blocks deque is empty
+		Because verification is long process, avoid large deque verification inside of write transaction */
+		bool start_verification (blocks.empty () && !state_blocks.empty () && state_blocks.size () < 2048);
 		lock_a.unlock ();
+		if (start_verification)
+		{
+			verify_state_blocks (lock_a);
```
rkeene: Instead of making it conditional upon being below a certain threshold, can we instead modify `verify_state_blocks()` to accept a maximum number of blocks to verify (defaulting to all).

That way we maximally be within our quanta for the DB transaction (+ up to verification spill).

```suggestion
			verify_state_blocks (lock_a, 2048);
```

With the appropriate updates to `verify_state_blocks()`

rai/node/node.cpp
```diff
@@ -1303,7 +1303,14 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> &
 			forced.pop_front ();
 			force = true;
 		}
+		/* Verify more state blocks if blocks deque is empty
+		Because verification is long process, avoid large deque verification inside of write transaction */
+		bool start_verification (blocks.empty () && !state_blocks.empty () && state_blocks.size () < 2048);
```
rkeene:
```suggestion
		bool start_verification (blocks.empty () && !state_blocks.empty ());
```

(see below for reasoning)

rai/node/node.cpp
rkeene: Proposed `verify_state_blocks()` looks something like:

```
void rai::block_processor::verify_state_blocks (std::unique_lock<std::mutex> & lock_a, _sizetype_ max_count = ...::limits<inf>)
{
	lock_a.lock ();
	std::deque<std::pair<std::shared_ptr<rai::block>, std::chrono::steady_clock::time_point>> items;
	if (max_count == ...::limits<inf> || max_count >= state_blocks.size ())
	{
		items.swap (state_blocks);
	}
	else
	{
		auto keep_size (state_blocks.size () - max_count);
		items.resize (keep_size);
		std::swap_ranges (state_blocks.end () - keep_size, state_blocks.end (), items.begin ());
		state_blocks.resize (max_count);
		items.swap (state_blocks);
	}
	lock_a.unlock ();
...
```

rai/node/node.cpp
```diff
@@ -1231,11 +1231,21 @@ bool rai::block_processor::have_blocks ()
 	return !blocks.empty () || !forced.empty () || !state_blocks.empty ();
 }
 
-void rai::block_processor::verify_state_blocks (std::unique_lock<std::mutex> & lock_a)
+void rai::block_processor::verify_state_blocks (std::unique_lock<std::mutex> & lock_a, size_t max_count)
 {
-	lock_a.lock ();
```
rkeene: Let's add an assertion here that the lock is currently held

rai/node/node.cpp
```diff
@@ -1273,12 +1284,16 @@ void rai::block_processor::verify_state_blocks (std::unique_lock<std::mutex> & l
...
-	lock_a.unlock ();
 }
 
 void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> & lock_a)
 {
-	verify_state_blocks (lock_a);
+	lock_a.lock ();
+	if (!state_blocks.empty ())
+	{
+		verify_state_blocks (lock_a);
```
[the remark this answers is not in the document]
rkeene: Even that high creates a time where nothing is really happening.  Is there an optimal number of signatures to verify at once ?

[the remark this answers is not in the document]
SergiySW: Actually what do you mean by nothing is really happening ?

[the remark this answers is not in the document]
SergiySW: It doesn't seem effective to verify less than can be processed in 1 write transaction (5 seconds)
1000 signatures verification time (us) is ~ ./rai_node --debug_verify_profile_batch (average 40-70 ms for sample Linode with SIMD). So 30 seconds are around 300-400-500k signatures at once?
Optimal batch size is n*64 iirc, but irrelevant for thousands of signatures

PR #1454, opened 2018-12-14
Add epoch timepoint for election end to confirmation history

argakiig:


rai/node/rpc.cpp
```diff
@@ -1454,6 +1454,7 @@ void rai::rpc_handler::confirmation_history ()
 			boost::property_tree::ptree election;
 			election.put ("hash", i->winner->hash ().to_string ());
 			election.put ("duration", i->election_duration.count ());
+			election.put ("time", i->election_end.time_since_epoch().count());
```
rkeene: Formatting:
```suggestion
			election.put ("time", i->election_end.time_since_epoch ().count ());
```

rai/node/node.cpp
```diff
@@ -2891,6 +2891,7 @@ void rai::election::confirm_once (rai::transaction const & transaction_a)
 {
 	if (!confirmed.exchange (true))
 	{
+		status.election_end = std::chrono::steady_clock::now();
```
rkeene: Formatting:
```suggestion
		status.election_end = std::chrono::steady_clock::now ();
```

PR #1458, opened 2018-12-15
Remove virtual function specifiers where override is used

wezrule:
Virtual function specifiers have always been implicitly inherited from their base classes. So the use of the virtual keyword in these cases has been redundant but was recommended in pre-modern C++ so that it was more easily distinguishable from non-virtual functions when reading derived class definitions. However since C++11, the override specifier has superseded this with compile time safety, and so explicitly writing both in a derived class is not required. 

I propose that the virtual specifier is removed in any function declaration where both the override and virtual function specifiers are used. 

The following command was used in the root directory to find and replace all instances where virtual and override specifiers were used.
find -type f -print0 | xargs -0 sed -ri 's/(virtual\s+)(.*override)/\2/g'

There were 2 files which were affected, one was rai/lib/expected.hpp but as this is third party code it has been left as is.

This change has not checked whether there are places override should be used where it is currently not.

### DIFF
--- a/rai/node/rpc_secure.hpp
+++ b/rai/node/rpc_secure.hpp
@@ -14,7 +14,7 @@ public:
 	rpc_secure (boost::asio::io_service & service_a, rai::node & node_a, rai::rpc_config const & config_a);
 
 	/** Starts accepting connections */
-	virtual void accept () override;
+	void accept () override;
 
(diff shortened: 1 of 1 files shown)


clemahieu:
Agreed on this change.  Is this something that would be picked up by clang-tidy --modernize?  We should probably have that in the pipeline in addition to clang-format just to catch coding practices.

PR #1459, opened 2018-12-15
 Disable batch signature check for unchecked state blocks

SergiySW:
Because all of these blocks should be verified  before insertion into unchecked table
Improving write block_processor performance with state blocks chains at bootstrap

### DIFF
--- a/rai/core_test/node.cpp
+++ b/rai/core_test/node.cpp
@@ -1842,11 +1842,6 @@ TEST (node, block_processor_signatures)
 	auto send5 (std::make_shared<rai::state_block> (rai::test_genesis_key.pub, send3->hash (), rai::test_genesis_key.pub, rai::genesis_amount - 5 * rai::Gxrb_ratio, key3.pub, rai::test_genesis_key.prv, rai::test_genesis_key.pub, 0));
 	node1.work_generate_blocking (*send5);
 	send5->signature.bytes[31] ^= 0x1;
-	// Invalid signature to unchecked
-	{
-		auto transaction (node1.store.tx_begin_write ());
-		node1.store.unchecked_put (transaction, send5->previous (), send5);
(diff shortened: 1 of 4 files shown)


rai/node/node.cpp
```diff
@@ -1220,15 +1220,15 @@ bool rai::block_processor::full ()
 	return (blocks.size () + state_blocks.size ()) > 16384;
 }
 
-void rai::block_processor::add (std::shared_ptr<rai::block> block_a, std::chrono::steady_clock::time_point origination)
+void rai::block_processor::add (std::shared_ptr<rai::block> block_a, std::chrono::steady_clock::time_point origination, bool unchecked)
```
rkeene: The "`unchecked`" variable needs a better name

rai/core_test/node.cpp
```diff
@@ -1842,11 +1842,6 @@ TEST (node, block_processor_signatures)
 	auto send5 (std::make_shared<rai::state_block> (rai::test_genesis_key.pub, send3->hash (), rai::test_genesis_key.pub, rai::genesis_amount - 5 * rai::Gxrb_ratio, key3.pub, rai::test_genesis_key.prv, rai::test_genesis_key.pub, 0));
 	node1.work_generate_blocking (*send5);
 	send5->signature.bytes[31] ^= 0x1;
-	// Invalid signature to unchecked
```
rkeene: Let's find a way to keep this test

rkeene:
I think a better approach to this one might be to add a `bool signature_verified` member to the various block types, which can be serialized to the unchecked table

rkeene:
After talking to @SergiySW we will look back into this after #1483 (part of #1478) is merged in

rkeene:
Replaced by #1492, closing this one

PR #1461, opened 2018-12-15
Pessimisation fix for condition_variable notify_*

### LINKED ISSUE #1284: Avoid lock pessimisation
Locks on a given mutex shouldn't cover notifications on conditional variables over the same mutex as it leads to pessimization.

Reference: https://en.cppreference.com/w/cpp/thread/condition_variable/notify_all

One example is going from:
```
    free.push_back (data_a);
    condition.notify_one ();
```

to 

```
    {
        free.push_back (data_a);
    }
    condition.notify_one ();
```

An alternative is to `unlock` instead of scoping.

termoose:
This should fix #1284. Some of these locks are for setting `stopped` and `started` flags, those functions could get a cleaner look by using `std::atomic_bool` on those flags and avoiding the explicit lock/unlock.

### DIFF
--- a/rai/core_test/processor_service.cpp
+++ b/rai/core_test/processor_service.cpp
@@ -54,8 +54,10 @@ TEST (alarm, one)
 	std::mutex mutex;
 	std::condition_variable condition;
 	alarm.add (std::chrono::steady_clock::now (), [&]() {
-		std::lock_guard<std::mutex> lock (mutex);
-		done = true;
+		{
+			std::lock_guard<std::mutex> lock (mutex);
+			done = true;
+		}
 		condition.notify_one ();
 	});
 	boost::asio::io_service::work work (service);
@@ -76,8 +78,10 @@ TEST (alarm, many)
 	for (auto i (0); i < 50; ++i)
 	{
 		alarm.add (std::chrono::steady_clock::now (), [&]() {
-			std::lock_guard<std::mutex> lock (mutex);
-			count += 1;
+			{
+				std::lock_guard<std::mutex> lock (mutex);
+				count += 1;
+			}
 			condition.notify_one ();
 		});
 	}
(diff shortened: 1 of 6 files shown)


clemahieu:
I was trying to apply lock pessimisation avoidance before but in some case it seemed to create a race condition though how that would happen escapes me now.

As long as all shared variables that are modified are protected by the same mutex it should be safe to notify the condition outside the lock right?

Does this pass with a tsan run?

cryptocode:
Should be safe and beneficial: "The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock."

termoose:
Yes this is the way to go and should increase performance in some cases. It has been discussed [here](https://stackoverflow.com/questions/17101922/do-i-have-to-acquire-lock-before-calling-condition-variable-notify-one).

Unfortunately I'm not able to run Clang's ThreadSanitizer here on my Mac as it's not supported for some reason.

What do you think about using `std::atomic_bool`s for those flags? Should that be a separate issue? I'd be happy to look into it. Will for example turn code like this from `node.cpp`:

```cpp
  started = true;

  lock.unlock ();
  condition.notify_all ();
  lock.lock ();
```

into this

```cpp
  started.load(true); // can use std::atomic::load to make it explicit that we're dealing with atomics
  condition.notify_all ();
```

clemahieu:
Since the condition variable is tied to the mutex we can't avoid holding a mutex over setting the variable, for instance

T1 Check started (false)
T2 Set started (true)
T2 Notify listeners
T1 Wait on condition variable (stalled)

clemahieu:
Looks great, thanks!

PR #1462, opened 2018-12-15
Remove function template specializations

wezrule:
Function template specializations are not candidates for overload resolution, and should be avoided where possible. Herb Sutter describes the Dimov/Abrahams example explaining why:
http://www.gotw.ca/publications/mill17.htm

A recent detailed talk summarising the issues during CppCon2018 can be found [here](https://www.youtube.com/watch?v=NIDEjY5ywqU)

Typed template parameters can just have the type substituted directly using standard function overloading.


clemahieu:
Good to learn!

PR #1468, opened 2018-12-16
Add more support for building boost with different options

rkeene:
Support more ways to build Boost, which may be required on some platforms.

This adds "-C" to build with clang's "libc++", as well as "-f" to build a PIC version of all the objects, which will be required for PIE executables on newer versions of Ubuntu.

Additionally a "-v" option is added so you can specify greater debug levels to, for example, see the compile strings.


cryptocode:
Could we have a --help option for this one?

PR #1472, opened 2018-12-17
Remove random items from the vote uniquer set

### LINKED ISSUE #1471: Track down potential memory leak with rai::vote
A possible memory leak has been observed.  When bootstrapping a node on the live network, the vote uniquer `votes` container exceeds 1M entries, which is a bit unusual.

rkeene:
When removing items from the vote uniquer `votes` set, use a random item rather than always the first item.

Fixes #1471


rkeene:
Replaced by #1473

Issue #1474, opened 2018-12-18
Investigate periodic bootstrap slowness


rkeene:
Some users have reported issues bootstrapping V17.0RC3 -- investigate and resolve before V17.0 final

rkeene:
Improvements to bootstrap block processing:

- In `process_receive_many()`, in addition to `should_log()` log the first time through the `while` loop
- Ensure `process_receive_many()` is called no more frequently than 100ms
- Pull Request #1450 
- Pull Request #1459 (with revisions)
- Don't perform (block) signature checking on a separate thread -- it's not helpful (unless the work is split up and multithreaded)

rkeene:
This should be resolved.

PR #1476, opened 2018-12-18
Improve process_receive_many timing logging

rkeene:
Part of #1474 

Add logging to see where time is spent

### DIFF
--- a/rai/node/logging.cpp
+++ b/rai/node/logging.cpp
@@ -20,6 +20,7 @@ log_rpc_value (true),
 bulk_pull_logging_value (false),
 work_generation_time_value (true),
 upnp_details_logging_value (false),
+timing_logging_value (false),
 log_to_cerr_value (false),
 flush (true),
 max_size (16 * 1024 * 1024),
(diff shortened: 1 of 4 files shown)


rai/node/logging.cpp
```diff
@@ -83,6 +85,7 @@ bool rai::logging::upgrade_json (unsigned version_a, boost::property_tree::ptree
 			result = true;
 		case 4:
 			tree_a.put ("upnp_details", "false");
+			tree_a.put ("timing", "false");
```
rkeene: I'll move this to `case 5` if we decide to put this in V18 instead of V17

rai/node/node.cpp
```diff
@@ -1367,11 +1376,14 @@ void rai::block_processor::process_receive_many (std::unique_lock<std::mutex> &
...
+	auto first_time (true);
+	unsigned number_of_blocks_processed (0);
 	while ((!blocks.empty () || !forced.empty ()) && std::chrono::steady_clock::now () - start_time < node.config.block_processor_batch_max_time)
 	{
-		if ((blocks.size () + state_blocks.size ()) > 64 && should_log ())
+		if (should_log (first_time))
 		{
-			BOOST_LOG (node.log) << boost::str (boost::format ("%1% blocks in processing queue") % (blocks.size () + state_blocks.size ()));
+			first_time = false;
+			BOOST_LOG (node.log) << boost::str (boost::format ("%1% blocks (+ %2% state blocks) in processing queue") % blocks.size () % state_blocks.size ());
```
rkeene: Let's add `forced.size ()` here as well

PR #1477, opened 2018-12-18
Ensure that process_receive_many is buffered

rkeene:
Ensure that process_receive_many is given atleast 100ms between successive calls, to allow buffering additional blocks

Part of #1474

### DIFF
--- a/rai/node/node.cpp
+++ b/rai/node/node.cpp
@@ -1272,6 +1272,14 @@ void rai::block_processor::process_blocks ()
 			if (elapsed_time < min_time_between_calls)
 			{
 				lock.unlock ();
+				if (node.config.logging.timing_logging ())
+				{
+					auto sleep_time_ms (std::chrono::duration_cast<std::chrono::milliseconds> (min_time_between_calls - elapsed_time));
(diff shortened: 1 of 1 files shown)


rkeene:
```
[ RUN      ] node.vote_republish
/workspace/rai/core_test/node.cpp:1668: Failure
Value of: system.poll ()
  Actual: Deadline expired
Expected: 
[  FAILED  ] node.vote_republish (5179 ms)
[ RUN      ] node.vote_by_hash_republish
/workspace/rai/core_test/node.cpp:1710: Failure
Value of: system.poll ()
  Actual: Deadline expired
Expected: 
[  FAILED  ] node.vote_by_hash_republish (5181 ms)
```

SergiySW:
Some tests failing because vote received before blocks processing. So vote not included in any election & ignored

SergiySW:
block_processor.flush () can solve this. Or condition.wait_for () probably

rkeene:
Ultimately this is due to std::mutex fairness not being guaranteed and a more direct approach might be feasible

zhyatt:
@SergiySW Is this still valid?

Issue #1478, opened 2018-12-18
Profile and improve process_receive_one


rkeene:
Beta on tmpfs is processing one below expectations

rkeene:
Profiling indicates `unchecked_put` is to blame

rkeene:
I believe this work is complete with the improvements made in the two referenced pull requests -- good job team !

PR #1483, opened 2018-12-18
Remove dupsort from unchecked table

SergiySW:
Currently without upgrade path (clear unchecked before or after using patch)

Part of work on #1478

### DIFF
--- a/rai/node/lmdb.cpp
+++ b/rai/node/lmdb.cpp
@@ -1072,7 +1072,7 @@ void rai::mdb_store::upgrade_v10_to_v11 (rai::transaction const & transaction_a)
 
 void rai::mdb_store::upgrade_v11_to_v12 (rai::transaction const & transaction_a)
 {
-	//version_put (transaction_a, 12);
+	version_put (transaction_a, 12);
 	mdb_drop (env.tx (transaction_a), unchecked, 1);
 	mdb_dbi_open (env.tx (transaction_a), "unchecked", MDB_CREATE, &unchecked);
 }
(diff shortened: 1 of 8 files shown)


cryptocode:
Works great, significant speedup on my ssd test machine.

SergiySW:
Great! Then adding upgrade path

rai/node/lmdb.cpp
```diff
@@ -1770,18 +1770,17 @@ std::shared_ptr<rai::vote> rai::mdb_store::vote_get (rai::transaction const & tr
 std::vector<std::shared_ptr<rai::block>> rai::mdb_store::unchecked_get (rai::transaction const & transaction_a, rai::block_hash const & hash_a)
```
rkeene: An `unchecked_exists()` could also be helpful

SergiySW: unchecked_exists for key hash? Or by unchecked_key (uint256+uint256)?
If by block own hash, then we need extra lookup table

rkeene: unchecked_key (successor/alternate and block hash)

rkeene:
Getting some issue with this on the beta network:

```
[beta] nano-rpc 17.0> show stats -bootstrap
Blocks:
  Count     = 3116
  Unchecked = 6584909
  Total     = 6588025
[beta] nano-rpc 17.0>
```

But there are only 5 million blocks on the beta network 😄

rkeene:
It looks like my issue was that I didn't delete the DB beforehand (and before the upgrade path existed), dupsort=1 is still set

PR #1484, opened 2018-12-18
Faster block existence check when the block type is known

cryptocode:
After https://github.com/nanocurrency/raiblocks/pull/1483, `block_exists (stateblock)` is the next function that shows up in the profiling of `process_receive_one`. Currently when checking state blocks, four other tables are checked first. With this, the correct table is checked directly if the block type is known. In current `block_exists` (where the block type is unknown) is reimplemented in terms of the new one.

### DIFF
--- a/rai/secure/ledger.cpp
+++ b/rai/secure/ledger.cpp
@@ -217,7 +217,7 @@ void ledger_processor::state_block (rai::state_block const & block_a)
 void ledger_processor::state_block_impl (rai::state_block const & block_a)
 {
 	auto hash (block_a.hash ());
-	auto existing (ledger.store.block_exists (transaction, hash));
+	auto existing (ledger.store.block_exists (transaction, block_a.type (), hash));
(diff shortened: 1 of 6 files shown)


rai/secure/ledger.cpp
```diff
@@ -217,7 +217,7 @@ void ledger_processor::state_block (rai::state_block const & block_a)
 void ledger_processor::state_block_impl (rai::state_block const & block_a)
 {
 	auto hash (block_a.hash ());
-	auto existing (ledger.store.block_exists (transaction, hash));
+	auto existing (ledger.store.block_exists (transaction, block_a.type (), hash));
```
SergiySW: Wondering if we can directly put types here & for other similar ledger_processor items like rai::block_type::state

cryptocode: Yeah, I couldn't decide on that :)

rkeene: I think it'll be okay -- we can always change it if it shows up in profiling

PR #1485, opened 2018-12-18
Periodically wake up the block processor

rkeene:
Periodically wake up the block processor even if it has not been notified.
This can occur if a notify is missing, or if a notify was sent out while we were not waiting for it because there is a lot of lock/unlock activity in process_receive_many


cryptocode:
I wonder if `condition_variable` conditions (!) is a good approach in these cases. It's supposed to handle both lost and spurious wakeups [1]

[1] http://www.modernescpp.com/index.php/c-core-guidelines-be-aware-of-the-traps-of-condition-variables

PR #1490, opened 2018-12-19
Slight improvements of block_exists usage in bootstrap & ledger

SergiySW:


clemahieu:
Can a function on this be extracted?  Send_exists

SergiySW:
Sure we can extract it

PR #1492, opened 2018-12-21
 Store timestamp & verified signature status in unchecked table

SergiySW:
Additionally periodic unchecked_cleaning ()

This replaces #1459


SergiySW:
--debug_profile_bootstrap shows 47% improvements for beta network with no significant changes on live network due to slow write/read for large tables

SergiySW:
Additional unchecked lookup table with hashes. Preventing multiple inserting attempts in unchecked for same blocks

nano/secure/ledger.cpp
```diff
@@ -183,6 +183,19 @@ void ledger_processor::state_block (nano::state_block const & block_a)
 			{
 				prev_balance = ledger.balance (transaction, block_a.hashables.previous);
 			}
+			else if (result.verified == nano::signature_verification::unknown)
+			{
+				// Check for possible regular state blocks with epoch link (send subtype)
+				result.verified = validate_message (block_a.hashables.account, block_a.hash (), block_a.signature) ? nano::signature_verification::unknown : nano::signature_verification::valid;
```
rkeene: This would be more readable as a `if` statement rather than a ternary and I think we should optimize for readability here

```suggestion
				if (validate_message (block_a.hashables.account, block_a.hash (), block_a.signature))
				{
					result.verified = nano::signature_verification::unknown
				}
				else
				{
					result.verified = nano::signature_verification::valid;
				}
```

nano/secure/ledger.cpp
```diff
@@ -209,13 +222,15 @@ void ledger_processor::state_block_impl (nano::state_block const & block_a)
 	result.code = existing ? nano::process_result::old : nano::process_result::progress; // Have we seen this block before? (Unambiguous)
 	if (result.code == nano::process_result::progress)
 	{
-		// Revalidate blocks with epoch links
-		if (!valid_signature || ledger.is_epoch_link (block_a.hashables.link))
+		// Validate block if not verified outside of ledger
+		if (result.verified != nano::signature_verification::valid)
```
rkeene: If `result.verified` is `nano::signature_verification::invalid`, could we return `nano::process_result::bad_signature` without checking the signature again ?

```suggestion
		if (result.verified == nano::signature_verification::invalid)
		{
			result.code = nano::process_result::bad_signature;
		}
		if (result.code == nano::process_result::progress && result.verified != nano::signature_verification::valid)
```

nano/secure/common.cpp
```diff
@@ -304,6 +305,59 @@ nano::block_hash nano::pending_key::key () const
...
+verified (verified_a)
+{
+}
+
+void nano::unchecked_info::serialize (nano::stream & stream_a) const
+{
+	assert (block != nullptr);
+	nano::serialize_block (stream_a, *block);
+	nano::write (stream_a, account.bytes);
+	nano::write (stream_a, boost::endian::native_to_big (modified));
```
rkeene: There's not a lot of value in storing these as a specific endian-ness since we are going to delete this table on startup

nano/secure/common.cpp
```diff
@@ -304,6 +305,59 @@ nano::block_hash nano::pending_key::key () const
...
+	nano::write (stream_a, boost::endian::native_to_big (static_cast<uint8_t> (verified)));
```
rkeene: uint8_t is endian-free

nano/node/node.cpp
```diff
@@ -1365,26 +1366,33 @@ void nano::block_processor::flush ()
 
 bool nano::block_processor::full ()
 {
+	size_t full_size (node.flags.fast_bootstrap ? 1024 * 1024 : 65536);
```
rkeene: This is a good idea.  Ideally this would be a different PR, but let's leave the set of changes in this one.  We can build something different on top of this.  Maybe fast(1M)/normal(64K)/custom(user-selectable).

nano/secure/ledger.cpp
SergiySW: This is check if signature_verification is unknown or valid_epoch. Then we need to check it again
Invalid should't go this far

nano/secure/common.cpp
SergiySW: Seems valid, thanks! Removing endian stuff

nano/secure/common.cpp
SergiySW: Removing endian stuff

nano/node/node.cpp
SergiySW: Yeah, selectable is good idea, will make issue to track this

nano/secure/ledger.cpp
```diff
@@ -538,11 +577,17 @@ void ledger_processor::open_block (nano::open_block const & block_a)
...
+		{
+			result.code = validate_message (block_a.hashables.account, hash, block_a.signature) ? nano::process_result::bad_signature : nano::process_result::progress; // Is the signature valid (Malformed)
+		}
 		if (result.code == nano::process_result::progress)
 		{
-			result.code = nano::validate_message (block_a.hashables.account, hash, block_a.signature) ? nano::process_result::bad_signature : nano::process_result::progress; // Is the signature valid (Malformed)
+			assert (!validate_message (block_a.hashables.account, hash, block_a.signature));
+			result.verified = nano::signature_verification::valid;
+			auto source_missing (!ledger.store.block_exists (transaction, nano::block_type::send, block_a.hashables.source) && !ledger.store.block_exists (transaction, nano::block_type::state, block_a.hashables.source));
```
wezrule: I think it may be more readable to factor out the boolean operands into 2 variables then do something like:
`auto source_missing (not_send_block && not_state_block);`

It could also be done here but because it is a boolean OR || it means the right operand is called regardless, so maybe just refactor the whole boolean expression `block_a.hashables.source) || ledger.store.block_exists (transaction, nano::block_type::state, block_a.hashables.source` into a variable: https://github.com/nanocurrency/nano-node/blob/ab62b094e5d81a598da3f8b7842d10d2f4efddd7/nano/secure/ledger.cpp#L528

SergiySW: Thanks! Made new nano::block_store::source_exists () function for this

PR #1499, opened 2018-12-23
Remove ip_address_hash from std namespace

wezrule:
In similar vein to #1460  the ip_address_hash struct and specializations have been removed from polluting the std namespace.

### DIFF
--- a/nano/node/common.hpp
+++ b/nano/node/common.hpp
@@ -69,19 +69,6 @@ struct endpoint_hash<4>
 		return result;
 	}
 };
-}
-
-namespace std
-{
-template <>
-struct hash<::nano::endpoint>
-{
-	size_t operator() (::nano::endpoint const & endpoint_a) const
-	{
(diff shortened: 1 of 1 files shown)


rkeene:
After merging #1504, I force pushed an update to this branch.

nano/node/common.hpp
```diff
@@ -104,6 +91,19 @@ struct ip_address_hash<4>
 		return result;
 	}
 };
+}
+
+namespace std
```
rkeene: As far as I can tell, this remains in the namespace `::std`

wezrule: The diff itself is not very clear about what has moved into what namespace, might be easier to view the whole file instead. There are only certain things which are allowed to be added to the std namespace, std::hash specializations being one of them.

rkeene: Got it, thanks !

PR #1500, opened 2018-12-26
Disable pool_connection () for stopped bootstrap or bootstrap_client

### LINKED ISSUE #1496: File descriptor leak
File descriptor leak in bulk pull, bulk push, or frontier client.

SergiySW:
Fixes #1496


zhyatt:
@SergiySW Can you find out when this issue first showed up? And also if anything we have done recently has made this worse?  Trying to evaluate where to include this fix.

SergiySW:
Ptobably it's related to lazy_bootstrap implementation (https://github.com/nanocurrency/raiblocks/pull/1332/) especially https://github.com/nanocurrency/raiblocks/blob/master/rai/node/bootstrap.cpp#L594
But it also happening with previous versions or with disabled lazy bootstrap, just less often.

zhyatt:
Looking at including this in V18 unless there is an urgency to push earlier.

PR #1503, opened 2018-12-27
Improve requesting frontier performance

wezrule:
On an MSVC Release build I profiled bootstrapping a fresh node for 30 seconds (chosen arbitrarily) to see if there was anything obvious which easily be improved. I found that 7% of the time was spent inside rai::uint256_union::number() when requesting frontiers.

There is a boost function boost::multiprecision::import_bits() which takes a sequence of bytes and constructs the appropriate multiprecision number (and conversely export_bits() to do the opposite). I tried using this instead and there was a small performance improvement. Next I looked at the callers and the majority were from the less than operator rai::uint256_union::operator<. Inside it, it calls number() on each operand and does the comparison with the result. However I think this step is unnecessary, and the bytes can be compared directly with std::memcmp. Results shown below (time spent inside operator<):

Original: **7%**
Using boost::import_bits() inside number (): **3.18%**
Replacing number () with memcmp: **0.3%**

The numbers are even greater on a debug build (sometimes as high as 20% in the current build). While the import_bits() function change no longer affects this specific result, I have left it in as it is still beneficial. I also modified the *_union constructors to call export_bits to modify the bytes array with the passed in number. Some tests have been added to cover the changes to operator<.

### DIFF
--- a/rai/core_test/uint256_union.cpp
+++ b/rai/core_test/uint256_union.cpp
@@ -532,7 +532,7 @@ namespace
 template <typename Union, typename Bound>
 void assert_union_types ()
 {
-	static_assert ((std::is_same_v<Union, rai::uint128_union> && std::is_same_v<Bound, rai::uint128_t>) || (std::is_same_v<Union, rai::uint256_union> && std::is_same_v<Bound, rai::uint256_t>) || (std::is_same_v<Union, rai::uint512_union> && std::is_same_v<Bound, rai::uint512_t>),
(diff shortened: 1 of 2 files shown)


clemahieu:
Excellent find.

wezrule:
Thanks guys. I haven't tried this on a big-endian system though, I wonder if there could be some problems?

cryptocode:
The protocol implementation (at least partially) assumes little endian hosts as-is, so that should be fine.

wezrule:
@cryptocode ok, good to know, thanks!

clemahieu:
The byte format of uint256_union is big-endian and this is tested in uint256_union.big_endian_union_constructor.  The memcmp comparison here will sort them in the same order so this change is good.

cryptocode:
Ah yeah, that's the real reason why this is fine

Issue #1305, opened 2018-10-15
Rename repository "nano-node"


rkeene:
During the V18 development cycle let's rename the "raiblocks" repository to "nano-node".  We'll include all the milestones, issues, pull requests, code, wiki, etc in the rename.

The old raiblocks repository's code should be kept up-to-date by syncing it for a few more releases.  We'll disable issues and pull requests and milestones.  The wiki can either be updated to include references to the new page or synchronized from the new repository for a period of time like the code.

argakiig:
It should just be a matter of renaming the repo, or are you proposing we have two repos? and then phase one out?

cenkalti:
Please do not make 2 separate repos 🙏

argakiig:
when 18 is released updates to the wiki will be pushed again closing out the ticket

PR #1504, opened 2018-12-27
Rename internal project bits from RaiBlocks to Nano

### LINKED ISSUE #627: Update Repository name, description, and url to Nano
**Description of bug:**
Repository is currently called "raiblocks"

Topic line is "RaiBlocks is a cryptocurrency"

Website url for project is "http://raiblocks.net"

Repository should be called "nano"

Topic line should be "Nano is a cryptocurrency", maybe add key features of the cryptocurrency such as DAG / Block Lattice / Feeless / Instant

Website url for project should be "https://nano.org"

This should help with SEO when searching for nano source code / github

rkeene:
Rename internal project bits from RaiBlocks to Nano.  We've held off on this disruptive change for a long time to avoid breaking our velocity.  During the V18 release cycle we also plan to rename to the repository (#1305), which along with this change should complete the RaiBlocks -> Nano rename.

Partially fixes #627 -- the remaining work is scheduled to be fixed by #1305.


nano/lib/config.hpp
```diff
@@ -3,21 +3,24 @@
...
 /**
  * Network variants with different genesis blocks and network parameters
  * @warning Enum values are used for comparison; do not change.
  */
-enum class rai_networks
+enum class nano_networks
 {
 	// Low work parameters, publicly known genesis key, test IP ports
+	nano_test_network = 0,
 	rai_test_network = 0,
```
cryptocode: Is keeping both the rai and nano enum values intentional?

rkeene: I was leaving it in for backwards compatibility, but the option name that is specified was also renamed with no backwards compatibility so maybe it's not needed -- or maybe more backwards compatibility is needed.

cryptocode:
This builds and runs, and tests passes, on macOS 👍

PR #1507, opened 2018-12-28
Only build essential boost libraries in docker

wezrule:
In the Dockerfile used for continuous integration all the boost libraries are built; however we only need 4: thread, log, filesystem & program_options. Now this isn't a big issue as we use the option "--cache-from nanocurrency/nano-ci:latest", which picks up the boost files from the intermediate cache layer. But it would be a bit quicker getting the cache and also result in a smaller docker image.

I've made another adjustment as per the docker best practices ["Always combine RUN apt-get update with apt-get install in the same RUN statement"](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get). This is because, if we only changed the second RUN then we would not be getting the latest repositories from apt even though apt-get update is called in the first one, because it is taken from the cache.

It is also recommend to clear the apt cache with rm -rf /var/lib/apt/lists/* after calling apt-get install, however this caused problems getting ccache from travis-ci, so I have left this as is.

### DIFF
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -9,16 +9,8 @@ RUN apt-get update -qq && apt-get install -yqq \
     g++ \
     wget
 
-WORKDIR "/tmp"
-
-RUN wget -qO ${BOOST_BASENAME}.tar.gz ${BOOST_URL} && \
-    tar xzf ${BOOST_BASENAME}.tar.gz && \
-    cd ${BOOST_BASENAME} && \
-    ./bootstrap.sh --with-libraries=thread,log,filesystem,program_options && \
-    ./b2 -d0 link=static install && \
-    rm -rf ${BOOST_BASENAME} && \
-    rm -f ${BOOST_BASENAME}.tar.gz && \
-    cd ..
+ADD util/build_prep/bootstrap_boost.sh bootstrap_boost.sh
+RUN ./bootstrap_boost.sh -m
 
 ENV DEBIAN_FRONTEND=noninteractive
 
@@ -28,4 +20,3 @@ RUN apt-get update -qq && apt-get install -yqq \
     xorg xvfb xauth xfonts-100dpi xfonts-75dpi xfonts-scalable xfonts-cyrillic \
     cargo
 
-RUN rm -rf /tmp/*
--- a/docker/ci/Dockerfile
+++ b/docker/ci/Dockerfile
@@ -20,8 +20,6 @@ RUN wget -qO ${BOOST_BASENAME}.tar.gz ${BOOST_URL} && \
     rm -f ${BOOST_BASENAME}.tar.gz && \
     cd ..
 
-RUN apt-get remove wget
-
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update -qq && apt-get install -yqq \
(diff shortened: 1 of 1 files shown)


docker/ci/Dockerfile
```diff
@@ -14,16 +14,15 @@ WORKDIR "/tmp"
 RUN wget -qO ${BOOST_BASENAME}.tar.gz ${BOOST_URL} && \
     tar xzf ${BOOST_BASENAME}.tar.gz && \
     cd ${BOOST_BASENAME} && \
-    ./bootstrap.sh && \
+    ./bootstrap.sh --with-libraries=thread,log,filesystem,program_options && \
```
rkeene: A better change would probably be to use the "`bootstrap_boost.sh`" script in the `util/build_prep` directory which has a `-m` option to take care of this

wezrule: Thanks for the tip about using the -m option, exactly what I needed.

docker/ci/Dockerfile
```diff
@@ -9,24 +9,14 @@ RUN apt-get update -qq && apt-get install -yqq \
     g++ \
```
rkeene: Looks like we need Python as well:

```
-- Could NOT find PythonInterp (missing:  PYTHON_EXECUTABLE) 
-- Configuring incomplete, errors occurred!
```

wezrule: Boost is also installed in a slightly different location now so cannot be found.

In ci/build-travis.sh it uses cmake -DBOOST_ROOT=/usr/local, but bootstrap_boost.sh installs it in /usr/local/**boost**

rkeene: The environment variable BOOST_ROOT, if set, will override where `bootstrap_boost.sh` puts the boost install.

wezrule: > The environment variable `BOOST_ROOT`, if set, will override where `bootstrap_boost.sh` puts the boost install.

Didn't know that, thanks! I did try changing build-travis.sh to use /usr/local/boost which worked ok, but tried your suggestion instead which also works, and that is what I have now done.

There are however some warnings "New Boost version may have incorrect or missing dependencies and imported targets". I think this is because it is now using boost 1.69 instead of 1.66 but our version of FindBoost.cmake has not been updated to support it. I don't want to extend the scope of this task more than I need to so am using additional arguments "**-B 1.66**" to bootstrap_boost.sh to use the version of Boost that Docker was originally using before.

> Looks like we need Python as well:
> 
> ```
> -- Could NOT find PythonInterp (missing:  PYTHON_EXECUTABLE) 
> -- Configuring incomplete, errors occurred!
> ```
Actually the missing python message is an existing error in the travis-ci Linux build, but I have modified the Dockerfile to install python to resolve it.

rkeene: Thanks !

docker/ci/Dockerfile
```diff
@@ -1,32 +1,22 @@
...
-    cd ..
+    wget \
+    python
 
+ENV BOOST_ROOT=/usr/local
+ADD util/build_prep/bootstrap_boost.sh bootstrap_boost.sh
+RUN ./bootstrap_boost.sh -m -B 1.66
+RUN rm bootstrap_boost.sh
 
 ENV DEBIAN_FRONTEND=noninteractive
```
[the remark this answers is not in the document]
wezrule: Done

PR #1509, opened 2018-12-28
Disable voting by default

### LINKED ISSUE #1369: Disable voting by default
Disable voting by default (but don't alter existing configs) since most nodes don't contribute much by voting.

argakiig:
... unless user action taken. This will not affect a current config, only newly generated ones
fixes #1369


SergiySW:
Guess for test_network it should be true

qwahzi:
Are there any concerns that "Disable voting by default [#1509]" will reduce decentralization? Will new node operators (merchants, whales, exchanges, etc) know that they need to manually enable voting once they hit .1% voting weight?

Worst case, what if a bunch of node operators clean install a new node version at the same time and forget to enable voting (similar to the issues we had with the v17 upgrade)? Would it make more sense to have the default be no voting unless you have .1%+ voting weight?

PR #1512, opened 2018-12-28
Migrate existing data from ~/RaiBlocks to ~/Nano

rkeene:
As a part of the rename from RaiBlocks to Nano (#1504), we now default to a different data path.

This adds a migration from the old/legacy path to the new path.

### DIFF
--- a/nano/nano_node/entry.cpp
+++ b/nano/nano_node/entry.cpp
@@ -58,6 +58,18 @@ int main (int argc, char * const * argv)
 	}
 	boost::program_options::notify (vm);
 	int result (0);
+
+	if (!vm.count ("data_path"))
+	{
+		std::string error_string;
+		if (!nano::migrate_working_path (error_string))
+		{
+			std::cerr << error_string << std::endl;
+
+			return 1;
+		}
+	}
+
 	boost::filesystem::path data_path = vm.count ("data_path") ? boost::filesystem::path (vm["data_path"].as<std::string> ()) : nano::working_path ();
 	auto ec = nano::handle_node_options (vm);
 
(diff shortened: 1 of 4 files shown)


arzarif:
How would this affect Dockerized nodes? If the migration is occurring within the container’s namespace, wouldn’t operators need to accommodate this change with an additional host-mounted volume for persistence?

arzarif:
I'm running the latest docker image and, while you were correct regarding about not needing to adjust the volume mappings, I noticed the migration of RaiBlocks to Nano did not occur as expected. My node began bootstrapping from scratch from an empty Nano directory while the contents of RaiBlocks were left hanging.

```
[root@xrb ~]# ls -l RaiBlocks/
total 5932508
drwx------. 2 root root         83 Oct 27 02:00 backup
-rw-------. 1 root root       2689 Dec 19 15:55 config.json
-rw-------. 1 root root       2657 Nov 29 02:21 config.json.bak
-rw-r--r--. 1 root root 6074867712 Jan  3 17:26 data.ldb
-rw-------. 1 root root       8192 Jan  3 17:26 data.ldb-lock
drwxr-xr-x. 2 root root        201 Jan  3 16:52 log
[root@xrb ~]# ls -l Nano
total 9401736
-rw-------. 1 root root       2720 Jan  3 17:26 config.json
-rw-------. 1 root root 9261748224 Jan  3 19:20 data.ldb
-rw-------. 1 root root       8192 Jan  3 19:20 data.ldb-lock
drwx------. 2 root root         43 Jan  3 17:26 log
```

Oddly, backing up Nano and replacing it with the old contents of RaiBlocks resulted in some kind of issue with the node (even after blowing away config.json and letting it pull a fresh config). Calls to the RPC port were failing.  

```
[root@xrb ~]# docker stop enn_nanonode_1
enn_nanonode_1
[root@xrb ~]# mv Nano Nano.bak
[root@xrb ~]# cp -r RaiBlocks/ Nano
[root@xrb ~]# docker start enn_nanonode_1
enn_nanonode_1
[root@xrb Nano]#  curl -g -d '{ "action": "block_count" }' '[::1]:7076'
curl: (56) Recv failure: Connection reset by peer
```

Issue #1133, opened 2018-08-31
Tool to build local beta-style network


rkeene:
Tool to build local beta-style network with full ledger and a few nodes

rkeene:
I did not get to this, trying again in the next release.

zhyatt:
Related to #918

dsiganos:
This issue does not have a good enough description and is possibly done by @gr0vity-dev's work. Closing.

qwahzi:
For future readers, two of the related local-network projects by gr0vity are:

- [Nano-Local](https://github.com/gr0vity-dev/nano-local)
- [Nano-Bench](https://github.com/gr0vity-dev/nano-bench)

PR #1518, opened 2018-12-30
Request confirmation faster for new blocks

SergiySW:


nano/node/node.cpp
```diff
@@ -23,6 +23,7 @@ std::chrono::seconds constexpr nano::node::search_pending_interval;
 int constexpr nano::port_mapping::mapping_timeout;
 int constexpr nano::port_mapping::check_timeout;
 unsigned constexpr nano::active_transactions::announce_interval_ms;
+unsigned constexpr nano::active_transactions::confirmation_request_delay_ms;
```
cryptocode: Would be nice to use chrono units here to avoid use-site casts (like with arrival_time_min), but since this is consistent with announce_interval_ms etc, maybe we can do a separate cleanup patch for all uses.

SergiySW: Good point!

nano/node/node.hpp
```diff
@@ -124,6 +124,8 @@ class active_transactions
 	// Threshold to start logging blocks haven't yet been confirmed
 	static unsigned constexpr announcement_long = 20;
 	static unsigned constexpr announce_interval_ms = (nano::nano_network == nano::nano_networks::nano_test_network) ? 10 : 16000;
+	// Delay required for average network propagartion before requesting confirmation
+	static std::chrono::milliseconds constexpr confirmation_request_delay = std::chrono::milliseconds (1500);;
```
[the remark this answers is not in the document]
SergiySW: Thanks!

zhyatt:
Plan to test performance once #1133 is available

nano/node/node.cpp
```diff
@@ -345,8 +346,8 @@ void nano::network::broadcast_confirm_req (std::shared_ptr<nano::block> block_a)
 	auto list (std::make_shared<std::vector<nano::peer_information>> (node.peers.representatives (std::numeric_limits<size_t>::max ())));
 	if (list->empty () || node.peers.total_weight () < node.config.online_weight_minimum.number ())
 	{
-		// broadcast request to all peers
-		list = std::make_shared<std::vector<nano::peer_information>> (node.peers.list_vector (100));
+		// broadcast request to all peers (with max limit 2 * sqrt (peers count))
+		list = std::make_shared<std::vector<nano::peer_information>> (node.peers.list_vector (std::min ((size_t)100, 2 * node.peers.size_sqrt ())));
```
cryptocode: minor: (size_t) -> static_cast<size_t>

SergiySW: Thx!

SergiySW:
Should be rebased after https://github.com/nanocurrency/nano-node/pull/1657

PR #1521, opened 2018-12-31
Emit "nano_" prefixed addresses

rkeene:
This changeset changes the generated names to begin with "nano_" rather than "xrb_", as a follow-on to #1504 and the renaming of RaiBlocks to Nano.


wezrule:
I think both `xrb_uint256_to_address ()` and the buffer in `TEST (interface, xrb_uint256_to_address)` will need a buffer of 66 for this to work correctly. The latter was changed by me in a previous task to match `xrb_uint256_to_address` as only 65 were required then.

wezrule:
Closed in favour of #1771

PR #1523, opened 2018-12-31
Block sideband

rkeene:
Keep track of block meta data in the same LMDB record as the block, in a consistent and stable way.


zhyatt:
Closing as duplicate of #1554

PR #1524, opened 2018-12-31
Fix tests broken with #1510

argakiig:
Neglected to update tests to match status, fixes tests broken in #1510

### DIFF
--- a/nano/qt_test/qt.cpp
+++ b/nano/qt_test/qt.cpp
@@ -43,7 +43,7 @@ TEST (wallet, status)
 	auto wallet_has = [wallet](nano_qt::status_types status_ty) {
 		return wallet->active_status.active.find (status_ty) != wallet->active_status.active.end ();
 	};
-	ASSERT_EQ ("Status: Disconnected, Block: 1", wallet->status->text ().toStdString ());
+	ASSERT_EQ ("Status: Disconnected, Blocks: 1", wallet->status->text ().toStdString ());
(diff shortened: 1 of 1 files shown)


rkeene:
The Qt tests failed with:

```
[ RUN      ] wallet.startup_balance
/workspace/nano/qt_test/qt.cpp:74: Failure
Value of: wallet->self.balance_label->text ().toStdString ()
  Actual: ""
Expected: "Balance: 0 NANO"
[  FAILED  ] wallet.startup_balance (44 ms)
[ RUN      ] wallet.select_account
/workspace/nano/qt_test/qt.cpp:96: Failure
Value of: key4
  Actual: 32-byte object <EC-09 BE-40 0D-85 35-B0 F8-AA 7A-E5 87-6E FF-A4 96-94 D8-37 93-70 BC-F1 70-8E D2-9F 83-41 DD-B1>
Expected: key2
Which is: 32-byte object <53-97 54-B7 13-5A 26-9B A5-10 F4-92 6D-C5 E3-05 AC-86 AF-BE DD-18 97-54 01-21 E4-C9 F0-85 45-F8>
[  FAILED  ] wallet.select_account (48 ms)
```

Is this related, or an intermittently failing test ?

argakiig:
intermittent, what's happening there is since it uses a random seed, the index 0 may be ordered before or after index 1 when it's generated and so the selection will sometimes not change to the other account and has been around for a.while

argakiig:
although startup balance is a new one, I thought I had caused.with another PR but it's intermittent as well apparently

PR #1525, opened 2018-12-31
Split wallet from ledger

### LINKED ISSUE #210: Separate DB file for wallets
like wallets.ldb
To make easy backup-restore + easy ledger replacement for slow PC / 1st sync

argakiig:
Splits existing wallet away from the ledger into a separate wallets.ldb closes #210 
Things currently remaining
- [ ]  Locked wallets will not unlock until nano_node/wallet is restarted after migration
- [ ]  provide automated/rpc/cli based removal of wallet databases in the ledger db

### DIFF
--- a/nano/node/wallet.cpp
+++ b/nano/node/wallet.cpp
@@ -165,6 +165,8 @@ bool nano::wallet_store::attempt_password (nano::transaction const & transaction
 			case version_3:
 				upgrade_v3_v4 (transaction_a);
 			case version_4:
+				upgrade_v4_v5 (transaction_a);
+			case version_5:
 				break;
 			default:
 				assert (false);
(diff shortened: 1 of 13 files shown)


argakiig:
```[  FAILED  ] 3 tests, listed below:
[  FAILED  ] wallet.status
[  FAILED  ] wallet.startup_balance
[  FAILED  ] wallet.enter_password```

First and third fixed with #1524 
wallet.startup_balance may be related to an inability to unlock without restarting the node.

File: nano/node/wallet.cpp [hunk 1]
```diff
@@ -1507,7 +1566,10 @@ nano::store_iterator<nano::uint256_union, nano::wallet_value> nano::wallet_store
 {
 	return nano::store_iterator<nano::uint256_union, nano::wallet_value> (nullptr);
 }
-
+nano::mdb_wallets_store::mdb_wallets_store (bool & error_a, boost::filesystem::path const & path_a, int lmdb_max_dbs) :
+environment (error_a, path_a, lmdb_max_dbs, 1ULL * 1024 * 1024 * 1024)
```
wezrule: The same GB logic is referenced here as in the mdb_env() constructor in lmdb.cpp. I wonder if it would be beneficial to add our own user defined literal? Then it literally could be written as: 1GB. There is Boost.Units but it only has stuff like length, weight etc.. which is a bit annoying. It's probably a bit out of scope to do it here, but if you are feeling adventurous could always add our own! :)

argakiig:
closing in favor of #1610

PR #1526, opened 2019-01-01
Timer utilities

cryptocode:
Documentation started here: https://github.com/cryptocode/notes/wiki/Timer

I've introduced it into a couple of node.cpp functions; seems to simplify the timing code a fair amount.

### DIFF
--- a/nano/lib/timer.hpp
+++ b/nano/lib/timer.hpp
@@ -2,7 +2,7 @@
 
 namespace nano
 {
-enum timer_state
+enum class timer_state
 {
 	started,
 	stopped
@@ -102,6 +102,14 @@ public:
 		return ticks;
 	}
 
+	/**
+	 * Return current units.
+	 */
+	UNIT value ()
+	{
+		return ticks;
+	}
+
 	/** Returns the duration in UNIT since the timer was last started. */
 	UNIT since_start ()
 	{
@@ -171,8 +179,7 @@ public:
 	/** Returns the SI unit string */
 	std::string unit () const
 	{
-		UNIT u;
-		return typed_unit (u);
+		return typed_unit<UNIT> ();
 	}
 
 	nano::timer_state current_state () const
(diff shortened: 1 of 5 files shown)


nano/lib/timer.hpp
```diff
@@ -0,0 +1,250 @@
...
+	std::vector<timer> children;
+	nano::timer_state state{ nano::timer_state::stopped };
+	std::string desc;
+	std::chrono::time_point<std::chrono::high_resolution_clock> begin;
+	UNIT ticks{ 0 };
+	UNIT minimum{ UNIT::zero () };
+	unsigned measurements{ 0 };
+
+	template <typename U = UNIT>
+	std::string typed_unit (typename std::enable_if<std::is_same<U, std::chrono::nanoseconds>::value, std::chrono::nanoseconds>::type dummy) const
```
wezrule: I quite like this approach. There is a helper alias [std::enable_if_t](https://en.cppreference.com/w/cpp/types/enable_if) in C++14 to shorten this syntax. You can also avoid the "dummy" variable by using it in the return type, i.e:
`std::enable_if_t<std::is_same<U, std::chrono::microseconds>::value, std::string> typed_unit () { return "microseconds";}`

Or even better inside the template argument to clean up the function interface:
```
template <typename U,
          std::enable_if_t<std::is_same<U, std::chrono::nanoseconds>::value>* = nullptr >
std::string typed_unit ()
{
	return "nanoseconds";
}

```

You could also create your own template alias to reduce some of the duplication and make it a bit tidier, something like this:

```
template<typename U, typename V>
using MatchesType = std::enable_if_t<std::is_same<U, V>::value>;

template <typename U, MatchesType<U, std::chrono::nanoseconds>* = 0>
std::string typed_unit ()
{
	return "nanoseconds";
}

template <typename U, MatchesType<U, std::chrono::microseconds>* = 0>
std::string typed_unit ()
{
	return "microseconds";
}

...
```

Then your unit () function would just contain:
 `return typed_unit<UNIT>();`

nano/lib/timer.hpp
```diff
@@ -0,0 +1,250 @@
+#pragma once
+
+namespace nano
+{
+enum timer_state
```
wezrule: This should use the modern **enum class**

nano/lib/timer.hpp
```diff
@@ -0,0 +1,250 @@
...
+{
+	started,
+	stopped
+};
+
+/** Timer utility with nesting support */
+template <typename UNIT = std::chrono::milliseconds>
+class timer
```
wezrule: Would be nice to have unit tests for this class

nano/lib/timer.hpp
cryptocode: Yep, thanks

nano/lib/timer.hpp
cryptocode: I've added a few tests.

nano/lib/timer.hpp
cryptocode: Yeah, the C++14 ...if_t is nicer - fixed.

clemahieu:
Would this be a replacement for the other timing we do, like RPC calls? Where else are you eyeing up to place it?

cryptocode:
@clemahieu I think so, we use std::chrono directly a few places and it's pretty verbose. I didn't want to change everything at once though. The other use is as an ad-hoc debugging tool when profilers tools don't cut it (e.g when a function is is sometimes slow due to locking or whatever, profilers just average it out - with the timer you can set a limit for when to output measurements)

nano/lib/timer.hpp
wezrule: Looks good! In C++17 there is also a helper variable template **std::is_same_v<U, V**>as well

PR #1531, opened 2019-01-02
"update_existing_accounts" option for RPC wallet_representative_set

SergiySW:
to change representative for existing accounts (https://github.com/nanocurrency/raiblocks/issues/1019, @renesq)

### DIFF
--- a/nano/node/rpc.cpp
+++ b/nano/node/rpc.cpp
@@ -3486,8 +3486,36 @@ void nano::rpc_handler::wallet_representative_set ()
 		nano::account representative;
 		if (!representative.decode_account (representative_text))
 		{
-			auto transaction (node.store.tx_begin_write ());
-			wallet->store.representative_set (transaction, representative);
+			{
+				auto transaction (node.store.tx_begin_write ());
+				wallet->store.representative_set (transaction, representative);
+			}
+			// Change representative for all wallet accounts
+			if (request.get<bool> ("force", false))
+			{
+				std::deque<nano::account> accounts;
+				{
+					auto transaction (node.store.tx_begin_read ());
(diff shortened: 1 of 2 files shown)


nano/node/rpc.cpp
```diff
@@ -3486,8 +3486,36 @@ void nano::rpc_handler::wallet_representative_set ()
...
 		if (!representative.decode_account (representative_text))
 		{
-			auto transaction (node.store.tx_begin_write ());
-			wallet->store.representative_set (transaction, representative);
+			{
+				auto transaction (node.store.tx_begin_write ());
+				wallet->store.representative_set (transaction, representative);
+			}
+			// Change representative for all wallet accounts
+			if (request.get<bool> ("force", false))
```
cryptocode: Since current the semantics of `wallet_representative_set` is to set representative for new accounts only, what do you think about calling this parameter `update_existing_accounts` ?

SergiySW: makes sense

nano/node/rpc.cpp
```diff
@@ -3486,8 +3486,36 @@ void nano::rpc_handler::wallet_representative_set ()
...
-			auto transaction (node.store.tx_begin_write ());
-			wallet->store.representative_set (transaction, representative);
+			{
+				auto transaction (node.store.tx_begin_write ());
+				wallet->store.representative_set (transaction, representative);
+			}
+			// Change representative for all wallet accounts
+			if (request.get<bool> ("update_existing_accounts", false))
+			{
+				std::deque<nano::account> accounts;
```
wezrule: I'm not sure std::deque is the right data structure to use here, it is generally only used when popping from the front is required, I think std::vector should be used.

PR #1538, opened 2019-01-02
Set outdated-version status on livenet as well

cryptocode:
As a result, the network visitor never deals with messages from outdated peers. There's an outdated-version stat counter though.


cryptocode:
@wezrule the check above that line is a special case for beta (protocol_version_reasonable_min is currently the same as min-version though, not sure if that check will survive)

argakiig:
closing and opening to re-init travis for this PR

PR #1541, opened 2019-01-02
Fix too-small buffer

rkeene:
Missed in change #1223


cryptocode:
Was `TEST (uint256_union, decode_account_variations)` updated likewise in any of the PRs?

PR #976, opened 2018-07-19
Local votes cache to reduce CPU usage for representatives

SergiySW:

PR #1542, opened 2019-01-03
Local votes cache to reduce CPU usage for representatives

SergiySW:
Replacing https://github.com/nanocurrency/nano-node/pull/976


nano/node/voting.hpp
```diff
@@ -30,4 +36,27 @@ class vote_generator
...
+
+private:
+	std::mutex cache_mutex;
+	boost::multi_index_container<
+	nano::cached_votes,
+	boost::multi_index::indexed_by<
+	boost::multi_index::ordered_non_unique<boost::multi_index::member<nano::cached_votes, std::chrono::steady_clock::time_point, &nano::cached_votes::time>>,
+	boost::multi_index::hashed_unique<boost::multi_index::member<nano::cached_votes, nano::block_hash, &nano::cached_votes::hash>>>>
+	cache;
+	size_t max_cache = (nano::nano_network == nano::nano_networks::nano_test_network) ? 2 : 1000;
```
wezrule: `max_cache` can be constexpr static

SergiySW: Good catch! Done

PR #1553, opened 2019-01-04
Use builder in CLI debug commands

cryptocode:
We have a builder, so let's use it. The debug commands seems like a good place to start.


nano/nano_node/entry.cpp
```diff
@@ -402,10 +403,30 @@ int main (int argc, char * const * argv)
...
 					blocks.push_back (std::move (send));
-					auto open (std::make_shared<nano::state_block> (keys[i].pub, 0, keys[i].pub, balances[i], genesis_latest, keys[i].prv, keys[i].pub, work.generate (keys[i].pub)));
+
+					auto open = builder.state ()
+					            .account (keys[i].pub)
+					            .previous (0)
+					            .representative (keys[i].pub)
+					            .balance (balances[i])
+					            .link (genesis_latest)
+					            .sign (nano::test_genesis_key.prv, nano::test_genesis_key.pub)
```
wezrule: Should signing use the same keys as was used before? `keys[i].prv, keys[i].pub`

cryptocode: Yeah it should, thanks!

PR #1554, opened 2019-01-04
Add sideband information to database.

clemahieu:
This commit adds block-type-specific sideband information which eliminates expensive value computations.
This information is stored in-line with the block itself, instead of in a separate table, for efficient retrieval.
Since this involves rewriting all block entries the process is run iteratively in the background.
Once the upgrade is complete the blocks_info table is dropped from the database.


SergiySW:
Seems faling on
https://github.com/nanocurrency/nano-node/blob/block_sideband_squash/nano/secure/ledger.cpp#L977

nano/node/lmdb.cpp
```diff
@@ -1344,6 +1491,24 @@ std::shared_ptr<nano::block> nano::mdb_store::block_get (nano::transaction const
...
+				assert (!error);
+			}
+			else
+			{
+				// Reconstruct sideband data for block.
+				sideband_a->account = block_account_computed (transaction_a, hash_a);
+				sideband_a->balance = block_balance_computed (transaction_a, hash_a);
+				sideband_a->successor = block_successor (transaction_a, hash_a);
+				sideband_a->height = std::numeric_limits<uint64_t>::max ();
+				sideband_a->timestamp = std::numeric_limits<uint64_t>::max ();
```
SergiySW: May be set default timestampt to 0?

cryptocode:
I noticed the new tests use iterations - does the deadline approach not fit in this case? Not that it matters much, just wondering if deadline_set needs improvement.

nano/node/lmdb.cpp
clemahieu: Does anyone else have an opinion on it?  I used height = max as a sentinel because 0 is a valid height. I could set the timestamp sentinel to 0 instead of max, this just matches height.

cryptocode:
Ah, there's no nano::system involved here. Could add freestanding deadline support one day to keep things consistent, but not important.

nano/node/lmdb.cpp
SergiySW: @rkeene @argakiig @cryptocode ?

rkeene: Either is fine with me.  `0` is falsey, which may be useful in some contexts, it's more commonly used to represent "time unavailable", and it's also easier to type.  `::max()` will never be a valid time, even if an unreasonable epoch is used for the system time keeping, and it's consistent with the existing members.

PR #1555, opened 2019-01-04
Sideband RPC & open blocks adjustment

SergiySW:
Disable sideband account & height for open blocks
Add sideband information to RPC blocks_info & block(_info)
Add local timestamp to RPC account_history


nano/node/node.cpp
```diff
@@ -337,7 +337,7 @@ void nano::network::republish_vote (std::shared_ptr<nano::vote> vote_a)
 void nano::network::broadcast_confirm_req (std::shared_ptr<nano::block> block_a)
 {
 	auto list (std::make_shared<std::vector<nano::peer_information>> (node.peers.representatives (std::numeric_limits<size_t>::max ())));
-	if (list->empty () || node.peers.total_weight () < node.config.online_weight_minimum.number ())
```
clemahieu: This should be done in a separate PR.

PR #1557, opened 2019-01-05
RPC node_config

guilhermelawless:
Useful for users/services to not have to parse `config.json` themselves.

Usage:
```json
{"action": "node_config"}
```

### DIFF
--- a/nano/core_test/rpc.cpp
+++ b/nano/core_test/rpc.cpp
@@ -3937,15 +3937,42 @@ TEST (rpc, node_config)
 	nano::system system (24000, 1);
 	nano::rpc rpc (system.io_ctx, *system.nodes[0], nano::rpc_config (true));
 	rpc.start ();
-	boost::property_tree::ptree request;
-	request.put ("action", "node_config");
-	test_response response (request, rpc, system.io_ctx);
+	const unsigned work_threads (rpc.node.config.work_threads + 10);
+	rpc.node.config.work_threads += 10;
+	boost::property_tree::ptree request0;
+	request0.put ("action", "node_config");
+	test_response response0 (request0, rpc, system.io_ctx);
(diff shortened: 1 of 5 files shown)


rkeene:
How do you feel about adding an additional optional attribute to fetch a subset of the config, or a single attribute ?

guilhermelawless:
@rkeene would be useful. How about:

```json
{"action": "node_config", "field": "logging.version"}
```

Not sure what to name the optional. Is it still an attribute if it's not a leaf?

guilhermelawless:
@rkeene Some design decisions and constraints to allow getting subset of config:

- To accept `logging` that case had to be specified to serialize. Will need to do the same if similar cases appear in the future
- Without looking for last separator (`.`) and finding the substring, the response would be e.g. `{"logging": {"log_rpc": "true"}"}` when asking for "logging.log_rpc", defeating the purpose of specifying a subset. The actual response is `{"log_rpc": "true"}`

If you think the current design is fine then i'll do some unit tests.

guilhermelawless:
Much more elegant @cryptocode thank you. But i find it a bit inconsistent that asking for nothing gets a field "node", asking for logging gets a field "logging" and asking for a leaf, like "logging.version", gets a field "value".

Is there an easy way to get the last part of the path?

cryptocode:
@guilhermelawless see https://en.cppreference.com/w/cpp/string/basic_string/find_last_of for how to extract the last part of the path, if you want that instead of "value".

guilhermelawless:
@cryptocode Ok I had it that before, put it back. Should be ready now, tests included.

zhyatt:
After further discussion with the team and @guilhermelawless it was determine the usefulness of this approach vs. directly parsing the config file did not warrant including the change, so we are closing this one out.

PR #1561, opened 2019-01-06
WIP: factory for blocks and votes

cryptocode:
Not really a PR yet; it's a hash consing experiment as an alternative to the uniquers - putting the approach up for discussion. Through custom deleters, this will evict objects as soon as the last ref goes away instead of small GC passes.


cryptocode:
The current uniquer seems to work well enough for now. We can revive this idea if needed.

Issue #1516, opened 2018-12-29
Lazy bootstrap part 2


SergiySW:
Priority bootstrap of local wallet accounts and unknown pending blocks

SergiySW:
Merged

PR #1565, opened 2019-01-08
Priority lazy bootstrap for wallet accounts

SergiySW:
https://github.com/nanocurrency/nano-node/issues/1516

### DIFF
--- a/nano/node/bootstrap.cpp
+++ b/nano/node/bootstrap.cpp
@@ -1393,6 +1393,17 @@ bool nano::bootstrap_attempt::lazy_finished ()
 	return result;
 }
 
+void nano::bootstrap_attempt::lazy_clear ()
+{
+	assert (!lazy_mutex.try_lock ());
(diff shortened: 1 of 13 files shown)


nano/node/bootstrap.cpp
```diff
@@ -743,6 +743,129 @@ void nano::bulk_push_client::push_block (nano::block const & block_a)
...
 }
 
+nano::bulk_pull_account_client::bulk_pull_account_client (std::shared_ptr<nano::bootstrap_client> connection_a, nano::account const & account_a) :
+connection (connection_a),
+account (account_a),
+total_blocks (0)
+{
+	std::lock_guard<std::mutex> mutex (connection->attempt->mutex);
+	connection->attempt->condition.notify_all ();
+}
```
cryptocode: No need to lock over the notify here right? (pessimization)

SergiySW: Seems correct

PR #1567, opened 2019-01-08
Accept seed in RPC wallet_create

### LINKED ISSUE #1341: Proposal Import seed/wallet without exsiting wallet
It would be perfect to have a rpc command that imports a wallet from a seed.

At the moment all import functions need an existing wallet that has to be created on the node before.

guilhermelawless:
Fixes #1341 

Since wallet::change_seed() inserts and returns the first key, it is also included in the response. If seed is not provided, the first account is not generated.

One could argue this is inconsistent behavior, and that doing the same when the seed is not provided would not break anything for users/services, so we could add that too.

### DIFF
--- a/nano/core_test/rpc.cpp
+++ b/nano/core_test/rpc.cpp
@@ -733,6 +733,10 @@ TEST (rpc, wallet_create_seed)
 		wallet->second->store.seed (seed0, transaction);
 		ASSERT_EQ (seed.pub, seed0.data);
 	}
+	auto account_text (response.json.get<std::string> ("account"));
+	nano::uint256_union account;
+	ASSERT_FALSE (account.decode_account (account_text));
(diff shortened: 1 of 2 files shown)


guilhermelawless:
Will clash with #775

zhyatt:
@SergiySW Before we merge #775 do we need to consider the changes in this PR?

guilhermelawless:
After checking again, it seems that it won't clash, because the new "count" field in nano::wallet::change_seed is optional and defaults to `0`.

The question is if we should use enable work generation for the created account with this RPC enhancement.

SergiySW:
It seems this PR will not affect https://github.com/nanocurrency/nano-node/pull/775, default change_seed () path was not changed

guilhermelawless:
Tests succeed after merge of #775

PR #1566, opened 2019-01-08
Add new swap method to verify_state_blocks ()

SergiySW:
For max_count = 256 || 2048 items.push_back (state_blocks.front ()) becoming more effective around (state_blocks.size () > 4 * max_count)
Following https://github.com/nanocurrency/nano-node/issues/1475

PR #1571, opened 2019-01-09
Check state blocks existence in ledger before verification

SergiySW:
Alternative to https://github.com/nanocurrency/nano-node/pull/1566


nano/node/node.cpp
```diff
@@ -1316,45 +1316,48 @@ void nano::block_processor::verify_state_blocks (nano::transaction const & trans
...
-	std::promise<void> promise;
-	nano::signature_check_set check = { size, messages.data (), lengths.data (), pub_keys.data (), signatures.data (), verifications.data (), &promise };
-	node.checker.add (check);
-	promise.get_future ().wait ();
-	lock_a.lock ();
-	for (auto i (0); i < size; ++i)
-	{
-		assert (verifications[i] == 1 || verifications[i] == 0);
-		if (verifications[i] == 1)
+	if (!items.empty ())
```
clemahieu: Is this needed after you added
`while (!state_blocks.empty () && timer_l.before_deadline (std::chrono::seconds (2)))` in the other commit?
Can we add an assert at the top that items is not empty instead?

PR #1576, opened 2019-01-10
Add -Werror=shadow with required changes

cryptocode:
There's a fair amount of shadowing going on. Doesn't seem to be any actual issues, though I wonder if someone could look at the nested uses of mdb transactions.

Mostly resolved compile errors using the `_a` and `_l` suffixes, otherwise tried to find reasonable names. For lmdb statues, the variable is reused.

- [ ] Fix gcc errors


cryptocode:
-Werror=shadow yields a large number of false positives on some compilers (complaining about unrelated lambda params being shadowed), and also unnecessary complains shadowing global type declarations.

We probably want -Wshadow=local but this is a too recent addition to gcc.

PR #1577, opened 2019-01-11
RPC debug command line option

cryptocode:
Allows for running RPC actions when you don't want to enable RPC in the config/have a running node, or when curl or similar tools aren't conveniently available (minimal-install air-gapped machines, etc)

Example: `echo '{"action": "block_count"}' |nano_node --debug_rpc` or feed it a file. 

Prefixed as debug as this is not a stable part of the command line interface/may be removed.


cryptocode:
@SergiySW Good call. Put it in the CLI command description ?

Issue #1529, opened 2019-01-01
Add size field to bootstrap message types


cryptocode:
In a similar vein to https://github.com/nanocurrency/raiblocks/pull/1528

Then reference that field instead of e.g. `size_t size_l (sizeof (nano::uint256_union) + sizeof (nano::uint256_union));` at call sites.

Some requests have variable size (based on header fields) so it'll be a function rather than a constexpr.

PR #1580, opened 2019-01-13
Add size expressions to bootstrap types

cryptocode:
Closes https://github.com/nanocurrency/nano-node/issues/1529


nano/node/bootstrap.cpp
```diff
@@ -1739,20 +1743,10 @@ void nano::bootstrap_server::receive_header_action (boost::system::error_code co
...
-						extended_size = nano::bulk_pull::extended_parameters_size;
-					}
-					else
-					{
-						extended_size = 0;
-					}
-
 					auto this_l (shared_from_this ());
-					socket->async_read (receive_buffer, sizeof (nano::uint256_union) + sizeof (nano::uint256_union) + extended_size, [this_l, header](boost::system::error_code const & ec, size_t size_a) {
+					socket->async_read (receive_buffer, nano::bulk_pull::size (header.bulk_pull_is_count_present ()), [this_l, header](boost::system::error_code const & ec, size_t size_a) {
```
rkeene: Why not make the header size a method of the header object ?

cryptocode: extended_parameters_size / count_present_flag was already part of the bulk_pull class.

cryptocode: added `payload_length_bytes`

nano/node/common.hpp
```diff
@@ -175,15 +175,17 @@ class message_header
 	uint8_t version_min;
 	nano::message_type type;
 	std::bitset<16> extensions;
-	//static size_t constexpr ipv4_only_position = 1;  // Not in use, deprecated, was conflicting
-	//static size_t constexpr bootstrap_server_position = 2;  // Not in use, deprecated
+
 	/*
 	 * A better approach might be to return the size of the message
 	 * payload based on the header
 	 */
```
rkeene: We can probably get rid of this suggestion now

nano/node/bootstrap.cpp
```diff
@@ -14,6 +14,11 @@ constexpr double bootstrap_minimum_termination_time_sec = 30.0;
 constexpr unsigned bootstrap_max_new_connections = 10;
 constexpr unsigned bulk_push_cost_limit = 200;
 
+size_t constexpr nano::frontier_req::size;
+size_t constexpr nano::bulk_pull_blocks::size;
+size_t constexpr nano::bulk_pull_account::size;
+size_t constexpr nano::frontier_req_client::size_frontier;
```
rkeene: Are these forward declarations needed if we `#include <node/common.hpp>` which includes the definition ?

nano/node/bootstrap.cpp
```diff
@@ -277,13 +281,13 @@ void nano::frontier_req_client::received_frontier (boost::system::error_code con
 {
 	if (!ec)
 	{
-		assert (size_a == sizeof (nano::uint256_union) + sizeof (nano::uint256_union));
+		assert (size_a == nano::frontier_req_client::size_frontier);
 		nano::account account;
-		nano::bufferstream account_stream (connection->receive_buffer->data (), sizeof (nano::uint256_union));
+		nano::bufferstream account_stream (connection->receive_buffer->data (), sizeof (nano::account));
```
rkeene: Why not change this to `sizeof(account)` if we're going to tie it to a type whose name doesn't express its on-wire representation ?

nano/node/bootstrap.cpp
cryptocode: You need a definition since we odr-use these (you'll get a linker error if not) We do this several places. The alternative is an inline constexpr function (or C++17 inline vars), but I think this is fine for now.

nano/node/bootstrap.cpp
```diff
@@ -239,19 +244,18 @@ nano::frontier_req_client::~frontier_req_client ()
 void nano::frontier_req_client::receive_frontier ()
 {
 	auto this_l (shared_from_this ());
-	size_t size_l (sizeof (nano::uint256_union) + sizeof (nano::uint256_union));
-	connection->socket->async_read (connection->receive_buffer, size_l, [this_l, size_l](boost::system::error_code const & ec, size_t size_a) {
+	connection->socket->async_read (connection->receive_buffer, nano::frontier_req_client::size_frontier, [this_l](boost::system::error_code const & ec, size_t size_a) {
```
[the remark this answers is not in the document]
cryptocode: It's not captured?

nano/node/bootstrap.cpp
cryptocode: Agreed, will fix

nano/node/bootstrap.cpp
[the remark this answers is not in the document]
rkeene: Oops, I was reading this wrong :+1:

nano/node/bootstrap.cpp
```diff
@@ -1838,7 +1832,7 @@ void nano::bootstrap_server::receive_bulk_pull_account_action (boost::system::er
 	if (!ec)
 	{
 		auto error (false);
-		assert (size_a == (sizeof (nano::uint256_union) + sizeof (nano::uint128_union) + sizeof (uint8_t)));
+		assert (size_a == nano::bulk_pull_account::size);
```
rkeene: Missed ?

nano/node/bootstrap.cpp
rkeene: After the most recent change, all that should be required here is `size_frontier`, correct ?

cryptocode: Yeah, that's true. I'll fix that.

PR #1595, opened 2019-01-16
Fix asan issue (illegal downcast) in test

cryptocode:


SergiySW:
So, now it's casting state block to send?

cryptocode:
Before this patch, yes, at least according to asan, and that complaint seems correct right?

PR #1599, opened 2019-01-17
json config: fix upgrade of required child nodes

cryptocode:
This fixes a bug introduced by https://github.com/nanocurrency/nano-node/pull/1416 (upgrade of required nodes)
Would be good to get this merged as master is partially broken wrt json upgrades.


nano/lib/jsonconfig.hpp
```diff
@@ -163,6 +164,12 @@ class jsonconfig : public nano::error_aware<>
 		return child_config;
 	}
 
+	jsonconfig & operator= (jsonconfig const & other_a)
+	{
+		tree = other_a.tree;
```
wezrule: I don't think you need to define this as a default one will be provided for you by the compiler. Defining it also prevents move constructors being implicitly generated.

cryptocode: I tried that but got "Object of type 'nano::jsonconfig' cannot be assigned because its copy assignment operator is implicitly deleted", probably by not following rule of 5. Also a defaulted impl yields the same error on clang. I think it's fine to keep this for now, I want to rewrite parts of this.

wezrule:
Could any more tests be added to make sure this is fixing what it should and won't be broken in the future?

nano/lib/jsonconfig.hpp
wezrule: Hmm weird, I did some more digging and actually the rules are more strict than I thought: http://eel.is/c++draft/class.copy.assign#7 looks like the tree& will stop them being generated.

PR #1600, opened 2019-01-17
Expanding root to uint512_union

rkeene:


SergiySW:
May be we can replace nano::uint512_union (block->previous (), block->root ()) with something like block->qualified_root () later to simplify code

PR #1601, opened 2019-01-17
Indicate active network in Qt wallet

cryptocode:
Adds network next to the version number, e.g, "Version 18.0 Beta network"


cryptocode:
I didn't think anyone would run the qt wallet when compiling for test, but I guess we can add it for completeness.

PR #1611, opened 2019-01-21
Added threading code for signature_checker::verify

termoose:
For >1000 verifications we spawn one thread per available core. We then collect the result of each in a similar way to the old single-threaded code, just by `future.get()`-ing them all in the `std::all_of`.

New test case added for the multi-threaded code, getting around a 2x speedup on my laptop.

### DIFF
--- a/nano/core_test/signing.cpp
+++ b/nano/core_test/signing.cpp
@@ -44,6 +44,40 @@ TEST (signature_checker, many)
 	promise.get_future ().wait ();
 }
 
+TEST (signature_checker, many_threaded)
+{
+	nano::keypair key;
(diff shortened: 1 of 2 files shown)


wezrule:
Thanks! I've not looked in detail, but could you elaborate how the 1000 check cut-off point was chosen?
We have done some performance tests and found that a batch size of 256 seemed most appropriate so were going to create a number of threads (same as your `nr_threads`) but have a loop inside doing batches of 256 signatures. Could you try this and see how it performs against your implementation on your machine? Thanks again.

termoose:
Thread pool is a good idea, I almost have a working implementation for that. Will submit a new pull request once it's finished.

PR #1614, opened 2019-01-22
Thread pool for checking verifications

termoose:
Added a `thread_pool` class for submitting work into one of the `hardware_concurrency` number of threads (default value). Class is somewhat general and can be used for other threading use-cases.
Added a test case for doing 100k verifications multi-threaded.
Batching `256` verifications in one thread, hard cut-off of `1000` for running the single threaded version.


cryptocode:
Could `boost::asio::thread_pool` work here? Since we're already using boost, it's good to use that one if possible.

termoose:
I actually didn't know there was a thread pool in boost, always heard that they wouldn't add one. Guess they did! Although it looks pretty basic, there's no support for futures so I'd have to roll my own way of getting results from the async operations in the thread pool (the true/false from all those async `std::all_of`).

I'll see if I can get away with a smaller code footprint by using it.

wezrule:
I will take a look at it today as it is something of high priority and is needed asap, thanks!

termoose:
Thanks, seems like it works as well as the custom thread pool. I changed the cut-off to 10k based on the feedback from @guilhermelawless. I was considering adding this to `node_config` as a setting, but decided not to since it was a bit awkward getting a hold of this object from `nano::signature_checker`.

nano/node/node.cpp
```diff
@@ -1179,20 +1177,27 @@ void nano::signature_checker::verify (nano::signature_check_set & check_a)
...
 			size = overflow;
 
-		auto result = pool.submit ([=]() -> bool {
-			auto code (nano::validate_message_batch (check_a.messages, check_a.message_lengths, check_a.pub_keys, check_a.signatures, size, check_a.verifications + index));
-			(void)code;
+		boost::asio::post (thread_pool, [=, &results]	{
+			bool result = verify_batch (check_a, index, size);
 
-			return std::all_of (check_a.verifications + index, check_a.verifications + index + size, [](int verification) { return verification == 0 || verification == 1; });
+			std::lock_guard<std::mutex> lock (results_mutex);
```
guilhermelawless: Mutex/push_back can be avoided:
```cpp
std::vector<bool> results (batches + 1, false);
...
results[batch] = result;
```

guilhermelawless:
Getting the same results with the new thread pool as well. Also a good thing is that results don't change much with higher batch sizes, so that's one less thing to parametrize.

Not sure I agree with the cut-off change. Setting to 512 (2x batch_size) already gets me a 1.7x speedup, so why not enjoy that?

wezrule:
Have you checked on live whether it goes into your `verify_threaded` function? For me the max signature size I got in 1 chunk was 2048. 256 was common, but so are very low numbers (many are 1). I think a different approach may be needed to get good results on the live network. I think it may be useful having a buffer of 256 which collates smaller signature sizes into 1 and then dishes them out to an available thread, rather than operating on only large (unreachable?) check sizes like 10k. I think our benchmark tests should use the `nano::signature_checker::add ()` function many times with different sizes to reach a total of something like 100k and then compare how long that takes to complete vs the master branch.

guilhermelawless:
@wezrule @termoose 
Did not catch this in boost documentation (didn't go through the .cpp, only headers), but this sanity check is failing:

```cpp
TEST (sanity, thread_pool)
{
	boost::asio::thread_pool thread_pool (4);
	std::vector<bool> v(4, false);
	for (int i=0; i<=1; ++i)
		boost::asio::post (thread_pool, [i, &v] { v[i] = true; });

	thread_pool.join ();

	for (int i=2; i<=3; ++i)
		boost::asio::post (thread_pool, [i, &v] { v[i] = true; });

	thread_pool.join ();

	ASSERT_EQ (4, v.size ());
	ASSERT_EQ (true, v[0]);
	ASSERT_EQ (true, v[1]);
	ASSERT_EQ (true, v[2]);
	ASSERT_EQ (true, v[3]);
}
```

This test fails:
```
Value of: v[2]
  Actual: 0
Expected: true
```

It seems to me that once a thread_pool has been joined, it cannot be reused. But to solve we only need to instance it inside verify_threaded, not class wide.

guilhermelawless:
See https://github.com/guilhermelawless/nano-node/pull/1 for a more realistic case with the following adds (repeated until 100k): `1, 256, 257, 1, 1, 512, 1, 2048, 1024, 1, 64, 1, 1, 1, 1, 1, 1, 1, 1`

It gets me a 2x speedup in total. Also fixed the thread_pool re-use, and the number of batches sometimes being 1 extra due to overflow = 0. Cut-off lowered to 512.

@termoose  feel free to take code from there and modify this PR, i won't be PRing myself.

nano/core_test/signing.cpp
```diff
@@ -70,3 +70,70 @@ TEST (signature_checker, one)
...
+			signatures.push_back (block.signature.bytes.data ());
+		}
+
+		nano::signature_check_set check = { size, messages.data (), lengths.data (), pub_keys.data (), signatures.data (), verifications.data (), &promise };
+		checker.add (check);
+		promise.get_future ().wait();
+	}
+}
+
+TEST (sanity, thread_pool)
```
guilhermelawless: Forgot to remove this in that commit, will fail

termoose:
Made it so that the thread pool does not have to be instantiated inside `verify_threaded` every time, rather we're `get()`ing the futures as explained above. Not sure why that syntax checker is failing though, it looks correct to me on that line.

guilhermelawless:
Do you notice a difference? I don't think with a minimum of 512 signature checks the overhead is significant.

wezrule:
Hi @termoose @guilhermelawless thanks for your contributions so far! Now I have a few comments:

- There was mention of using std::vector<bool> for concurrent modification in different threads on different elements. Normally this would be ok, however std::vector<bool> is an anomaly, this is because std::vector<bool> has a special specialization in which some implementations squeeze bool values into bits.
> Notwithstanding (17.6.5.9), implementations are required to avoid data races when the contents of the contained object in different elements in the same container, excepting vector<bool>, are modified concurrently.
- `validate_message_batch ()` had a bug where it doesn't check the correct range after the first batch size.
- I wasn't entirely sure why we had another thread for signature_checking because the calling thread just blocks until the future was finished, so I've got rid of it now.
- If threads is set to 1 I just use the calling thread to do the work.
- I have added the config option so that the user can specify the number of cores (we do similar things with work, io and network threads)
- The newly added test is now run over multiple threads and promises are checked after all signature checks are added to the thread pool rather than waiting for each one to finish. Using this test and changing the cores I got the following results:
1 core - 5.9s
2 cores - 5.4s
4 cores - 3.4s
8 cores - 2.5s

When more messages of a greater size are used it greatly favours the higher cores, however we want tests to run fairly quickly irrespective of the number of cores so this test is not meant to be representative of a benchmark, it just checks functionality is working correctly for a wide range of sizes.
- There were some other things, but I'm far too sleepy to remember :)

Can you guys have a look here and give any comments: https://github.com/wezrule/raiblocks/tree/multithread_signature_checker @termoose feel free to modify your PR with any of these changes, otherwise I will mine as a PR some point tomorrow.

termoose:
- Aware of this specialisation, we don't use `std::vector<bool>` anywhere but instead have futures and promises in our vectors.
- Thanks for finding that off by 1 on batch sizes!
- Modified my PR to include the ability to set number of threads from the config

There are a number of changes here though that I find confusing but I don't have time to figure out all the details until you merge it in tomorrow.

- You use a `shared_ptr` to the vector as well as on your `Task` which causes a heap allocation every time `verify_threaded` is called. Is this an overhead that's needed?
- You have the `tasks_remaining` and `task->pending` counters which looks like spin-your-own `std::future`s for checking when the threads are finished. Myself would prefer a set of `std::promise<bool>`s that communicate these results as return values from `future.get()`.
- Casting to and from `char`, is this to keep the `std::vector` thread safe? This overhead would also not be needed with our original `std::vector<std::future<bool>>`. It also makes things look strange, for example when initialising a `vector` of `char`s to `false`.
- If `pending_mutex` is only used for the counter then it could be made `std::atomic_int` and you could rather do `if(--task->pending == 0)` where it decrements and loads atomically. But again I don't see these counters are needed if threading primitives are being used.

Waiting for each signature check to finish before `post`ing new ones should not incur any great overhead as the cores are all fully exhausted anyway. Would be interesting to see some benchmarks on the different implementations if someone has time to run them.

guilhermelawless:
It looks good, `add` was confusing me as well since there was never background tasks happening, every `verify` was blocking.

All I would add is a `release_assert (pending == 0);` on a Task destructor.

wezrule:
@guilhermelawless done. I will close this off and we can use: 1651

PR #1616, opened 2019-01-22
RPC uptime

guilhermelawless:
Tracks the node uptime in seconds.

Sample

```json
{"action": "uptime"}

{"seconds": "73"}
```

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1622,6 +1622,7 @@ node (init_a, io_ctx_a, application_path_a, alarm_a, nano::node_config (peering_
 }
 
 nano::node::node (nano::node_init & init_a, boost::asio::io_context & io_ctx_a, boost::filesystem::path const & application_path_a, nano::alarm & alarm_a, nano::node_config const & config_a, nano::work_pool & work_a) :
+startup_time (std::chrono::system_clock::now ()),
 io_ctx (io_ctx_a),
 config (config_a),
 alarm (alarm_a),
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1622,7 +1622,6 @@ node (init_a, io_ctx_a, application_path_a, alarm_a, nano::node_config (peering_
 }
 
 nano::node::node (nano::node_init & init_a, boost::asio::io_context & io_ctx_a, boost::filesystem::path const & application_path_a, nano::alarm & alarm_a, nano::node_config const & config_a, nano::work_pool & work_a) :
-startup_time (std::chrono::system_clock::now ()),
 io_ctx (io_ctx_a),
 config (config_a),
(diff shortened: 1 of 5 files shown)


wezrule:
Thanks, although I'm not sure if this is necessary. Can you give us a scenario where this gives information that can't be supplied by asking the operation system?

guilhermelawless:
@wezrule i guess it's possible using something like `ps -eo pid,etime,args | grep node` and parsing that, but having it in the RPC is pretty useful, and makes it consistent.

One use case is node monitors, currently the [most used one](https://github.com/NanoTools/nanoNodeMonitor) uses system uptime.

zhyatt:
The team is considering this but would prefer it as unformatted monotonically increasing seconds instead of sec/min/hr/day splits, so if you get the time to update the PR for that change it would be very helpful.

guilhermelawless:
Sure thing, done and thanks for considering it!

Left the response field as `seconds` but can change to something else.

nano/node/node.hpp
```diff
@@ -534,6 +534,7 @@ class node : public std::enable_shared_from_this<nano::node>
 	nano::keypair node_id;
 	nano::block_uniquer block_uniquer;
 	nano::vote_uniquer vote_uniquer;
+	const std::chrono::system_clock::time_point startup_time;
```
wezrule: Use the `std::chrono::steady_clock` to get an accurate time difference, as the `system_clock` can be easily manipulated.

nano/node/node.cpp
```diff
@@ -1622,6 +1622,7 @@ node (init_a, io_ctx_a, application_path_a, alarm_a, nano::node_config (peering_
 }
 
 nano::node::node (nano::node_init & init_a, boost::asio::io_context & io_ctx_a, boost::filesystem::path const & application_path_a, nano::alarm & alarm_a, nano::node_config const & config_a, nano::work_pool & work_a) :
+startup_time (std::chrono::system_clock::now ()),
```
wezrule: Effective C++ (Scott Meyers) - Item 13 

> "List members in an initialization list in the order in which they are declared"
> Reason for rule: Class members are initialized in the order of their declaration in the class, not by the order in which they are listed in a member initialization list.

If you want `startup_time` to be called before other members then it should be listed at the top in the class definition.

guilhermelawless: Done, moved it to the bottom of the initialization order.

wezrule:
Haven't tested it, but LGTM. Can you update the description of this PR so it reflects the latest changes (i.e just seconds)?

guilhermelawless:
Done, thank you!

zhyatt:
@cryptocode Can you give this a quick test and merge if it's working as expected?

Issue #615, opened 2018-02-13
Chain RPC call suggestions


eirikrye:
Hi,

I would like to propose two additions to the `chain` RPC call:

1) Add optional `offset` parameter, which allows you to return the account chain hashes offset by the specified number of blocks. That is, the call would skip the `offset` number of blocks before returning block hashes. I'm not too sure about the data structures in the database whether this would significantly speed up the call compared to doing the list slicing yourself. However, it would at the very least offer significant bandwidth savings when you are interested in a large offset.

```
http POST :7076 action=chain block=ECCB8CB65CD3106EDA8CE9AA893FEAD497A91BCA903890CBD7A5C59F06AB9113 count=5 offset=5
HTTP/1.1 200 OK

{
    "blocks": [
        "7D327B5D86FA2488DA39DD4110FBE9A7D5391414DDA26697CCE7E70577F459A9",
        "AD374A55F31C467699D951409656B046C16864013FBA467B97931AA4A09AC7D2",
        "3845877A97CBD7CE71F7F7154F560ECE65AFB33FE5876E4B4719B36AF31928FF",
        "0FA0F8A826F02ADE7496C99DE21759B6ED4637DFDE983B0CEC02E8CC016D4CA4",
        "FF84533A571D953A596EA401FD41743AC85D04F406E76FDE4408EAED50B473C5"
    ]
}
```

2) Add optional `reverse` parameter to allow to you move up the account chain instead of only down.

```
http POST :7076 action=chain block=991CF190094C00F0B68E2E5F75F6BEE95A2E0BD93CEAA4A6734DB9F19B728948 count=5 reverse=1
HTTP/1.1 200 OK

{
    "blocks": [
        "991CF190094C00F0B68E2E5F75F6BEE95A2E0BD93CEAA4A6734DB9F19B728948",
        "A170D51B94E00371ACE76E35AC81DC9405D5D04D4CEBC399AEACE07AE05DD293",
        "28129ABCAB003AB246BA22702E0C218794DFFF72AD35FD56880D8E605C0798F6",
        "BA369AC2E725AE8A9F6F3F024ADFF03EBB063FD29FEDE83D235EA0CC6F580B4F",
        "021C4EFC9CE0EB4B67E1D7510091371CA70F4522CF24FDD2323FC4D73FE92521"
    ]
}

```

PlasmaPower:
The block storage is similar to a linked list, so I think these parameters would be faster than applying them manually.

[...]

PR #1622, opened 2019-01-22
Add optional "offset" & "reverse" parameters for RPC chain

SergiySW:
https://github.com/nanocurrency/nano-node/issues/615


nano/node/rpc.cpp
```diff
@@ -1391,8 +1404,10 @@ void nano::rpc_handler::bootstrap_status ()
 
 void nano::rpc_handler::chain (bool successors)
 {
+	successors = successors || request.get<bool> ("reverse", false);
```
rkeene: Maybe this should be XOR instead of OR ?

RPC successors(reverse=true) == chain
RPC chain(reverse=true) == successors

`chain(true)`, reverse=false == successors
`chain(true)`, reverse=true == chain
`chain(false)`, reverse=false == chain
`chain(false)`, reverse=true == successors

| successors | XOR | reverse | = | successors' |
|:----------:|:---:|:-------:|:-:|:------:|
|    true    |  ^  |  false  | = |  true  |
|    true    |  ^  |   true  | = |  false |
|    false   |  ^  |  false  | = |  false |
|    false   |  ^  |   true  | = |  true  |

SergiySW: I thought about it. Well, if you think it's logical, let's do it

cryptocode: Not to bike-shed, but I think `successors = successors != request.get<bool> ("reverse", false);` is clearer when reading that code down the road.

PR #1625, opened 2019-01-23
RPC to clear stats

### LINKED ISSUE #1620: RPC to clear stats
Add an rpc action to clear stats and possibly include timestamp of last clear (initialized to startup time) in the result.

cryptocode:
Closes #1620 

`stats_clear` clears all existing stats, and `stats` is  updated to include seconds since last clear (or node startup if never cleared)

### DIFF
--- a/nano/core_test/rpc.cpp
+++ b/nano/core_test/rpc.cpp
@@ -4152,6 +4152,8 @@ TEST (rpc, stats_clear)
 	{
 		ASSERT_NO_ERROR (system.poll ());
 	}
+	std::string success (response.json.get<std::string> ("success"));
+	ASSERT_TRUE (success.empty ());
 	ASSERT_EQ (0, system.nodes[0]->stats.count (nano::stat::type::ledger, nano::stat::dir::in));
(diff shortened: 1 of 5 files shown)


nano/core_test/rpc.cpp
```diff
@@ -4136,6 +4136,26 @@ TEST (rpc, node_id_delete)
...
+	ASSERT_EQ (1, system.nodes[0]->stats.count (nano::stat::type::ledger, nano::stat::dir::in));
+	boost::property_tree::ptree request;
+	request.put ("action", "stats_clear");
+	test_response response (request, rpc, system.io_ctx);
+	system.deadline_set (5s);
+	while (response.status == 0)
+	{
+		ASSERT_NO_ERROR (system.poll ());
+	}
+	ASSERT_EQ (0, system.nodes[0]->stats.count (nano::stat::type::ledger, nano::stat::dir::in));
```
SergiySW: May be also add check for "success" in response?

cryptocode:
Submitting wiki changes now

PR #1627, opened 2019-01-23
Add wallet locked check to RPC wallet_representative_set

SergiySW:
Required after https://github.com/nanocurrency/nano-node/pull/1531 if update existing accounts option is used


nano/node/rpc.cpp
```diff
@@ -3531,6 +3531,7 @@ void nano::rpc_handler::wallet_representative_set ()
 {
 	rpc_control_impl ();
 	auto wallet (wallet_impl ());
+	bool update_existing_accounts (request.get<bool> ("update_existing_accounts", false));
```
cryptocode: Very minor: this line could be moved to a more narrow scope, inside `if (!representative.decode_account (representative_text))`

Other than that, LGTM

SergiySW: Thx

Issue #1112, opened 2018-08-28
Break up and document node source


cryptocode:
A common feedback from new devs looking at the code is the large and mostly undocumented source files.

Breaking it up makes navigating easier and will reduce compile times, esp. when editing headers which many files depend on due to their scope. `secure.cpp` is already done.

(This issue is work in progress, suggestions welcome)

**documentation**
- [ ] Add class/member docs where needed
- [ ] Move existing docs to doxygen format

**node.cpp**

- [x] break up bootstrap #2274
- [x] port mapping #1038 
- [x] configuration #1296
- [x] peering #1315 
- [x] signature checker #1700
- [x] rep crawler #1803 
- [x] election #2030 
- [x] network (network and udp_buffer classes)
- [x] gap_cache/online_reps #2070
- [x] vote processor #2041 
- [x] logging #1262 
- [x] block_processor #1633
- [x] move port/address parsing code to common.cpp (declared in common.hpp already) #1263 
- [x] alarm/operation #2173

PR #1633, opened 2019-01-25
Factor out block processor

cryptocode:
Part of #1112


cryptocode:
Just moving-around code, can wait until v19

PR #1638, opened 2019-01-25
Better handling of TCP accept overflow

rkeene:
Currently when we overflow the number of concurrent TCP sockets
we wish to service we accept them then immediately disconnect.
This change causes us to defer accepting new connections until we
are ready for them, relying instead on the OS socket backlog to
queue TCP handshakes until full, then stop processing new
connections when the backlog is full.

We use the I/O threads for scheduling all but one accept, so this
keeps the same thread affinity.


cryptocode:
Seems like we should be checking `ec != boost::asio::error::operation_aborted` when calling acceptor#is_open everywhere (like in the RPC shutdown fix), but that is probably for a different PR since is_open is called a few places.

PR #1643, opened 2019-01-26
PATCH to indicate RC, Updated Version responses

argakiig:
Use Patch Version to indicate if RC or not. If Patch version == 0 Final Release else RC<Patch> appended


wezrule:
The same version strings are created in 4 different ways. It might be nice to construct both strings in one place (probably using macro stringification) and use those instead. There will be less duplication and less chance of mistakes. Something like:

```
#define xstr(a) str(a)
#define str(a) #a

#define NANO_MAJOR_MINOR_VERSION xstr(NANO_VERSION_MAJOR) "." xstr(NANO_VERSION_MINOR)
#define NANO_MAJOR_MINOR_RC_VERSION NANO_MAJOR_MINOR_VERSION "RC" xstr(NANO_VERSION_PATCH)

// Or
//const char* NANO_MAJOR_MINOR_VERSION = xstr(NANO_VERSION_MAJOR) "." xstr(NANO_VERSION_MINOR);
//const char* NANO_MAJOR_MINOR_RC_VERSION = NANO_MAJOR_MINOR_VERSION "RC" xstr(NANO_VERSION_PATCH);

// Example use
BOOST_LOG (log) << "Node starting, version: " << NANO_MAJOR_MINOR_VERSION;
```

argakiig:
> The same version strings are created in 4 different places (and some in different ways). It might be nice to construct both strings in one place (probably using macro stringification) and use those instead. There will be less duplication and less chance of mistakes. Something like:
> 
> ```
> #define xstr(a) str(a)
> #define str(a) #a
> 
> #define NANO_MAJOR_MINOR_VERSION xstr(NANO_VERSION_MAJOR) "." xstr(NANO_VERSION_MINOR)
> #define NANO_MAJOR_MINOR_RC_VERSION NANO_MAJOR_MINOR_VERSION "RC" xstr(NANO_VERSION_PATCH)
> 
> // Example use
> BOOST_LOG (log) << "Node starting, version: " << NANO_MAJOR_MINOR_VERSION;
> ```

what about instead just adding the definitions for NANO_MAJOR_MINOR_VERSION and NANO_MAJOR_MINOR_RC_VERSION that's what I updated the PR to reflect

wezrule:
I think it probably could have been done once in a common header file, but yeh that works as well

clemahieu:
Would it be worth constructing this in the globals data structure? Something to avoid the use of macros?

PR #1648, opened 2019-01-27
Long-term online rep weight tracking

clemahieu:


nano/node/node.cpp
```diff
@@ -2937,80 +2956,75 @@ bool nano::block_arrival::recent (nano::block_hash const & hash_a)
...
-	std::weak_ptr<nano::node> node_w (node.shared ());
-	node.alarm.add (now + std::chrono::minutes (5), [node_w]() {
-		if (auto node_l = node_w.lock ())
-		{
-			node_l->online_reps.recalculate_stake ();
-		}
-	});
+	std::sort (items.begin (), items.end ());
+	// Pick median value for our target vote weight
+	nano::uint128_t result (items[items.size () / 2]);
```
wezrule: `std:nth_element` can be used which finds what a certain index's value would be if the whole collection is sorted (but is much quicker):
```
    auto i = items.size() / 2;
    nth_element (items.begin (), items.begin () + i, items.end ());
    return nano::uint128_t{ items[i] };
```

nano/node/node.hpp
```diff
@@ -214,31 +214,24 @@ class block_arrival
...
-	nano::uint128_t online_stake_total;
 	std::vector<nano::account> list ();
-	boost::multi_index_container<
-	nano::rep_last_heard_info,
-	boost::multi_index::indexed_by<
-	boost::multi_index::ordered_non_unique<boost::multi_index::member<nano::rep_last_heard_info, std::chrono::steady_clock::time_point, &nano::rep_last_heard_info::last_heard>>,
-	boost::multi_index::hashed_unique<boost::multi_index::member<nano::rep_last_heard_info, nano::account, &nano::rep_last_heard_info::representative>>>>
-	reps;
+	static uint64_t constexpr weight_period = 5 * 60; // 5 minutes
+	static uint64_t constexpr weight_samples = 4032; // 2 weeks
```
wezrule: Is this comment correct?

nano/node/lmdb.cpp
```diff
@@ -717,8 +760,9 @@ blocks_info (0),
 representation (0),
 unchecked (0),
 vote (0),
-meta (0),
-stopped (false)
+online_weight (0),
+stopped (false),
+meta (0)
```
wezrule: meta and stopped have swapped order, so they now no longer match the same order in the class definition.

cryptocode: Probably happened during my rebase.

nano/node/node.hpp
cryptocode: I think it means two weeks worth of samples - the math checks out at least. Maybe just add that to the comment.

wezrule: ah, perhaps it could be made more clear then :)

nano/node/lmdb.cpp
SergiySW: I rebased after your PR to master, @wezrule . online_weight { 0 } is now in header file

wezrule: ok thanks

nano/node/lmdb.cpp
```diff
@@ -699,6 +712,36 @@ nano::store_iterator<nano::account, std::shared_ptr<nano::vote>> nano::mdb_store
...
+{
+	return nano::store_iterator<uint64_t, nano::amount> (std::make_unique<nano::mdb_iterator<uint64_t, nano::amount>> (transaction_a, online_weight));
+}
+
+nano::store_iterator<uint64_t, nano::amount> nano::mdb_store::online_weight_end ()
+{
+	return nano::store_iterator<uint64_t, nano::amount> (nullptr);
+}
+
+size_t nano::mdb_store::online_weight_count (nano::transaction const & transaction_a)
```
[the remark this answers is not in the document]
wezrule: Done

nano/node/node.cpp
wezrule: Done

nano/node/node.hpp
wezrule: Added a more descriptive comment

wezrule:
As discussed it would be nice to have a way to query/clear the data from CLI, but this can be another PR.

PR #1650, opened 2019-01-27
Fix race conditions reported by TSAN

cryptocode:
@wezrule Wanna see if this fixes the TSAN traces you noticed during bootstrap? Should also fix a shutdown race.


wezrule:
I still get some TSAN errors, but as discussed it makes sense to not do io on a closed socket anyway. I can have a look at this later as my system seems to reproduce it a bit easier.

Issue #1606, opened 2019-01-18
Multithread signature_checker


zhyatt:
Create hardware_concurrency() number of threads and divide each signature set between the threads.

We probably don't want to divide trivial number of verifications, so some good lower-level cutoff where we'd only use one thread.

termoose:
Added a pull request for this #1611, it's threading the `nano::signature_checker::verify` function based on `std::thread::hardware_concurrency()` number of threads.

Not sure what should be set as a trivial number of verifications, I just used the number of the `many` test case. Added a new test case as well with `+1` verifications to trigger the multi-threaded version of the code.

Seeing around a 2x speedup on my 4 core laptop.

```
./core_test --gtest_filter=signature*
Running main() from gtest_main.cc
Note: Google Test filter = signature*
[==========] Running 4 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 4 tests from signature_checker
[ RUN      ] signature_checker.empty
[       OK ] signature_checker.empty (0 ms)
[ RUN      ] signature_checker.many
[       OK ] signature_checker.many (35 ms)
[ RUN      ] signature_checker.many_threaded
[       OK ] signature_checker.many_threaded (17 ms)
[ RUN      ] signature_checker.one
[       OK ] signature_checker.one (0 ms)
[----------] 4 tests from signature_checker (52 ms total)

[----------] Global test environment tear-down
[==========] 4 tests from 1 test case ran. (52 ms total)
[  PASSED  ] 4 tests.
```

wezrule:
@termoose thanks for the effort! This is something we would like to get in soon, so let me know if you need any help with the PR remarks. In essence we want a thread pool, where each thread is operating on 256 batched chunks.

wezrule:
Hard to say. When bootstrapping from 0 there will be a lot, but in real-time probably not that many (at least yet). I think this task is to improve bootstrapping so I would give that priority.

termoose:
Closed the old PR, opened a new one #1614. It might be a bit faster than spawning threads each time, around 2.2x speedup on my 4 cores. We could experiment with the `batch_size` as well as the cut-off value for doing the single-threaded version. The current values seem fine from the tests I ran locally.

guilhermelawless:
8 cores results with #1614 :   2x from 1000 (single) to 1001 (16 threads). Of course with the 256 batch size it's not using all of them.

Modified the cut-off to 100k just for testing and it yields a 7.5x speedup.

[...]

PR #1651, opened 2019-01-27
Multi-thread the signature checker

wezrule:
This solves #1606 and provides a finished implementation of #1614 which which was created from the work of @termoose @guilhermelawless so thanks a lot!

Currently the vote and block processors add checks to the `signature checker` which runs on a separate thread and sets the value of a promise which the calling threads are blocked on until it is set. Having it in a separate thread doesn't really do much because the calling threads are blocked straight away so the caller might as well do the work themselves. I have removed that thread and added a thread pool which can be user assigned on the config.json file and is called if the batch size of the signatures to check are > 512. This value was heuristically chosen as the largest size seemed to be 2048 (hardcoded) and 256 size chunks were found to be optimal in terms of memory size and speed benchmarked by Colin in 57d819b which is what each thread on a thread pool works on.

A new test was added which runs over multiple threads and the promises are checked after all signature checks are added to the thread pool rather than waiting for each one to finish (which checks overloading the boost thread pool). Using this test and changing the cores I got the following results:
1 core - 5.9s
2 cores - 5.4s
4 cores - 3.4s
8 cores - 2.5s

This is by no means a benchmark test as it needs to run quickly on any number of cores. A more biased test which has more messages of a larger size greatly favours the higher amount of cores used. 

I ran TSAN for a few hours on the live network and it showed no new errors.

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1100,11 +1100,8 @@ void nano::vote_processor::verify_votes (std::deque<std::pair<std::shared_ptr<na
 		pub_keys.push_back (vote.first->account.bytes.data ());
 		signatures.push_back (vote.first->signature.bytes.data ());
 	}
-	std::promise<void> promise;
-	nano::signature_check_set check = { size, messages.data (), lengths.data (), pub_keys.data (), signatures.data (), verifications.data (), &promise };
-	std::future<void> future = promise.get_future ();
+	nano::signature_check_set check = { size, messages.data (), lengths.data (), pub_keys.data (), signatures.data (), verifications.data () };
 	node.checker.verify (check);
-	future.wait ();
 	std::remove_reference_t<decltype (votes_a)> result;
 	auto i (0);
 	for (auto & vote : votes_a)
@@ -1247,11 +1244,12 @@ bool nano::rep_crawler::exists (nano::block_hash const & hash_a)
 
 nano::signature_checker::signature_checker (unsigned num_threads) :
 thread_pool (num_threads),
-single_threaded (num_threads == 1)
+single_threaded (num_threads == 0),
+num_threads (num_threads)
 {
 	if (!single_threaded)
 	{
-		set_name_threads (num_threads);
+		set_thread_names (num_threads);
 	}
 }
 
@@ -1264,7 +1262,7 @@ void nano::signature_checker::verify (nano::signature_check_set & check_a)
 {
 	{
 		// Don't process anything else if we have stopped
-		std::lock_guard<std::mutex> guard (mutex);
+		std::lock_guard<std::mutex> guard (stopped_mutex);
 		if (stopped)
 		{
 			return;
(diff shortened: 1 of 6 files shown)


nano/node/node.cpp
```diff
@@ -1120,72 +1118,119 @@ nano::signature_checker::~signature_checker ()
...
+		int start_index = batch * batch_size;
+
+		// Clamp it to the max number of verifications
+		if (start_index + batch_size > check_a.size)
 		{
-			condition.wait (lock);
+			size = overflow;
 		}
+
+		boost::asio::post (thread_pool, [this, task, size, start_index, batch] {
```
cryptocode: `batch` is captured but not used

nano/node/node.cpp
```diff
@@ -1120,72 +1118,119 @@ nano::signature_checker::~signature_checker ()
...
+	bool ready = false;
+	int pending = num_threads;
+	std::condition_variable cv;
+	std::mutex mutex_l;
+	std::unique_lock<std::mutex> lk (mutex_l);
+
+	for (int i = 0; i < num_threads; ++i)
+	{
+		boost::asio::post (thread_pool, [&cv, &ready, &pending, &mutex_l]() {
+			nano::thread_role::set (nano::thread_role::name::signature_checking);
```
cryptocode: This doesn't appear to work properly, only 2 out of 4 threads are named in my test run - probably because ::post picks the same two threads in the loop.

cryptocode: Would be nice to make it work, thread tracking was a v17 feature and it may help diagnose issues in production.

nano/node/node.cpp
wezrule: Good spot, removed

nano/node/node.cpp
wezrule: Ah ok, thanks for informing me 👍. Can you try the latest code and see if it works on yours now? (and more importantly actually exits the function); I tried it on windows/mac and it works ok even with 128 threads. Each thread now waits on a condition variable when reached, and the last one wakes them up which sets a promise so that we now when all threads have actually finished. 

I couldn't find anything that says calling `get_future()` on a `promise` is thread safe so I now get it before any threads are called.

I changed some inconsistent size_t/unsigned types which were causing build warnings

wezrule:
I've changed it so that the calling thread does work verifying signatures instead of just blocking and made the new test (and an existing one) check for an incorrect signature as well.

nano/node/node.cpp
cryptocode: Works fine here

nano/node/node.cpp
```diff
@@ -1120,72 +1117,152 @@ nano::signature_checker::~signature_checker ()
...
+			promise.set_value ();
+		});
+	}
+
+	// Wait until all threads have finished
+	for (auto & future : futures)
+	{
+		future.wait ();
 	}
+	release_assert (pending == 0);
```
cryptocode: Maybe a normal debug assert is sufficient here as it's not critical to the node's operation that threads have names.

nano/node/node.cpp
wezrule: @cryptocode Thanks for checking

nano/node/node.cpp
wezrule: done

SergiySW:
I suppose we can increase hardcoded values as
`verify_state_blocks (transaction, lock_a, 2048 * signature_checker_threads);`
or something like that for more efficency

nano/node/node.cpp
```diff
@@ -1120,72 +1117,152 @@ nano::signature_checker::~signature_checker ()
...
+			return;
+		}
 	}
-	condition.notify_all ();
+
+	if (check_a.size < multithreaded_cutoff || single_threaded)
+	{
+		// Not dealing with many signatures so just use the calling thread for checking signatrure
+		verify_batch (check_a, 0, check_a.size);
+		return;
```
guilhermelawless: Missing a release_assert on results of this verify_batch

wezrule: Good spot, thanks!

wezrule:
@SergiySW this was noted in the other PR but I wasn't sure if there was a specific reason 2048 was chosen so I was hesitant to change it. It does make sense though as this change would help greatly for large message sizes. I have made the change and done it for forced state blocks as well. I tested it on mac and worked ok (can process 8192 messages on 4 core cpu fine :)). It may be a problem if they choose a thread size which is a lot greater than their actual number of CPU cores (as it will try and process a lot of messages! with no gain as the threads will interleave), but then they deserve what they get tbh :).

SergiySW:
@wezrule are there downsides for 1 core with 4 default checker threads?

wezrule:
@SergiySW good spot! Changed to account for this.

nano/node/node.cpp
```diff
@@ -1261,72 +1258,153 @@ nano::signature_checker::~signature_checker ()
...
+			{
+				--tasks_remaining;
+				promise.set_value ();
+			}
+		});
+	}
+}
+
+// Set the names of all the threads in the thread pool for easier identification
+void nano::signature_checker::set_thread_names (unsigned num_threads)
```
rkeene: It might be simpler to just call:

```
nano::thread_role::set (nano::thread_role::name::signature_checking);
```

within the lambda in `verify_async()` (above) and update `thread_role::set()` to be idempotent

wezrule: yeh that's possible but it might pollute/distract readers from the hot path code in verify_async

renesq:
The `config.signature_checker_threads` variable being set to 0 by default made me a bit anxious, until I found this code line
`		verify_state_blocks (transaction, lock_a, 2048 * (node.config.signature_checker_threads + 1));`
Which means that the config documentation in the wiki apparently is not very helpful. It sais  `"signature_checker_threads": "1", // Number of threads to use for verifying signatures`
I didn't quite grasp the numeral concept that kicks in here. It seems like these "threads" are different from regular boost threads as they are used in the other sections of the config. Does the setting even matter?

clemahieu:
The intention for the thread classes we use are wher N = hardware_concurrency:
N signature verification threads at normal priority for signature verification 
N work-generation threads at idle priority
N IO threads at normal priority

IO threads are supposed to be very short-lived operations and thus, we don’t need more than the number of cores or there’s an issue where the long operations are happening in IO threads. This is currently the case and we want to remove all synchronous disk operations from IO threads since that’s a bottleneck.

Work generation threads are at idle priority since if the node or other processes are using priority, we don’t need to fight for it, generating work for new transactions isn’t essential to node operation.

Signature processing threads are at normal priority and one per CPU to make sure we’re using as much of the computational resources as are needed to process vote and block traffic.

guilhermelawless:
If it says `"signature_checker_threads": "1"` in the config wiki and that's the default, it's because it was generated from a dual core machine. When creating the config it defaults to number of cores minus one.

When verifying signatures, N-1 threads are created to do batches of [by default] 256 signatures, and the remainder goes to the calling thread (thus in total, N threads doing work).

If `config.signature_checker_threads = 0` that's the default for a single core machine.

wezrule:
The vote/block processor threads currently just wait for the result from the signature checker thread pool, so instead of waiting they do work as well. In the case of a small number of verifications (less than 256 if I recall) all the work is done on the calling thread, so no scheduling is necessary. So there is an "implicit" signature checker thread if you will, but I can see the confusion from a users point of view though.

PR #1661, opened 2019-01-29
Migrate Gtest to submodule

argakiig:
https://github.com/google/googletest/ checked out at https://github.com/google/googletest/commit/dea0216d


SergiySW:
Builds failing right now, I guess some Cmake & include headers paths changes required

argakiig:
yeah looks like my cmake cache still had the path, lemme see what i can do to fix it

PR #1663, opened 2019-01-30
Fix windows tests for ipc

cryptocode:
Which was actually a string lifetime issue spotted by msvc's iterator debug support.


wezrule:
Working for me now, good job!

PR #1674, opened 2019-01-31
Fix scope issue when setting thread names in the multi-threaded signature checker

wezrule:
There are some intermittent test failures (I could only reproduce them on a MAC), caused by a SIGABRT in the ~std::unique_ptr destructor. It seems that the mutex is going out of scope before the thread pool has fully finished. When the last promise.set_value () is called it notifies the main thread that the thread work is finished, and then exits the function (set_thread_names), but the threads haven't fully finished (i.e they still need to call destructors). I'm now (re)using the class member mutex so that it stays in the scope during this cleanup.

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1374,7 +1374,7 @@ void nano::signature_checker::verify (nano::signature_check_set & check_a)
 {
 	{
 		// Don't process anything else if we have stopped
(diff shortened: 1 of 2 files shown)


wezrule:
Thanks! Yeh this was changed because the calling thread now does work, so num_threads is 'extra' threads doing work.

PR #1676, opened 2019-02-01
Remove hardcoded FindBoost.cmake and instead rely on local default

### LINKED ISSUE #1597: Update FindBoost.cmake
Newer versions of Boost need newer FindBoost.cmake, apparently -- verify and update.

argakiig:
closes #1597


cryptocode:
Approved based on it working on my systems, could have the issues Serg mentions.

Issue #1658, opened 2019-01-29
Fix all warnings


wezrule:
We currently have a few warnings being generated in our build. It would be good to fix/suppress 3rd party ones so that new ones don't sneak in,

wezrule:
Warnings caused directly by us should now be resolved in #1677. No more should be added here on out 👍

PR #1677, opened 2019-02-01
Fix our own compiler warnings

wezrule:
This solves the "Internal warnings" part of #1658

A *full clean build was done on:
Windows (Visual Studio 2017)
Mac (XCode + Clang)
Fedora (gcc)

*The wallet was only build on Windows

All warnings were analysed and ones which were caused by us (i.e not third party ones) should now be fixed, with an exception detailed at the end. These include:

- 'peer_put' overrides a member function but is not marked 'override'. Solution - add override
- clGetPlatformIDs' is deprecated. Solution - add `CL_SILENCE_DEPRECATION` above the opencl includes inside `#define __APPLE__`
- Field 'pull' will be initialized after field 'known_account'. Solution - put them in matching order
- Lambda capture 'block_a' is not used. Solution - remove it
- Array index -8388608 is before the beginning of the array. This was a somewhat special one, as the test was checking the address sanitizer, but currently it is DISABLED_. It only appears with gcc/clang so I have suppressed the warning for these compilers.  
- Unused variable 'value'  Solution - remove if not used, otherwise add `(void)`
'argument': conversion from 'size_t' to 'uint32_t', possible loss of data. Solution - do appropriate casts

warning C4996: 'localtime': This function or variable may be unsafe. Consider using localtime_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
warning C4996: 'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
warning C4996: 'strncpy': This function or variable may be unsafe. Consider using strncpy_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.

- The suggested functions above are not standard (MSVC specific) and are only optional with the current compilers we support. Solution - Add -D_CRT_SECURE_NO_WARNINGS to cmake file when using Windows.

- warning: zero-length gnu_printf format string [-Wformat-zero-length] printf (""); . Solution - remove it

The only warnings left are:
warning C4996: 'xrb_generate_random': was declared deprecated

We recently deprecated them, so wasn't sure if we should suppress the warnings, as it will be nice to have a constant reminder about them so that they don't get forgotten.

Other changes: 
interface.cpp - The array only needs a buffer of 65.
promise.get_future () is called before the thread is created as it is not guaranteed to be thread safe (only in areas where code was changed, I haven't checked them all).


nano/node/rpc.cpp
```diff
@@ -708,7 +708,8 @@ void nano::rpc_handler::account_representative_set ()
 			{
 				bool generate_work (work == 0); // Disable work generation if "work" option is provided
 				auto response_a (response);
-				wallet->change_async (account, representative, [response_a](std::shared_ptr<nano::block> block) {
+				// clang-format off
+                wallet->change_async (account, representative, [response_a](std::shared_ptr<nano::block> block) {
```
cryptocode: Something happened to indentation here it seems, and with clang-format off there're no complaints :D

wezrule: What should I do here? :) This is what happens when I tab it to the correct place and run clang-format

wezrule: Using my mac for checking in for the first time, my Xcode preferences were changing tabs to spaces. Fixed :)

nano/node/testing.cpp
```diff
@@ -234,7 +235,8 @@ void nano::system::generate_activity (nano::node & node_a, std::vector<nano::acc
 
 nano::account nano::system::get_random_account (std::vector<nano::account> & accounts_a)
 {
-	auto index (random_pool.GenerateWord32 (0, accounts_a.size () - 1));
+	assert (std::numeric_limits<CryptoPP::word32>::max () > accounts_a.size ());
```
cryptocode: release_assert seems to be used other places for std::numeric_limits<CryptoPP::word32>::max () checks. Maybe it should be `assert` everywhere?

wezrule: `assert` was used here because it is only used by the tests. I think `release_assert` makes sense in the other places as bad things will happen in production if it overflows regardless.

PR #1680, opened 2019-02-01
Create and use new network helper constants

wezrule:
We do a lot of conditional checking like `nano::nano_network == nano::nano_networks::nano_test_network` I have created some helper constexpr variables which can be used instead e.g `nano::is_test_network`.

It especially helps in situations like this:
`(nano::nano_network == nano::nano_networks::nano_test_network ? nano_test_account : nano::nano_network == nano::nano_networks::nano_beta_network ? nano_beta_account : nano_live_account)`

Which becomes:
`(nano::is_test_network ? nano_test_account : nano::is_beta_network ? nano_beta_account : nano_live_account)`

### DIFF
--- a/nano/lib/config.hpp
+++ b/nano/lib/config.hpp
@@ -21,6 +21,10 @@ enum class nano_networks
 	nano_live_network = 2,
 	rai_live_network = 2,
 };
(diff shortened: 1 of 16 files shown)


wezrule:
Thanks!

PR #1684, opened 2019-02-01
add CLI switch for adjusting Sideband upgrade batch sizes

argakiig:
to improve UI and RPC responsiveness


nano/nano_node/entry.cpp
```diff
@@ -78,9 +79,18 @@ int main (int argc, char * const * argv)
 		}
 	}
 
+	size_t batch_size;
+	if (vm.count ("batch_size"))
+	{
+		batch_size = (vm["batch_size"].as<size_t> ());
```
wezrule: You are ruining my work :P (See #1)
https://github.com/nanocurrency/nano-node/pull/1514

nano/nano_wallet/entry.cpp
```diff
@@ -358,13 +358,23 @@ int main (int argc, char * const * argv)
...
 						data_path = nano::working_path ();
 					}
+					size_t batch_size;
+					if (vm.count ("batch_size"))
+					{
+						batch_size = (vm["batch_size"].as<size_t> ());
+					}
+					else
+					{
+						batch_size = 512;
```
SergiySW: I guess no need to set 512 as this is sideband_batch_size default value

SergiySW: Probably all this can be shortened to something like
`if (vm.count ("batch_size"))
{
	flags.sideband_batch_size = (vm["batch_size"].as<size_t> ());
}`

nano/node/lmdb.cpp
```diff
@@ -1205,10 +1205,10 @@ void nano::mdb_store::upgrade_v11_to_v12 (nano::transaction const & transaction_
 	mdb_drop (env.tx (transaction_a), checksum, 1);
 }
 
-void nano::mdb_store::upgrade_v12_to_v13 ()
+void nano::mdb_store::upgrade_v12_to_v13 (size_t const batch_size)
 {
 	size_t cost (0);
-	size_t const max (16384);
+	size_t const max (512);
```
SergiySW: Is batch_size used here?

nano/node/node.cpp
```diff
@@ -2060,7 +2060,7 @@ io_ctx (io_ctx_a),
 config (config_a),
 alarm (alarm_a),
 work (work_a),
-store_impl (std::make_unique<nano::mdb_store> (init_a.block_store_init, config.logging, application_path_a / "data.ldb", config_a.lmdb_max_dbs, drop_unchecked)),
+store_impl (std::make_unique<nano::mdb_store> (init_a.block_store_init, config.logging, application_path_a / "data.ldb", config_a.lmdb_max_dbs, drop_unchecked, flags.sideband_batch_size)),
```
SergiySW: I guess then we have to pass flags at node creation like `nano::node::node (nano::node_init & init_a, boost::asio::io_context & io_ctx_a, boost::filesystem::path const & application_path_a, nano::alarm & alarm_a, nano::node_config const & config_a, nano::work_pool & work_a, **nano::node_flags & flags_a**, bool drop_unchecked) :`

SergiySW: That will also allow make drop_unchecked a flag

nano/node/lmdb.cpp
argakiig: sent the wrong branch up good catch

nano/node/lmdb.cpp
```diff
@@ -1205,10 +1205,10 @@ void nano::mdb_store::upgrade_v11_to_v12 (nano::transaction const & transaction_
 	mdb_drop (env.tx (transaction_a), checksum, 1);
 }
 
-void nano::mdb_store::upgrade_v12_to_v13 ()
+void nano::mdb_store::upgrade_v12_to_v13 (size_t const batch_size)
 {
 	size_t cost (0);
-	size_t const max (16384);
+	size_t const max (batch_size);
```
SergiySW: You could pass max here `void nano::mdb_store::upgrade_v12_to_v13 (size_t const max)` or make `if (cost >= batch_size)`

PR #1687, opened 2019-02-02
fixing bootstrap_max being overwritten

argakiig:
It would appear that bootstrap_connections_max is being overwritten to 4 when run from the wallet regardless of the existing config settings, looks to have been part of
https://github.com/nanocurrency/nano-node/commit/e5c2b66959168da2140a22c146fa965a88c225a8

@SergiySW was this intentional or a side effect of something else

### DIFF
--- a/nano/nano_wallet/entry.cpp
+++ b/nano/nano_wallet/entry.cpp
@@ -135,7 +135,7 @@ public:
 		json.put ("account", account.to_account ());
 		nano::jsonconfig node_l;
 		node.enable_voting = false;
-		node.bootstrap_connections_max = 4;
+		json.put ("bootstrap_connections_max", node.bootstrap_connections_max);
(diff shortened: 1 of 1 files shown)


SergiySW:
Is wallet updating config each launch?
Anyway no reason for desktop wallets allow 64 incoming connections, it can kill HDD wallet or poor connection easily

nano/nano_wallet/entry.cpp
```diff
@@ -135,7 +135,7 @@ class qt_wallet_config
 		json.put ("account", account.to_account ());
 		nano::jsonconfig node_l;
 		node.enable_voting = false;
```
SergiySW: No reason to keep this line as we canged default config to false

argakiig:
seems to be working as intended I am not sure what I did to create my issue. closing

PR #1696, opened 2019-02-04
Build/debug beta

### LINKED ISSUE #1470: When building for Beta or Test default to a Debug build
When building for Beta or Test, default to a Debug build

argakiig:
closes #1470

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,7 @@ after_build:
 - cmd: >-
     cpack -C %CONFIGURATION% --verbose --config ./CPackConfig.cmake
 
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -26,11 +26,11 @@ build:
   verbosity: minimal
 after_build:
 - cmd: >-
-    cpack -NSIS --verbose
+    cpack -C %CONFIGURATION% --verbose --config ./CPackConfig.cmake
```
SergiySW: Is there file CPackConfig.cmake in repo?

argakiig: it gets generated as part of the cmake process when `include (CPack)` is called in the CMakeLists.txt file

PR #1697, opened 2019-02-04
Sign windows

### LINKED ISSUE #575: Signing executables
The Windows installer and binary should be digitally signed to avoid SmartScreen/AV issues.

argakiig:
fixes #575 automating signing of binaries


argakiig:
openned after updating appveyor settings

PR #1704, opened 2019-02-05
 State blocks subtype check for RPC "process"

SergiySW:


argakiig:
windows failures are expected

PR #1708, opened 2019-02-07
ipc: Handle rpc actions doing async ops

cryptocode:
Thanks to dotcom for finding this


cryptocode:
@guilhermelawless Thanks for checking, much appreciated!

PR #1711, opened 2019-02-08
Make work faster

DanielMazurFL4RE:
Current PoW take too much time to calculate on smartphones or regular machines, I've spent few days writing own optimized "miner", checked few PoW scripts (like this written in WASM) and they takes approx. 40 seconds to mine on PC and approx. 120 seconds on regular smartphone.
How this coin may be adopted by masses If guy who want to pay for coffee must stay by shop counter 2 minutes?

We need to find other way to prevent spam attacks!

### DIFF
--- a/nano/lib/work.hpp
+++ b/nano/lib/work.hpp
@@ -45,7 +45,7 @@ public:
 	nano::observer_set<bool> work_observers;
 	// Local work threshold for rate-limiting publishing blocks. ~5 seconds of work.
 	static uint64_t const publish_test_threshold = 0xff00000000000000;
-	static uint64_t const publish_full_threshold = 0xffffffc000000000;
+	static uint64_t const publish_full_threshold = 0xffffff0000000000; //make it faster dudes, it takes 2 to 200 seconds to calculate, do the real 5 seconds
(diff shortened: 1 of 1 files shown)


wezrule:
PoW can be precomputed for the next block so for everyday transactions like getting coffee it should not be problem. For services requiring high throughput and also done server side using dPoW.

DanielMazurFL4RE:
Alright but what if I have few pending blocks or want to send NANOs few times?
As you can see, then it may take long time if user have no luck and won't time valid work in few first seconds

rotilho:
How avoid spam in this case? Just reduce the difficult can create a even bigger problem.

Keep the wallet open for while (and using dPoW or Work Server) sounds quite reasonable for me and payment processors will have good hardware to compute pow and they can also shard their wallet to increase TPS.

zhyatt:
@DanielMazurFL4RE Thanks for submitting. The PoW is meant to be a spam reduction mechanism so reducing difficulty decreases protection against spam. At this time we are not interested in doing so, so closing this ticket. In the future if we decide this is an option we can reopen and evaluate.

Issue #1656, opened 2019-01-28
Not all versions of GCC support -fsanitize-blacklist


rkeene:
TSAN blacklist support (added in #1637) is not available in all versions of GCC.

rkeene:
Ideally the build system would detect whether or not the compiler supports this.

wezrule:
gcc does not support this option, but you can use a suppressions file in the TSAN_OPTIONS environment variable:
https://github.com/nanocurrency/nano-node/pull/1941
Documentation: https://github.com/nanocurrency/nano-node/wiki/Testing

PR #1713, opened 2019-02-08
Only blacklist the TSAN file when using clang compiler

wezrule:
This solves #1656

When enabling TSAN on my system using gcc my build fails because `-fsanitize-blacklist` is not supported. I'm not sure if gcc currently supports this at all, at least the latest version (8.2.1) on the Fedora 29 repository doesn't. I tried using `check_cxx_compiler_flag` to check if `-fsanitize-blacklist` but this didn't work on MAC, so I'm now just checking for the only supported compiler (Clang).

In my searching I have found:
https://github.com/google/sanitizers/wiki/ThreadSanitizerSuppressions which allows a file to be specified with special key:value pair options using the environment variable: TSAN_OPTIONS=suppressions:_filename_ .This may be useful in future

Tested on a system which supports (Mac/Clang) and doesn't (Linux/gcc).

Unrelated: exception specifiers are deprecated in c++11 so display a warning, this is removed.


wezrule:
Checking in for now, until we have support for the suppressions TSAN_OPTIONS file

PR #1714, opened 2019-02-08
Find pthread in CMakeLists.txt

wezrule:
When using gcc 8.2 with Eclipse generated Makefiles on Linux I get unresolved pthread symbols. It appears to only be a gcc error where -lpthread is explicitly required. The recommendation is to use `find_package (Threads)`:
https://cmake.org/pipermail/cmake/2016-February/062729.html

Tested on Mac/clang, Fedora/gcc, and appveyor (Windows) seems to build it ok as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,9 +30,7 @@ endif()
 
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
-# TODO check if any library requires these
-#find_package(Threads REQUIRED)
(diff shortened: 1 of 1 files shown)


devinus:
I think we might actually need those two lines just uncommented for cross-platform builds.

wezrule:
I'm not sure as we only call `${PLATFORM_LIBS}` in nano_node so the `target_link_libraries` would need changing elsewhere. `boost::thread` has a dependency on the import target `Thread::Thread` which is called in the `target_link_libraries` of the node library's `CMakeList.txt` file, so I think that should resolve it for us (could be wrong)

PR #1721, opened 2019-02-12
[WIP] Endpoint abstraction

cryptocode:
In prep for tcp support, this PR provides an abstraction for the asio endpoint types, and a uniform socket api. 

As the nano namespace is getting overloaded, a `nano::net` namespace is introduced for core networking. The `nano::net::socket_addr` type replaces all uses of asio's udp and tcp specific endpoints (our internal aliases are removed as well). The hashing and ordering behavior should be retained. The  `nano::net::client` and subtypes replaces the socket type previously in bootstrap.

A few ideas are explored at the bottom half of network_generic.cpp.

Tests are passing locally and the node is syncing.

### DIFF
--- a/nano/core_test/CMakeLists.txt
+++ b/nano/core_test/CMakeLists.txt
@@ -10,6 +10,7 @@ add_executable (core_test
 	gap_cache.cpp
 	ledger.cpp
 	network.cpp
+	network_generic.cpp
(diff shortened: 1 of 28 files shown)


cryptocode:
To be replaced with a sink based api

PR #1724, opened 2019-02-12
 Cleanup unchecked table after each bootstrap attempt

SergiySW:


SergiySW:
@wezrule or after every bootstrap attempt if automatic legacy bootstrap is not disabled

Issue #1336, opened 2018-10-29
PoW Client Compute and Resend Work based on network load


rkeene:
Follow up to #1298 

Update the work computation system on the node to increase and resend blocks with greater work based on the current network load and the likelyhood that the block will be processed.  This may be exposed to the user as an RPC knob.

rkeene:
1. OpenCL + Work Peers
  2. Track Locally Generated Blocks from Generation to Confirmation, resubmit if not confirmed in some epsilon

argakiig:
- [ ] calculate median active_transaction difficulty
- [ ] calculate work generation based on max(threshold, median active_transaction difficulty)

guilhermelawless:
How will the cutoff point be decided for the node using the pre-computed work vs. computing new work with higher difficulty?

I'm thinking tracking other percentiles apart from the median might help in that decision.

zhyatt:
@guilhermelawless Do you mean if a node has pre-computed work handy which is below the current median difficulty value, how will we decide whether to attempt using it vs. dumping it and computing a new higher value?

guilhermelawless:
Exactly that, which is why I suggested tracking other percentiles.

Example to start the discussion. If the median difficulty has risen by 70% above the current precomputed work, should the node compute new work or not? Well, if we go check the 75th percentile difficulty and it is actually below our current work value, then it might not make sense to.

Just an idea to start the discussion. It doesn't look like something that will matter much, but incorrect tuning of parameters could lead to greedy nodes that are competing against each other in the event of a very high tx throughput.

PR #1725, opened 2019-02-13
track avg work difficulty of active_transactions containers

argakiig:
related to #1336 will satisfy first part

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -3534,6 +3534,28 @@ bool nano::active_transactions::active (nano::block const & block_a)
 	return roots.find (nano::uint512_union (block_a.previous (), block_a.root ())) != roots.end ();
 }
 
+uint64_t nano::active_transactions::active_difficulty ()
+{
+	nano::network_params params;
+	uint64_t result (params.publish_threshold);
+	{
+		std::lock_guard<std::mutex> lock (mutex);
+		auto size_l (roots.size ());
+		switch (size_l)
+		{
+			case 0:
+				break;
+			case 1:
+				result = roots.get<1> ().begin ()->difficulty;
+				break;
+			default:
+				result = roots.get<1> ().find ((roots.get<1> ().begin ()->difficulty + (--roots.get<1> ().end ())->difficulty) / 2)->difficulty;
+				break;
+		}
+	}
+	return result;
+}
+
 void nano::active_transactions::update_difficulty (nano::block const & block_a)
 {
 	std::lock_guard<std::mutex> lock (mutex);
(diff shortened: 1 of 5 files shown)


nano/node/node.cpp
```diff
@@ -3520,6 +3520,7 @@ void nano::active_transactions::stop ()
 bool nano::active_transactions::start (std::shared_ptr<nano::block> block_a, std::function<void(std::shared_ptr<nano::block>)> const & confirmation_action_a)
 {
 	std::lock_guard<std::mutex> lock (mutex);
+	active_total_work += block_a->block_work ();
```
clemahieu: I think this needs to be adding in the difficulty rather than the random nonce.

nano/node/node.cpp
```diff
@@ -3603,9 +3604,11 @@ void nano::active_transactions::update_difficulty (nano::block const & block_a)
 		uint64_t difficulty;
 		auto error (nano::work_validate (block_a, &difficulty));
 		assert (!error);
+		active_total_work -= existing->difficulty;
 		roots.modify (existing, [difficulty](nano::conflict_info & info_a) {
 			info_a.difficulty = difficulty;
 		});
+		active_total_work += block_a.block_work ();
```
clemahieu: Difficulty instead of nonce.

nano/node/node.hpp
```diff
@@ -114,6 +114,8 @@ class active_transactions
 	bool vote (std::shared_ptr<nano::vote>, bool = false);
 	// Is the root of this block in the roots container
 	bool active (nano::block const &);
+	//calculate median difficulty;
```
SergiySW: Average?

nano/node/node.cpp
```diff
@@ -3534,6 +3534,28 @@ bool nano::active_transactions::active (nano::block const & block_a)
...
+		auto size_l (roots.size ());
+		switch (size_l)
+		{
+			case 0:
+				break;
+			case 1:
+				result = roots.get<1> ().begin ()->difficulty;
+				break;
+			default:
+				result = roots.get<1> ().find ((roots.get<1> ().begin ()->difficulty + (--roots.get<1> ().end ())->difficulty) / 2)->difficulty;
```
SergiySW: Not sure if such estimation makes sense

argakiig: if not median difficulty, what would you think to be better?

nano/node/node.cpp
```diff
@@ -3534,6 +3534,28 @@ bool nano::active_transactions::active (nano::block const & block_a)
 	return roots.find (nano::uint512_union (block_a.previous (), block_a.root ())) != roots.end ();
 }
 
+uint64_t nano::active_transactions::active_difficulty ()
+{
+	nano::network_params params;
+	uint64_t result (params.publish_threshold);
```
cryptocode: This can use `node.network_params.publish_threshold` instead of a new params instance.

argakiig:
replaced with #1858

PR #1727, opened 2019-02-13
RPC unopened

### LINKED ISSUE #1166: Enhancement Request: Update Ledger RPC action to show unopened accounts with pending balance
**Description of bug:**
When using the RPC action "ledger" it appears to not show unopened accounts.  Is it possible to enhance the action to return all accounts including those that do not have frontier blocks?  Or create another action to return all accounts with a pending balance that are not opened yet?

1. Parsing through all 560k+ accounts from the ledger action results in a total about 2 million NANO less than the available supply.  Best guess is these are made up of unopened accounts since I was able to verify that an unopened account without a frontier block does not show up from the ledger action.  
2. Running the action "account_balance" on an unopened account returns the correct pending balance.

Expected to see the total balance+pending from accounts returned by the "ledger" action to match the available supply.  Results of all accounts from "ledger" is  131262162524572716751492490483419331982.

Expected to see the total of balance+pending from the "ledger" action to match the available supply of 133248289196499221154116917710445381553

If it doesn't make sense to update the "ledger" action to show unopened accounts then could another action be added that shows all accounts that are unopened with a pending balance?

guilhermelawless:
Fixes #1166 .

Wiki update:

## Unopened

_enable_control required, version 19.0+_

Returns the total pending balance for unopened accounts in the local database, starting at **account** (optional) up to **count** (optional), sorted by account number. _**Notes:**_ By default excludes the burn account.

Request:
  ```
  {  
    "action": "unopened",  
    "account": "xrb_1111111111111111111111111111111111111111111111111111hifc8npp",   
    "count": "1"
  }
  ```

Response:
  ```
  {
    "accounts": {
      "xrb_1111111111111111111111111111111111111111111111111111hifc8npp": "207034077034226183413773082289554618448"
    }
  }
  ```

### DIFF
--- a/nano/node/rpc.cpp
+++ b/nano/node/rpc.cpp
@@ -3316,10 +3316,11 @@ void nano::rpc_handler::unchecked_keys ()
 
 void nano::rpc_handler::unopened ()
 {
-	std::unordered_map<nano::account, nano::uint128_t> unopened;
 	auto transaction (node.store.tx_begin_read ());
-	auto iterator (node.store.pending_begin (transaction));
+	auto iterator (node.store.pending_begin (transaction, nano::pending_key (1, 0))); // exclude burn account
 	auto end (node.store.pending_end ());
+	nano::account current_account;
+	nano::uint128_t current_account_sum{ 0 };
 	while (iterator != end)
 	{
 		nano::pending_key key (iterator->first);
@@ -3332,23 +3333,17 @@ void nano::rpc_handler::unopened ()
 		}
 		else
 		{
-			auto pending_block_amount (info.amount.number ());
-			auto u (unopened.find (account));
-			if (u == unopened.end ())
+			if (account != current_account)
 			{
-				unopened.insert ({ account, pending_block_amount });
+				response_l.put (current_account.to_account (), current_account_sum.convert_to<std::string> ());
+				current_account = account;
+				current_account_sum = 0;
 			}
-			else
-			{
-				u->second += pending_block_amount;
-			}
+			current_account_sum += info.amount.number ();
 			++iterator;
 		}
 	}
-	for (const auto & kv : unopened)
-	{
-		response_l.put (kv.first.to_account (), kv.second.convert_to<std::string> ());
-	}
+	response_l.put (current_account.to_account (), current_account_sum.convert_to<std::string> ()); // last one
 	response_errors ();
 }
 
--- a/nano/node/rpc.cpp
+++ b/nano/node/rpc.cpp
@@ -3317,47 +3317,50 @@ void nano::rpc_handler::unchecked_keys ()
 void nano::rpc_handler::unopened ()
 {
 	rpc_control_impl ();
-	auto transaction (node.store.tx_begin_read ());
-	auto iterator (node.store.pending_begin (transaction, nano::pending_key (1, 0))); // exclude burn account
-	auto end (node.store.pending_end ());
-	nano::account current_account;
-	nano::uint128_t current_account_sum{ 0 };
-	boost::property_tree::ptree accounts;
-	while (iterator != end)
+	if (!ec)
 	{
-		nano::pending_key key (iterator->first);
-		nano::account account (key.account);
-		nano::pending_info info (iterator->second);
-		if (node.store.account_exists (transaction, account))
+		auto transaction (node.store.tx_begin_read ());
+		auto iterator (node.store.pending_begin (transaction, nano::pending_key (1, 0))); // exclude burn account
+		auto end (node.store.pending_end ());
+		nano::account current_account;
+		nano::uint128_t current_account_sum{ 0 };
+		boost::property_tree::ptree accounts;
+		while (iterator != end)
 		{
-			if (account.number () == std::numeric_limits<nano::uint256_t>::max ())
+			nano::pending_key key (iterator->first);
+			nano::account account (key.account);
+			nano::pending_info info (iterator->second);
+			if (node.store.account_exists (transaction, account))
 			{
-				break;
-			}
-			// Skip existing accounts
-			iterator = node.store.pending_begin (transaction, nano::pending_key (account.number () + 1, 0));
-		}
-		else
-		{
-			if (account != current_account)
-			{
-				if (current_account_sum > 0)
+				if (account.number () == std::numeric_limits<nano::uint256_t>::max ())
 				{
-					accounts.put (current_account.to_account (), current_account_sum.convert_to<std::string> ());
-					current_account_sum = 0;
+					break;
 				}
-				current_account = account;
+				// Skip existing accounts
+				iterator = node.store.pending_begin (transaction, nano::pending_key (account.number () + 1, 0));
 			}
-			current_account_sum += info.amount.number ();
-			++iterator;
+			else
+			{
+				if (account != current_account)
+				{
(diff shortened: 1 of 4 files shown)


SergiySW:
It would be more efficient to iterate pending table keys & check each account existence

guilhermelawless:
So the node is aware of all pending blocks? Could you show me an example of where that is done?

SergiySW:
https://github.com/nanocurrency/nano-node/blob/master/nano/nano_node/entry.cpp#L721
Something like
```   
auto iterator (node.store.pending_begin (transaction));  
auto end (node.store.pending_end (transaction));
while (iterator != end)
{
	nano::pending_key key (i->first);
	nano::account account (key.account);
	nano::pending_info info (i->second);
	if (node.store.account_exists (transaction, account))
	{
		// Skip existing accounts
		iterator = node.store.pending_begin (transaction, nano::pending_key (account.number () + 1, 0));
	}
	else
	{
		auto pending_block_amount (info.amount.number ());
		// ...
		// summ for account
		// ...
		iterator++;
	}
}
```

guilhermelawless:
Yep that's much better.

guilhermelawless:
@SergiySW  i'm getting this weird result on beta, the unordered total is larger than before.

```
Supply (available_supply): 340272367920938463463374607431768151531
Total balance with pending (ledger): 232484364820933402871612082323207277732                                                                                                                                                                                                 Total unopened (unopened): 107798002100005060591762525108560933723                                                                                                                                                                                                 Unopened + balance + pending (unopened + ledger): 340282366920938463463374607431768211455
```

SergiySW:
@guilhermelawless could you check it with disconnected node?

nano/node/rpc.cpp
```diff
@@ -3314,6 +3314,44 @@ void nano::rpc_handler::unchecked_keys ()
 	response_errors ();
 }
 
+void nano::rpc_handler::unopened ()
+{
+	std::unordered_map<nano::account, nano::uint128_t> unopened;
```
SergiySW: I guess we need something ordered to receive oredered final result. Also now it's always modifying only last record of set, we don't need to search for random keys

guilhermelawless: The pending iterator guarantees it loops through accounts first? I'll change it

nano/node/rpc.cpp
```diff
@@ -3314,6 +3314,44 @@ void nano::rpc_handler::unchecked_keys ()
...
+	auto end (node.store.pending_end ());
+	while (iterator != end)
+	{
+		nano::pending_key key (iterator->first);
+		nano::account account (key.account);
+		nano::pending_info info (iterator->second);
+		if (node.store.account_exists (transaction, account))
+		{
+			// Skip existing accounts
+			iterator = node.store.pending_begin (transaction, nano::pending_key (account.number () + 1, 0));
```
SergiySW: Guess we need extra check for possible overflow like
iterator = (account.number () == std::numeric_limits<nano::uint128_t>::max ()) ? end : node.store.pending_begin (transaction, nano::pending_key (account.number () + 1, 0));

nano/node/rpc.cpp
SergiySW: Yeah, pending keys ordered by accounts first

guilhermelawless:
> 
> 
> @SergiySW i'm getting this weird result on beta, the unopened total is larger than before (notice 34027... vs 34028...).
> 
> ```
> Supply (available_supply): 340272367920938463463374607431768151531
> 
> Total balance with pending (ledger): 232484364820933402871612082323207277732
> 
> Total unopened (unopened): 107798002100005060591762525108560933723
> 
> Unopened + balance + pending (unopened + ledger): 340282366920938463463374607431768211455
> ```
> 
> Was matching exactly with the previous approach.

@SergiySW 
The mismatching account is the burn: xrb_1111111111111111111111111111111111111111111111111111hifc8npp

In the first method i checked and skipped hash zero so it didn't show up. Would you like to skip it for this RPC or not?

nano/node/rpc.cpp
SergiySW: uint256_t::max () of course..

SergiySW:
This is correct result for total theoretical supply, but incorrect for alvailable_supply ( - burned amount)
Yes, probably better exclude burning account with `auto iterator (node.store.pending_begin (transaction, nano::pending_key (1, 0)));`

nano/node/rpc.cpp
guilhermelawless: Weirdly not working due to a deleted function.

```
‘nano::store_iterator<nano::pending_key, nano::pending_info>::store_iterator(const nano::store_iterator<nano::pending_key, nano::pending_info>&)’ is implicitly declared as deleted because ‘nano::store_iterator<nano::pending_key, nano::pending_info>’ declares a move constructor or move assignment operator 
```

guilhermelawless: I can just do normal if

SergiySW: Yeah, I didn't test this draft. Nornal if or break of while is fine

guilhermelawless: Problem was not being able to copy the end, needed to get another pending_end ()

guilhermelawless:
Ready

SergiySW:
I guess few things remain we may want.
Similar to "ledger" & other RPC we return list of results with some key i.e.
```
{
    "accounts": {
       ....
   }
}
```
And because it's one of the slowest RPC, we may want to limit it with rpc control enabled check

nano/node/rpc.cpp
```diff
@@ -3314,6 +3314,43 @@ void nano::rpc_handler::unchecked_keys ()
...
+			{
+				response_l.put (current_account.to_account (), current_account_sum.convert_to<std::string> ());
+				current_account = account;
+				current_account_sum = 0;
+			}
+			current_account_sum += info.amount.number ();
+			++iterator;
+		}
+	}
+	response_l.put (current_account.to_account (), current_account_sum.convert_to<std::string> ()); // last one
```
SergiySW: What if there are no unopened accounts? Could we make test for such case?

guilhermelawless: Makes sense, see now

nano/node/rpc.cpp
```diff
@@ -3314,6 +3314,53 @@ void nano::rpc_handler::unchecked_keys ()
 	response_errors ();
 }
 
+void nano::rpc_handler::unopened ()
+{
+	rpc_control_impl ();
```
SergiySW: All lines between rpc_control_impl () & response_errors () in `if  (!ec) { ... }`
Other than that, LGTM

guilhermelawless:
Can do, will take a few days. As a git patch or simply the markdown code?

SergiySW:
I guess it could be useful to allow specifying starting "account" (default 1) & max "count" similar to ledger

guilhermelawless:
Should be done, wiki entry in the initial comment.

guilhermelawless:
It's not that slow anymore with the new approach, so you could consider to leave it without control required.

SergiySW:
How fast is it on live network?
I guess even if it's not slow now with ledger growth (with millions of pending & existing accounts ?) it will become slower & slower

guilhermelawless:
You're right. 5 seconds with 27k results, linear scaling means it will indeed become slower.

PR #1728, opened 2019-02-13
Message sink api

clemahieu:


SergiySW:
lmdb changes seems not required

PR #1729, opened 2019-02-15
Network selector

### LINKED ISSUE #1099: Easy way to switch between beta/live network
In Bitcoin allow the user to switch between live/testnet change the configuration of `bitcoin.conf`, setting `testnet=1`. It's easier and simple. This behavior are present in many others cryptocurrencies.

Why not create a new field in [`config.json`](https://github.com/nanocurrency/raiblocks/wiki/config.json):

```
{
   "node": {
          ....
          "network_mode": "C", // Switch between network modes. A = Test, B = Beta, C = Live
          ....
   }
}
```

It's possible?

cryptocode:
Closes #1099

A new CLI option is added:

`--network live|beta|test`

Example of using a livenet-compiled wallet to run on betanet:

`./nano_wallet --network beta --data_path MyBetaDirectory`

This is a CLI option instead of a config option because the config-file may not exist yet.

The ACTIVE_NETWORK build flag is still there to define the default value if no network option is present. As a result, RC builds for beta don't require a network option, so startup scripts, etc should work as before.

As discussed in #1099, some other projects use flags such as -testnet, but those mean sightly different things. A --network flag makes it clear that we don't follow any other projects' network naming standard and it mirrors how we choose network at compile-time (this doesn't seem like an option regular users will employ)

Implementation details:
* The main function of tests forces testnet. Hence, tests can be run even when compiled for live- or betanet.
* Any option whose value depends on network is now placed in config.hpp in the appropriate class.
* Ledger constants are moved to config.hpp as well. Reason is these constants depend on the selected network, and initialization-order of globals in different translation units is unspecified.
* The node and a few other classes have a `nano::network_params` member where the per-network values can be queried.
* For freestanding functions, a static local is used for the params. A singleton could have been used, but we may want to inject params everywhere in the future (and perhaps remove freestanding functions) and a singleton would make that harder.
* For network constants used as default function arguments (bound at compile-time), an overload without the argument is created which forwards to the original. One such example is `work_generate_blocking (nano::uint256_union const &, uint64_t = nano::work_pool::publish_threshold);`
* Globals such as `nano::genesis_amount` used to depend on the compile-time network value. These are now only available for tests, which enforces the testnet. Since only tests include testutil.hpp, any attempts to use these in the node results in a compile error. As a result, this PR contains almost no changes to tests (initially there was a network_params in `system`, etc, but that meant a huge amount of changes with no gain)

Tests pass and it appears to sync fine on both networks.

### DIFF
--- a/gtest/src/gtest_main.cc
+++ b/gtest/src/gtest_main.cc
@@ -29,10 +29,12 @@
 
 #include <stdio.h>
 
-#include "gtest/gtest.h"
+extern void force_nano_test_network ();
 
+#include "gtest/gtest.h"
 GTEST_API_ int main(int argc, char **argv) {
   printf("Running main() from gtest_main.cc\n");
+  force_nano_test_network ();
   testing::InitGoogleTest(&argc, argv);
   return RUN_ALL_TESTS();
 }
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -1,4 +1,5 @@
 #include <gtest/gtest.h>
+#include <nano/core_test/testutil.hpp>
 #include <nano/lib/utility.hpp>
 #include <nano/node/common.hpp>
 #include <nano/node/node.hpp>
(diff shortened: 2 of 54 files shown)


guilhermelawless:
Does the data path change to /NanoBeta by default if a live net node is compiled and only --network beta is given?

What are the consequences if an already complete beta database is provided by accident to a live net node?

cryptocode:
@guilhermelawless If you pick the wrong database, and error is logged and the node exits (that's already the case)

PR #1734, opened 2019-02-16
Ability to limit CPU rate for POW

wezrule:
After a transaction some CPUs works overtime trying to pre-cache the next POW. It may not be necessary for them to pre-cache as soon as possible, for example if transactions are made infrequently. This add the ability to do a thread sleep in the work threads using an amount of nanoseconds specified in the new config option `pow_sleep_interval` (default is 0). Add the same string as a CLI sub-option to `--debug_profile_generate` to be able to test the timings there.


guilhermelawless:
Maybe a debug CLI to PoW with CPU and make it easy to check if the current option is appropriate? Sorry if it already exists.

wezrule:
@guilhermelawless your comment seems to have been removed during the merges 🐙. There is an existing CLI option `--debug_profile_generate` which should time how long work takes.

guilhermelawless:
Huh, strange. Thank you that will be useful, and this PR especially is great for Distributed PoW clients. I hope you don't mind I take the idea and implement in that one and nano-work-server

wezrule:
Although looking at `--debug_profile_generate` it doesn't take into account the config file, so I will add a sub-option to the CLI to specify the sleep amount so that users can test it to find a suitable value. Thanks for point that out!

nano/lib/work.cpp
```diff
@@ -99,6 +101,12 @@ void nano::work_pool::loop (uint64_t thread)
 					blake2b_final (&hash, reinterpret_cast<uint8_t *> (&output), sizeof (output));
 					blake2b_init (&hash, sizeof (output));
 					iteration -= 1;
+
```
clemahieu: I think this should be moved up one loop so it's not doing a branch check in the innermost loop.

wezrule: thanks, have moved it to the outside loop.

zhyatt:
@wezrule Does the documentation need to be updated for the CLI pow_sleep_interval sub-option? https://docs.nano.org/commands/command-line-interface/#-debug_profile_generate

wezrule:
@zhyatt done. https://github.com/nanocurrency/nano-docs/pull/46

PR #1736, opened 2019-02-16
 block_store.sideband_height test

SergiySW:


wezrule:
Can more state blocks be checked (in particular state open)? to make sure they are correct

SergiySW:
ledger+processor part replaced with https://github.com/nanocurrency/nano-node/pull/1737

SergiySW:
@wezrule updated with state open, state epoch & satte open epoch

PR #1737, opened 2019-02-17
Make the sideband open block start at a height of 1

wezrule:
Account block count + sideband block height now match at the head
Sideband/confirmation height now start at 0 to indicate that they are uninitialized

### DIFF
--- a/nano/node/lmdb.cpp
+++ b/nano/node/lmdb.cpp
@@ -1226,7 +1226,7 @@ void nano::mdb_store::upgrade_v12_to_v13 (size_t const batch_size)
 		if (!first.is_zero ())
 		{
 			auto hash (second.open_block);
-			uint64_t height (0);
+			uint64_t height (1);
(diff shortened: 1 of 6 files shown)


nano/node/lmdb.cpp
```diff
@@ -1243,9 +1243,9 @@ void nano::mdb_store::upgrade_v12_to_v13 (size_t const batch_size)
 				}
 				auto block (block_get (transaction, hash, &sideband));
 				assert (block != nullptr);
-				if (sideband.height == std::numeric_limits<uint64_t>::max ())
+				if (sideband.height == 0)
 				{
-					sideband.height = height;
+					sideband.height = height + 1;
```
SergiySW: Probably just make start height = 1?
`uint64_t height (1);`

wezrule: Sure, if you think that is clearer

PR #1738, opened 2019-02-17
Initialize peer_information members

cryptocode:
rep_weight / probable_rep_account had random values.


wezrule:
Are you sure `rep_weight` has a garbage value? It should be initialized here:
https://github.com/nanocurrency/nano-node/blob/0031cd004e16cb9cb3da2c41bb014eda188a17ea/nano/node/peers.hpp#L56

I used in-class member initializers here as there are multiple constructors so saves some duplication in the source file (as well as reducing line count).

cryptocode:
I'll move probable_rep_account to in-class since rep_weight is done there (which I assumed had the same issue, didn't catch that some initialization was done in ctor and some in-class now...)

PR #1740, opened 2019-02-18
RPC: option to render block as JSON

cryptocode:
A new option `json_block` (bool) is added to `create_block` and a few other RPCs stringifying blocks. 

Default is "false" (the old behavior)

If set to "true", the returned block will be proper JSON instead of being stringified with embedded escapes of newlines and quotes. This makes it easier to move offline-signed blocks for processing via the Qt wallet or websites with process functionality.

(A future PR could make `process` support non-stringified blocks.)

### DIFF
--- a/nano/core_test/rpc.cpp
+++ b/nano/core_test/rpc.cpp
@@ -3397,6 +3397,36 @@ TEST (rpc, account_info)
 	ASSERT_EQ (nano::test_genesis_key.pub.to_account (), representative2);
 }
 
+TEST (rpc, json_block)
+{
+	nano::system system (24000, 1);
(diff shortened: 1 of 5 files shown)


SergiySW:
Could we add same option to other RPCs returning block content?

cryptocode:
@SergiySW that sounds like a good idea

SergiySW:
Probably add json check to some tests?

zhyatt:
@cryptocode Would including this option for https://docs.nano.org/commands/rpc-protocol/#unchecked be needed?

cryptocode:
That option is not supported for the unchecked RPC, only unchecked_get.

PR #1742, opened 2019-02-18
Respond to keepalive messages within a TCP socket

clemahieu:


nano/node/bootstrap.cpp
```diff
@@ -2178,6 +2186,28 @@ void nano::bootstrap_server::receive_bulk_pull_account_action (boost::system::er
...
+void nano::bootstrap_server::receive_keepalive_action (boost::system::error_code const & ec, size_t size_a, nano::message_header const & header_a)
+{
+	if (!ec)
+	{
+		auto error (false);
+		nano::bufferstream stream (receive_buffer->data (), header_a.payload_length_bytes ());
+		std::unique_ptr<nano::keepalive> request (new nano::keepalive (error, stream, header_a));
+		if (!error)
+		{
+			add_request (std::unique_ptr<nano::message> (request.release ()));
```
wezrule: can't this just be: `add_request (std::move (request));` ?

nano/core_test/network.cpp
```diff
@@ -1392,3 +1392,31 @@ TEST (bulk_pull_account, basics)
...
+	socket->async_connect (system.nodes[0]->bootstrap.endpoint (), [&input, socket](boost::system::error_code const & ec) {
+		ASSERT_FALSE (ec);
+		socket->async_write (input, [&input](boost::system::error_code const & ec, size_t size_a) {
+			ASSERT_FALSE (ec);
+			ASSERT_EQ (input->size (), size_a);
+		});
+	});
+
+	auto output (keepalive.to_bytes ());
+	bool done (false);
```
SergiySW: Probably should be atomic ?

nano/node/bootstrap.cpp
clemahieu: Yes, a lot of these need to be fixed to use c++13 semantics.

nano/node/bootstrap.cpp
```diff
@@ -2235,9 +2265,35 @@ class request_response_visitor : public nano::message_visitor
...
+		connection->node->stats.inc (nano::stat::type::message, nano::stat::detail::keepalive, nano::stat::dir::in);
+		connection->node->network.merge_peers (message_a.peers);
+		nano::keepalive message;
+		connection->node->peers.random_fill (message.peers);
+		auto bytes = message.to_bytes ();
+		if (connection->node->config.logging.network_keepalive_logging ())
+		{
+			BOOST_LOG (connection->node->log) << boost::str (boost::format ("Keepalive req sent to %1%") % connection->socket->remote_endpoint ());
+		}
+		connection->socket->async_write (bytes, [connection = connection](boost::system::error_code const & ec, size_t size_a) {
```
SergiySW: connection = connection (?)

wezrule: connection is a member variable, so if you want to pass it in without passing `this` this is the way to do it (and [&var = var] for reference)

PR #1744, opened 2019-02-18
Cli debug online weight

argakiig:
add cli debugging for online_weights table to dump and clear the table


nano/nano_node/entry.cpp
```diff
@@ -165,6 +167,25 @@ int main (int argc, char * const * argv)
...
+		{
+			nano::inactive_node node (data_path);
+			auto current (node.node->online_reps.online_stake ());
+			std::cout << boost::str (boost::format ("Online Weight %1%\n") % current);
+			auto transaction (node.node->store.tx_begin_read ());
+			for (auto i (node.node->store.online_weight_begin (transaction)), n (node.node->store.online_weight_end ()); i != n; ++i)
+			{
+				std::string weight;
+				i->second.encode_dec (weight);
+				std::cout << boost::str (boost::format ("Date %1% Weight %2%\n") % i->first % weight);
```
SergiySW: May be timestamp instead of date?

nano/nano_node/entry.cpp
```diff
@@ -34,6 +34,8 @@ int main (int argc, char * const * argv)
 		("batch_size",boost::program_options::value<std::size_t> (), "Increase sideband batch size, default 512")
 		("debug_block_count", "Display the number of block")
 		("debug_bootstrap_generate", "Generate bootstrap sequence of blocks")
+		("debug_clear_online_weight", "clear online_weights table")
```
SergiySW: We have cli --online_weight_clear in cli.cpp

argakiig: k will drop that we can use it instead, i must have missed it

PR #1748, opened 2019-02-19
RPC pending - sort by amount

guilhermelawless:
Per request of @bbedward . This PR serves to add a `sorting` option to RPC "pending". It helps services which will now only need to get the first X blocks ordered by amount, to prioritize receiving the most important blocks instead of dust.

I decided to use `ptree::sort()` due to how cleaner it makes the code vs. having two branches as in `ledger ()`.

===== Wiki entry update =====

### Optional "sorting"

Boolean, false by default. Additionally sorts the blocks by their amounts in descending order.

### DIFF
--- a/nano/core_test/rpc.cpp
+++ b/nano/core_test/rpc.cpp
@@ -1752,74 +1752,98 @@ TEST (rpc, pending)
 	request.put ("action", "pending");
 	request.put ("account", key1.pub.to_account ());
 	request.put ("count", "100");
-	test_response response (request, rpc, system.io_ctx);
-	system.deadline_set (5s);
-	while (response.status == 0)
 	{
-		ASSERT_NO_ERROR (system.poll ());
+		test_response response (request, rpc, system.io_ctx);
+		system.deadline_set (5s);
+		while (response.status == 0)
+		{
+			ASSERT_NO_ERROR (system.poll ());
+		}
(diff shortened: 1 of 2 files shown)


guilhermelawless:
To note: can still add the "sorting" option without breaking the RPC, simply by maintaining the same behavior when "threshold" is not given.

As such, this PR could be split in 2: one that adds the sorting option, and one that breaks the RPC to increase consistency. Let me know if this would be the preferred way to go about it.

SergiySW:
Not sure if we want to break existing rpc. It may be confusing, but all libraries rework... Also accounts_pending & wallet pending will return only hashes, even more confusing
Probably we can add "amount" bool field

guilhermelawless:
Makes sense @SergiySW , i think i achieved something more sensible now.

- The simple case is maintained to not break RPC.
- If any additional argument is given ( threshold, min_version, source or sorting (**new**) ) then the amounts are returned, like before.
- If only "sorting" or "threshold" is given, then the response is "hash": "amount"
- In the specific case of min_version or source given, the complex version is returned (with different fields)

In summary, RPC not broken and "sorting" added.

guilhermelawless:
Done. Btw, the wiki is missing the optional "min_version"

guilhermelawless:
Label should change to rpc non breaking.

PR #1754, opened 2019-02-19
Make confirm_back () non-recursive

SergiySW:


cryptocode:
confirm_if_quorum is still using depth instead of bool

PR #1757, opened 2019-02-21
check for legacy directory

argakiig:
move data_path for docker to prevent resync of ledger
add symlink for legacy support to docker


SergiySW:
Hm, but commit is "create ci image for clang"

argakiig:
I had pushed up the wrong branch somehow, corrected to appropriate commit

PR #1758, opened 2019-02-21
Use a single synchronised random pool across all threads

wezrule:
thread_local version seems to cause TSAN error with `s_TeFilled` variable inside CryptoPP

### DIFF
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -666,7 +666,10 @@ TEST (block_store, large_iteration)
 	{
 		auto transaction (store.tx_begin (true));
 		nano::account account;
-		nano::random_pool.GenerateBlock (account.bytes.data (), account.bytes.size ());
+		{
+			std::lock_guard<std::mutex> lk (nano::random_pool_mutex);
+			nano::random_pool.GenerateBlock (account.bytes.data (), account.bytes.size ());
+		}
 		accounts1.insert (account);
 		store.account_put (transaction, account, nano::account_info ());
 	}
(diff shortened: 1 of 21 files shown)


cryptocode:
Any areas we need to profile with this change, as only one thread can generate random data at a time?

wezrule:
store.load creates loads of threads and uses the random_pool but I'm not sure it accurately reflects what will happen on the network. I don't think it should affect it much but I haven't run any benchmarks. However I'm not sure what an appropriate alternative is currently, so not sure we have a choice? @SergiySW wants to remove the cryptopp library entirely but that will probably have to happen at a later date.

nano/core_test/block_store.cpp
```diff
@@ -666,7 +666,10 @@ TEST (block_store, large_iteration)
 	{
 		auto transaction (store.tx_begin (true));
 		nano::account account;
-		nano::random_pool.GenerateBlock (account.bytes.data (), account.bytes.size ());
+		{
```
clemahieu: Can we encapsulate this in a function so if we want to change it against we don’t have to change all call sites?

wezrule: Good idea to wrap these to prevent forgetting to lock as well. I wanted to make the mutex/cryptopp variables inaccessible so made them local to the source file which worked until I realised shuffle () was a template function, so I've made a static class instead where those members are private.

wezrule:
Encapsulated the random_pool functions
Moved the local static variables from `not_an_account ()` to be a member variables as cryptocode is changing it so that multiple ledger_constants can exist.

PR #1765, opened 2019-02-22
Improve block_exists table lookup order

cryptocode:
Same as was done for block_raw_get, i.e order table lookups according to match probability.


nano/node/lmdb.cpp
```diff
@@ -1709,13 +1709,14 @@ bool nano::mdb_store::block_exists (nano::transaction const & transaction_a, nan
 
 bool nano::mdb_store::block_exists (nano::transaction const & tx_a, nano::block_hash const & hash_a)
 {
+	// Table lookups are ordered by match probability
```
wezrule: should it be "**to** match"?

PR #1770, opened 2019-02-23
Add confirmation height

wezrule:
These are summarized notes given before the start of this work:
> When we confirm a block what we want to set the account's confirmation height to be equal to the height of the block just confirmed. So if the confirmation height was 2 and the block height is 7, we set the confirmation height to 7. For each block that was implicitly confirmed; 7-3, we check if it’s a receive block. If it’s a receive block we update the confirmation height of the sender in the same fashion. The algorithm needs to be iterative, not recursive since this could be a deep confirmation process. When we update the database, all confirmation heights start as zero and when the first confirmed block comes off the live network it’s going to confirm a ton of accounts.

The sideband upgrade has been made synchronous (approved by @clemahieu) to make the process simpler for future upgrades. This means nodes will be very slow to start if going from v17.1 -> v19 however the big services/exchanges should have been made aware of this during the v18 release announcement so it should not affect many nodes.

New RPC `block_confirmed` has been added to check if a block has been confirmed and the `account_info` has been modified to output the confirmation height for the height as well. MSVC only compiles up to 128 if/elseif blocks (well 128 nested blocks) which was reached after I added 2 RPC calls. I have combined the password/payment RPC calls into another one temporarily until a better solution can be found. A possible solution is to use a std::map<std::string, std::function<void(rpc_handler*)>> which maps action => handler for functions with no arguments.

--snapshot now accepts an optional `confirmation_height_clear` option for clearing the confirmation heights (setting to 0) for all accounts.

Updated nano/core_test/versioning.cpp to have tests for all account_info_v* classes

lmdb.cpp - nano::mdb_store::account_get() didn't check the error condition from info_a.deserialize (stream); and same in pending_get(), so I have added those.

There are still some questions around pending blocks, but this can always be done in a separate PR.


nano/node/node.cpp
```diff
@@ -2668,7 +2668,7 @@ void nano::node::process_confirmed (std::shared_ptr<nano::block> block_a, uint8_
 	auto hash (block_a->hash ());
 	if (ledger.block_exists (block_a->type (), hash))
 	{
-		auto transaction (store.tx_begin_read ());
+		auto transaction (store.tx_begin_write ());
```
cryptocode: Would it be possible to to scope everything until add_confirmation_heights in a read tx (as before), and then only do add_confirmation_heights in a write transaction? We want to keep write tx's open as short as possible. confirmed_visitor can take some time if there are many wallets, etc.

nano/node/rpc.cpp
```diff
@@ -391,6 +391,25 @@ void nano::rpc_handler::account_block_count ()
 	response_errors ();
 }
 
+void nano::rpc_handler::account_confirmation_height ()
```
SergiySW: Probably it can be part of account_info instead?

nano/node/node.cpp
```diff
@@ -3748,6 +3750,89 @@ std::unique_ptr<seq_con_info_component> collect_seq_con_info (active_transaction
...
+	{
+		if (!open_receive_blocks.empty ())
+		{
+			current = open_receive_blocks.top ();
+			open_receive_blocks.pop ();
+
+			auto block (store.block_get (transaction, current));
+			if (block != nullptr)
+			{
+				nano::block_hash source_hash = block->source ();
```
SergiySW: Probably it have to be a bit more complex. We don't have to check block existence if source_hash .is_zero ()
For state blocks need to check link () instead if link is not 0, not epoch link & that current state block !is_send ()
Instead of check block_get + (source_block != nullptr) can use source_exists (...) function

nano/node/node.cpp
```diff
@@ -3748,6 +3750,89 @@ std::unique_ptr<seq_con_info_component> collect_seq_con_info (active_transaction
...
+			// Start from the most recent one and work our way through
+			for (uint64_t i = 0; i < newly_confirmed_num_blocks; ++i)
+			{
+				nano::block_sideband sideband;
+				auto block (store.block_get (transaction, current, &sideband));
+				if (block != nullptr)
+				{
+					// First check legacy receive/open
+					auto should_process = (sideband.type == nano::block_type::receive || (sideband.type == nano::block_type::open && !current.is_zero ()));
+					if (!should_process)
```
SergiySW: Probably check block type == state before cast?

SergiySW: Hm, why check sideband type instead of block type?

nano/node/node.cpp
SergiySW: Update: considering balance check for state blocks below expensive !is_send () not required

Actually why not merge this 2 parts? `if (!open_receive_blocks.empty ()) { ... }` & `// First check legacy receive/open`
I guess it can save several iops. If put to stack source () for receive/open blocks & link () for state blocks after balance check `if (state->hashables.balance > ledger.balance (transaction, previous))`

nano/node/node.cpp
wezrule: Good point, done both

nano/node/rpc.cpp
wezrule: I did know about that :) yes makes sense to go there, thanks!

nano/node/node.cpp
wezrule: The `dynamic_pointer_cast`? It does a `dynamic_cast` under the hood and checks for null, so it should be ok to just check for `nullptr` after the cast. I have made it more explicit now though.

Good point, done and that now saves getting the `sideband` object now.

nano/node/node.cpp
wezrule: Think I understand, please check to confirm though, thanks :)

SergiySW: Seems great!

nano/core_test/network.cpp
```diff
@@ -1423,3 +1423,165 @@ TEST (bootstrap, keepalive)
...
+		{
+			auto transaction = node->store.tx_begin_write ();
+			ASSERT_FALSE (node->ledger.rollback (transaction, node->latest (key2.pub)));
+			ASSERT_FALSE (node->ledger.rollback (transaction, node->latest (key1.pub)));
+		}
+		{
+			// These rollbacks should fail
+			auto transaction = node->store.tx_begin_write ();
+			ASSERT_TRUE (node->ledger.rollback (transaction, node->latest (key2.pub)));
+			ASSERT_TRUE (node->ledger.rollback (transaction, node->latest (key1.pub)));
```
SergiySW: Could we add check for latest hashes probably? send4 & send6 I guess

nano/node/cli.cpp
```diff
@@ -306,6 +313,45 @@ std::error_code nano::handle_node_options (boost::program_options::variables_map
...
+			nano::account account;
+			if (!account.decode_account (account_str))
+			{
+				nano::account_info account_info;
+				auto transaction (node.node->store.tx_begin_read ());
+				if (!node.node->store.account_get (transaction, account, account_info))
+				{
+					auto transaction (node.node->store.tx_begin_write ());
+					node.node->store.confirmation_height_clear (transaction, account, account_info);
+					std::cerr << "Confirmation height of account " << account_str << " is set to 0" << std::endl;
```
SergiySW: std::cout ?

wezrule: I copied the other `*_clear` ones which used `std::cerr` as well, but I agree it doesn't seem correct, so I'm going to change those as well.

nano/core_test/network.cpp
wezrule: Good idea, I'm testing rollbacks on the latest blocks of the other accounts and a few other cemented blocks.

nano/core_test/block_store.cpp
```diff
@@ -1587,3 +1517,149 @@ TEST (block_store, online_weight)
...
+		modify_account_info_to_v13 (store, transaction, nano::genesis_account);
+
+		// Add many accounts
+		for (auto i = 0; i < total_num_accounts - 1; ++i)
+		{
+			nano::account account (i);
+			nano::open_block open (1, 2, 3, nullptr);
+			nano::account_info_v13 account_info_v13 (open.hash (), open.hash (), open.hash (), 3, 4, 1, nano::epoch::epoch_1);
+			auto status (mdb_put (store.env.tx (transaction), store.accounts_v1, nano::mdb_val (account), nano::mdb_val (account_info_v13), 0));
+			release_assert (status == 0);
```
clemahieu: In a test this can be an ASSERT_EQ

wezrule: yep, good spot

nano/node/node.cpp
```diff
@@ -3737,6 +3742,86 @@ std::unique_ptr<seq_con_info_component> collect_seq_con_info (active_transaction
 	return composite;
 }
 }
+
+/**
+ * For all the blocks below this height which have been implicitly confirmed check if they
+ * are open/receive blocks, and if so follow the source blocks and iteratively repeat to genesis.
+ */
+void nano::node::add_confirmation_heights (nano::transaction const & transaction, nano::block_hash const & hash)
```
cryptocode: When starting on a synced v18 ledger, this function appears to take a long time for some hashes right after startup. I noticed this because RPC was stuck. This call is in a write tx and all other IO threads ended up waiting for it to complete (specifically  `online_reps::sample`, `ongoing_store_flush`, `ongoing_peer_store` and thus RPC is blocked as well with 4 IO threads)

Example hash: F0394C8F7521DF8586BECAF2DA4868DA18E9D2DF702EAA89AFCABDB8AFF07EEB
This took ~30 minutes to complete, all other IO threads are eventually stuck on the write tx mutex.

Since `upgrade_v13_to_v14` sets height to 0, is the idea that add_confirmation_heights is part of a lazy upgrade, which may take a while for long chains the first time they are encountered?

One solution might be to process `open_receive_blocks` in time slices, like we do in `process_batch` - not sure if that fits with the stack processing logic.

wezrule: The initial ones being confirmed are expected to take a while. The issue we have is that for a confirmation height of an account to be accurate, all the receive blocks of the account also need to be confirmed, which if it hits a long account chain will take a very long time. I'm not sure if this can be broken up as any state in-between ends up invalidating it all completely. Perhaps there is a cleverer way this can be achieved, but how much of a problem is it for the initial upgrades? RPCs were down for people doing the sideband upgrades (days for some).

cryptocode: >  but how much of a problem is it for the initial upgrades?

What confused me was that the node was operational for several minutes, until RPC stopped working.

It might be fine, I just wasn't aware of it. It may be a matter of documenting this effect on upgrade + clearing confirmation height.

wezrule: Ah right, I think I'm using term upgrade incorrectly here, as setting the confirmation height is really a continuous update. So the actual upgrade itself (v13_v14 adding 0 confirmation height) shouldn't take too long, at that point RPCs etc should work fine. It is only when blocks are confirmed off the network that the confirmation height is "updated", with the conditions mentioned earlier. I have added the documentation label to make sure this is not forgotten, thanks.

nano/node/node.cpp
```diff
@@ -2669,7 +2669,12 @@ void nano::node::process_confirmed (std::shared_ptr<nano::block> block_a, uint8_
 	auto hash (block_a->hash ());
 	if (ledger.block_exists (block_a->type (), hash))
 	{
-		auto transaction (store.tx_begin_read ());
+		{
+			auto transaction (store.tx_begin_write ());
+			add_confirmation_heights (transaction, hash);
+		}
+
+		auto transaction (store.tx_begin_write ());
```
cryptocode: I think this should be tx_begin_read  (like it was before)

wezrule: ahh good point, I must have changed that by accident again at some point, it's not necessary to write. Thanks

wezrule:
Stein has ran it for 12 hours and found most blocks spend < 1ms inside the iterative function add_confirmation_heights with a few outliers when there are long chains but they don't lock long enough to cause rpc lockups. Merging this in now so that any problems can be found sooner. Zach has some further questions about functionality but these can be investigated separately as this is a building block for further enhancements.

Issue #793, opened 2018-04-11
RPC/CLI inconsistency with wallet_change_seed


cgerikj:
The parameters for this command is different for RPC and CLI.

For RPC it's wallet & seed (correct):
https://github.com/nanocurrency/raiblocks/wiki/RPC-protocol#wallet-change-seed

For CLI it's wallet & key ("key" incorrect?):
https://github.com/nanocurrency/raiblocks/wiki/Command-line-interface#--wallet_change_seed---walletwallet---keykey

SergiySW:
Resolved with https://github.com/nanocurrency/nano-node/pull/1772

PR #1772, opened 2019-02-24
Add "seed" option to CLI

SergiySW:
Supported in wallet_change_seed & wallet_create
Following https://github.com/nanocurrency/nano-node/issues/793


zhyatt:
@SergiySW Was the optional "seed" available for wallet_create in V18 before? Or is it new for V19? See docs here indicating it was in V18: https://docs.nano.org/commands/rpc-protocol/#wallet_create . Also does response when seed is included in wallet_create return "last restored account from given seed & restored count" as it indicates?

zhyatt:
I was getting RPC and CLI crossed up. Updating documentation to reflect change to seed option from key.

Issue #1723, opened 2019-02-12
Deprecate `payments_` RPC calls


zhyatt:
Deprecate the following RPC calls ahead of removal in a future version:

- payment_begin
- payment_init
- payment_end
- payment_wait

Removing because they are not commonly used and other RPC calls provide better options for managing.

SergiySW:
Deprecated with https://github.com/nanocurrency/nano-node/pull/1773 & wiki updates. For v20 plan is to disable deprecated calls without config.json modification

PR #1773, opened 2019-02-24
Deprecate `payments_` RPC calls

SergiySW:
https://github.com/nanocurrency/nano-node/issues/1723

### DIFF
--- a/nano/node/rpc.cpp
+++ b/nano/node/rpc.cpp
@@ -2356,6 +2356,7 @@ void nano::rpc_handler::payment_begin ()
 				} while (account.is_zero ());
 				if (!account.is_zero ())
 				{
+					response_l.put ("deprecated", "1");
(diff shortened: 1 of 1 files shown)


cryptocode:
An alternative I was thinking about is to add deprecation in Wiki docs only and then in version + 1 require a flag ("allow_deprecated": "true") to allow the deprecated action (the response would be something like "error": "deprecated"). In version + 2 or 3, the action is removed. The flag could be either an RPC param or a config option.

This would give users ample warning and still opt-in for a version or two, and then there's a way to actually remove actions and be confident that users know it'll happen.

SergiySW:
I guess we can add it now in rpc response & wiki, in v20 require ("allow_deprecated": "true") & remove ~ v22

cryptocode:
@SergiySW sounds like a good plan

Issue #809, opened 2018-04-16
Impossible to distinct between a send/receive/change using the `blocks` in RPC


Inkeliz:
Using the state-block can be impossible to distinct between send/receive with the `blocks` or `blocks_info`.

---------------------

If someone creates a `receive` (legacy), then use the `state-block` (for send), is impossible to know if it is a send or receive, because the previous block (receive, non-state) don't includes the `balance`. 

Usually, it's possible to check the `send`/`receive` comparing against the `balance` from the previous block, but in some cases don't exist a `balance`. 

Also, is impossible to know if it's a `change block`. If someone uses a legacy block and do a `send and change` (both operations in one block) is impossible to know the change, the RPC doesn't include that information too, the previous block might not include it.

-----------------------

I think the `blocks_info` should include the `balance` and `representative` of each block. Alternatively, ignoring performance aspect, can include a `SubType` array, like `SubType: ["receive"]` or `SubType: ["receive", "change"]`.

But it isn’t my preferred method… If we go futher we can do the opposite. If have a new optional parameter like `force_state_block` (in complement with "pending", "source" already existent). If this parameter is used it will retrieve the legacy blocks as a state-block, fixing all issues. So, the “send block” (legacy one) will be a “state-block” with all correct information (the `Balance`, the `Representive` used up this block, the `Account`, and so on).

In this case the `SubType` is dropped, since all types can be identified, and all types are state-block. For signature check we need to know if it’s a really state or a _converted_ one. In my mind, creating a `is_legacy` as a Boolean response is enough. It should help the usage of the RPC, since all blocks can be treated as a state-block, if someone need to check the signature they can hash the legacy equivalent of the content and compare it.

renesq:
`blocks_info` actually has an option to display the balance, just not the representative. 

Other alternatives to distinguish between sends and receives: 
- try to retrieve the block referenced in the "`link`" field. If it works, it's a send or send type block. 
- use `account_history` instead, which shows the emulted subtype for legacy reasons (But it only shows one subtype at a time, i think. The command also has a "`raw`" option to show the real blocks. )
- The main reason why e.g. callback doesn't list subtypes but has a "`is_send`" flag instead is because **representative changes are rather irrelevant for most app development**. The is_send flag also is rather precise, since there oould be two subtypes at the same time, as you already noticed. 
[...]

PR #1774, opened 2019-02-24
Add subtype field for state blocks in RPC block_info, blocks_info

SergiySW:
https://github.com/nanocurrency/nano-node/issues/809

### DIFF
--- a/nano/node/rpc.cpp
+++ b/nano/node/rpc.cpp
@@ -834,7 +834,7 @@ void nano::rpc_handler::block_info ()
 				block->serialize_json (contents);
 				response_l.put ("contents", contents);
 			}
-			
+
 			// Subtype check
 			if (block->type () == nano::block_type::state)
 			{
--- a/nano/node/rpc.cpp
+++ b/nano/node/rpc.cpp
@@ -802,6 +802,31 @@ void nano::rpc_handler::available_supply ()
 	response_errors ();
 }
 
+void state_subtype (nano::transaction const & transaction_a, nano::node & node_a, std::shared_ptr<nano::block> block_a, nano::uint128_t const & balance_a, boost::property_tree::ptree & tree_a)
+{
+	// Subtype check
+	auto previous_balance (node_a.ledger.balance (transaction_a, block_a->previous ()));
+	if (balance_a < previous_balance)
+	{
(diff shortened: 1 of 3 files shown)


nano/node/rpc.cpp
```diff
@@ -834,6 +834,31 @@ void nano::rpc_handler::block_info ()
...
+				{
+					response_l.put ("subtype", "send");
+				}
+				else
+				{
+					if (block->link ().is_zero ())
+					{
+						response_l.put ("subtype", "change");
+					}
+					else if (balance == previous_balance && !node.ledger.epoch_link.is_zero () && node.ledger.is_epoch_link (block->link ()))
```
cryptocode: We don't seem to be checking if epoch_link is zero anywhere else but subtype checking - is that necessary?

SergiySW: We check it in ledger.cpp processing as well

nano/node/rpc.cpp
```diff
@@ -941,7 +966,30 @@ void nano::rpc_handler::blocks_info ()
 						block->serialize_json (contents);
 						entry.put ("contents", contents);
 					}
-
+					// Subtype check
+					if (block->type () == nano::block_type::state)
+					{
```
wezrule: This block (20+ lines) is identical to that in `block_info ()` can it be refactored into a separate function?

severalif:
I think it would be useful to include this information in the RPC callback as well.

SergiySW:
@severalif callback is updated as well

renesq:
Callback didn't have a subtype field, because a state block can have multiple purposes at the same time, so the "is_send" flag was introduced instead, which is less misleading. I don't think the "subtype" makes any sense here at all, because it will give you no further info that you could get by having the "is_send" flag and looking at the callback content ("amount" and "is_send"). It will just add more traffic.

SergiySW:
@renesq unfortunantely some services not willing to understand many things in Nano. The simpler, the better

renesq:
In that case, revise the current block concept as a whole. Details in #1832

SergiySW:
Yes, several heavy issues exists with state blocks design, but making new types again with a lot of legacy types support

Issue #1749, opened 2019-02-19
Support difficulty for OpenCL PoW


SergiySW:

SergiySW:
Closed with https://github.com/nanocurrency/nano-node/pull/1775

PR #1775, opened 2019-02-24
Support difficulty for OpenCL generate_work ()

SergiySW:
https://github.com/nanocurrency/nano-node/issues/1749


SergiySW:
Test requires OpenCL device on test machine, most servers don't have it

wezrule:
true but we can add a check to see if opencl is supported, and if not then the test logic can be ignored for those systems. Our dev systems can then still run the tests locally

SergiySW:
@wezrule test updated, if you have opencl system

PR #1789, opened 2019-02-26
Time limited logging

wezrule:
The constraint on this task was "Arguments should be forwarded so there's no performance penalty for calling this function in the event the logging entry is skipped due to timing."

This is achieved using forward references and perfect forwarding.

A new object `nano::logger` has been created with 2 new functions: `try_log` and `always_log`. The constructor for `nano::logger` contains a reference to a time object which denotes the minimum time that successive log calls can be made. So if it is set to 1ms, and a log call is made, another one won't be output (with `try_log`) until 1ms has passed, any made prior are ignored. `always_log` overrides this behaviour and always outputs. This is because there could be matching start/stop entries which can look strange if only 1 is output and there are important log entries we do not want to miss out on. Stuff in testing.cpp. rpc & ipc are also always logged.

There is a new config option `min_time_between_output`, which is defaulted to 5ms (200 log entries per second maximum)..
BOOST_LOG is a convience macro but there isn't a way to use it with variadic templates, so it was substituted following "Or, if you prefer logging without macros:"
https://www.boost.org/doc/libs/1_67_0/libs/log/doc/html/log/detailed/sources.html
I thought about make the new `logging` class template where the template parameter is the boost logger to use, but as we only use logger_mt everywhere I decided against the added complexity.
Travis does not output log files so there's not currently a way to check them. I now temporarily override  the cerr stream in those tests and check that output is correct.

There's a couple ideas I thought of which have not been implemented:
- Having a fixed queue of things to output. This would be useful in situations where there a burst of log messages then a quite period, so the old log entries would still be output but over a longer period of time. This would require copying objects so would not satisfy the performance constraint above.
- Categories of messages. If there are 100 network messages and a rpc log message inbetween, the rpc message might get missed because a network log message was done right before, so the time-space constraint could be made against categories instead rather than globally. This has not been done.

Notes for reviewer: Check that try/always are appropriate where used.

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -478,7 +478,7 @@ nano::process_return nano::block_processor::process_one (nano::transaction const
 		}
 		case nano::process_result::opened_burn_account:
 		{
-			node.logger.try_log (boost::str (boost::format ("*** Rejecting open block for burn account ***: %1%") % hash.to_string ()));
+			node.logger.always_log (boost::str (boost::format ("*** Rejecting open block for burn account ***: %1%") % hash.to_string ()));
 			break;
 		}
 		case nano::process_result::balance_mismatch:
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -268,7 +268,7 @@ void nano::block_processor::process_batch (std::unique_lock<std::mutex> & lock_a
 		if (log_this_record)
 		{
 			first_time = false;
-			node.logger.try_log (boost::str (boost::format ("%1% blocks (+ %2% state blocks) (+ %3% forced) in processing queue") % blocks.size () % state_blocks.size () % forced.size ()));
+			node.logger.always_log (boost::str (boost::format ("%1% blocks (+ %2% state blocks) (+ %3% forced) in processing queue") % blocks.size () % state_blocks.size () % forced.size ()));
 		}
 		nano::unchecked_info info;
 		bool force (false);
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -77,7 +77,7 @@ void nano::block_processor::add (nano::unchecked_info const & info_a)
 	}
 	else
 	{
-		BOOST_LOG (node.log) << "nano::block_processor::add called for hash " << info_a.block->hash ().to_string () << " with invalid work " << nano::to_string_hex (info_a.block->block_work ());
+		node.logger.try_log ("nano::block_processor::add called for hash ", info_a.block->hash ().to_string (), " with invalid work ", nano::to_string_hex (info_a.block->block_work ()));
 		assert (false && "nano::block_processor::add called with invalid work");
 	}
 }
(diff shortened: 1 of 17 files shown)


nano/node/blockprocessor.cpp
```diff
@@ -472,36 +472,36 @@ nano::process_return nano::block_processor::process_one (nano::transaction const
...
 			{
-				BOOST_LOG (node.log) << boost::str (boost::format ("Fork for: %1% root: %2%") % hash.to_string () % info_a.block->root ().to_string ());
+				node.logger.try_log (boost::str (boost::format ("Fork for: %1% root: %2%") % hash.to_string () % info_a.block->root ().to_string ()));
 			}
 			break;
 		}
 		case nano::process_result::opened_burn_account:
 		{
-			BOOST_LOG (node.log) << boost::str (boost::format ("*** Rejecting open block for burn account ***: %1%") % hash.to_string ());
+			node.logger.try_log (boost::str (boost::format ("*** Rejecting open block for burn account ***: %1%") % hash.to_string ()));
```
SergiySW: I guess this should be always_log

wezrule: done

nano/node/node.cpp
```diff
@@ -1554,19 +1555,19 @@ startup_time (std::chrono::steady_clock::now ())
 	});
 	if (NANO_VERSION_PATCH == 0)
 	{
-		BOOST_LOG (log) << "Node starting, version: " << NANO_MAJOR_MINOR_VERSION;
+		logger.try_log ("Node starting, version: ", NANO_MAJOR_MINOR_VERSION);
```
SergiySW: Guess 2 Node starting, Work pool, "Genesis block not found", Node ID also should be logged. "Using bootstrap rep weight" likely too

nano/node/node.cpp
```diff
@@ -1620,7 +1621,7 @@ nano::node::~node ()
 {
 	if (config.logging.node_lifetime_tracing ())
 	{
-		BOOST_LOG (log) << "Destructing node";
+		logger.try_log ("Destructing node");
```
SergiySW: And this probably as well

nano/node/node.cpp
```diff
@@ -1736,7 +1737,7 @@ void nano::node::process_fork (nano::transaction const & transaction_a, std::sha
 				    }
 			    }))
 			{
-				BOOST_LOG (log) << boost::str (boost::format ("Resolving fork between our block: %1% and block %2% both with root %3%") % ledger_block->hash ().to_string () % block_a->hash ().to_string () % block_a->root ().to_string ());
+				logger.try_log (boost::str (boost::format ("Resolving fork between our block: %1% and block %2% both with root %3%") % ledger_block->hash ().to_string () % block_a->hash ().to_string () % block_a->root ().to_string ()));
```
SergiySW: Probably this too, especially after confirmation height (forks for confirmed blocks can be ignored I guess or just increase some stats)

nano/node/node.cpp
```diff
@@ -1953,7 +1954,7 @@ void nano::node::start ()
 
 void nano::node::stop ()
 {
-	BOOST_LOG (log) << "Node stopping";
+	logger.try_log ("Node stopping");
```
SergiySW: I guess this also should be always logged

nano/node/node.cpp
wezrule: Yep seems I missed adding any to node.cpp. Thanks

guilhermelawless:
This one wasn't squash merged @wezrule

wezrule:
> This one wasn't squash merged @wezrule

yep was accidental but we decided not to try rewrite the history, so not doing anything about it

PR #1790, opened 2019-02-26
Add TCP server timeout

SergiySW:
Make timeouts configurable

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -1430,10 +1430,8 @@ TEST (bootstrap, tcp_listener_timeout_empty)
 	auto node0 (system.nodes[0]);
 	node0->config.tcp_server_timeout = std::chrono::seconds (1);
 	auto socket (std::make_shared<nano::socket> (node0));
-	nano::keepalive keepalive;
-	auto input (keepalive.to_bytes ());
 	std::atomic<bool> connected (false);
-	socket->async_connect (node0->bootstrap.endpoint (), [&input, socket, &connected](boost::system::error_code const & ec) {
+	socket->async_connect (node0->bootstrap.endpoint (), [&connected](boost::system::error_code const & ec) {
 		ASSERT_FALSE (ec);
 		connected = true;
 	});
@@ -1443,6 +1441,7 @@ TEST (bootstrap, tcp_listener_timeout_empty)
 		ASSERT_NO_ERROR (system.poll ());
 	}
 	bool disconnected (false);
+	system.deadline_set (std::chrono::seconds (5));
 	while (!disconnected)
 	{
 		{
@@ -1480,6 +1479,7 @@ TEST (bootstrap, tcp_listener_timeout_keepalive)
 		ASSERT_EQ (node0->bootstrap.connections.size (), 1);
 	}
 	bool disconnected (false);
+	system.deadline_set (std::chrono::seconds (5));
 	while (!disconnected)
 	{
 		{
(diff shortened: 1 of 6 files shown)


cryptocode:
I don't think start() needs a param after this change, that function should just set cutoff to `now()` (maybe rename from cutoff to start_time or something?)

SergiySW:
@cryptocode good point!

nano/node/nodeconfig.cpp
```diff
@@ -246,6 +250,12 @@ bool nano::node_config::upgrade_json (unsigned version_a, nano::jsonconfig & jso
 			upgraded = true;
 		}
 		case 16:
+		{
+			json.put ("tcp_client_timeout", tcp_client_timeout.count ());
+			json.put ("tcp_server_timeout", tcp_server_timeout.count ());
+			upgraded = true;
```
wezrule: Should this have a test? I added one in a PR but it's not reviewed yet, feel free to copy and replace with your options: https://github.com/nanocurrency/nano-node/pull/1734/files

nano/node/bootstrap.cpp
```diff
@@ -2253,6 +2258,41 @@ void nano::bootstrap_server::finish_request ()
 	{
 		run_next ();
 	}
+	else
+	{
+		auto this_l (shared_from_this ());
+		std::weak_ptr<nano::bootstrap_server> this_w (shared_from_this ());
+		node->alarm.add (std::chrono::steady_clock::now () + node->config.tcp_server_timeout + std::chrono::seconds (1), [this_w]() {
```
wezrule: do we need this +1 second?

nano/node/bootstrap.cpp
```diff
@@ -2253,6 +2258,41 @@ void nano::bootstrap_server::finish_request ()
 	{
 		run_next ();
 	}
+	else
+	{
+		auto this_l (shared_from_this ());
```
wezrule: is this variable needed?

nano/core_test/network.cpp
```diff
@@ -1423,3 +1423,69 @@ TEST (bootstrap, keepalive)
...
+		ASSERT_EQ (node0->bootstrap.connections.size (), 1);
+	}
+	bool disconnected (false);
+	while (!disconnected)
+	{
+		{
+			std::lock_guard<std::mutex> guard (node0->bootstrap.mutex);
+			disconnected = node0->bootstrap.connections.empty ();
+		}
+		ASSERT_NO_ERROR (system.poll ());
```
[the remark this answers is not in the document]
SergiySW: Probably

nano/core_test/network.cpp
```diff
@@ -1423,3 +1423,69 @@ TEST (bootstrap, keepalive)
...
+TEST (bootstrap, tcp_listener_timeout_empty)
+{
+	nano::system system (24000, 1);
+	auto node0 (system.nodes[0]);
+	node0->config.tcp_server_timeout = std::chrono::seconds (1);
+	auto socket (std::make_shared<nano::socket> (node0));
+	nano::keepalive keepalive;
+	auto input (keepalive.to_bytes ());
+	std::atomic<bool> connected (false);
+	socket->async_connect (node0->bootstrap.endpoint (), [&input, socket, &connected](boost::system::error_code const & ec) {
```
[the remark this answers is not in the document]
SergiySW: Not needed, right

nano/node/nodeconfig.cpp
SergiySW: Thanks

nano/node/bootstrap.cpp
SergiySW: No need

nano/node/bootstrap.cpp
SergiySW: I.e. now = 100, tcp_server_timeout = 1. Then we should check after 1 second if 100 + 1 <= now ()
But what if there was some action at timestamp 100.999? Then it still will be timed out at 101 even efter 1 ms left

PR #1800, opened 2019-03-03
Improve CLI --wallet_import

SergiySW:
Return error if new wallet is encrypted & provided password invalid
Full import for new wallet with --force option instead of private keys import


zhyatt:
@SergiySW Can you explain more about how the --force option functions? I don't see any documentation for it now.

zhyatt:
If the provided wallet does not exist and --force is included, a new wallet will be created with the provided wallet id value, and the json file will be imported as is with existing seed and password (instead of a set of private keys without a change of seed).

PR #1803, opened 2019-03-05
Improve representative crawler

cryptocode:
Overall changes:

* Reps are tracked by account instead of endpoint. This thus adds support for multiple reps per host, with an increase of total weight on livenet.
* Rep crawling is currently partly in peer container and partly in free standing functions. This patch puts all the functionality into a class.

Functional changes compared to v18:

* Stored peers are crawled at startup. In my tests, this obtains sufficient weight within a couple of seconds instead of up to a minute.
* Since the rep container is now separate from the peer container, the latter no longer has a `last_rep_request` index. The rep crawler grabs random peers in addition to iterating known reps periodically. Along with the changes mentioned below, this quickly finds reps on startup, even if there are no stored peers.
* When total weight < minimum, reps are crawled slightly more aggressively than before (higher frequency, and potentially a few more peers)
* When total weight >= minimum, reps are crawled at a slightly lower frequency than before (7 vs 4 seconds). This reduces traffic a bit.

These changes appears to improve the process of obtaining peer weight in my tests.

Part of https://github.com/nanocurrency/nano-node/issues/1112

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1213,7 +1213,7 @@ TEST (node, DISABLED_fork_stale)
 	auto & node1 (*system1.nodes[0]);
 	auto & node2 (*system2.nodes[0]);
 	node2.bootstrap_initiator.bootstrap (node1.network.endpoint ());
-	node2.peers.rep_response (node1.network.endpoint (), nano::test_genesis_key.pub, nano::genesis_amount);
+	node2.rep_crawler.response (node1.network.endpoint (), nano::test_genesis_key.pub, nano::genesis_amount);
 	nano::genesis genesis;
 	nano::keypair key1;
 	nano::keypair key2;
@@ -1501,17 +1501,17 @@ TEST (node, rep_list)
 	nano::keypair key1;
 	// Broadcast a confirm so others should know this is a rep node
 	wallet0->send_action (nano::test_genesis_key.pub, key1.pub, nano::Mxrb_ratio);
-	ASSERT_EQ (0, node1.peers.representatives (1).size ());
+	ASSERT_EQ (0, node1.rep_crawler.representatives (1).size ());
 	system.deadline_set (10s);
 	auto done (false);
 	while (!done)
 	{
-		auto reps (node1.peers.representatives (1));
+		auto reps (node1.rep_crawler.representatives (1));
 		if (!reps.empty ())
 		{
 			if (reps[0].endpoint == node0.network.endpoint ())
 			{
-				if (!reps[0].rep_weight.is_zero ())
+				if (!reps[0].weight.is_zero ())
 				{
 					done = true;
 				}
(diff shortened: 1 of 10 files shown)


nano/core_test/node.cpp
```diff
@@ -1521,6 +1521,33 @@ TEST (node, rep_list)
...
+	nano::amount amount50 (50);
+	node.peers.insert (endpoint2, nano::protocol_version);
+	node.peers.insert (endpoint0, nano::protocol_version);
+	node.peers.insert (endpoint1, nano::protocol_version);
+	nano::keypair keypair1;
+	nano::keypair keypair2;
+	node.rep_crawler.response (endpoint0, keypair1.pub, amount100);
+	node.rep_crawler.response (endpoint1, keypair2.pub, amount50);
+	// Make sure we get the rep with the most weight first
+	auto reps (node.rep_crawler.representatives (1));
```
SergiySW: Could we also check total representatives size == 2 ?

nano/node/node.cpp
```diff
@@ -1498,7 +1452,6 @@ startup_time (std::chrono::steady_clock::now ())
 	}
 	observers.endpoint.add ([this](nano::endpoint const & endpoint_a) {
 		this->network.send_keepalive (endpoint_a);
-		rep_query (*this, endpoint_a);
```
SergiySW: Why not query new peers?

cryptocode: The rep_crawler class observers this.

nano/core_test/node.cpp
cryptocode: Sure. The rep container isn't exposed, but I could add a total_size function for tests.

nano/node/node.cpp
SergiySW: I see

nano/node/repcrawler.hpp
```diff
@@ -0,0 +1,131 @@
...
+	{
+	}
+	nano::account account{ 0 };
+	nano::amount weight{ 0 };
+	nano::endpoint endpoint;
+	std::chrono::steady_clock::time_point last_request{ std::chrono::steady_clock::time_point () };
+	std::chrono::steady_clock::time_point last_response{ std::chrono::steady_clock::time_point () };
+};
+
+using PROBABLE_REP_TYPE = boost::multi_index_container<
```
clemahieu: It looks like this container is only accessed within rep_crawler now which is an improvement in encapsulation. Can we move this type and the tags privately inside the rep_crawler class so they aren't in the nano namespace.

clemahieu: All uppercase here looks like a constant.  probably_rep_t seems good.

nano/node/repcrawler.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+	using PROBABLE_REP_ITR = PROBABLE_REP_TYPE::index<nano::tag_endpoint>::type::iterator;
+	PROBABLE_REP_TYPE::index<nano::tag_endpoint>::type & endpoint_index = probable_reps.get<nano::tag_endpoint> ();
+
+	// Find and update the timestamp on all reps available on the endpoint (a single host may have multiple reps)
+	std::vector<PROBABLE_REP_ITR> view;
+	auto itr_pair = probable_reps.get<nano::tag_endpoint> ().equal_range (endpoint_a);
+	for (; itr_pair.first != itr_pair.second; itr_pair.first++)
+	{
+		view.push_back (itr_pair.first);
```
clemahieu: Is storing this list to avoid modifying the container while iterating?

Reading the documentation it seems safe to do a modification while iterating since we're modifying a non-unique index
https://www.boost.org/doc/libs/1_57_0/libs/multi_index/doc/reference/ord_indices.html#modify

clemahieu:
LGTM after some small nits.

nano/node/repcrawler.hpp
cryptocode: Good points, will fix.

nano/node/repcrawler.cpp
cryptocode: Didn't know that, thanks!

PR #1691, opened 2019-02-03
Use random shuffle for active_transactions::request_confirm ()

SergiySW:
for blocks rebroadcast & confirmation request in overload mode

PR #1810, opened 2019-03-07
Use adjusted difficulty for trees in active transactions

SergiySW:
* add adjusted difficulty calculation for active transactions
* make list of dependent blocks in election

Better alternative to https://github.com/nanocurrency/nano-node/pull/1691


nano/node/node.cpp
```diff
@@ -3477,9 +3519,94 @@ void nano::active_transactions::update_difficulty (nano::block const & block_a)
...
+			adjust_difficulty (block_a.hash ());
+		}
+	}
+}
+
+void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash_a)
+{
+	assert (!mutex.try_lock ());
+	std::deque<std::pair<nano::block_hash, int64_t>> remaining_blocks;
+	remaining_blocks.push_back (std::make_pair (hash_a, 0));
```
wezrule: Rather than make a pair then copy/move it to the deque (which push_back will do) you can use remaining_blocks.emplace_back (previous, item.second + 1); which creates the pair directly inside the deque using _placement new_. The same can be done with the other uses in this function

nano/node/node.cpp
```diff
@@ -3477,9 +3519,94 @@ void nano::active_transactions::update_difficulty (nano::block const & block_a)
...
+	std::unordered_set<nano::block_hash> processed_blocks;
+	std::vector<std::pair<nano::uint512_union, int64_t>> elections_list;
+	uint128_t sum (0);
+	while (!remaining_blocks.empty ())
+	{
+		auto item (remaining_blocks.front ());
+		remaining_blocks.pop_front ();
+		auto hash (item.first);
+		// item.second is level
```
wezrule: perhaps use a variable named `level` instead of this comment

nano/node/node.cpp
```diff
@@ -3477,9 +3519,94 @@ void nano::active_transactions::update_difficulty (nano::block const & block_a)
...
+	assert (!mutex.try_lock ());
+	std::deque<std::pair<nano::block_hash, int64_t>> remaining_blocks;
+	remaining_blocks.push_back (std::make_pair (hash_a, 0));
+	std::unordered_set<nano::block_hash> processed_blocks;
+	std::vector<std::pair<nano::uint512_union, int64_t>> elections_list;
+	uint128_t sum (0);
+	while (!remaining_blocks.empty ())
+	{
+		auto item (remaining_blocks.front ());
+		remaining_blocks.pop_front ();
```
wezrule: You could move the `pop_front ()` to the end of the loop and then `item` can be a reference which saves a copy each iteration.

nano/node/node.cpp
```diff
@@ -3477,9 +3519,94 @@ void nano::active_transactions::update_difficulty (nano::block const & block_a)
...
+		{
+			auto existing_root (roots.find (item.first));
+			uint64_t difficulty_a (average + (item.second / divider));
+			roots.modify (existing_root, [difficulty_a](nano::conflict_info & info_a) {
+				info_a.adjusted_difficulty = difficulty_a;
+			});
+		}
+	}
+	// Return difficulty instead of adjusted value
+	else if (elections_list.size () == 1)
```
wezrule: Is `Return` the correct word to use here?

nano/node/node.cpp
SergiySW: Thx

nano/node/node.cpp
SergiySW: Good point

nano/node/node.cpp
SergiySW: Thanks

nano/node/node.cpp
SergiySW: Excellent find!

PR #1814, opened 2019-03-09
Valgrind fixes & suppressions

SergiySW:
* Fix detelcted by Valgrind compute_reps () issue
* Improve logging headers
* Minor tests improvements & fixes
* Valgrind suppression file (usage valgrind --suppressions=valgrind.supp)
Note: to run Valgrind reduce map_size in lmdb.hpp from 128GB to to several MB


nano/node/logging.hpp
```diff
@@ -79,9 +79,10 @@ class logger_mt
 		return error;
 	}
 
+	std::chrono::milliseconds min_log_delta{ 5 };
```
wezrule: Did valgrind complain when this was a reference? This was done so that it pointed to the value in the `logging` class and would be updated if that was, which I suppose is not a necessity. Although it doesn't need to be initialized in-class here as this is set in the constructor.

PR #1816, opened 2019-03-10
Add required stake to confirmation_quorum

cryptocode:
The rules for effective stake required for quorum may change and isn't obvious from the output of confirmation_quorum (it's the largest of online_weight_minimum and quorum_delta)

Adding `peers_stake_required` to make this obvious and allow for future changes to the rules without breaking clients.


SergiySW:
Should it also be converted with .convert_to<std::string> () ?

cryptocode:
@SergiySW it's not necessary, but I can add it to be consistent.

PR #1821, opened 2019-03-13
Periodically check for unconfirmed frontiers and start elections

SergiySW:
With max limit of elections to start: 250. More frequest checks for representative nodes


guilhermelawless:
How about having a "queued elections" buffer? For this PR, after the limit for active elections are reached, it could queue up the other elections to be started later.

SergiySW:
@guilhermelawless main reason for this limit here is to prevent large udp fraffic so nodes can passively wait for frontiers confirmation in usual circumstances

nano/node/node.cpp
```diff
@@ -2453,6 +2453,32 @@ void nano::node::block_confirm (std::shared_ptr<nano::block> block_a)
...
 
+void nano::node::confirm_frontiers ()
+{
+	// Limit maximum count of elections to start. max_broadcast_queue for regular nodes, x10 for representatives
+	bool representative (config.enable_voting && wallets.reps_count > 0);
+	size_t max_elections (representative ? active.max_broadcast_queue * 10 : active.max_broadcast_queue);
+	size_t elections_count (0);
+	auto transaction (store.tx_begin_read ());
+	for (auto i (store.latest_begin (transaction)), n (store.latest_end ()); i != n && elections_count < max_elections; ++i)
+	{
```
wezrule: I wonder if this will create some unnecessary traffic if it asks for confirmations for accounts which may get implicitly confirmed in a previous iteration from another account?
Do you think it would make sense to have a separate table (or something) which keeps track of any accounts which aren't fully confirmed rather than scan all 600k+ accounts every time? Most accounts are generally static I guess.
`block_confirm` calls `network.broadcast_confirm_req (block_a);` is that necessary here as well?

SergiySW: `block_confirm` replaced with `active_transactions::start`

SergiySW: Hm, I guess we can disable check for bootstraps that didn't process any blocks (most bootstraps for synced node)

guilhermelawless:
> @guilhermelawless main reason for this limit here is to prevent large udp fraffic so nodes can passively wait for frontiers confirmation in usual circumstances

Couldn't you say the same about all these frontiers that will get elections at once? It just seems strange to confirm some and not others.

SergiySW:
1k elections is not huge traffic. 600k is

SergiySW:
@clemahieu what delays between frontiers scans seems reasonable?

nano/node/node.cpp
```diff
@@ -2727,6 +2727,44 @@ size_t nano::election::last_votes_size ()
...
+					}
+				}
+				// Update last account
+				last_frontier_account = i->first;
+			}
+		}
+		// Update last account if all frontiers checked
+		if (elections_count <= max_elections)
+		{
+			last_frontier_account = 0;
```
clemahieu: int fully_confirmed_factor = elections_count <= max_elections ? 4 : 1;

nano/node/node.cpp
```diff
@@ -2727,6 +2727,44 @@ size_t nano::election::last_votes_size ()
...
+		}
+	}
```
clemahieu: next_frontier_check = std::chrono::steady_clock::now () + std::chrono::seconds(representative_factor * fully_confirmed_factor);

nano/node/node.cpp
```diff
@@ -2727,6 +2727,44 @@ size_t nano::election::last_votes_size ()
...
 }
 
+void nano::active_transactions::confirm_frontiers (nano::transaction const & transaction_a)
+{
+	// Limit maximum count of elections to start
+	bool representative (node.config.enable_voting && node.wallets.reps_count > 0);
+	/* Check less frequently for non-representative nodes and if previous checks show no unconfirmed frontiers
+	~15 minutes for non-representative nodes, 3.5 minutes for representatives
+	4 times slower if previously all frontiers were confirmed */
+	if (request_confirm_iteration % ((60 - (45 * representative)) * (3 * last_frontier_account.is_zero () + 1)) == 0)
```
clemahieu: Let's simplify this logic a bit by calculating a time point when we should check next.
if (std::chrono::steady_clock::now () > = next_frontier_check)

nano/node/node.hpp
```diff
@@ -154,6 +154,9 @@ class active_transactions
 	// clang-format on
 	void request_loop ();
 	void request_confirm (std::unique_lock<std::mutex> &);
+	void confirm_frontiers (nano::transaction const &);
+	nano::account last_frontier_account{ 0 };
+	uint64_t request_confirm_iteration{ 0 };
```
clemahieu: std::chrono::steady_clock next_frontier_check;

nano/node/node.cpp
```diff
@@ -2727,6 +2727,44 @@ size_t nano::election::last_votes_size ()
...
+				{
+					++elections_count;
+					// Calculate votes for local representatives
+					if (representative)
+					{
+						node.block_processor.generator.add (block->hash ());
+					}
+				}
+				// Update last account
+				last_frontier_account = i->first;
```
clemahieu: It looks like this will check the same account the next time through the loop.  last_frontier_account = i->first + 1

nano/node/node.hpp
```diff
@@ -154,6 +154,9 @@ class active_transactions
 	// clang-format on
 	void request_loop ();
 	void request_confirm (std::unique_lock<std::mutex> &);
+	void confirm_frontiers (nano::transaction const &);
+	nano::account last_frontier_account{ 0 };
```
clemahieu: This should probably be next_frontier_account since we start at 0 and check the one this points to.

nano/node/node.cpp
```diff
@@ -2727,6 +2727,44 @@ size_t nano::election::last_votes_size ()
 	return last_votes.size ();
 }
 
+void nano::active_transactions::confirm_frontiers (nano::transaction const & transaction_a)
+{
+	// Limit maximum count of elections to start
+	bool representative (node.config.enable_voting && node.wallets.reps_count > 0);
+	/* Check less frequently for non-representative nodes and if previous checks show no unconfirmed frontiers
```
clemahieu: int representative_factor = representative ? 3 * 60 : 15 * 60;

nano/node/node.hpp
SergiySW: Right

nano/node/node.cpp
SergiySW: Right

nano/node/node.hpp
SergiySW: next_frontier_check seems better than now

nano/node/node.cpp
SergiySW: it's more clear

nano/node/node.cpp
SergiySW: Makes sense. Could you check new version?

PR #1822, opened 2019-03-13
RPC account_history add block height and optional reverse

### LINKED ISSUE #1820: Get account history starting at a block and ending in head
Reported by Natrium.
Currently it is cumbersome for integrators to build an incremental account history internally. Scenario:

- Initially, request full `account_history` to cache it internally.
- When updating the history, to **prevent getting too much data unecessarily** (some accounts have hundreds of thousands of blocks), one would expect to use `account_history` starting at the internally cached head block, and up to the new head block.

Currently I see no easy way of doing this, and it seems like a pretty important thing to have. `successors` doesn't let you **start** at a block, only end. If that option existed, one could use `successors` followed by `blocks_info` as an alternative.

Ideally, `account_history` would have an optional `reverse` and could also start at a block. However, the semantics of the returned values stop making sense in that RPC, namely, the returned field "previous" would have to be "next" instead.

What would be the best way to approach this? I don't mind coding it once it's agreed upon.

----

Also, maybe we could add (block) "height" to the blocks in `account_history`?

guilhermelawless:
Fixes #1820 .

**TODO**
- [x] Wait for #1355 
- [x] Merge those changes here
- [x] Modify tests to account for block height and reverse

==== Wiki update ====

## Account history  

(...)

Response:  
```
{  
    "account": "xrb_1ipx847tk8o46pwxt5qjdbncjqcbwcc1rrmqnkztrfjy5k7z4imsrata9est",   
    "history": [   
        {   
            "type": "send",   
            "account": "xrb_38ztgpejb7yrm7rr586nenkn597s3a1sqiy3m3uyqjicht7kzuhnihdk6zpz",   
            "amount": "80000000000000000000000000000000000",   
            "local_timestamp": "1551532723",  
            "height": "60", 
            "hash": "80392607E85E73CC3E94B4126F24488EBDFEB174944B890C97E8F36D89591DC5"   
        }   
    ],   
    "previous": "8D3AB98B301224253750D448B4BD997132400CEDD0A8432F775724F2D9821C72"   
}   
```

(...)

- `offset`: skips a number of blocks starting from head (if given). Not often used.
- `reverse`: if true, the response starts from head (if given, otherwise the first block of the account), and lists all blocks up to the frontier. **Note**: the field `previous` in the response changes to `next`.


nano/node/rpc.cpp
```diff
@@ -1963,6 +1964,15 @@ void nano::rpc_handler::account_history ()
 		if (!ec)
 		{
 			hash = node.ledger.latest (transaction, account);
+			if (reverse)
```
SergiySW: You can just use account_info open block

guilhermelawless:
So the documentation says 

> Reports send/receive information for an account. Change blocks are skipped, open blocks will appear as receive  unless raw is set to true - see optional parameters below

I think now with #1774 the open block is showing even with raw: false, and shows up as a receive.

nano/node/rpc.cpp
guilhermelawless: Done, good point. Ignore the other comment I made (deleted), was confused about legacy vs state blocks.

guilhermelawless:
Merge and tests done.

nano/core_test/rpc.cpp
```diff
@@ -1146,73 +1146,101 @@ TEST (rpc, account_history)
...
+		boost::property_tree::ptree other_account;
+		other_account.put ("", account2.to_account ());
+		boost::property_tree::ptree filtered_accounts;
+		filtered_accounts.push_back (std::make_pair ("", other_account));
+		request.add_child ("account_filter", filtered_accounts);
+		request.put ("count", 100);
+		test_response response (request, rpc, system.io_ctx);
+		while (response.status == 0)
+		{
+			system.poll ();
```
SergiySW: I guess `ASSERT_NO_ERROR (system.poll ());`

Issue #1826, opened 2019-03-14
Timestamp import feature


BitDesert:
Now that we have local timestamps it would be great if we could import/export older timestamps to the node. That way we would only have to query one source for everything when building an application that shows timestamps rather than keeping a a seperate callback app and Redis.

schenkty:
I requested the exact same thing only a few hours ago on discord.

renesq:
I don't think we will be able to get reliable timestamps, back to the inception of Raiblocks. But I see how an import/export could indeed be a use case, even tho it's not super reliable. Also keep in mind that there have been "account chain modified" timestamps for quite some time now, but most of this information got lost when spam attacks made node operators do a re-sync, that's why nanode will show 
"This account was last updated December 17, 2018 1:18 AM." for the Landing account that hasn't been used since mid '17. 

SergSW's  prebuilt ledger file has block timestamps as of ~1550832662 (22nd of february '19)
Nanode probably is the earliest authority to deliberately record block timestamps in a database. Wonder if they delete older data as they will only list the newest transactions. Nanocrawler is another one to record timestamps but about 10% of callbacks failed on there, so a lot of timestamps are missing. 

I guess that you could also use a LMDB python library to achieve your goal. Reading is a bit easier than updating AFAIK.

zhyatt:
After discussing internally we consider importing and exporting of timestamps to be beyond the scope of node operations. As @renesq mentioned, using LMDB python libraries or other similar tools can provide methods for accomplishing this with external scripts. And although we haven't tested the updates, @guilhermelawless has a patch here that has the potential to be built in the node for performing these actions via CLI commands: https://gist.github.com/guilhermelawless/545e4490bac4632e9a9d5bca6379b439.

schenkty:
Extremely unfortunate. The node should have had that basic functionality from day one of production local timestamp support.

guilhermelawless:
@schenkty  let me know if you need help applying the patch to get the functionality.

clemahieu:
The block store was written to allow storage plugins.  Another option is if you write a new storage backend, maybe SQLite, we can work on taking in your patch.

PR #1831, opened 2019-03-15
CLI export timestamps

guilhermelawless:
Part of #1826 .

Currently outputs to `std::cout` as pairs "hash,timestamp" ready for CSV parsing. Can probably be optimized further, maybe by saving hash-timestamp pairs in memory and only writing in the end.

Because it takes some time, also outputs to `std::cerr` checkpoints (no more than 100 checkpoints, but depends on the number of frontiers).

Looking for some suggestions/help.


guilhermelawless:
Now reads to memory first, which made it faster. Bottleneck is CPU. Is there a faster way to iterate through blocks in the DB without using previous and block_get?

Changed to write a CSV file in `data_path` instead.

PR #1837, opened 2019-03-16
Add the concept of message channels to the node.

clemahieu:
All transport related code is in the node/transport directory.
There is a generic base class nano::transport::channel that represents a bi-directional messaging channel that can be subclassed by different transport layer types.
UDP-channels are managed by a class nano::transport::udp_channels which largely replaces peer_container.
Generic network related functionality is handled by nano::network and the nano::peer_container class has been removed.
UDP-specific functionality such as keepalives and UDP handshaking is handled by the udp_channels class
TCP channels have been added and the bootstrapping functionality has been rewritten to push messages through these channels.


cryptocode:
Tests and node on live runs fine in my testing so far.

A couple of questions: 

a) Since the type is now named `transport#channel`, should the parameters and locals be called `channel` instead of `sink` as well?

b) For code like below, I wonder if `send_buffer` for TCP needs to be queued, as it calls `async_write`, and there can only be one write operation in flight at the same time per stream.

```
	for (auto & vote : votes)
	{
		nano::confirm_ack confirm (vote);
		auto vote_bytes = confirm.to_bytes ();
		sink_a.send_buffer (vote_bytes, nano::stat::detail::confirm_ack);
	}
```

(These cases are currently UDP-only, so no issue right now)

clemahieu:
Yes I'll change the references from sink to channel.

The TCP operations will need to be queued yes, the existing code self-limited so there were never overlapping write operations but to make it more generic it should be queued and queue length limited.

We'll also have to devise some way for the bootstrap operations to lock the channel while their operation is being performed since they're not message-length operations.

PR #1838, opened 2019-03-16
CLI import and export timestamps using CSV files

### LINKED ISSUE #1826: Timestamp import feature

guilhermelawless:
Closes #1826 .
Alternative to #1831 with both import and export.

Would it be worth it to build a way to upgrade only the timestamp, without using store.block_put? As it is right now, it pretty much rewrites the whole database. Still, using a fast SSD it's bottlenecked by CPU and takes less than an hour. Exporting is fast.

### DIFF
--- a/nano/node/cli.cpp
+++ b/nano/node/cli.cpp
@@ -53,6 +53,8 @@ void nano::add_node_options (boost::program_options::options_description & descr
 	("wallet_representative_get", "Prints default representative for <wallet>")
 	("wallet_representative_set", "Set <account> as default representative for <wallet>")
 	("vote_dump", "Dump most recent votes from representatives")
+	("timestamps_import", "Imports a CSV file, overwriting the timestamps recorded in the database (warning: high resource usage).")
+	("timestamps_export", "Writes a CSV file with the local timestamp recorded for each hash with timestamp in the database.")
 	("account", boost::program_options::value<std::string> (), "Defines <account> for other commands")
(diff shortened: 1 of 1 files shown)


nano/node/cli.cpp
```diff
@@ -950,6 +952,167 @@ std::error_code nano::handle_node_options (boost::program_options::variables_map
...
+					std::cout << "Upgrading database..." << std::endl;
+
+					auto block_count (pairs.size ());
+					size_t count{ 0 };
+					size_t step (std::max<size_t> (10, std::pow (10.0f, std::floor (std::log10 (block_count / 10.0)))));
+					auto transaction (node.node->store.tx_begin_write ());
+					nano::block_sideband sideband;
+					for (auto i (pairs.begin ()), n (pairs.end ()); i != n; ++i, ++count)
+					{
+						auto block (node.node->store.block_get (transaction, i->first, &sideband));
```
SergiySW: Block can be nullptr

guilhermelawless: Handled in 5dab4cc

guilhermelawless:
For future reference here's a git patch for this PR: https://gist.github.com/guilhermelawless/545e4490bac4632e9a9d5bca6379b439

zhyatt:
See note in #1826 about timestamp import/export out of scope.

PR #1840, opened 2019-03-17
Websocket support

cryptocode:
Adds support for notifications via Websockets. This offers a more efficient callback mechanism for block confirmations, and allows for multiple listeners. As it's topic-based, filtering and additional events can be added in the future. Proper JSON is returned (the http callback returns stringified blocks). Comments on message formatting are most welcome. Note that the http callback is still supported.

Preliminary docs: https://github.com/cryptocode/notes/wiki/Websocket-support
Sample nodejs client: https://github.com/cryptocode/nano-websocket-sample-nodejs

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1050,7 +1050,8 @@ startup_time (std::chrono::steady_clock::now ())
 {
 	if (config.websocket_config.enabled)
 	{
-		websocket_server = std::make_shared<nano::websocket::listener> (*this, nano::tcp_endpoint (boost::asio::ip::address_v6::any (), config.websocket_config.port));
+		auto endpoint_l (nano::tcp_endpoint (config.websocket_config.address, config.websocket_config.port));
+		websocket_server = std::make_shared<nano::websocket::listener> (*this, endpoint_l);
 		this->websocket_server->run ();
 	}
 
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1048,6 +1048,12 @@ stats (config.stat_config),
 vote_uniquer (block_uniquer),
 startup_time (std::chrono::steady_clock::now ())
 {
+	if (config.websocket_config.enabled)
+	{
+		websocket_server = std::make_shared<nano::websocket::listener> (*this, nano::tcp_endpoint (boost::asio::ip::address_v6::any (), config.websocket_config.port));
+		this->websocket_server->run ();
+	}
+
 	wallets.observer = [this](bool active) {
 		observers.wallet.notify (active);
 	};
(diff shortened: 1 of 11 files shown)


nano/node/websocket.cpp
```diff
@@ -0,0 +1,310 @@
...
+	ws.close (reason, ec_ignore);
+}
+
+void nano::websocket::session::write (nano::websocket::message message_a)
+{
+	// clang-format off
+	std::lock_guard<std::mutex> lk (subscriptions_mutex);
+	if (message_a.topic == nano::websocket::topic::ack || subscriptions.find (message_a.topic) != subscriptions.end ())
+	{
+		subscriptions_mutex.unlock ();
```
clemahieu: This should be a unique_lock::unlock instead of a lock_guard and then manually unlocking the mutex. When unlock is called and then lk goes out of scope, effectively unlock is called again which is UB.
https://en.cppreference.com/w/cpp/thread/mutex/unlock

nano/node/websocket.cpp
```diff
@@ -0,0 +1,310 @@
...
+	auto msg (send_queue.front ());
+	auto msg_str (msg.to_string ());
+
+	std::lock_guard<std::mutex> lk (io_mutex);
+	ws.async_write (boost::asio::buffer (msg_str.data (), msg_str.size ()),
+	boost::asio::bind_executor (write_strand,
+	[msg, self_l = shared_from_this ()](boost::system::error_code ec, std::size_t bytes_transferred) {
+		if (!ec)
+		{
+			self_l->send_queue.pop_front ();
```
clemahieu: We need to pop_front when there's an error also right?

nano/node/node.cpp
```diff
@@ -1048,6 +1048,12 @@ stats (config.stat_config),
 vote_uniquer (block_uniquer),
 startup_time (std::chrono::steady_clock::now ())
 {
+	if (config.websocket_config.enabled)
+	{
+		websocket_server = std::make_shared<nano::websocket::listener> (*this, nano::tcp_endpoint (boost::asio::ip::address_v6::any (), config.websocket_config.port));
```
clemahieu: The bind address should be configurable.

cryptocode: Good idea, adding the same way as is done for RPC

nano/node/websocket.cpp
cryptocode: Ah, yeah that was bonkers. Thanks for spotting.

nano/node/websocket.cpp
cryptocode: Wasn't sure. If there's an intermittent error, then it'll be sent on the next try. But yeah, if keeps err'ing, the queue will fill up, so seems safest to always pop it. Changing.

nano/node/websocket.cpp
```diff
@@ -0,0 +1,310 @@
...
+
+nano::websocket::session::session (nano::websocket::listener & listener_a, boost::asio::ip::tcp::socket socket_a) :
+ws_listener (listener_a), ws (std::move (socket_a)), write_strand (ws.get_executor ())
+{
+	ws.text (true);
+}
+
+nano::websocket::session::~session ()
+{
+	ws_listener.get_node ().logger.try_log ("websocket session ended");
```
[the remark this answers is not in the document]
cryptocode: Added

PR #1841, opened 2019-03-18
Fix assert assignment for Debug builds

guilhermelawless:


guilhermelawless:
Strike of luck. Also, `grep -r 'assert (' | grep ' = '` doesn't catch anything else :+1:

Issue #1660, opened 2019-01-29
Allow block_processor variables tuning with launch flags


SergiySW:
block_processor::full () size
and max processed blocks in block_processor::process_batch ()

SergiySW:
Closed via https://github.com/nanocurrency/nano-node/pull/1492

PR #1842, opened 2019-03-18
Allow block_processor variables tuning with launch flags

SergiySW:
--block_processor_batch_size
--block_processor_full_size
--block_processor_verification_size

Can be useful for fast nodes and database profiling
https://github.com/nanocurrency/nano-node/issues/1660


nano/nano_node/entry.cpp
```diff
@@ -32,6 +32,9 @@ int main (int argc, char * const * argv)
 		("disable_unchecked_drop", "Disables drop of unchecked table at startup")
 		("fast_bootstrap", "Increase bootstrap speed for high end nodes with higher limits")
 		("batch_size",boost::program_options::value<std::size_t> (), "Increase sideband batch size, default 512")
+		("block_processor_batch_size",boost::program_options::value<std::size_t> (), "Increase block processor transaction batch write size, default 0 (limited by config block_processor_batch_max_time)")
+		("block_processor_full_size",boost::program_options::value<std::size_t> (), "Increase block processor allowed blocks queue size before dropping live network packets and holding bootstrap download, default 65536")
```
cryptocode: "default 65536"

Not sure if it's important to state, but block_processor_full_size has a different default for fast_bootstrap.

SergiySW: Done

nano/node/nodeconfig.hpp
```diff
@@ -69,14 +71,18 @@ class node_flags
 {
 public:
 	node_flags ();
```
wezrule: This constructor can be `= default` now that you are using in-class member initialization for all members. Although depending on the result below you might not need any constructors other than the implicitly defined default one.

nano/node/nodeconfig.hpp
```diff
@@ -69,14 +71,18 @@ class node_flags
...
-	bool disable_backup;
-	bool disable_lazy_bootstrap;
-	bool disable_legacy_bootstrap;
-	bool disable_wallet_bootstrap;
-	bool disable_bootstrap_listener;
-	bool disable_unchecked_cleanup;
-	bool disable_unchecked_drop;
-	bool fast_bootstrap;
-	size_t sideband_batch_size;
+	node_flags (boost::program_options::variables_map const &);
```
wezrule: `boost::program_options::variables_map` is quite a high level construct so I don't think `node_flags` should have a dependency on it. I think you can just move the logic contained in the constructor somewhere higher up (entry.cpp) and set it similarly as before.

nano/nano_node/entry.cpp
```diff
@@ -32,6 +32,9 @@ int main (int argc, char * const * argv)
 		("disable_unchecked_drop", "Disables drop of unchecked table at startup")
 		("fast_bootstrap", "Increase bootstrap speed for high end nodes with higher limits")
 		("batch_size",boost::program_options::value<std::size_t> (), "Increase sideband batch size, default 512")
+		("block_processor_batch_size",boost::program_options::value<std::size_t> (), "Increase block processor transaction batch write size, default 0 (limited by config block_processor_batch_max_time), 256k for fast_bootstrap")
+		("block_processor_full_size",boost::program_options::value<std::size_t> (), "Increase block processor allowed blocks queue size before dropping live network packets and holding bootstrap download, default 65536, 1 million for fast_bootstrap")
+		("block_processor_verification_size",boost::program_options::value<std::size_t> (), "Increase batch signature verification size in block processor, default 0 (2048 * signature checker threads + 1), unlimited for fast_bootstrap")
```
wezrule: Should it say `default 0` here? looks a bit confusing.
Also following BODMAS we should put: `2048 * (signature checker threads + 1)`

nano/node/nodeconfig.hpp
SergiySW: Right

nano/nano_node/entry.cpp
SergiySW: Is new definition better?
Config is created in `daemon.run ()` after this

wezrule: I misread it actually :P, yeh looks good.

But there is some duplication in the setting of the `node_flags` now in nano_node/nano_wallet entry.cpp now, can this be put into a function accessible to both .exes? (the new `if (flags.fast_bootstrap)` etc..` stuff). Alternatively just create a new function in `nano_nano` for it, as the wallet is being separated out later so these options will be removed anyway I'd imagine. Actually does this even work as we don't set the `description.add_options ()` for nano_wallet anyway?

SergiySW: You right, it's not working in wallet *shoked*

SergiySW: You right, it's not even working there..

SergiySW: Made separate function for node

wezrule: Looks good

nano/node/nodeconfig.hpp
wezrule: `node_flags()` default constructor can be removed

Issue #247, opened 2017-12-20
Official wallet doesn't support decimals


frankh:
I've been doing some testing send small amounts of XRB between the web wallet and the desktop wallet.

First i tried sending 0.000001 from web to desktop. It worked fine, except the desktop wallet shows the balance as 0 still and amount on the receive as 0. I thought fine maybe it just rounds down. I tried sending it back, and it sent a transaction sending 0.000000XRB back.

I tried the same process with 0.1 and got the same problem.

In the end i fixed it by going to advanced and changing the scale to xrb instead of Mxrb. Imo out of the box it should display enough decimal places when necessary

XHyperY:
Advanced ->Scale -> Mxrb -> kxrb -> xrb   You can change the scale however you want.

tnkmt:
Why is this? It's a very anti-intuitive.

cryptocode:
@tnkmt Consider it a developer's wallet. A new one for mere mortals is underway - sign up for the beta on nano.co

zhyatt:
Resolved by #1843

PR #1843, opened 2019-03-20
 Support sending decimals from QT wallet

SergiySW:
https://github.com/nanocurrency/nano-node/issues/247


nano/lib/numbers.cpp
```diff
@@ -564,6 +564,79 @@ bool nano::uint128_union::decode_dec (std::string const & text)
 	return error;
 }
 
+bool nano::uint128_union::decode_dec (std::string const & text, nano::uint128_t scale)
+{
+	bool error (text.size () > 40 || (!text.empty () && text.front () == '-'));
+	if (!error)
+	{
+		auto decimal_point = std::use_facet<std::numpunct<char>> (std::locale ()).decimal_point ();
```
cryptocode: I think this will always use the C locale because locale's default ctor ends up using `locale::classic` (C). The tests assume "." so maybe this is intentional? (in which case . could be hardcoded)

SergiySW: I'm not sure if we should suppor other locales later (as it can cause confusions). Hardcoding back dot for now

PR #1845, opened 2019-03-20
Use RAII when stopping daemon to cleanup on SIGINT/SIGTERM

devinus:
This gracefully cleans up resources and frees ports using RAII and adding a signal handler to the daemon.

I expect a discussion around this, but I have yet to experience issues around ports being held on to after the process has died on macOS with this patch which previously has not been the case.

### DIFF
--- a/nano/nano_node/daemon.hpp
+++ b/nano/nano_node/daemon.hpp
@@ -2,13 +2,27 @@
 
 namespace nano
 {
+class node;
+class rpc;
 class node_flags;
+namespace ipc
+{
+	class ipc_server;
+}
 }
 namespace nano_daemon
 {
 class daemon
 {
 public:
-	void run (boost::filesystem::path const &, nano::node_flags const & flags);
+	daemon (boost::filesystem::path const & data_path, nano::node_flags const & flags);
+	~daemon ();
+	void run ();
+	void stop ();
+	boost::filesystem::path const & data_path;
+	nano::node_flags const & flags;
+	std::shared_ptr<nano::node> node;
+	std::shared_ptr<nano::rpc> rpc;
+	std::shared_ptr<nano::ipc::ipc_server> ipc;
 };
 }
--- a/nano/nano_node/entry.cpp
+++ b/nano/nano_node/entry.cpp
@@ -1,3 +1,7 @@
+#include <atomic>
+#include <csignal>
+#include <cstdlib>
+#include <functional>
 #include <nano/lib/utility.hpp>
 #include <nano/nano_node/daemon.hpp>
 #include <nano/node/cli.hpp>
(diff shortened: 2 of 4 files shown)


nano/node/ipc.cpp
```diff
@@ -398,7 +398,7 @@ node (node_a), rpc (rpc_a)
 
 nano::ipc::ipc_server::~ipc_server ()
 {
-	node.logger.always_log ("IPC: server stopped");
```
wezrule: should we check that `stop ()` hasn't already been called in cases like this?:
```
{
ipc_server server;
server.stop ();
}
```
Could cause issues I guess.

nano/nano_node/daemon.hpp
```diff
@@ -2,13 +2,27 @@
...
+}
 }
 namespace nano_daemon
 {
 class daemon
 {
 public:
-	void run (boost::filesystem::path const &, nano::node_flags const & flags);
+	daemon (boost::filesystem::path const & data_path, nano::node_flags const & flags);
+	~daemon ();
```
wezrule: As you are not using a virtual destructor (which I think is correct), the class should be `final` so it cannot be subclassed.

nano/nano_node/daemon.hpp
```diff
@@ -2,13 +2,27 @@
...
+	void run ();
+	void stop ();
+	boost::filesystem::path const & data_path;
+	nano::node_flags const & flags;
+	std::shared_ptr<nano::node> node;
+	std::shared_ptr<nano::rpc> rpc;
+	std::shared_ptr<nano::ipc::ipc_server> ipc;
```
wezrule: `node`, `rpc` and `ipc` can all be `std::unique_ptr` (I know node is unrelated to your change)

nano/nano_node/entry.cpp
```diff
@@ -115,7 +133,16 @@ int main (int argc, char * const * argv)
 			flags.disable_unchecked_cleanup = (vm.count ("disable_unchecked_cleanup") > 0);
 			flags.disable_unchecked_drop = (vm.count ("disable_unchecked_drop") > 0);
 			flags.fast_bootstrap = (vm.count ("fast_bootstrap") > 0);
-			daemon.run (data_path, flags);
+
+			nano_daemon::daemon daemon (data_path, flags);
+			on_shutdown = [&](int) {
```
wezrule: You should just capture what you need i.e `[&daemon]`

nano/node/ipc.cpp
```diff
@@ -407,6 +407,7 @@ void nano::ipc::ipc_server::stop ()
 	{
 		transport->stop ();
 	}
+	node.logger.always_log ("IPC: server stopped");
```
clemahieu: Can we also transports.clear () here so multiple calls to stop won't have an effect?

Issue #1135, opened 2018-08-31
Proposal: Compute work differently on Test and Beta networks


rkeene:
To ensure blocks from the various networks do not get propagated to nodes (i.e., beta network blocks sent to live network hosts) where they can never be validated, consider computing work differently on each network.

One possibility is:
```
uint64_t work = compute_work(work_hash ^ genesis_hash ^ live_network_genesis_hash, work_min);
bool valid = validate_work(work, work_hash ^ genesis_hash ^ live_network_genesis_hash);
```

rkeene:
Mitigations for this added in #1124

PR #1847, opened 2019-03-20
Compute work differently on Beta networks

SergiySW:
root = root ^ ~~genesis_account~~ network number (for non-live networks)
Advantages: malicious users cannot spam live network with beta blocks
Disadvantages: harder to use remote work generation for both live & beta networks

https://github.com/nanocurrency/nano-node/issues/1135


nano/node/json_handler.cpp
```diff
@@ -4452,7 +4452,8 @@ void nano::json_handler::work_generate ()
 		}
 		else
 		{
-			node.work_generate (hash, callback, difficulty);
+			// Double XOR with nano::node::work_generate () function
+			node.work_generate (node.network_params.network.is_live_network () ? hash : hash ^ static_cast<uint64_t> (node.network_params.network.current_network), callback, difficulty);
```
clemahieu: It seems like what's applied here would be undone inside work_generate.

Should the xoring be done in the wallet when work is queued?

SergiySW: Wallet also uses work_generate () & many tests contain work_generate_blocking etc.
So I guess easier to XOR in single function (work_generate)

But because "work_generate" RPC should receive already XORed value, XOR will be undone by work_generate. With extra XOR here we can pass root to work generator unchanged

clemahieu: Can we add an overload which accepts the 64-bit value to mix, which then calls the other work_generate overload? Then the RPC would just directly call the overload without the 64bit value.

SergiySW: Done

PR #1848, opened 2019-03-21
 rpc.online_reps test udpate

SergiySW:
* send to non existing account to prevent receiving & change of representative weight
* remove genesis representative test collision


wezrule:
Fixes the intermittent failures for me

PR #1849, opened 2019-03-22
qualified_root type alias

cryptocode:
A type alias in the same vein as nano::signature to improve readability and a new qualified_root 
 function in `block` to construct the value.

### DIFF
--- a/nano/core_test/conflicts.cpp
+++ b/nano/core_test/conflicts.cpp
@@ -17,7 +17,7 @@ TEST (conflicts, start_stop)
 	auto root1 (send1->root ());
 	{
 		std::lock_guard<std::mutex> guard (node1.active.mutex);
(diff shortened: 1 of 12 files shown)


nano/core_test/conflicts.cpp
```diff
@@ -44,7 +44,7 @@ TEST (conflicts, add_existing)
 	ASSERT_EQ (1, node1.active.size ());
 	{
 		std::lock_guard<std::mutex> guard (node1.active.mutex);
-		auto votes1 (node1.active.roots.find (nano::uint512_union (send2->previous (), send2->root ()))->election);
+		auto votes1 (node1.active.roots.find (nano::qualified_root (send2->previous (), send2->root ()))->election);
```
clemahieu: We should add a function nano::block::qualified_root so these are always constructed correctly.

cryptocode: Added function and updated applicable call sites

PR #1716, opened 2019-02-09
Add wallet_seed RPC

devinus:
An alternative to https://github.com/nanocurrency/nano-node/pull/949 by @SergiySW.

This only adds a `wallet_seed` action and includes a test.

My rationale can be found here: https://github.com/nanocurrency/nano-node/pull/949#issuecomment-462010711

PR #1851, opened 2019-03-23
Unsafe encoding in IPC

cryptocode:
Enables wallet_seed over a new IPC encoding. In addition to clients using the new encoding, a config flag must be set on each transport the client wants to use for this.

Based on #949 (only wallet_seed is used for now) and #1716

### DIFF
--- a/nano/rpc/rpc_handler.cpp
+++ b/nano/rpc/rpc_handler.cpp
@@ -4138,6 +4138,27 @@ void nano::rpc_handler::wallet_republish ()
 	response_errors ();
 }
 
+void nano::rpc_handler::wallet_seed ()
+{
+	rpc_control_impl ();
+	auto wallet (wallet_impl ());
+	if (!ec)
+	{
+		auto transaction (node.wallets.tx_begin_read ());
+		if (wallet->store.valid_password (transaction))
+		{
+			nano::raw_key seed;
+			wallet->store.seed (seed, transaction);
+			response_l.put ("seed", seed.data.to_string ());
+		}
+		else
+		{
+			ec = nano::error_common::wallet_locked;
+		}
+	}
+	response_errors ();
+}
+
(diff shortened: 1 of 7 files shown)


nano/rpc/rpc_handler.cpp
```diff
@@ -4137,6 +4137,27 @@ void nano::rpc_handler::wallet_republish ()
 	response_errors ();
 }
 
+void nano::rpc_handler::wallet_seed ()
```
clemahieu: Implementation 2?

nano/node/rpc.cpp
```diff
@@ -4235,6 +4235,27 @@ void nano::rpc_handler::wallet_republish ()
 	response_errors ();
 }
 
+void nano::rpc_handler::wallet_seed ()
```
clemahieu: Two implementations of this function?

nano/rpc/rpc_handler.cpp
cryptocode: rebase-fail (there was some duplication in the rpc refactoring I think. I believe that has been fixed, so I'll rebase it and fix)

zhyatt:
@cryptocode Do we need to make updates to the documentation here for this additional encoding: https://docs.nano.org/integration-guides/advanced/#ipc-requestresponse-format?

cryptocode:
@zhyatt I updated the wiki, but it doesn't seem to have been moved over - was probably done during the transition?

The text is here: https://github.com/nanocurrency/nano-node/wiki/IPC/473895c22b6f34bce5847bad0f7cab88e4a01845#ipc-requestresponse-format

Do you want me to make a doc PR of it?

zhyatt:
@cryptocode Thanks for clarifying. I have added in to my changes for the docs so no need for a separate PR.

PR #1855, opened 2019-03-24
Generalize installer path

### LINKED ISSUE #821: Windows: Old version is not removed after upgrading to a new release
**Description of bug:**
The windows installer of Nano does not replace or remove the old version of the node software. I had this problem on v11 -> v12 and v12 -> v12.1.
It's not a huge deal to deinstall it manually, it's just a bit annoying.

1. Install a old version of nano
2. Install the current one

Both version are installed afterwards. Present in the file system, start menu and app list.
Image
Image

Since it's the same software I expect that the old version is removed or upgraded.

**Environment**:
Windows 10, Version 1709

argakiig:
will require uninstall to remove old versions after
which new installs will use the same path
closes #821


CMakeLists.txt
```diff
@@ -418,7 +418,7 @@ if (NANO_GUI OR RAIBLOCKS_GUI)
 		set (CPACK_NSIS_MUI_ICON ${CMAKE_SOURCE_DIR}/Nano.ico)
 		set (CPACK_NSIS_DISPLAY_NAME "Nano")
 		set (CPACK_PACKAGE_NAME "Nano_Installer")
-		set (CPACK_NSIS_PACKAGE_NAME "Nano ${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}")
+		set (CPACK_NSIS_PACKAGE_NAME "Nanocurrency")
```
clemahieu: Can we use just Nano here?

argakiig: yup

PR #1856, opened 2019-03-24
Support release_assert failures in tests

cryptocode:
Replacement of https://github.com/nanocurrency/nano-node/pull/1117 which works with `network_params``

Not convinced the added dependency on gtest is worth it (release asserts should be very rare and easy to spot in tests), but putting it up for discussion.


cryptocode:
With the recent work done on reducing dependencies, adding gtest to nanolib seems like a non-starter. release_assert's are very rare and it doesn't seem like much of a problem if the test suit aborts on them - the source and location is still reported.

Issue #1827, opened 2019-03-14
Move RPC out of node process


zhyatt:
Remove RPC operations out of node process

zhyatt:
Resolved in #1874

PR #1857, opened 2019-03-25
Move RPC server to a new process

wezrule:
This solves #1827

The RPC server has now been moved to a new executable. If `rpc_enable` is set to true a new `rpc_path` config option will be added which points to the `nano_rpc` executable, and this will be launched as a child process of the `nano_node` executable. When `rpc_enable` is false the RPC executable will not be started, however it can be started manually with `nano_rpc --daemon` either before or after `nano_node --daemon`. In the event that the node crashes or needs upgrading but the RPC process doesn't, the RPC process can try and reconnect to the node on the next RPC command fired so it does not require restarting as well.

There are no API changes required as the current JSON requests are simply passed via IPC transport mechanism to be handled inside the node now. This required removing anything related to the node from the RPC process as they are now separate services and don't know about each other (theoretically).

A new json file has been created (rpc_config.json) for it and upon upgrade the settings from config.json are migrated. `enable_sign_hash` and `max_work_generate_difficulty` have been moved to the ipc config as the node itself needed to know about it rather than the RPC server. `log_rpc` has currently just been removed and no explicit logging config options have been added to the RPC server, it now has a log_rpc.*.log file in similar format to the existing log file.

The dependencies of a few things were a bit messed up, causing me all sorts of issues.
Circular dependency between and lib <-> ed25519. We declare our own functions in nano/lib/interface.cpp but ed25519 is dependent on this, but lib is also dependent on ed25519 functions, so to resolve this I have created a new library crypto_lib for lack of a better name, which includes the interface.cpp functions. Annoyingly it only showed on travis linux, and I could not reproduce it locally with MSVC, Mac/Clang and Ubuntu Clang Gcc versions 5.4 (used by travis), 5.5 or 7.3. It's not clear why some executables required the ed25519 library when linking in the new crypto_lib when they don't use the ed25519 functions (static libs should only pull in symbols they need). I ended up just adding the link libraries in each executable to resolve the travis Linux linker errors. There were also circular references with the `secure` and `lib` libraries which has been refactored out as well.

A few IPC connections are made, defaulting to 8 on the live network which is user configurable with the rpc_config.json option "num_ipc_connections". The new class `rpc_request_processor` queues up requests and dispatches them to any available connections.

Notes:
- At the time of writing there there are some TSAN issues with udp_channel which can affect this, most noticeably the `nano-load-test` which makes simultaneous requests, it worked fine before merging but something to keep in mind.
- log_ipc has been disabled in the load-tests as there was an assertion when logging ipc calls in timer::stop () `assert (state == nano::timer_state::started);`. Not sure if this is related to the above problem or something completely different.
- rpc_secure.cpp didn't compile before and some changes were made to make it compile, but have not been tested

When testing:
Try stop RPC command. It should close down both nano-node and rpc programs. If the RPC server is started successfuly and the node goes down, restarting the node should allow the RPC server to connect to it again (it started separately).
Try it with the gui wallet.
Try config changes and upgrades with both node and gui.
Test with secure RPC (-DNANO_SECURE_RPC=ON)


cryptocode:
Seems like all handlers are now in `ipc.cpp` file. In my opinion, IPC's responsibility should be restricted to be a gateway over transports and manage those connections.

Would it be possible to factor out all the handler stuff and detach that from IPC? json_handler.cpp or something like that, and then IPC forwards the request there if it has the proper encoding.

In the future we may have non-JSON gateways, like protobuf, flatbuffer and whatnot.

nano/rpc/rpc_handler.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+	std::stringstream ss;
+	boost::property_tree::write_json (ss, response_l);
+	response_a (ss.str ());
+}
+}
+
+namespace
+{
+std::unordered_set<std::string> create_rpc_control_impls ()
```
cryptocode: I wonder if moving this to the RPC process is problematic, because there are/will be other IPC clients. Instead of controlling RPCs, I suspect we should control *actions* in the node. Then it doesn't matter how the action is called (rpc, raw ipc, future protobufs, ec). 

I'm thinking the external RPC process should do as little such logic as possible - basically be Web server frontend to IPC.

nano/node/common.cpp
```diff
@@ -207,7 +207,7 @@ status (parse_status::success)
 
 void nano::message_parser::deserialize_buffer (uint8_t const * buffer_a, size_t size_a)
 {
-	static nano::network_params network_params;
+	static nano::network_constants network_constants;
```
cryptocode: Is the network_params changes unrelated to this PR and possible to put in a separate PR? Not important, but could reduce the size of this patch a bit if that's the case.

wezrule:
This was split in various PRs, the main functionality being in: #1857

PR #1858, opened 2019-03-25
Trending active_transactions difficulty for an average

argakiig:
Provide active_transactions.active_difficulty for 5 min trended avg of adjusted_difficulty in active_transactions 
alternative to #1725


argakiig:
This is part of #1336

nano/core_test/node.cpp
```diff
@@ -2428,6 +2428,47 @@ TEST (node, unchecked_cleanup)
...
+	nano::work_validate (*send1, &difficulty1);
+	// Process as local block
+	node1.process_active (send1);
+	system.deadline_set (2s);
+	while (node1.active.empty ())
+	{
+		ASSERT_NO_ERROR (system.poll ());
+	}
+	std::unique_lock<std::mutex> lock (node1.active.mutex);
+	auto existing (node1.active.roots.find (nano::uint512_union (send1->previous (), send1->root ())));
```
cryptocode: `send1->qualified_root ()` can be used now now

nano/node/node.cpp
```diff
@@ -3280,6 +3281,39 @@ void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash
...
+	// Rough average difficulty estimation
+	switch (size_l)
+	{
+		case 0:
+			break;
+		case 1:
+			difficulty = roots.get<1> ().begin ()->adjusted_difficulty;
+			break;
+		default:
+			difficulty = roots.get<1> ().find ((roots.get<1> ().begin ()->adjusted_difficulty + (--roots.get<1> ().end ())->adjusted_difficulty) / 2)->adjusted_difficulty;
```
clemahieu: This whole switch statement can be replaced with an if(!roots.empty()) check and then this line inside the block.

nano/node/node.cpp
```diff
@@ -3280,6 +3281,39 @@ void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash
...
+			break;
+	}
+	difficulty_cb.push_front (difficulty);
+	{
```
clemahieu: This scoping isn't needed.

nano/node/node.cpp
```diff
@@ -3328,6 +3362,7 @@ nano::active_transactions::active_transactions (nano::node & node_a) :
 node (node_a),
 started (false),
 stopped (false),
+difficulty_cb (20),
```
clemahieu: Can we initialize all the values with the publish_threshold with this constructor https://www.boost.org/doc/libs/1_61_0/doc/html/boost/circular_buffer.html#idp9066352-bb

Then we'll always have a minimum of the publish_threshold on all the samples.

nano/node/node.cpp
```diff
@@ -3280,6 +3281,39 @@ void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash
...
+	}
+	difficulty_cb.push_front (difficulty);
+	{
+		uint64_t result (node.network_params.publish_threshold);
+		uint128_t sum (0);
+		for (auto & i : difficulty_cb)
+		{
+			sum += i;
+		}
+		if (!difficulty_cb.empty ())
```
clemahieu: We don't need to check if the circular buffer is empty, it never will be.

nano/node/node.cpp
```diff
@@ -3280,6 +3281,39 @@ void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash
...
+		case 1:
+			difficulty = roots.get<1> ().begin ()->adjusted_difficulty;
+			break;
+		default:
+			difficulty = roots.get<1> ().find ((roots.get<1> ().begin ()->adjusted_difficulty + (--roots.get<1> ().end ())->adjusted_difficulty) / 2)->adjusted_difficulty;
+			break;
+	}
+	difficulty_cb.push_front (difficulty);
+	{
+		uint64_t result (node.network_params.publish_threshold);
```
clemahieu: We don't really need this, we can just assign to active_difficulty directly when dividing sum down.

nano/core_test/node.cpp
argakiig: cleaned up

nano/node/node.cpp
argakiig: simplified, thanks I kept forgetting

nano/node/node.cpp
argakiig: removed

nano/node/node.cpp
argakiig: initialized with the threshold

nano/node/node.cpp
argakiig: cleaned up

nano/node/node.cpp
argakiig: cleaned up

nano/node/node.cpp
```diff
@@ -3280,6 +3281,24 @@ void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash
...
 
+void nano::active_transactions::update_active_difficulty ()
+{
+	assert (!mutex.try_lock ());
+	uint64_t difficulty (node.network_params.publish_threshold);
+	if (!roots.empty ())
+	{
+		difficulty = roots.get<1> ().find ((roots.get<1> ().begin ()->adjusted_difficulty + (--roots.get<1> ().end ())->adjusted_difficulty) / 2)->adjusted_difficulty;
+	}
+	difficulty_cb.push_front (difficulty);
```
clemahieu: We should assert here that difficulty >= publish_threshold.

nano/node/node.cpp
```diff
@@ -3280,6 +3281,24 @@ void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash
...
+	uint128_t sum (0);
+	for (auto & i : difficulty_cb)
+	{
+		sum += i;
+	}
+	active_difficulty = static_cast<uint64_t> (sum / difficulty_cb.size ());
+	;
```
clemahieu: Extra semicolon.

nano/core_test/node.cpp
```diff
@@ -2428,6 +2428,47 @@ TEST (node, unchecked_cleanup)
...
+	while (node1.active.empty ())
+	{
+		ASSERT_NO_ERROR (system.poll ());
+	}
+	std::unique_lock<std::mutex> lock (node1.active.mutex);
+	auto existing (node1.active.roots.find (send1->qualified_root ()));
+	ASSERT_NE (node1.active.roots.end (), existing);
+	auto election (existing->election);
+	// Fake history records to force work recalculation
+	for (auto i (0); i < 20; i++)
```
clemahieu: Lets test the active_difficulty numerically before and this loop in addition to testing the side effect of higher difficulty generation.

clemahieu: This should count up to the size of the circular buffer in case we change the size later.

nano/node/node.hpp
```diff
@@ -151,6 +152,8 @@ class active_transactions
 	static unsigned constexpr announcement_long = 20;
 	static size_t constexpr election_history_size = 2048;
 	static size_t constexpr max_broadcast_queue = 1000;
+	std::atomic<uint64_t> active_difficulty;
```
wezrule: Is there an appropriate default to initialize this with?

nano/node/node.cpp
```diff
@@ -3280,6 +3281,24 @@ void nano::active_transactions::adjust_difficulty (nano::block_hash const & hash
...
+	uint64_t difficulty (node.network_params.publish_threshold);
+	if (!roots.empty ())
+	{
+		difficulty = roots.get<1> ().find ((roots.get<1> ().begin ()->adjusted_difficulty + (--roots.get<1> ().end ())->adjusted_difficulty) / 2)->adjusted_difficulty;
+	}
+	difficulty_cb.push_front (difficulty);
+	uint128_t sum (0);
+	for (auto & i : difficulty_cb)
+	{
+		sum += i;
```
wezrule: This can be replaced with the `std::accumulate` algorithm

nano/node/node.hpp
```diff
@@ -151,6 +152,8 @@ class active_transactions
...
+	boost::circular_buffer<uint64_t> difficulty_cb;
```
wezrule: This member is declared before started/stopped in the class but is initialized after them in the constructor member initialization list. This will generate warnings on some compilers as they should be in the same order. Easiest way to prevent this (and save duplication/extra lines) is to just initialize default values in-class.

nano/core_test/node.cpp
```diff
@@ -2428,6 +2428,47 @@ TEST (node, unchecked_cleanup)
...
+	system.deadline_set (5s);
+	while (!done)
+	{
+		lock.lock ();
+		node1.work_generate_blocking (*send1);
+		uint64_t difficulty2;
+		nano::work_validate (*send1, &difficulty2);
+		done = (difficulty2 > difficulty1);
+		lock.unlock ();
+		ASSERT_NO_ERROR (system.poll ());
```
wezrule: Is this `system.poll ()` necessary? (and deadline_set above), I don't think we are relying on the io thread here.

nano/node/node.cpp
cryptocode: Does that work with boost's uint128_t? I've tried that in the past and failed. Very nice if there's a way.

wezrule: `auto sum = std::accumulate (difficulty_cb.begin (), difficulty_cb.end (), uint128_t (0));` should work. Perhaps you just used a plain 0, std::accumulate does it's automatic type deduction based on the `init` variable.

cryptocode: The init type could well've been wrong. Cool.

PR #1860, opened 2019-03-25
recalculate work for long unconfirmed local blocks

argakiig:
using a difficulty threshold of the long unconfirmed block
#1858 allows for recalculating with a threshold of active_difficulty instead
part of #1336

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -2435,7 +2435,48 @@ TEST (node, unchecked_cleanup)
 	}
 }
 
-TEST (active_difficulty, recalculate_work)
+TEST (active_difficulty, recalculate_low_work)
+{
+	{
+		nano::system system (24000, 1);
+		auto & node1 (*system.nodes[0]);
+		nano::genesis genesis;
+		nano::keypair key1;
+		auto send1 (std::make_shared<nano::send_block> (genesis.hash (), key1.pub, 0, nano::test_genesis_key.prv, nano::test_genesis_key.pub, 0));
+		node1.work_generate_blocking (*send1);
+		uint64_t difficulty1;
+		nano::work_validate (*send1, &difficulty1);
+		// Process as local block
+		node1.process_active (send1, true);
+		system.deadline_set (2s);
+		while (node1.active.empty ())
+		{
+			ASSERT_NO_ERROR (system.poll ());
+		}
+		std::unique_lock<std::mutex> lock (node1.active.mutex);
+		auto existing (node1.active.roots.find (nano::uint512_union (send1->previous (), send1->root ())));
+		ASSERT_NE (node1.active.roots.end (), existing);
+		auto election (existing->election);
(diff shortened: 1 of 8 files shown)


nano/core_test/node.cpp
```diff
@@ -2428,6 +2428,41 @@ TEST (node, unchecked_cleanup)
...
+	nano::work_validate (*send1, &difficulty1);
+	// Process as local block
+	node1.process_active (send1, true);
+	system.deadline_set (2s);
+	while (node1.active.empty ())
+	{
+		ASSERT_NO_ERROR (system.poll ());
+	}
+	std::unique_lock<std::mutex> lock (node1.active.mutex);
+	auto existing (node1.active.roots.find (nano::uint512_union (send1->previous (), send1->root ())));
```
cryptocode: `send1->qualified_root ()`

argakiig:
rebased to dd46609 and adjusted to calculate new work above active_difficulty every announcement period.

nano/node/node.cpp
```diff
@@ -2906,6 +2907,28 @@ void nano::active_transactions::request_confirm (std::unique_lock<std::mutex> &
...
+							if (work_a && !nano::work_validate (block_l->root (), work_a.get ()))
+							{
+								block_l->block_work_set (work_a.get ());
+								node_l->active.update_difficulty (*block_l);
+								node_l->network.flood_block (block_l);
+								election_l->ongoing_work_update = false;
+							}
+						};
+						node.work_generate (block_l->root (), callback, difficulty);
+						election_l->ongoing_work_update = true;
```
wezrule: It seems possible (although very unlikely) that the work is calculated and the callback gets called and sets `election_l->ongoing_work_update = false;` before `election_l->ongoing_work_update = true;` is set as they will be on different threads. Will that cause any issues?

argakiig: it can be set after diff is checked against active_difficulty

wezrule: looks good

nano/node/node.cpp
```diff
@@ -2906,6 +2907,28 @@ void nano::active_transactions::request_confirm (std::unique_lock<std::mutex> &
...
+					uint64_t difficulty (0);
+					nano::work_validate (*block_l, &difficulty);
+					if (difficulty < active_difficulty)
+					{
+						election_l->ongoing_work_update = true;
+						auto node_l (node.shared ());
+						auto callback = [node_l, block_l, election_l](boost::optional<uint64_t> const & work_a) {
+							if (work_a && !nano::work_validate (block_l->root (), work_a.get ()))
+							{
+								block_l->block_work_set (work_a.get ());
```
cryptocode: This updates a block's work field without a lock / it's not atomic.
Because of the block uniquer, I believe the same block can be accessed or have its work validated by other threads (such as the same block arriving again over bootstrap or realtime traffic)

PR #1867, opened 2019-03-27
Remove circular dependency between secure <-> lib libraries

wezrule:
This PR removes any mention of secure in the lib library. The secure library has a few dependencies on the lib library, while the `lib` library currently has a dependency on the `keypair` class which is part of the secure library (which also currently has circular dependencies with the node library), this was part of a larger class `network_params`. A lot of this code used to be in the secure library before, so I've just moved it back. However it is necessary to know logic like which network we are on (test/live/beta) in the future RPC stuff so a new class network_constants has been created which incorporates this in the lib library.


wezrule:
Thanks!

PR #1872, opened 2019-03-27
Refactor preparation for moving RPC out of process

wezrule:
Move RPC classes to a new subfolder nano/rpc and also create a few different files to hold the related clases in. It currently depends on the `node` library (and vice versa) but this dependency will be removed in the main RPC task. Also moved the rpc tests to their own executable.
Extracted a few other things like thread_runner, logging, ipc_client/transport etc.. to the lib library to make them accessible to other libraries which won't have access to the `node` library (e.g the new RPC server process).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@
 *.DS_Store
 core_test
 !core_test/
+rpc_test
+!rpc_test/
 qt_test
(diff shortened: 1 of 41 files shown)


wezrule:
Thanks :)

Issue #1750, opened 2019-02-19
Investigate high memory usage


cryptocode:
Some RPC's takes a ton of memory, i.e. "ledger". Probably due to serializing ptree to string before writing the response. Investigate if it's possible to stream ptree straight to socket.
RPC and IPC uses the same technique and a solution should apply both places.

renesq:
Can confirm, on a 4GB RAM VPS, these commands will crash the node due to lack of free memory:

`curl -o "ledger.txt" -g -d '{ "action": "ledger", "account": "xrb_1111111111111111111111111111111111111111111111111111hifc8npp", "count": "-1"}' '[::1]:7076'
curl -o "binance.txt" -g -d '{ "action": "account_history", "account": "xrb_3jwrszth46rk1mu7rmb4rhm54us8yg1gw3ipodftqtikf5yqdyr7471nsg1k", "count": "-1"}' '[::1]:7076'`

CURL responds `Empty reply from server` and the node dies completely. 
Up until 150k-220k entries it will work fine on this machine. 

Node should abort the RPC query when running out of RAM, and give a error message - maybe even tell the user how many entries it could process so far, but not putting those out .

cryptocode:
The reason for highmem usage in RPC's like "ledger" is understood and there's no easy fix. The upcoming JSON library switch (most likely v20) will help a bit, but the endgame is either a better paging API or streamy responses. Thus moving to v20.

PR #1874, opened 2019-03-29
RPC server is moved to a new process

wezrule:
This solves #1827

The RPC server has now been moved to a new executable. If `rpc_enable` is set to true a new `rpc_path` config option will be added which points to the `nano_rpc` executable, and this will be launched as a child process of the `nano_node` executable. When `rpc_enable` is false the RPC process will not be started, however it can be started manually with `nano_rpc --daemon` either before or after `nano_node --daemon` or starting the gui wallet. In the event that the node crashes or needs upgrading but the RPC process doesn't, the RPC process can try and reconnect to the node on the next RPC command fired so it does not require restarting as well.

There are no API changes required as the current JSON requests are simply passed via IPC transport mechanism to be handled inside the node now. This required removing anything related to the node from the RPC process (which has been done largely in other PRs) as they are now separate services and don't know about each other (theoretically).

A new json file has been created (rpc_config.json) for it and upon upgrade the settings from config.json are migrated. `enable_sign_hash` and `max_work_generate_difficulty` have been moved to the ipc config as the node itself needs to know about it deep in the logic. `log_rpc` has currently just been removed and no explicit logging config options (can be added in the future if required) have been added to the RPC server, it now has a log_rpc.*.log file in similar format to the existing log file.

A few IPC connections are made, defaulting to 8 on the live network which is user configurable with the rpc_config.json option `num_ipc_connections`. The new class `rpc_request_processor` queues up requests and dispatches them to any available connections.

When testing:
Try the `stop` RPC command, it should close down both nano_node and nano_rpc processes, whether they are separate or as a child process with `rpc_enable`. If the RPC server is started successfuly and the node goes down, restarting the node should allow the RPC server to connect to it again when a new RPC command is issued.
Try the RPC server with the gui wallet.
Try config changes and upgrades with both node and gui.
Check that the new log file is working as expected.

There are now 3 different ways to run the RPC server:
**In process**
- rpc_enable = **true**
- rpc_in_process = **true** (default)

**Child process**
- rpc_enable = **true**
- rpc_path = *path to nano_rpc*
- rpc_in_process = **false**
- ipc -> tcp -> enable = **true**
- ipc -> tcp -> port == **ipc_port** of rpc_config.json

**Out of process**
As above but rpc_enable = **false**


cryptocode:
Seems like I'm hitting https://github.com/boostorg/process/issues/55 on macOS/latest clang/boost 1.69

wezrule:
> Seems like I'm hitting [boostorg/process#55](https://github.com/boostorg/process/issues/55) on macOS/latest clang/boost 1.69

Well that's annoying. I was able to reproduce it (works ok with 1.68), according to that link it will be fixed in 1.70. There is actually a std::system () function call but it is only synchronous and the standard doesn't really have a concept of process yet, so there's no way to explicitly terminate it. This becomes a problem with the nano_wallet when closed by the user as the child process continues to exist. I could send an HTTP request to the RPC server from the wallet, but it currently doesn't know about the rpc_config.json file so would require a lot more changes. I've now set conditional compilation depending on the boost version, so it uses the boost interprocess stuff if not using 1.69 otherwise if rpc_enable is true it errors out with a warning which states to start the RPC process manually. I've not done this for nano_node which is either forcefully closed from the command line or closed using stop () RPC command, which seems to clean up both processes properly, but it's possible it could have issues as well in some circumstances.

cryptocode:
Given how hard cross-platform process mgmt can be to get right, I wonder if (at least for a version or two) we should offer a fallback to the old in-process RPC  🤔 Is that hard to get going architecturally? If the core of the externalized RPC was a shared library, the node could link it but only enable if config says "inprocess" or something like that. Once external RPC and IPC is battle tested, we can remove the inprocess stuff. 

Another reason for an inprocess option is that we end up using more memory because two processes now holds data during the reques. This might be a problem on small VPS's when the RPC process runs on the same machine as the node (there have been some reports on Discord about high memory usage, and there's https://github.com/nanocurrency/nano-node/issues/1750)

nano/nano_wallet/entry.cpp
```diff
@@ -10,10 +10,15 @@
 #include <nano/qt/qt.hpp>
 
 #include <boost/make_shared.hpp>
-#include <boost/process.hpp>
 #include <boost/program_options.hpp>
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
+#include <boost/version.hpp>
+
+#define USE_BOOST_PROCESS BOOST_VERSION != 106900
```
cryptocode: Does this problem happen on anything but macOS? If it's mac only, we could add that to the constraint so that 1.69 works on Linux (which most services use) and Windows.

wezrule: This is just for the gui wallet, I don't think many people build this from source do they? (could be wrong) so I played it safe as didn't check all platforms to see whether it failed to build with boost 1.69.

cryptocode: Oh ok, so daemon doesn't have the same check (IOW daemon won't compile with 1.69 on macOS?)

wezrule: daemon will compile as it doesn't use the boost process, it does things a bit differently (see my first comment). It has to do with proper shutdown of the child RPC process. When clicking the X (close) button on the gui wallet, we have no way to terminate the RPC process as the standard doesn't know about processes yet, so it leaves dangling invisible processes. We could probably set up a connection with the RPC process from the daemon and send the stop RPC command, but I just prevent the child process being opened if using boost 1.69 and say they have to start it separately. With the node daemon, it is either closed by sending `stop` RPC command or forcefully quitting, which seems to clean up the child process, although I can imagine in some circumstances this might not do it correctly.

cryptocode: OK, I noticed the #include <boost/process.hpp> in the daemon while reviewing. I haven't compiled the latest yet - but the include is needed?

wezrule: it should be removed in the latest (I can't see it at least), including that header or pretty much any sub-header will indeed fail to compile on mac with 1.69

wezrule: Boost interprocess stuff works with Windows + 1.69 so I have added an additional compilation check and allow it for windows. Have raised #1879 to check for linux. I also found under some circumstances (like forcefully exiting a debug session), the child rpc process is not cleaned up on windows with the CLI daemon, so I will add similar checks there as well (after #1879) to know which platforms are supported.

wezrule:
> Putting this up for debate:
> 
> Given how hard cross-platform process mgmt can be to get right, I wonder if (at least for a version or two) we should offer a fallback to the old in-process RPC 🤔 Is that hard to get going architecturally? If the core of the externalized RPC was a shared library, the node could link it but only enable if config says "inprocess" or something like that. Alternatively, keep the original RPC connection code (default disabled) while keeping the external RPC as-is (probably the least amount of headache, but probably some things to think about config-wise)
> 
> Once external RPC and IPC is battle tested, we can remove the inprocess stuff. Of course the inprocess variation would go straight to rpc handlers, not via IPC.
> 
> Another reason for an inprocess option is that we end up using more memory because two processes now holds data during the requests. This might be a problem on small VPS's when the RPC process runs on the same machine as the node (there have been some reports on Discord about high memory usage, and there's #1750)

I have added an RPC in_process option now. @SergiySW commented in another PR that filenames shouldn't have underscores _ so I have changed all RPC related files to remove them. I no longer update the daemon config when migrating the RPC, as it doesn't make sense to put the options in the IPC anymore, I have stripped down the old RPC config and reused that instead. It does mean that the IPC port is not migrated to the new rpc_config.json file, but I don't think this is much of an issue as the same default port will be used and we are defaulting to in-process now so users should have no issues unless they explicitly want to use the out of process RPC which is an active choice (it saves a lot of complexity for little gain imo). I've appended some information to the end of this PR as well.

load-tester/src/launch_node_and_rpc.rs
```diff
@@ -0,0 +1,182 @@
...
+    let ipc_port = IPC_PORT_START + i;
+
+    let config = json!({
+    "version": "2",
+    "rpc_enable": "false",
+    "rpc": {
+        "version": "1",
+        "enable_sign_hash": "false",
+        "max_work_generate_difficulty": "ffffffffc0000000",
+        "rpc_path": "C:\\Users\\Wesley\\Documents\\nano-node\\build\\Debug\\nano_rpc.exe",
```
cryptocode: Probably want a placeholder here

nano/node/noderpcconfig.cpp
```diff
@@ -0,0 +1,77 @@
...
+		migrate (json, data_path);
+
+		json.erase ("io_threads");
+
+		json.put ("enable_sign_hash", enable_sign_hash);
+		json.put ("max_work_generate_difficulty", nano::to_string_hex (max_work_generate_difficulty));
+		json.put ("rpc_path", get_default_rpc_filepath ());
+		bool rpc_in_process_l;
+		json.get_optional<bool> ("rpc_in_process", rpc_in_process_l);
+		if (!rpc_in_process)
```
cryptocode: I noticed `rpc_in_process` doesn't end up in the upgraded json file. If that was the intention here, maybe use the single-parameter `get_optional` overload which returns an boost::optional. If a value is not set, then write it out.

nano/node/noderpcconfig.cpp
```diff
@@ -0,0 +1,77 @@
...
+	{
+		nano::from_string_hex (max_work_generate_difficulty_text, max_work_generate_difficulty);
+	}
+	json.get_optional<std::string> ("rpc_path", rpc_path);
+	json.get_optional<bool> ("rpc_in_process", rpc_in_process);
+	return json.get_error ();
+}
+
+void nano::node_rpc_config::migrate (nano::jsonconfig & json, boost::filesystem::path const & data_path)
+{
```
cryptocode: Should `migrate` remove unused json properties, like "port", from the old config? 

By the way, I had a non-standard RPC port number in config.json, but the new rpc_config.json contains the default 7076 after the upgrade.

nano/nano_node/entry.cpp
```diff
@@ -716,7 +717,7 @@ int main (int argc, char * const * argv)
 				std::cout.write (reinterpret_cast<const char *> (seed.data.bytes.data ()), seed.data.bytes.size ());
 			}
 		}
-		else if (vm.count ("debug_rpc"))
+		else if (vm.count ("debug_ipc"))
```
cryptocode: This change broke a script I have. `--debug_rpc` was a v18 feature so we probably shouldn't change it. I feel `debug_rpc ` is the correct option name here since it receives an RPC json string, not an IPC sequence.

nano/nano_rpc/entry.cpp
```diff
@@ -0,0 +1,139 @@
...
+{
+	nano::set_umask ();
+
+	boost::program_options::options_description description ("Command line options");
+
+	// clang-format off
+	description.add_options ()
+		("help", "Print out options")
+		("version", "Prints out version")
+		("daemon", "Start node daemon")
```
cryptocode: Should probably say "Start RPC deamon" so people don't think this will start a node when you start the rpc process.

nano/lib/rpcconfig.cpp
```diff
@@ -44,44 +47,93 @@ nano::error nano::rpc_config::serialize_json (nano::jsonconfig & json) const
 	json.put ("port", port);
 	json.put ("enable_control", enable_control);
 	json.put ("max_json_depth", max_json_depth);
-	json.put ("enable_sign_hash", enable_sign_hash);
 	json.put ("max_request_size", max_request_size);
-	json.put ("max_work_generate_difficulty", nano::to_string_hex (max_work_generate_difficulty));
+	json.put ("io_threads", io_threads);
+	json.put ("ipc_port", ipc_port);
+	json.put ("ipc_path", ipc_path);
```
cryptocode: Is this used? I don't think the external RPC process supports using domain sockets yet.

nano/rpc/rpc.cpp
```diff
@@ -1,62 +1,50 @@
...
 	}
+	acceptor.listen ();
+	accept ();
 }
 
 void nano::rpc::accept ()
 {
-	auto connection (std::make_shared<nano::rpc_connection> (node, *this));
+	auto connection (std::make_shared<nano::rpc_connection> (config, network_constants, io_ctx, logger, rpc_handler_interface));
 	acceptor.async_accept (connection->socket, [this, connection](boost::system::error_code const & ec) {
```
cryptocode: Ideally the external RPC process should establish persistent connection with its HTTP clients to reduce file descriptor use and improve performance. Now it's one connection per call (i.e same as before). 

One of the benefits of IPC is persistent connections and it would be nice if that extended to http clients. Not sure if such a change is in scope for v19 though.

cryptocode:
Both in-process (currently default which I think is correct) and out-of-process is working great on macOS. Would be nice to ship with boost 1.70 so we could get the process mgmt fix for macOS as well (must start manually now), but not sure if it'll be released in time.

load-tester/src/launch_node_and_rpc.rs
wezrule: Yeh that looks weird :D, it's not actually used here as we are using out of process RPC (passed in via command line), so I've just removed it.

nano/node/noderpcconfig.cpp
wezrule: Yes it should! Thanks, done.

I wonder if you still had "version" : "1" in the old rpc config (I have made that mistake). We have done an upgrade to it this release which I am hijacking, so it won't upgrade correctly if the version is present.

nano/node/noderpcconfig.cpp
wezrule: Yeh was using the wrong constructs, thanks.

nano/nano_node/entry.cpp
wezrule: ok done

nano/node/noderpcconfig.cpp
cryptocode: I removed the version field as part of the test.

nano/lib/rpcconfig.cpp
wezrule: Nope was an artifact when I was also migrating the values in the ipc config. Removed

nano/node/noderpcconfig.cpp
wezrule: hmm and the rpc_config.json file was deleted beforehand? The port seems to migrate correctly for me, perhaps you could try again after I push all the changes.

nano/nano_rpc/entry.cpp
wezrule: yep :D, thanks.

nano/rpc/rpc.cpp
wezrule: Not sure it is possible to maintain connections with HTTP as it is stateless, but I think it would be the responsibility for another PR regardless (> v19).

nano/node/noderpcconfig.cpp
cryptocode: yeah, will do, i'll try migrating from a proper v18 installation, seems like a more reliable way to test

cryptocode:
Upgraded a v18 installation using a non-default rpc port. Seems like the created `rpc_config.json` always contains the default rpc port (7076) - the same applies to address and possibly other settings.

nano/node/node_rpc_config.cpp
```diff
@@ -0,0 +1,82 @@
...
+	json.get_optional<std::string> ("rpc_path", rpc_path);
+	json.get_optional<bool> ("rpc_in_process", rpc_in_process);
+	return json.get_error ();
+}
+
+void nano::node_rpc_config::migrate (nano::jsonconfig & json, boost::filesystem::path const & data_path)
+{
+	// Copy RPC config if the file doesn't exist already
+	nano::network_constants network_constants;
+	auto default_ipc_port = network_constants.is_live_network () ? 7077 : 24077;
```
cryptocode: default_ipc_port was recently added to network constants (https://github.com/nanocurrency/nano-node/pull/1891)
(Note that it differentiates between beta and test to allow tests to run while running a beta node.)

wezrule: good spot. That looks like something that was left over after I did some restructuring, the variable is not actually used so I have removed it, thanks

nano/nano_node/entry.cpp
```diff
@@ -95,13 +96,13 @@ int main (int argc, char * const * argv)
...
 		("debug_random_feed", "Generates output to RNG test suites")
-		("debug_rpc", "Read an RPC command from stdin and invoke it. Network operations will have no effect.")
 		("debug_validate_blocks", "Check all blocks for correct hash, signature, work value")
 		("debug_peers", "Display peer IPv6:port connections")
+		("debug_ipc", "Read an IPC command in JSON from stdin and invoke it. Network operations will have no effect")
 		("platform", boost::program_options::value<std::string> (), "Defines the <platform> for OpenCL commands")
 		("device", boost::program_options::value<std::string> (), "Defines <device> for OpenCL command")
 		("threads", boost::program_options::value<std::string> (), "Defines <threads> count for OpenCL command")
-		("difficulty", boost::program_options::value<std::string> (), "Defines <difficulty> for OpenCL command, HEX")
+		("difficulty", boost::program_options::value<std::string> (), "Defines <difficulty> for OpenCL command, HEX");
```
clemahieu: Is the semicolon at the end of this making the next line have no effect, does the pow_sleep_interval switch still work?

wezrule: nope it breaks it!
Good spot. Thanks

SergiySW:
RPC docs update :)

PR #1875, opened 2019-03-29
When clearing confirmation heights set the genesis account to be 1

wezrule:
In mdb_store::initialize we now set the genesis account's open block as being confirmed (which makes sense). When clearing the confirmation heights we probably want the genesis account to go to 1 rather than 0 which is what this PR does.


wezrule:
whoops, closed in error

PR #1877, opened 2019-04-02
Update confirmation height in another thread

wezrule:
Currently the confirmation height is updated on io threads, and if it touches a long chain it can take a very long time (30 minutes) because it can update pretty much the whole ledger. A different approach is now taken. When a block being confirmed, we go down to the lowest block in the chain which has not been confirmed and store each receive block along the way. We now start at the lowest (before it would have been the highest) receive and iterate to the source accounts until we get to a chain where there are no more unconfirmed receives, storing this in a collection of pending writes , working our way from the bottom upwards.

API changes:
**RPC** stats -> counters. `confirmed_block_count` shows the number of blocks confirmed during node start-up
**RPC** stats -> objects. `confirmation_height_processor` -> `receive_source_pairs`. The current number of receive-source pairs that still await iterative confirmation for the current block being confirmed. `pending_confirmation_height` -> `pending` - The number of top level block hashes in the queue awaiting confirmation
**RPC** `{"action":"confirmation_height_currently_processing"}` this will return the hash if any which is currently being processed. To be marked as unstable/debug.
**CLI**: `--debug_cemented_block_count` iterates over all account chains and sums up the confirmation heights (can take a little bit). For the first time the node is launched this will be 1 more than that in stats -> counters due to the implicitly confirmed genesis block.


nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,238 @@
...
+		auto error = store.account_get (transaction, pending.first, account_info);
+		release_assert (!error);
+		if (pending.second.height > account_info.confirmation_height)
+		{
+			// Check that the confirmation height and block hash still match, as there
+			// may have been changes outside this processor
+			nano::block_sideband sideband;
+			auto block = store.block_get (transaction, pending.second.hash, &sideband);
+			release_assert (block != nullptr);
+			release_assert (sideband.height == pending.second.height);
```
cryptocode: This release_assert hit after a few minutes on a live ledger that was almost in sync.

`sideband.height` == 10
`pending.second.height` == 11

Block hash 66B06DDC0D40AC86522CEEF5747FDFE7DD503B1D108B0B00432FB5E714F8BEC5

wezrule: The off by one appears to originate from the `confirmation_height + num_to_confirm` logic, which was done as an optimization to prevent having to get the sideband data from the block. I can't think why it would be an issue in some cases though, but I'm just going to use the sideband data directly. I'm also doing another pessimistic approach as in-between the reads and writes of the confirmation heights the chains could be altered possibly forking and removing blocks I was in the process of confirming, or perhaps this is a non-issue?

nano/node/CMakeLists.txt
```diff
@@ -21,6 +21,8 @@ add_library (node
 	cli.cpp
 	common.cpp
 	common.hpp
+	confirmation_height_processor.hpp
```
SergiySW: Probably consistent name for files without _

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,254 @@
...
+	return false;
+}
+
+void nano::confirmation_height_processor::collect_unconfirmed_receive_and_sources_for_account (uint64_t block_height, uint64_t confirmation_height, nano::block_hash & current, const nano::block_hash & genesis_hash, std::stack<open_receive_source_pair> & open_receive_source_pairs, nano::account const & account, nano::transaction & transaction)
+{
+	// Get the last confirmed block in this account chain
+	auto num_to_confirm = block_height - confirmation_height;
+	while (num_to_confirm > 0 && !current.is_zero ())
+	{
+		active.confirm_block (current);
```
SergiySW: Why make votes for each confirmed block? For all 22 millions?

SergiySW: All corerect

nano/node/CMakeLists.txt
wezrule: Thanks, changed

cryptocode: I don't think confirmationheightprocessor.cpp is readable. We use underscores a few other places already, like in tests and ipc_client.hpp in the lib.

nano/node/node.cpp
```diff
@@ -2268,7 +2271,7 @@ void nano::node::process_confirmed (std::shared_ptr<nano::block> block_a, uint8_
 	auto hash (block_a->hash ());
 	if (ledger.block_exists (block_a->type (), hash))
 	{
-		add_confirmation_heights (hash);
+		confirmation_height_processor.add (hash);
```
cryptocode: In the previous version, I believe this blocked until confirmation height was set. Now it schedules the block for processing. 

The confirmation callbacks may thus occur before the height is set. Is that a potential issue for callback receivers?  E.g using `block_info` to double check confirmation status.

During the initial confirmation height update (which may take a long time due to long chains), I think this non-blocking design means that calling `block_confirm` wont update height until the long chains are done with.

(Not sure if all this is just an upgrade documentation issue or not, but seems like a topic worth discussing as the block height updating necessarily takes a while.)

nano/node/CMakeLists.txt
SergiySW: @cryptocode probably you right...

wezrule: gladly reverted back :)

nano/node/node.cpp
clemahieu: I think this is a good observation, we want our confermation to be coherent from an external observer i.e. we shouldn't allow a case where a confirmation comes through yet the confirmation height hasn't been committed.

nano/rpc/rpc_handler.cpp
```diff
@@ -3970,6 +3974,7 @@ void nano::rpc_handler::wallet_pending ()
 	const bool source = request.get<bool> ("source", false);
 	const bool min_version = request.get<bool> ("min_version", false);
 	const bool include_active = request.get<bool> ("include_active", false);
+	const bool include_only_confirmed = request.get<bool> ("include_only_confirmed", false);
```
clemahieu: We'll need an RPC test for this flag.

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,294 @@
...
+		}
+		else
+		{
+			condition.wait (lk);
+		}
+	}
+}
+
+void nano::confirmation_height_processor::add (nano::block_hash const & hash_a)
+{
```
clemahieu: We could store if a block is live here which is before the potentially long delay of updating confirmation height.

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,294 @@
...
+	std::unique_lock<std::mutex> lk (pending_confirmations.mutex);
+	while (!stopped)
+	{
+		if (!pending_confirmations.queue.empty ())
+		{
+			auto pending_confirmation = pending_confirmations.queue.front ();
+			pending_confirmations.queue.pop ();
+			lk.unlock ();
+			add_confirmation_height (pending_confirmation);
+			lk.lock ();
```
clemahieu: We can externally signal confirmation here if the block was marked as live when it was inserted in to pending_confirmations.queue.

zhyatt: If the plan is to manually trigger callback and add record in confirmation history for external signal, updates here may be useful for reference: https://github.com/nanocurrency/nano-node/pull/1898/files. The block_info/blocks_info RPC calls also need context of which blocks are confirmed during upgrade (since they return a "confirmed" bool), is that already covered by a check for the requested block in the pending_confirmations.queue?

nano/core_test/network.cpp
```diff
@@ -1775,3 +1776,13 @@ TEST (bootstrap, tcp_listener_timeout_keepalive)
 		ASSERT_NO_ERROR (system.poll ());
 	}
 }
+
+namespace
+{
+void prevent_frontier_confirmation_height_checking (std::vector<std::shared_ptr<nano::node>> & nodes)
+{
+	auto unreachable_time = std::chrono::steady_clock::now () + std::chrono::seconds (10000);
+	nodes.front ()->active.next_frontier_check = unreachable_time;
```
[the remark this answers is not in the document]
wezrule: Good point, I only used 2 so had a bit of tunnel vision. However this function has been made obsolete in #1899 so has been removed after merging.

nano/lib/utility.cpp
```diff
@@ -96,6 +96,8 @@ namespace thread_role
 			case nano::thread_role::name::signature_checking:
 				thread_role_name_string = "Signature check";
 				break;
+			case nano::thread_role::name::confirmation_height_processing:
+				thread_role_name_string = "Conf height";
```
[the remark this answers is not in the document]
wezrule: Yep!

nano/node/node.cpp
wezrule: `block_info` and related RPCs should return that the block is confirmed after a callback is done. It now checks the `confirmation_height_processor` to see if the block hash is being processed. This does not work iteratively (until the confirmation height is actually set), it will only return true for the top most block being confirmed.

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,294 @@
...
+	// Get the last confirmed block in this account chain
+	auto num_to_confirm = block_height - confirmation_height;
+	while (num_to_confirm > 0 && !current.is_zero ())
+	{
+		active.confirm_block (current);
+		nano::block_sideband sideband;
+		auto block (store.block_get (transaction, current, &sideband));
+		if (block)
+		{
+			if (block->type () == nano::block_type::receive || (block->type () == nano::block_type::open && block->hash () != genesis_hash))
```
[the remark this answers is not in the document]
wezrule: I'll trust you on that one! I have implemented this, and added various tests. Seems to work ok, but more eyes @SergiySW would be beneficial.

nano/rpc/rpc_handler.cpp
wezrule: Added a test for it, currently there isn't one for the existing `include_active` though.

nano/node/confirmation_height_processor.cpp
wezrule: Is this still needed with the approach we have now taken? (i.e to know when that block's confirmation height has been finalised). If so perhaps can wait until RC2.

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,371 @@
...
+		auto block (store.block_get (transaction, hash, &sideband));
+		if (block)
+		{
+			auto source (block->source ());
+			if (source.is_zero ())
+			{
+				source = block->link ();
+			}
+
+			if (store.block_exists (transaction, source))
```
SergiySW: It can be `if (store.source_exists (transaction, source))`

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,371 @@
...
+				// Commit changes periodically to reduce time holding write locks for long chains
+				break;
+			}
+		}
+	}
+	return false;
+}
+
+// This function assumes receive_source_pairs_lk is not already locked
+void nano::confirmation_height_processor::collect_unconfirmed_receive_and_sources_for_account (uint64_t block_height, uint64_t confirmation_height, nano::block_hash const & current, const nano::block_hash & genesis_hash, std::vector<receive_source_pair> & receive_source_pairs, nano::account const & account, nano::transaction & transaction, std::unique_lock<std::mutex> & receive_source_pairs_lk)
```
SergiySW: I guess genesis_hash is not required anymore here & in other parts of confirmation_height_processor
Also isn't receive_source_pairs vector available for all confirmation_height_processor functions, so you shoudn't pass it, right?

wezrule: good points! I made it a member so that I could access it in the memory stats RPC, have removed unnecessary parameters from the function.

nano/node/confirmation_height_processor.cpp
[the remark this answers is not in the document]
wezrule: I'm not sure if this handles the scenario where it is a receive block but the `source` is just not in the ledger? For instance:
```
nano::send_block send (latest, key1.pub, nano::genesis_amount - nano::Gxrb_ratio, nano::test_genesis_key.prv, nano::test_genesis_key.pub, system.work.generate (latest));
auto open = std::make_shared<nano::open_block> (send.hash (), nano::test_genesis_key.pub, key1.pub, key1.prv, key1.pub, system.work.generate (key1.pub));

{
	auto transaction (store.tx_begin_write ());
	node->ledger.process (transaction, send);
	node->ledger.process (transaction, *open);
	store.block_del (transaction, send.hash ());   // delete the send (could happen from fork?)
	}

	node->block_confirm (open);

        // open gets confirmed, send doesn't. Does this make sense?
```

nano/node/confirmation_height_processor.cpp
wezrule: ok thanks

nano/node/confirmation_height_processor.cpp
wezrule: This is only called for blocks which are from the live network/have `block_confirm` called on them.

nano/node/confirmation_height_processor.cpp
zhyatt: I believe this should be covered by the approach we have now taken with block_info/blocks_info providing confirmation context for the requested block(s) only.

wezrule: Thought so too, thanks :)

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,372 @@
...
+	std::unique_lock<std::mutex> lk (pending_confirmations.mutex);
+	while (!stopped)
+	{
+		if (!pending_confirmations.pending.empty ())
+		{
+			current_original_pending_block = *pending_confirmations.pending.begin ();
+			pending_confirmations.pending.erase (current_original_pending_block);
+			lk.unlock ();
+			add_confirmation_height (current_original_pending_block);
+			current_original_pending_block = 0;
```
cryptocode: Is it really safe to update this member outside the lock?

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,372 @@
...
+		pending_confirmations.pending.insert (hash_a);
+	}
+	condition.notify_one ();
+}
+
+// This only check top-level blocks having their confirmation height sets, not anything below
+bool nano::confirmation_height_processor::is_processing_block (nano::block_hash const & hash_a)
+{
+	// First check the hash currently being processed
+	if (!current_original_pending_block.is_zero () && current_original_pending_block == hash_a)
```
cryptocode: Same question about locking here (though I think `is_processing_block` is only used in tests)

nano/node/confirmation_height_processor.cpp
wezrule: oh gosh no!! That needs to be protected, good spot. I haven't run TSAN on the most recent set of changes.

nano/node/confirmation_height_processor.cpp
wezrule: This is also used in the node. I have protected it now (moved the logic to the `pending_confirmation_height` class)

cryptocode:
Todays testing: --debug_cemented_block_count shows `16M` after ~4 hours, then it slowed down (the 1000-accounts-every-few-minutes logic probably kicked in - probably quite a few chains remain untouched at this point)

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,426 @@
...
+		auto block (store.block_get (transaction, hash, &sideband));
+		if (block)
+		{
+			auto source (block->source ());
+			if (source.is_zero ())
+			{
+				source = block->link ();
+			}
+
+			if (store.source_exists (transaction, source))
```
SergiySW: I guess we can add check if source is_zero () or epoch link as well

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,426 @@
...
+	{
+		assert (pending_writes.empty ());
+		write_remaining_unconfirmed_non_receive_blocks (store, hash_a);
+	}
+}
+
+/*
+ * This takes a block and confirms the blocks below it. It assumes that all other blocks below the first receive has already been confirmed.
+ */
+void nano::confirmation_height_processor::write_remaining_unconfirmed_non_receive_blocks (nano::block_store & store, nano::block_hash const & hash_a)
```
SergiySW: No need to pass store reference

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,426 @@
...
+#include <nano/node/active_transactions.hpp>
+#include <nano/node/stats.hpp>
+#include <nano/secure/blockstore.hpp>
+#include <nano/secure/common.hpp>
+#include <nano/secure/ledger.hpp>
+#include <numeric>
+
+nano::confirmation_height_processor::confirmation_height_processor (nano::pending_confirmation_height & pending_confirmation_height, nano::block_store & store, nano::ledger & ledger, nano::active_transactions & active, nano::logger_mt & logger) :
+pending_confirmations (pending_confirmation_height),
+store (store),
```
SergiySW: I guess store can be received from ledger.store without extra reference

wezrule: Looking again I don't actually need the `ledger` reference. It was only used to access the stats (`ledger.stats`) which I am now passing in directly and to call `ledger.account` which is just a wrapper around `store.block_account`. So this removes a rather large dependency.

nano/node/confirmation_height_processor.cpp
wezrule: Function now removed

nano/node/confirmation_height_processor.cpp
wezrule: Good point, done.

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,378 @@
...
+#include <nano/lib/logger_mt.hpp>
+#include <nano/lib/numbers.hpp>
+#include <nano/lib/utility.hpp>
+#include <nano/node/active_transactions.hpp>
+#include <nano/node/stats.hpp>
+#include <nano/secure/blockstore.hpp>
+#include <nano/secure/common.hpp>
+#include <numeric>
+
+nano::confirmation_height_processor::confirmation_height_processor (nano::pending_confirmation_height & pending_confirmation_height, nano::block_store & store, nano::stat & stats, nano::active_transactions & active, nano::block_hash const & epoch_link, nano::logger_mt & logger) :
```
SergiySW: May be better name references as store_a, stats_a etc. as it is in other parts of code?

wezrule: Sure, makes sense to be consistent. Changes all function parameters in that file.

nano/node/confirmation_height_processor.cpp
```diff
@@ -0,0 +1,378 @@
...
+ * To limit write locking and to keep the confirmation height ledger correctly synced, confirmations are
+ * written from the ground upwards in batches.
+ */
+void nano::confirmation_height_processor::add_confirmation_height (nano::block_hash const & hash_a)
+{
+	boost::optional<conf_height_details> receive_details;
+	auto current = hash_a;
+	nano::account_info account_info;
+	std::deque<conf_height_details> pending_writes;
+	release_assert (receive_source_pairs_size == 0);
```
cryptocode: Is this worth a release_assert, given it's only used for memory stats?

wezrule: Make it an `assert`

zhyatt:
@wezrule Can you help review the stats counters documentation and help determine needed updates based on details in this ticket? https://docs.nano.org/running-a-node/troubleshooting/#statistics-from-rpc and https://docs.nano.org/commands/rpc-protocol/#stats Seems like we should consolidate these and perhaps expand some details.

zhyatt:
@wezrule Can you help review the stats RPC documentation and help suggest updates? https://docs.nano.org/commands/rpc-protocol/#stats and https://docs.nano.org/running-a-node/troubleshooting/#statistics-from-rpc. Looks like we may want to combine and possibly expand some pieces, not sure. The other RPC and CLI command in this ticket look good in the docs.

wezrule:
@zhatt it was quite out of date, have added the missing stats

zhyatt:
Thanks!

PR #1882, opened 2019-04-07
Reduce work.eco_pow intermittent failure rate

wezrule:
Due to the random nature of a successful proof when generating pow this test can fail when comparing the time taken for 2 threads to find a valid proof even when 1 thread has a delay. I have made this delay more pronounced, which does increase the time taken to complete the test (100ms on my machine), but it seems worthwhile to do this to cause it to fail even more infrequently (pretty much 0)

### DIFF
--- a/nano/core_test/work_pool.cpp
+++ b/nano/core_test/work_pool.cpp
@@ -178,10 +178,10 @@ TEST (work, eco_pow)
 	std::future<std::chrono::nanoseconds> future2 = promise2.get_future ();
 
 	std::thread thread1 (work_func, std::ref (promise1), std::chrono::nanoseconds (0));
-	std::thread thread2 (work_func, std::ref (promise2), std::chrono::nanoseconds (10000000));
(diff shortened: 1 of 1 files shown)


nano/core_test/work_pool.cpp
```diff
@@ -178,10 +178,10 @@ TEST (work, eco_pow)
 	std::future<std::chrono::nanoseconds> future2 = promise2.get_future ();
 
 	std::thread thread1 (work_func, std::ref (promise1), std::chrono::nanoseconds (0));
-	std::thread thread2 (work_func, std::ref (promise2), std::chrono::nanoseconds (100000));
+	std::thread thread2 (work_func, std::ref (promise2), std::chrono::nanoseconds (10000000));
```
cryptocode: Maybe `std::chrono::milliseconds (10)` is a bit more readable. Other than that, LGTM and the test now survives 1000 iterations.

wezrule: Good point, done

PR #1889, opened 2019-04-09
Remove vote generator delay

cryptocode:
Colin realized we could remove the vote generator delay. Multiple hashes still got packed when I tested this on beta (with high tps it should still stuff 12 hashes into a v-b-h message)

Some confirmation height tests will fail intermittently until PR #1877 is merged.

### DIFF
--- a/nano/node/voting.cpp
+++ b/nano/node/voting.cpp
@@ -67,35 +67,15 @@ void nano::vote_generator::run ()
 	lock.unlock ();
 	condition.notify_all ();
 	lock.lock ();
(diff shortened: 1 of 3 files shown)


cryptocode:
Removing the delay seems to cause intermittent failures of 

* confirmation_height.multiple
* confirmation_height.gap_live

Removing the delay possibly exacerbates some timing issues.

cryptocode:
Some confirmation height tests will fail intermittently until PR #1877 is merged.

PR #1894, opened 2019-04-10
Fix TSAN warning in node.no_voting & wallet.update_work_action

wezrule:
In `active::confirmed_frontiers ()`, `enable_voting` is used but this is accessed at the same time it is set in  `node.no_voting`. It seemed a bit pointless to add a mutex to protect `enable_voting` as it is only changed in tests (other than during node initialization), and this creates other problems as the `node_config` will need an explicit copy constructor as the mutex implicitly deletes it, so will need to be a `mutable` mutex and all the data members (there is a lot) would need setting and maintaining, just for tests, it didn't seem worth it.


wezrule:
@argakiig I've updated this a bit, so that we can customize nodes in the tests a bit easier and don't have to have dummy ones sitting around.

PR #1895, opened 2019-04-11
Recalculate work from wallet

argakiig:
replaces #1860

blocks handled by the internal wallets will have work calculated using the active_difficulty if active_difficulty is greater
creates a thread in wallets to track these blocks and the time they were inserted, if they arent confirmed they are re inserted with an updated time.


guilhermelawless:
Is this scenario plausible?

If the work generation difficulty is an average of the active transactions, I can trigger work generation for many nodes by doing a new transaction every 4  ensuring that the difficulty only goes up very slightly each time (say 0.01%).

nano/node/wallet.cpp
```diff
@@ -1512,6 +1534,67 @@ void nano::wallets::reload ()
...
+					{
+						node.active.update_difficulty (*block_l.get ());
+						node.network.flood_block (block_l);
+					}
+				}
+			}
+			if (!confirmed)
+			{
+				regeneration_lock.unlock ();
+				queue_work_regeneration (now, block_l);
```
wezrule: I might be tempted to move `difficulty_reque.insert (std::make_pair (time_point, block_a));` to another private function which is not thread safe and call that here instead of lock/unlocking (and also call it in `queue_work_regeneration ()`), or even just call `difficulty_reque.insert (std::make_pair (time_point, block_a));` in place (however then you have duplication). These solutions are probably not any cleaner than what you are doing already, but thought I'd suggest some alternatives anyway. Feel free to ignore.

nano/core_test/wallet.cpp
```diff
@@ -1029,3 +1029,60 @@ TEST (wallet, deterministic_restore)
...
+	auto existing (node.active.roots.find (block_new->qualified_root ()));
+	ASSERT_NE (node.active.roots.end (), existing);
+	//election no longer using difficulty1
+	ASSERT_NE (existing->difficulty, difficulty1);
+	//election using difficulty2
+	ASSERT_EQ (existing->difficulty, difficulty2);
+	//active_difficulty2 after filling difficulty_cb with difficulty2 +10000 is greater than difficulty2
+	ASSERT_GT (active_difficulty2, difficulty2);
+	ASSERT_GT (active_difficulty2, active_difficulty1);
+}
```
wezrule: Needs a newline

nano/node/wallet.cpp
```diff
@@ -1512,6 +1534,67 @@ void nano::wallets::reload ()
...
+	{
+		if (!difficulty_reque.empty ())
+		{
+			auto const first (difficulty_reque.begin ());
+			auto now (std::chrono::steady_clock::now ());
+			auto queued (first->first);
+			auto block (first->second);
+			std::shared_ptr<nano::block> block_l (block);
+			bool confirmed (false);
+			difficulty_reque.erase (first);
```
wezrule: The `regeneration_lock` could be released here (just make sure to lock again before the scope leaves)

nano/node/wallet.cpp
```diff
@@ -1512,6 +1534,67 @@ void nano::wallets::reload ()
...
+			if ((now - queued) >= node.config.work_recalc_inverval)
+			{
+				std::unique_lock<std::mutex> lock (node.active.mutex);
+				auto existing (node.active.roots.find (block->qualified_root ()));
+				if (node.active.roots.end () != existing)
```
SergiySW: Probably else condition should trigger `confirmed = true;` or at least check with ledger.block_confirmed ()

nano/node/wallet.cpp
```diff
@@ -1512,6 +1534,67 @@ void nano::wallets::reload ()
...
+				nano::work_validate (*block, &difficulty);
+				if (!confirmed && node.active.active_difficulty < difficulty)
+				{
+					block_l = update_work_action (block);
+				}
+				if (!confirmed)
+				{
+					if (block != block_l)
+					{
+						node.active.update_difficulty (*block_l.get ());
```
SergiySW: nano::active_transactions::update_difficulty () should also replace block in election, now it's just updating difficulty

nano/node/wallet.cpp
```diff
@@ -1512,6 +1534,67 @@ void nano::wallets::reload ()
...
+				std::unique_lock<std::mutex> lock (node.active.mutex);
+				auto existing (node.active.roots.find (block->qualified_root ()));
+				if (node.active.roots.end () != existing)
+				{
+					confirmed = existing->election->confirmed.load ();
+				}
+				lock.unlock ();
+				uint64_t difficulty (0);
+				nano::work_validate (*block, &difficulty);
+				if (!confirmed && node.active.active_difficulty < difficulty)
```
SergiySW: May be need check if enough reps quorum is observed

nano/node/wallet.cpp
```diff
@@ -1433,6 +1440,21 @@ std::shared_ptr<nano::wallet> nano::wallets::create (nano::uint256_union const &
...
+	auto const active_difficulty1 (node.active.active_difficulty.load ());
+	auto state = std::dynamic_pointer_cast<nano::state_block> (block_a);
+	nano::block_builder builder;
+	std::error_code ec;
+	std::shared_ptr<nano::block> block_l = builder
+	                                       .state ()
+	                                       .from (*state.get ())
+	                                       .work (node.work_generate_blocking (state->previous (), active_difficulty1))
+	                                       .build (ec);
+	return block_l;
```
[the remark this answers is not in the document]
argakiig: yes yes it does, and it revealed that the block_builder flags werent being set during .from(), #1902 to fix that, Thanks :)

nano/lib/utility.cpp
```diff
@@ -87,6 +87,8 @@ namespace thread_role
 			case nano::thread_role::name::wallet_actions:
 				thread_role_name_string = "Wallet actions";
 				break;
+			case nano::thread_role::name::wallet_work_regen:
+				thread_role_name_string = "Wallet work regen";
```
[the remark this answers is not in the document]
argakiig: shortened

nano/node/wallet.cpp
```diff
@@ -1512,6 +1534,67 @@ void nano::wallets::reload ()
...
+	}
+}
+
+void nano::wallets::queue_work_regeneration (std::chrono::steady_clock::time_point const & time_point, std::shared_ptr<nano::block> const & block_a)
+{
+	{
+		std::lock_guard<std::mutex> regeneration_lock (difficulty_mutex);
+		difficulty_reque.insert (std::make_pair (time_point, block_a));
+	}
+	condition.notify_all ();
```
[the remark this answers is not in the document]
argakiig: updated

nano/node/wallet.cpp
```diff
@@ -1512,6 +1534,67 @@ void nano::wallets::reload ()
...
+			condition.wait (regeneration_lock);
+		}
+	}
+}
+
+void nano::wallets::queue_work_regeneration (std::chrono::steady_clock::time_point const & time_point, std::shared_ptr<nano::block> const & block_a)
+{
+	{
+		std::lock_guard<std::mutex> regeneration_lock (difficulty_mutex);
+		difficulty_reque.insert (std::make_pair (time_point, block_a));
```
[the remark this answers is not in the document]
argakiig: thanks :)

nano/node/nodeconfig.cpp
```diff
@@ -248,6 +249,7 @@ bool nano::node_config::upgrade_json (unsigned version_a, nano::jsonconfig & jso
 			json.put ("tcp_client_timeout", tcp_client_timeout.count ());
 			json.put ("tcp_server_timeout", tcp_server_timeout.count ());
 			json.put (pow_sleep_interval_key, pow_sleep_interval.count ());
+			json.put ("work_recalc_inverval", work_recalc_inverval.count ());
```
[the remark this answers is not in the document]
argakiig: v17 upgrade and v17 fresh variants done

nano/node/nodeconfig.hpp
```diff
@@ -60,6 +60,7 @@ class node_config
 	std::chrono::seconds tcp_client_timeout{ std::chrono::seconds (5) };
 	std::chrono::seconds tcp_server_timeout{ std::chrono::seconds (30) };
 	std::chrono::nanoseconds pow_sleep_interval{ 0 };
+	std::chrono::seconds work_recalc_inverval{ std::chrono::seconds (4) };
```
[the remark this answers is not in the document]
argakiig: updated

nano/node/wallet.cpp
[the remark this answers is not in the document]
argakiig: #1902 merged, and fixed to return block if there is an error

nano/core_test/wallet.cpp
argakiig: every time, thanks

nano/node/wallet.cpp
argakiig: updated

nano/node/wallet.cpp
argakiig: else now creates block transaction and sets confirmed to `ledger.block_confirmed(transaction,block->hash())`

nano/node/wallet.cpp
SergiySW: Also there is election status.winner (of course if winner hash == block hash)

nano/lib/utility.cpp
```diff
@@ -87,6 +87,9 @@ namespace thread_role
 			case nano::thread_role::name::wallet_actions:
 				thread_role_name_string = "Wallet actions";
 				break;
+			case nano::thread_role::name::wallet_work_regen:
+				thread_role_name_string = "Work regen";
+                break;
```
SergiySW: Seems like formatting?

nano/node/wallet.cpp
```diff
@@ -1512,6 +1536,85 @@ void nano::wallets::reload ()
...
+			std::shared_ptr<nano::block> block_l (block);
+			bool confirmed (false);
+			difficulty_reque.erase (first);
+			regeneration_lock.unlock ();
+			auto online = node.rep_crawler.total_weight () > (std::max (node.config.online_weight_minimum.number (), node.delta ()));
+
+			if (((now - queued) >= node.config.work_recalc_interval && online) || (node.network_params.network.current_network == nano_networks::nano_test_network || node.network_params.network.current_network == nano_networks::rai_test_network))
+			{
+				std::unique_lock<std::mutex> lock (node.active.mutex);
+				auto existing (node.active.roots.find (block->qualified_root ()));
```
cryptocode: This crashes `gap_cache.gap_bootstrap` locally because `block` is null.

nano/node/wallet.cpp
```diff
@@ -1512,6 +1536,85 @@ void nano::wallets::reload ()
...
+			auto now (std::chrono::steady_clock::now ());
+			auto queued (first->first);
+			auto block (first->second);
+			std::shared_ptr<nano::block> block_l (block);
+			bool confirmed (false);
+			difficulty_reque.erase (first);
+			regeneration_lock.unlock ();
+			auto online = node.rep_crawler.total_weight () > (std::max (node.config.online_weight_minimum.number (), node.delta ()));
+
+			if (((now - queued) >= node.config.work_recalc_interval && online) || (node.network_params.network.current_network == nano_networks::nano_test_network || node.network_params.network.current_network == nano_networks::rai_test_network))
```
cryptocode: `node.network_params.network.is_test_network()` can be used here

nano/node/wallet.cpp
```diff
@@ -1131,9 +1131,10 @@ bool nano::wallet::change_sync (nano::account const & source_a, nano::account co
 
 void nano::wallet::change_async (nano::account const & source_a, nano::account const & representative_a, std::function<void(std::shared_ptr<nano::block>)> const & action_a, uint64_t work_a, bool generate_work_a)
 {
-	wallets.node.wallets.queue_wallet_action (nano::wallets::high_priority, shared_from_this (), [source_a, representative_a, action_a, work_a, generate_work_a](nano::wallet & wallet_a) {
+	wallets.node.wallets.queue_wallet_action (nano::wallets::high_priority, shared_from_this (), [this, source_a, representative_a, action_a, work_a, generate_work_a](nano::wallet & wallet_a) {
```
cryptocode: Should probably do `auto this_l (shared_from_this ());` and pass/use that instead of `this`. Same a few other places.

nano/node/wallet.cpp
```diff
@@ -1433,6 +1440,23 @@ std::shared_ptr<nano::wallet> nano::wallets::create (nano::uint256_union const &
...
+	auto state = std::dynamic_pointer_cast<nano::state_block> (block_a);
+	nano::block_builder builder;
+	std::error_code ec;
+	std::shared_ptr<nano::block> block_l = builder
+	                                       .state ()
+	                                       .from (*state.get ())
+	                                       .work (node.work_generate_blocking (state->previous (), active_difficulty1))
+	                                       .build (ec);
+	if (ec)
+		return block_a;
```
cryptocode: Nitpicking: we tend to use braces even for single statement blocks.

SergiySW:
P.S. I guess for RC2 we may need to add some metrics to delay recalculation during network spam

nano/node/node.cpp
```diff
@@ -3320,10 +3318,16 @@ void nano::active_transactions::update_active_difficulty ()
...
+	auto sum (std::accumulate (node.active.difficulty_cb.begin (), node.active.difficulty_cb.end (), uint128_t (0)));
 	difficulty = static_cast<uint64_t> (sum / difficulty_cb.size ());
 	assert (difficulty >= node.network_params.network.publish_threshold);
-	active_difficulty.store (difficulty);
+	atomic_active_difficulty.store (difficulty);
+}
+
+uint64_t nano::active_transactions::active_difficulty ()
+{
+	std::lock_guard<std::mutex> lock (mutex);
```
cryptocode: Is this needed, as atomic_active_difficulty is atomic?

nano/core_test/wallet.cpp
```diff
@@ -1038,3 +1038,47 @@ TEST (wallet, deterministic_restore)
...
+		//if existing is junk the block has been confirmed already
+		ASSERT_NE (existing, node.active.roots.end ());
+		updated = existing->difficulty != difficulty1;
+		updated_difficulty = existing->difficulty;
+		lock.unlock ();
+		ASSERT_NO_ERROR (system.poll ());
+	}
+	lock.lock ();
+	ASSERT_GT (updated_difficulty, difficulty1);
+	lock.unlock ();
```
wezrule: do these local variables need protecting?

nano/node/node.cpp
argakiig: this is to keep a timing issue from arising, for example: 
`active_difficulty = threshold;`
Thread 1
``` update_active_difficulty ()
//has lock and updates local active_difficulty calculation
active_difficulty = 20 ```
meanwhile
Thread 2
``` do_work_regeneration ()
if (active_difficulty > block_difficulty)
{
update_work_action (block)
}
```

nano/node/wallet.cpp
```diff
@@ -1512,6 +1550,121 @@ void nano::wallets::reload ()
...
+					if (node.active.roots.end () != existing)
+					{
+						//block may not be in existing yet
+						confirmed = existing->election->confirmed.load ();
+					}
+					else
+					{
+						//and so we fall back to ledger confirmation
+						confirmed = node.ledger.block_confirmed (transaction, block->hash ());
+					}
```
wezrule: This if/else block (incl comments) is identical to one above it. Can it be refactored?

nano/node/node.cpp
cryptocode: I see, seems like a comment in the code there about this is all that is needed, so the lock doesn't get removed by accident (since it locally looks like protecting an atomic)

nano/node/wallet.cpp
```diff
@@ -1512,6 +1550,107 @@ void nano::wallets::reload ()
...
+			auto const first (difficulty_reque.begin ());
+			auto now (std::chrono::steady_clock::now ());
+			auto queued (first->first);
+			auto block (first->second);
+			std::shared_ptr<nano::block> block_l (block);
+			auto is_confirmed = [this, block]() -> bool {
+				bool confirmed_l (false);
+				std::unique_lock<std::mutex> lock (this->node.active.mutex);
+				auto existing (this->node.active.roots.find (block->qualified_root ()));
+				if (this->node.active.roots.end () != existing)
```
cryptocode: Could these three lines (including the lock) be replaced with a call to `this->node.active.active (*block)` ? To avoid using the mutex directly.

PR #1897, opened 2019-04-11
Delete unused rpc.cpp

SergiySW:


wezrule:
I did fix this in #1874, however it's not in yet and is good to get this removed as it's annoying. It should be removed from CMakeLists.txt file as well though

SergiySW:
@wezrule this file (nano/node/rpc.cpp) is not deleted in https://github.com/nanocurrency/nano-node/pull/1874. Also it's not referenced in any CMakeLists.txt

wezrule:
Oh, sorry I thought this was rpc.cpp in the core_test which is also duplicated, fancy removing it as well here?

SergiySW:
Well, removing core_test/rpc.cpp is already part of https://github.com/nanocurrency/nano-node/pull/1874

PR #1900, opened 2019-04-11
Fix vote::to_json()

guilhermelawless:
2 bugs:

First, it would crash at a `block_hash` variant:
```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_get> >'
what():  boost::bad_get: failed value get using boost::get
Aborted (core dumped) 
```

Now, the `block_hash` variant is correctly handled.

Second, only one block would be serialized (overwritten on top of the last one). Now it serializes a list of all blocks. Tested with `--vote_dump`

### DIFF
--- a/nano/secure/common.cpp
+++ b/nano/secure/common.cpp
@@ -400,26 +400,33 @@ bool nano::vote::operator!= (nano::vote const & other_a) const
 	return !(*this == other_a);
 }
 
-std::string nano::vote::to_json () const
+void nano::vote::serialize_json (boost::property_tree::ptree & tree) const
 {
-	std::stringstream stream;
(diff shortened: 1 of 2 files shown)


guilhermelawless:
Just noticed, it's not actually serializing all blocks (e.g. 12 in a vbh), just one block. I will fix this.

@cryptocode  do you mind if i split `to_json` to use a common `serialize_json` ?

cryptocode:
Doing changes to vote's `to_json` should be fine, seems low risk too if only `--vote_dump` uses it.

guilhermelawless:
Updated the first comment.

PR #1903, opened 2019-04-12
Add pulls cache for long bootstraps

SergiySW:


nano/node/bootstrap.cpp
```diff
@@ -3161,3 +3170,67 @@ void nano::frontier_req_server::next ()
...
+			});
+		}
+	}
+}
+
+void nano::pulls_cache::update_pull (nano::pull_info & pull_a)
+{
+	std::lock_guard<std::mutex> guard (pulls_cache_mutex);
+	nano::uint512_union head_512 (pull_a.account, pull_a.head_original);
+	auto existing (cache.get<1> ().find (head_512));
```
cryptocode: If it's not too much hassle, it would be nice with a multiindex tag here even if it's just one index for now. Makes the code a bit more readable.

SergiySW: Makes sense

SergiySW: And done

nano/node/bootstrap.cpp
```diff
@@ -3161,3 +3170,67 @@ void nano::frontier_req_server::next ()
...
+void nano::pulls_cache::remove (nano::pull_info const & pull_a)
+{
+	std::lock_guard<std::mutex> guard (pulls_cache_mutex);
+	nano::uint512_union head_512 (pull_a.account, pull_a.head_original);
+	cache.get<account_head_tag> ().erase (head_512);
+}
+
+namespace nano
+{
+std::unique_ptr<seq_con_info_component> collect_seq_con_info (pulls_cache & pulls_cache, const std::string & name)
```
wezrule: This should be added to the `collect_seq_con_info (bootstrap_initiator &, const std::string &)` function:

`composite->add_component (collect_seq_con_info (bootstrap_initiator.cache, "cache"));`

otherwise it isn't be used.

nano/node/bootstrap.cpp
```diff
@@ -3161,3 +3170,67 @@ void nano::frontier_req_server::next ()
...
+
+void nano::pulls_cache::add (nano::pull_info const & pull_a)
+{
+	if (pull_a.processed > 500)
+	{
+		std::lock_guard<std::mutex> guard (pulls_cache_mutex);
+		// Clean old pull
+		if (cache.size () >= max_cache)
+		{
+			cache.erase (cache.begin ());
```
wezrule: Is it possible for `cache.size ()` to ever be greater than `max_cache`? Perhaps that should be an assert (programmer error). If it can (perhaps I'm missing something) then it should erase a range of elements so that `cache.size` always becomes less than `max_cache`, rather than just erasing one.

nano/node/bootstrap.hpp
```diff
@@ -217,6 +224,35 @@ class bulk_pull_account_client final : public std::enable_shared_from_this<nano:
...
+
+private:
+	std::mutex pulls_cache_mutex;
+	boost::multi_index_container<
+	nano::cached_pulls,
+	boost::multi_index::indexed_by<
+	boost::multi_index::ordered_non_unique<boost::multi_index::member<nano::cached_pulls, std::chrono::steady_clock::time_point, &nano::cached_pulls::time>>,
+	boost::multi_index::hashed_unique<boost::multi_index::member<nano::cached_pulls, nano::uint512_union, &nano::cached_pulls::account_head>>>>
+	cache;
+	size_t max_cache = 10000;
```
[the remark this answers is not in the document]
SergiySW: Makes sense

nano/node/bootstrap.cpp
SergiySW: Adding assert, thanks

PR #1904, opened 2019-04-13
Remove delay exiting node in request processor

wezrule:
I noticed this when running this CLI command `nano_node --network=live --debug_account_count` (Debug/Windows) but I'd imagine it will be caused by other commands depending on timing. It can take a while for the node to actually stop when it should be pretty instant. It was caused by the request loop, so I now check if `stopped` is true to break early out of some unnecessary processing.

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -2777,6 +2777,14 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
 		size_t elections_count (0);
 		for (auto i (node.store.latest_begin (transaction_a, next_frontier_account)), n (node.store.latest_end ()); i != n && elections_count < max_elections; ++i)
 		{
+			{
+				std::lock_guard<std::mutex> guard (mutex);
+				if (stopped)
+				{
+					break;
+				}
+			}
+
 			nano::account_info info (i->second);
 			if (info.block_count != info.confirmation_height)
 			{
@@ -3070,6 +3078,13 @@ void nano::active_transactions::request_loop ()
 	while (!stopped)
 	{
 		request_confirm (lock);
+
+		// This prevents unnecessary waiting if stopped is set in-between the above check and now
+		if (stopped)
+		{
+			break;
+		}
+
 		const auto extra_delay (std::min (roots.size (), max_broadcast_queue) * node.network.broadcast_interval_ms * 2);
 		condition.wait_for (lock, std::chrono::milliseconds (node.network_params.network.request_interval_ms + extra_delay));
 	}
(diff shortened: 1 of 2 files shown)


guilhermelawless:
Had the same thing today with `--wallet_list` :)

nano/node/node.cpp
```diff
@@ -2777,6 +2777,14 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
 		size_t elections_count (0);
 		for (auto i (node.store.latest_begin (transaction_a, next_frontier_account)), n (node.store.latest_end ()); i != n && elections_count < max_elections; ++i)
 		{
+			{
+				std::lock_guard<std::mutex> guard (mutex);
+				if (stopped)
```
SergiySW: May be better make stopped atomic, then it can be just `for (auto i (node.store.latest_begin (transaction_a, next_frontier_account)), n (node.store.latest_end ()); i != n && && !stopped && elections_count < max_elections; ++i)`

SergiySW:
I guess request_loop shouldn't even start for inactive_node. As well as networking and some other threads

nano/node/node.cpp
wezrule: I like that it is protected by a mutex otherwise it can still have timing issues and delays:

```
std::unique_lock<std::mutex> lk (mutex);
...
const auto extra_delay (std::min (roots.size (), max_broadcast_queue) * node.network.broadcast_interval_ms * 2);
if (stopped) // read stopped = false
{
  break;
}
// [another thread] `stopped` = true (no lock)
condition.wait_for (lock, std::chrono::milliseconds (node.network_params.network.request_interval_ms + extra_delay));
```

wezrule:
> I guess request_loop shouldn't even start for inactive_node. As well as networking and some other threads
Yeh probably :), but the daemon will probably encounter similar issues when stopped via RPC.

nano/node/node.cpp
wezrule: Stopped still needs to be protected by a mutex in some scenarios otherwise you can still have delays, e.g:

```
std::unique_lock<std::mutex> lk (mutex);
...
const auto extra_delay (std::min (roots.size (), max_broadcast_queue) * node.network.broadcast_interval_ms * 2);
if (stopped) // read stopped = false
{
  break;
}
// [another thread] `stopped` = true (no lock)
// Still delay
condition.wait_for (lock, std::chrono::milliseconds (node.network_params.network.request_interval_ms + extra_delay));
``` 

However `stopped = true` currently is protected by a lock, so it should work as you've described and is what I've done.

PR #1906, opened 2019-04-14
Websockets - check for subscriptions before proceeding

guilhermelawless:
If the websocket server is active via config but there are no active subscriptions, currently the observer will still create and try to broadcast the block confirmation message.

A simple check would have been to go through all sessions and check if any topics are subscribed. However, since this would be done every time a new block arrives, I have added a map `topic -> subscription count` in the listener to do this lookup in constant time. I'm open to suggestions as I am not sure what the right pattern would be here.

Small test added for this as well.

### DIFF
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -70,7 +70,7 @@ TEST (websocket, confirmation)
 
 	// Start websocket test-client in a separate thread
 	std::atomic<bool> confirmation_event_received{ false };
-	ASSERT_FALSE (node1->websocket_server->any_subscription (nano::websocket::topic::confirmation));
+	ASSERT_FALSE (node1->websocket_server->any_subscribers (nano::websocket::topic::confirmation));
 	std::thread client_thread ([&system, &confirmation_event_received]() {
 		// This will expect two results: the acknowledgement of the subscription
 		// and then the block confirmation message
@@ -92,7 +92,7 @@ TEST (websocket, confirmation)
 	{
 		ASSERT_NO_ERROR (system.poll ());
 	}
-	ASSERT_TRUE (node1->websocket_server->any_subscription (nano::websocket::topic::confirmation));
+	ASSERT_TRUE (node1->websocket_server->any_subscribers (nano::websocket::topic::confirmation));
 
 	// Quick-confirm a block
 	nano::keypair key;
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -70,6 +70,7 @@ TEST (websocket, confirmation)
 
 	// Start websocket test-client in a separate thread
 	std::atomic<bool> confirmation_event_received{ false };
+	ASSERT_FALSE (node1->websocket_server->any_subscription (nano::websocket::topic::confirmation));
 	std::thread client_thread ([&system, &confirmation_event_received]() {
 		// This will expect two results: the acknowledgement of the subscription
 		// and then the block confirmation message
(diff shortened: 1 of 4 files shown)


nano/node/websocket.cpp
```diff
@@ -13,6 +13,11 @@ ws_listener (listener_a), ws (std::move (socket_a)), write_strand (ws.get_execut
 
 nano::websocket::session::~session ()
 {
+	for (auto & subscription : subscriptions)
```
cryptocode: This loop is missing a lock on subscriptions_mutex.

guilhermelawless: Thanks, fixed!

guilhermelawless:
@cryptocode it was definitely with the case of different subscriptions in mind. In that case this becomes crucial.

Most likely anyone with websocket enabled will be connecting to it, but maybe not always? I will have something that I will connect only occasionally to gather information about votes/blocks, so it wouldn't make sense to always go through the message building process.

The added complexity is to avoid the following for every block:
```cpp
if (this->block_arrival.recent (block_a->hash ()))
{
	std::string subtype;
	if (is_state_send_a)
	{
		subtype = "send";
	}
	else if (block_a->type () == nano::block_type::state)
	{
		if (block_a->link ().is_zero ())
		{
			subtype = "change";
		}
		else if (amount_a == 0 && !this->ledger.epoch_link.is_zero () && this->ledger.is_epoch_link (block_a->link ()))
		{
			subtype = "epoch";
		}
		else
		{
			subtype = "receive";
		}
	}
	nano::websocket::message_builder builder;
	auto msg (builder.block_confirmed (block_a, account_a, amount_a, subtype));
	this->websocket_server->broadcast (msg);
}
```

If we have other subscription types in the future, their processing might be more resource heavy so it should be avoided.

nano/node/websocket.hpp
```diff
@@ -143,12 +143,20 @@ namespace websocket
 			return node;
 		}
 
+		/** Per topic subscription check */
+		bool any_subscription (nano::websocket::topic const & topic_a);
```
cryptocode: Minor: `any_subscribers` or `has_subscribers` reads more naturally to me on call sites. WDYT?

cryptocode: Minor: any_subscribers reads more naturally to me on call sites. WDYT?

nano/core_test/websocket.cpp
```diff
@@ -91,6 +92,7 @@ TEST (websocket, confirmation)
 	{
 		ASSERT_NO_ERROR (system.poll ());
 	}
+	ASSERT_TRUE (node1->websocket_server->any_subscription (nano::websocket::topic::confirmation));
```
cryptocode: This needs the rename as well ;)

guilhermelawless: Appreciate it 👍

cryptocode:
@guilhermelawless makes sense, I think we should add this.

nano/node/websocket.hpp
```diff
@@ -143,12 +143,20 @@ namespace websocket
...
+		void decrease_subscription_count (nano::websocket::topic const & topic_a);
+
 	private:
 		nano::node & node;
 		boost::asio::ip::tcp::acceptor acceptor;
 		boost::asio::ip::tcp::socket socket;
-		std::mutex sessions_mutex;
+		std::mutex sessions_mutex, counts_mutex;
 		std::vector<std::weak_ptr<session>> sessions;
+		std::unordered_map<topic, std::size_t> topic_subscription_count;
```
clemahieu: Since the number of topic is small and this could be an array indexed by the enum value.

cryptocode: Doing so would also fix the CI build error (missing hash specialization for the enum class I think, on some compilers)

guilhermelawless: Good idea, also could change to atomic size_t and thus remove the mutex. Getting the length of the enum is cumbersome.

clemahieu:
You should be able to size and enum by creating a value at the end called something like topic_count which doesn’t represent an actual topic.

guilhermelawless:
Ok that's how I did it in 9e8471f

Issue #1901, opened 2019-04-11
Extensions to the websocket API


guilhermelawless:
The socket replacing HTTP callback is well built and can immediately support extensions with little development effort.

One way to extend will likely be to add account filters in order to minimize the number of unnecessary messages broadcasted to clients.

I would like to discuss if it is possible to add other kinds of extensions. One example I have already built is broadcasting all vote messages received:

```json
{
  "topic": "vote",
  "time": "1554995525343",
  "message": {
    "account": "xrb_1n5aisgwmq1oibg8c7aerrubboccp3mfcjgm8jaas1fwhxmcndaf4jrt75fy",
    "signature": "1950700796914893705657789944906107642480343124305202910152471520450456881722545967829502369630995363643731706156278026749554294222131169148120786048025353",
    "sequence": "855471574",
    "blocks": [
      "6FB9DE5D7908DEB8A2EA391AEA95041587CBF3420EF8A606F1489FECEE75C869"
    ]
  }
}
```

Other extensions in mind:

- Computed proofs of work
- Wallet lock/unlock or any kind of wallet use, for security reasons
- Using it for the RPC functionality

In my opinion, having the power of getting notifications of any kind from the node to any service is quite powerful.

-----

The discussion falls into the usual "Should we have addons in the node?". I would argue that developers have enjoyed the benefits of having a thorough RPC and by extension, this could help people analyze data by not requiring them to build a packet sniffer (and keep it updated with the protocol).

It is inherently optional as I have to actually subscribe to it through the websocket, specifying a topic.

cryptocode:
Thanks for testing the WebSockets feature and initiating the discussion! The possibility of using it as an RPC came up earlier. Given the full duplex nature of WS we'd need a common request/reply correlation system, or look into something like the WAMP sub protocol.

guilhermelawless:
For reference, I have the vote subscription here: https://github.com/guilhermelawless/nano-node/commit/5a40e4e379181fe896579656d6ad4f6de5d39611

Surprisingly easy to set it up; can PR whenever.

guilhermelawless:
The vote subscription only provides a subset of the votes in the node, and does not separate by vote type (e.g. a replay, a locally generated vote, etc).

One idea would be to have multiple subscriptions to votes, possibly using the same options for each one.

guilhermelawless:
Further ideas:

[...]

PR #1907, opened 2019-04-14
Websockets - filtering options

guilhermelawless:
This adds per-subscription filtering options via an extra field in a subscribe action:

```json
{
  "action": "subscribe",
  "topic": "confirmation",
  "options": {
    "all_local_accounts": "true",
    "accounts": [ "xrb_123...", "..." ]
  }
}
```

Filtering happens after the message is built and is sent for broadcasting, with a bunch of O(1) checks for existance of accounts. We would have to change the dynamics of the websocket server quite a bit to handle it differently. Open to suggestions as usually or just pickup from this PR and improve!

Can add something to `core_test` if this seems ok.

Related: #1901

### DIFF
--- a/nano/node/websocket.cpp
+++ b/nano/node/websocket.cpp
@@ -11,13 +11,13 @@ node (node_a)
 	auto accounts_l (options_a.get_child_optional ("accounts"));
 	if (accounts_l)
 	{
-		for (auto account : *accounts_l)
+		for (auto account_l : *accounts_l)
 		{
 			// Check if the account is valid, but no error handling if it's not, simply not added to the filter
-			nano::account result (0);
-			if (!result.decode_account (account.second.data ()))
+			nano::account result_l (0);
+			if (!result_l.decode_account (account_l.second.data ()))
 			{
-				accounts.insert (account.second.data ());
+				accounts.insert (account_l.second.data ());
 			}
 		}
 	}
@@ -25,30 +25,31 @@ node (node_a)
 
 bool nano::websocket::confirmation_options::should_filter (nano::websocket::message const & message_a) const
 {
-	auto source_text (message_a.contents.get<std::string> ("message.account"));
-	auto destination_opt (message_a.contents.get_optional<std::string> ("message.block.link_as_account"));
-	if (!destination_opt)
+	bool should_filter_l (true);
+	auto source_text_l (message_a.contents.get<std::string> ("message.account"));
+	auto destination_opt_l (message_a.contents.get_optional<std::string> ("message.block.link_as_account"));
+	if (!destination_opt_l)
 	{
-		destination_opt = message_a.contents.get_optional<std::string> ("message.block.destination");
+		destination_opt_l = message_a.contents.get_optional<std::string> ("message.block.destination");
 	}
-	assert (destination_opt);
-	auto destination_text (destination_opt.get ());
+	assert (destination_opt_l);
+	auto destination_text (destination_opt_l.get ());
 	if (all_local_accounts)
 	{
-		auto transaction (node.wallets.tx_begin_read ());
-		nano::account source (0), destination (0);
-		assert (!source.decode_account (source_text));
-		assert (!destination.decode_account (destination_text));
-		if (node.wallets.exists (transaction, source) || node.wallets.exists (transaction, destination))
+		auto transaction_l (node.wallets.tx_begin_read ());
+		nano::account source_l (0), destination_l (0);
+		assert (!source_l.decode_account (source_text_l));
+		assert (!destination_l.decode_account (destination_text));
+		if (node.wallets.exists (transaction_l, source_l) || node.wallets.exists (transaction_l, destination_l))
 		{
-			return false;
+			should_filter_l = false;
 		}
 	}
-	if (accounts.find (source_text) != accounts.end () || accounts.find (destination_text) != accounts.end ())
+	if (accounts.find (source_text_l) != accounts.end () || accounts.find (destination_text) != accounts.end ())
 	{
-		return false;
+		should_filter_l = false;
 	}
-	return true;
+	return should_filter_l;
 }
 
 nano::websocket::session::session (nano::websocket::listener & listener_a, boost::asio::ip::tcp::socket socket_a) :
--- a/nano/node/websocket.cpp
+++ b/nano/node/websocket.cpp
@@ -23,11 +23,11 @@ node (node_a)
 	}
 }
 
-bool nano::websocket::confirmation_options::should_filter (boost::property_tree::ptree const & message_a) const
+bool nano::websocket::confirmation_options::should_filter (nano::websocket::message const & message_a) const
 {
 	// If this fails, the message builder has been changed
-	auto source_text (message_a.get<std::string> ("message.account"));
-	auto destination_text (message_a.get<std::string> ("message.block.link_as_account"));
+	auto source_text (message_a.contents.get<std::string> ("message.account"));
+	auto destination_text (message_a.contents.get_optional<std::string> ("message.block.link_as_account"));
 	if (all_local_accounts)
 	{
 		auto transaction (node.wallets.tx_begin_read ());
@@ -97,7 +97,7 @@ void nano::websocket::session::write (nano::websocket::message message_a)
 	// clang-format off
 	std::unique_lock<std::mutex> lk (subscriptions_mutex);
 	auto subscription (subscriptions.find (message_a.topic));
-	if (message_a.topic == nano::websocket::topic::ack || (subscription != subscriptions.end () && !subscription->second->should_filter (message_a.contents)))
+	if (message_a.topic == nano::websocket::topic::ack || (subscription != subscriptions.end () && !subscription->second->should_filter (message_a)))
 	{
 		lk.unlock ();
 		boost::asio::post (write_strand,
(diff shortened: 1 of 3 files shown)


nano/node/websocket.hpp
```diff
@@ -63,6 +64,28 @@ namespace websocket
...
+	class confirmation_options final : public options
+	{
+	public:
+		confirmation_options ();
+		confirmation_options (boost::property_tree::ptree const & options_a);
+		virtual bool filter (boost::property_tree::ptree const & message_a, nano::node & node_a) const;
+
+	private:
+		bool all_local_accounts{ false };
+		std::unordered_set<nano::account> accounts;
```
guilhermelawless: Would it be better to have these as `std::strings` and not do the account decoding during `filter`?

cryptocode: Yeah that makes sense.

cryptocode:
#1906 is merged so this one will need a rebase @guilhermelawless

nano/node/websocket.cpp
```diff
@@ -42,11 +81,12 @@ void nano::websocket::session::close ()
 	ws.close (reason, ec_ignore);
 }
 
-void nano::websocket::session::write (nano::websocket::message message_a)
+void nano::websocket::session::write (nano::websocket::message message_a, nano::node & node_a)
```
cryptocode: No need to pass the node here, it's available through `ws_listener.get_node ()` everywhere in the session.

nano/node/websocket.hpp
```diff
@@ -63,6 +64,28 @@ namespace websocket
...
+			return false;
+		}
+	};
+
+	class confirmation_options final : public options
+	{
+	public:
+		confirmation_options ();
+		confirmation_options (boost::property_tree::ptree const & options_a);
+		virtual bool filter (boost::property_tree::ptree const & message_a, nano::node & node_a) const;
```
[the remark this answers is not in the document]
guilhermelawless: What do you think of `should_filter` ?

I will implement the other suggestions, absolutely right.

nano/node/websocket.cpp
```diff
@@ -4,6 +4,45 @@
...
+				accounts.insert (account.second.data ());
+			}
+		}
+	}
+}
+
+bool nano::websocket::confirmation_options::filter (boost::property_tree::ptree const & message_a, nano::node & node_a) const
+{
+	// If this fails, the message builder has been changed
+	auto account_text (message_a.get<std::string> ("message.account"));
```
guilhermelawless: Should this also check `message.block.link_as_account` aka destination?

nano/node/websocket.cpp
```diff
@@ -4,6 +4,45 @@
...
+	if (all_local_accounts)
+	{
+		auto transaction (node_a.wallets.tx_begin_read ());
+		nano::account account (0);
+		assert (account.decode_account (account_text));
+		if (node_a.wallets.exists (transaction, account))
```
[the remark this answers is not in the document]
guilhermelawless: I think this is a good idea, with the other use case being once you have a pruned node with only the local accounts, everything is kept in memory, blazing fast. 👍

nano/node/websocket.hpp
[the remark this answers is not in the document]
cryptocode: That works

nano/node/websocket.cpp
guilhermelawless: We reached the conclusion that it makes sense to and not use multiple options e.g. destination_only, source_and_destination etc to keep it simple; services can filter on their end.

guilhermelawless: @cryptocode is it worth it to check subtype and consequently only source/destination depending on the block type?

guilhermelawless: Right now it will get whatever block mentions the given account, be it epoch, change (source or delegation), send and receive. I think it's better to be inclusive.

cryptocode:
Could we get a test for confirmations being filtered / not filtered for both query types?

nano/node/websocket.cpp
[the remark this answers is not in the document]
cryptocode: @clemahieu any thoughts on this? If it's hotspot data, lmdb is likely to bring these to the page cache, but there's still some overhead. Should probably profile this to avoid unnecessary complexity of maintaining a cache.

guilhermelawless:
Some tests were failing due to the destination being in `block.destination` and not `block.link_as_account` for some blocks.

Other tests were incomplete or some things untested due to not using a true async read on the socket client test.

Everything should be ok now.

nano/node/websocket.cpp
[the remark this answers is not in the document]
guilhermelawless: Did some ad-hoc profiling in the `if (all_local_accounts)` scope, with `nano::autotimer` and instructions by @cryptocode . I have an NVME SSD.

- 1 wallet with 100k accounts: 14-16 µs, no observed trend for subsequent tries
- 20 wallets with 5000 accounts each: 38-50 µs, no observed trend
- 2000 wallets with 20 accounts each: 1100-1500 µs, no observed trend

All profiling done in a Debug build.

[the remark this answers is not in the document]
guilhermelawless: Release build:

- 20 wallets with 5000 accounts each: 1 µs
- 3000 wallets with 20 accounts each: 200-400 µs, no observed trend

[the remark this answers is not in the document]
cryptocode: Great data points! A huge amount of wallets should be rare (are there any examples?) and we can document that local-account filtering may affect performance if a large number of wallets are used. These numbers look acceptable to me, the limiting factor will be elsewhere.

nano/node/websocket.cpp
```diff
@@ -4,6 +4,54 @@
...
+	{
+		destination_opt_l = message_a.contents.get_optional<std::string> ("message.block.destination");
+	}
+	assert (destination_opt_l);
+	auto destination_text (destination_opt_l.get ());
+	if (all_local_accounts)
+	{
+		auto transaction_l (node.wallets.tx_begin_read ());
+		nano::account source_l (0), destination_l (0);
+		assert (!source_l.decode_account (source_text_l));
```
SergiySW: I guess this line & one below can work only in debug builds

nano/node/websocket.cpp
```diff
@@ -4,6 +4,54 @@
...
+{
+	bool should_filter_l (true);
+	auto source_text_l (message_a.contents.get<std::string> ("message.account"));
+	auto destination_opt_l (message_a.contents.get_optional<std::string> ("message.block.link_as_account"));
+	if (!destination_opt_l)
+	{
+		destination_opt_l = message_a.contents.get_optional<std::string> ("message.block.destination");
+	}
+	assert (destination_opt_l);
+	auto destination_text (destination_opt_l.get ());
```
SergiySW: destination_opt_l can be not initialized for some old blocks (open, receive, change). So we may add some if else condition or drop legacy blocks support probably

nano/node/websocket.cpp
```diff
@@ -4,6 +4,54 @@
...
+	auto accounts_l (options_a.get_child_optional ("accounts"));
+	if (accounts_l)
+	{
+		for (auto account_l : *accounts_l)
+		{
+			// Check if the account is valid, but no error handling if it's not, simply not added to the filter
+			nano::account result_l (0);
+			if (!result_l.decode_account (account_l.second.data ()))
+			{
+				accounts.insert (account_l.second.data ());
```
SergiySW: Probably it can be unordered_set of nano:account or insert result_l.to_account () to prevent nano_ / xrb_ prefixes issues (i.e. you can request array of xrb_accounts or even xrb-accounts, but node message will contain nano_accounts by default)

nano/node/websocket.cpp
guilhermelawless: Dropping support then. Note for documentation @cryptocode

cryptocode: @SergiySW @guilhermelawless I feel this should deliver the exact same blocks as the HTTP callback.

nano/node/websocket.cpp
SergiySW: So, if functions inside of assert () are ignored in Release build, then both source_l & destination_l remain 0 I guess

guilhermelawless: I'm switching to release_assert instead, or do you rather not have it?

nano/node/websocket.cpp
guilhermelawless: Encoding back is a great idea, thanks.

nano/node/websocket.cpp
cryptocode: auto error (...)
assert (!error)

nano/node/websocket.cpp
guilhermelawless: Would have been too hard given that we start from an encoded `nano::message`, so i did drop support in the end.

For instance, for receive blocks we would have to get block by its hash.

nano/node/websocket.cpp
guilhermelawless: Thanks, eheh

nano/node/websocket.cpp
guilhermelawless: Merged master, `xrb_` and `nano_` both working.

cryptocode: Good catch and good fix 👍

cryptocode:
@guilhermelawless Thanks! The Websocket Wiki page needs an update, feel free to attach markup here.

guilhermelawless:
@cryptocode  `md` is not supported so i zipped it, here you go!

[WebSockets.zip](https://github.com/nanocurrency/nano-node/files/3094211/WebSockets.zip)

cryptocode:
Docs merged into the Websockets wiki page

PR #1919, opened 2019-04-24
 Replace xxHash with blake2b

SergiySW:


nano/node/common.hpp
```diff
@@ -22,40 +21,57 @@ bool parse_tcp_endpoint (std::string const &, nano::tcp_endpoint &);
 
 namespace
 {
+uint64_t random_64_m{ 0 };
+uint64_t const & random_64 ()
+{
+	if (random_64_m == 0)
+	{
+		random_64_m = (uint64_t) nano::random_pool::generate_word32 (0, 0xffffffffUL) << 32 | nano::random_pool::generate_word32 (0, 0xffffffffUL);
```
cryptocode: Could we make a `nano::random_pool::generate_word64` ? That would lock the pool mutex only once and everything is in one place.

nano/node/common.hpp
```diff
@@ -22,40 +21,57 @@ bool parse_tcp_endpoint (std::string const &, nano::tcp_endpoint &);
 
 namespace
 {
+uint64_t random_64_m{ 0 };
+uint64_t const & random_64 ()
+{
+	if (random_64_m == 0)
```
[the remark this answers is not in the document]
clemahieu: Let's put this in to node/ledger constants and make it 128bits.

nano/node/common.hpp
```diff
@@ -22,40 +21,57 @@ bool parse_tcp_endpoint (std::string const &, nano::tcp_endpoint &);
...
 	address.bytes = endpoint_a.address ().to_v6 ().to_bytes ();
-	XXH64_state_t * const state = XXH64_createState ();
-	XXH64_reset (state, 0);
-	XXH64_update (state, address.bytes.data (), address.bytes.size ());
 	auto port (endpoint_a.port ());
-	XXH64_update (state, &port, sizeof (port));
-	auto result (XXH64_digest (state));
-	XXH64_freeState (state);
+	blake2b_state state;
+	blake2b_init (&state, sizeof (result));
```
clemahieu: Let's use the blake2b_init_key initializer which does the key-initialization for us.

SergiySW: It's much slower than blake2b_update, like 1.5-2 times

nano/node/common.hpp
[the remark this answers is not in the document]
SergiySW: @wezrule could mutex kill performance even more?

nano/node/common.hpp
SergiySW: Colin suggested 128 bit, then new function seems not required

SergiySW:
Can be modified with random_constants from https://github.com/nanocurrency/nano-node/pull/1928

wezrule:
I did an ASAN run on core_test/live and it didn't hit any of CryptoPP uninitialized global warnings, so LGTM.

PR #1929, opened 2019-04-25
[TSAN] Add modify_node_id function

SergiySW:


SergiySW:
Replacing with https://github.com/nanocurrency/nano-node/pull/1930

PR #1938, opened 2019-04-26
TCP socket rewrite with strand and queueing support

cryptocode:
Marking as incomplete as more docs and discussions related to the needs for implementing tcp live messages are needed.

* Factored out socket stuff from bootstrap
* Introduced server_socket since it's no longer bootstrap specific (though bootstrap still starts the tcp server, a different PR should refactor that)
* Socket multithreading support through write queues and strands
* Write queue means shared_ptr of buffers is required everywhere (to extend buffer lifetime). `send_buffer_raw` removed for that reason.
* Socket can be single-writer (default for now) or multi-writer. Former is slightly more efficient (suitable for bootstrap connections which is send/reply), while the latter allows concurrent writers (for live messages if we end up not awaiting replies like with UDP)
* Checkup and the two timer variables are replaced with a steady deadline timer
* Added stats for accept success/failures as an accept failure log entry may give the impression that the node has stopped accepting tcp connections even though a timeout will ensure stale connections are dropped.
* Added network_timeout as a logging option. This was previously tied to bulk pull logging which doesn't make sense for live messages.
* tcp_client_timeout renamed to tcp_io_timeout. Proposing to increase from 5 to 15 second in case tunnels need to be established etc (5 sec may be a tad too low, causing unnecessary connection timeouts)
* tcp_server_timeout removed, no longer needed as accept always starts a timer
* tcp_idle_timeout added, default socket idling timeout, can be overridden by constructor
* Removed `if (!ec)` from callback wrappers, in case callback needs notification of errors/do corrective actions/logging.
* Fix io_ctx#run race in wallet. Thread runner is now started first. This requires a work token, which is cancelled on join.
* To enable tests with thread_runner to finish more quickly, join takes an argument where event looping can be stopped
* `remote` is cached on connect/accept as it's not thread safe to call it directly (caching is more convenient than wrapping in a strand on use sites). For outgoing connections, the remote endpoint is known. For accept, there's an overload to receive it.
* In testutils, a `completion_signal` type is added. Some tests will benefit from using thread runner instead of polling (the new socket test is an example) and this class simplifies waiting for async operations to complete.


cryptocode:
If merged, update wiki docs wrt config changes

nano/node/stats.hpp
```diff
@@ -292,6 +293,10 @@ class stat final
 		invalid_node_id_handshake_message,
 		outdated_version,
 
+		// tcp
+		tcp_accept_success,
+		tcp_accept_failure,
```
wezrule: Can these stats be checked in the socket test at all?

nano/node/transport/udp.cpp
```diff
@@ -390,14 +392,29 @@ void nano::transport::udp_channels::stop ()
...
 	local_endpoint = nano::endpoint (boost::asio::ip::address_v6::loopback (), 0);
 
+	// On test-net, close directly to avoid address-reuse issues. On livenet, close
+	// through the strand as multiple IO threads may access the socket.
 	// clang-format off
-	boost::asio::post (strand, [this] {
-		boost::system::error_code ignored;
-		this->socket.close (ignored);
-	});
+	if (node.network_params.network.is_test_network())
```
wezrule: Should probably have a space before the parentheses of `is_test_network`

nano/core_test/network.cpp
```diff
@@ -1990,7 +2003,7 @@ TEST (bootstrap, tcp_listener_timeout_keepalive)
 		ASSERT_EQ (node0->bootstrap.connections.size (), 1);
 	}
 	bool disconnected (false);
-	system.deadline_set (std::chrono::seconds (5));
+	system.deadline_set (std::chrono::seconds (46));
```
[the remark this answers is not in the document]
cryptocode: Seems like I was testing something and didn't turn it back :) Will fix.

nano/node/stats.hpp
cryptocode: Added checks. It's a bit icky to check stats that's updated async', but it seems to work locally. Will revisit if tests starts failing on the added checks.

nano/core_test/testutil.hpp
```diff
@@ -36,4 +40,83 @@ extern nano::uint256_union const & nano_test_account;
...
+			while (!success)
+			{
+				success = count >= required_count;
+				if (!success && timer.before_deadline (deadline_duration_a))
+				{
+					std::unique_lock<std::mutex> lock (mutex);
+					cv.wait_for (lock, std::chrono::milliseconds (1));
+				}
+			}
+			return success && timer.before_deadline (deadline_duration_a);
```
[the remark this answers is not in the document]
cryptocode: Good spot. Yeah success should be removed from the return, I guess the loop condition was different at some point. Fixed.

nano/node/logging.cpp
```diff
@@ -38,6 +38,7 @@ nano::error nano::logging::serialize_json (nano::jsonconfig & json) const
 	json.put ("ledger_duplicate", ledger_duplicate_logging_value);
 	json.put ("vote", vote_logging_value);
 	json.put ("network", network_logging_value);
+	json.put ("network_timeout", network_timeout_logging_value);
```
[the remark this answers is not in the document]
cryptocode: Updated tests

nano/node/CMakeLists.txt
```diff
@@ -67,8 +67,10 @@ add_library (node
 	transport/udp.cpp
 	signatures.hpp
 	signatures.cpp
-	stats.hpp
+	socket.cpp
+	socket.hpp
 	stats.cpp
+	stats.hpp
```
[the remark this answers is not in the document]
cryptocode: Reordered

nano/core_test/testutil.hpp
```diff
@@ -36,4 +40,83 @@ extern nano::uint256_union const & nano_test_account;
...
+namespace util
+{
+	/**
+	 * Helper to signal completion of async handlers in tests.
+	 * Subclasses implement specific conditions for completion.
+	 */
+	class completion_signal
+	{
+	public:
+		~completion_signal ()
```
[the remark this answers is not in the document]
cryptocode: Fixed.

nano/node/bootstrap.cpp
```diff
@@ -1918,29 +1802,34 @@ std::unique_ptr<seq_con_info_component> collect_seq_con_info (bootstrap_initiato
 }
 
 nano::bootstrap_listener::bootstrap_listener (boost::asio::io_context & io_ctx_a, uint16_t port_a, nano::node & node_a) :
```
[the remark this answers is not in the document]
cryptocode: Removed from signature.

nano/node/socket.cpp
```diff
@@ -0,0 +1,334 @@
...
+	if (auto node_l = node.lock ())
+	{
+		start_timer (node_l->config.tcp_io_timeout);
+	}
+}
+
+void nano::socket::start_timer (std::chrono::seconds deadline_a)
+{
+	std::weak_ptr<nano::socket> this_w (shared_from_this ());
+	boost::asio::post (strand, [this_w, deadline_a]() {
```
clemahieu: The old way of starting and stopping timers was to queue actions on the alarm class which had issues when there were a high number of start/stops queued and memory usage would balloon, is this an issue here as well with queueing items on a strand?
This was replaced with an atomic ticket system where we could detect if a timer had been stopped without a memory allocation.

nano/node/socket.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+void nano::socket::set_max_idle_timeout (std::chrono::seconds max_idle_time_a)
+{
+	auto this_l (shared_from_this ());
+	boost::asio::dispatch (strand, [this_l, max_idle_time_a]() {
+		this_l->max_idle_time = max_idle_time_a;
+	});
+}
+
+// This must be called from a strand
```
clemahieu: Is there a way to assert if we're inside a specific executor context?

nano/node/socket.cpp
```diff
@@ -0,0 +1,334 @@
...
+							}
+						}
+					});
+				}
+			}
+		}
+	});
+}
+
+// This must be called from a strand
```
clemahieu: Can this be checked programatically in an assert?

nano/node/socket.cpp
```diff
@@ -0,0 +1,334 @@
...
+void nano::server_socket::evict_dead_connections ()
+{
+	auto it = connections.begin ();
+	for (; it != connections.end (); it++)
+	{
+		if (it->expired ())
+		{
+			it = connections.erase (it);
```
clemahieu: std::vector::erase returns the location after the item that was erased which also could be end (). At the tail of the loop when it calls it++ either an item is skipped or we're incrementing the iterator past end() which is UB. 

We should also use ++it instead the post-increment form unless post-increment is specifically needed.

nano/node/socket.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include <utility>
+#include <vector>
+
+namespace nano
+{
+class node;
+class server_socket;
+
+/** Socket class for tcp clients and newly accepted connections */
+class socket : public std::enable_shared_from_this<nano::socket>
```
clemahieu: Socket needs to have a virtual destructor or its children won't clean up correctly. Is it a problem to call close_internal within ~socket or should we be checking to make sure the socket isn't deconstructed outside a specific execution context?

nano/node/socket.cpp
cryptocode: I had asserts on `running_in_this_thread`, but that didn't do what I thought (still asserts when called from a strand). I couldn't find a way, but can look again.

nano/node/socket.cpp
cryptocode: Same as above.

nano/node/socket.hpp
cryptocode: Fixed virtual dtor. Calling from d'tor should be fine, but will verify.

nano/node/socket.cpp
cryptocode: Replacing with erase+remove_if instead of a loop

nano/node/socket.cpp
cryptocode: Good point about posting through a strand, reverting to using an atomic ticket system.

cryptocode: I'll see if I can adapt the atomic ticket system to work with concurrent writers.

PR #1942, opened 2019-04-27
Fix our warnings

wezrule:
Some more warnings have been introduced on our end


nano/core_test/websocket.cpp
```diff
@@ -383,7 +383,7 @@ TEST (websocket, vote)
 }
 
 /** Tests vote subscription options */
-TEST (websocket, vote_options)
+TEST (websocket, DISABLED_vote_options)
```
SergiySW: Why disabling?

wezrule: Ergh good spot, those changes should not have been included, they were remnants from some previous ASAN/TSAN runs I did to get past some errors

PR #1943, opened 2019-04-27
Work value in RPC work_validate response

guilhermelawless:
Especially useful now with dynamic PoW.

Needs a documentation update if merged.


guilhermelawless:
@cryptocode 

```markdown
# Work validate

Check whether work is valid for block. The response also includes the work value in hexadecimal format.
Request:

{  
    "action": "work_validate",  
    "work": "2bf29ef00786a6bc",  
    "hash": "718CC2121C3E641059BC1C2CFC45666C99E8AE922F7A807B7D07B62C995D79E2"  
}

Response:

{  
    "valid": "1",
    "value": "ffffffd21c3933f4"
}

## Optional "difficulty"

_version 19.0+_
Difficulty value (16 hexadecimal digits string, 64 bit). Uses **difficulty** value to validate work.
```

guilhermelawless:
Idea from @arranHarty to also output the value as a multiplier using `float((1 << 64) - base_difficulty) / float((1 << 64) - difficulty)` , what do you think?

cryptocode:
A multiplier seems useful (note that the use of 1 << 64 is from Python, which has arbitrary precision integers and 64-bit floats, so the expression won't map directly to C++)

guilhermelawless:
Added the multiplier, wiki markup in the previous comment updated.

guilhermelawless:
@cryptocode might make more sense for multiplier to be in relative to the given difficulty, what do you think?

cryptocode:
@zhyatt docs added to wiki

PR #1944, opened 2019-04-28
Cache resolved dependent elections to save a read in confirmation height processor later

wezrule:
Currently `active.confirm_block` (now renamed to `finalize_election` proposed by @cryptocode) will add the hash to the confirmation height processor pending queue if it confirmed an active election. This results in a read when it is popped off to compare if the block height is greater than the confirmation height which will never succeed. Probably not a big deal but it was an area I saw could be improved.

However if a block underneath a block which is being confirmed and has already been iterated over any future active elections will not have any performance improvement and still require voting. This can be alleviated by storing every single block that is traversed and comparing against this. However the collection alone could hold many millions of hashes if it hits long chains (so could reach many 100s of MBs of memory usage) when first upgrading not sure if this is worth it.

(Unrelated) Removed an erroneous comment in the `rpc.confirmation_height_currently_processing` test which was copied from another test.

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -1938,6 +1938,74 @@ TEST (confirmation_height, all_block_types)
 	ASSERT_EQ (node->ledger.stats.count (nano::stat::type::confirmation_height, nano::stat::detail::blocks_confirmed, nano::stat::dir::in), 15);
 }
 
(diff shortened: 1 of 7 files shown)


wezrule:
After speaking with @SergiySW saving a short lived read tansaction from occurring doesn't seem worth the extra complexity.

PR #1949, opened 2019-05-01
Fix regression introduced by 100-continue

cryptocode:
Reports of stuck RPC. Based on community reports, Wesley tracked it down to IO threads waiting for header reads to complete. All handlers run in IO threads, so lots of requests could get it stuck. Reorganized the 100-continue header handling (which solves the slow large-request problems people were having) so it no longer waits on a promise. Hopefully this fixes the issue (unable to reproduce before and after though, hopefully community members can try this PR before merge)


wezrule:
LGTM after removing `header_parser` from the lambda capture in `boost::beast::http::async_write`

PR #1950, opened 2019-05-02
Add blocks_not_found in RPC blocks_info response rather than an error

guilhermelawless:
Wiki markup:

**INCOMPLETE** - needs to be updated with latest changes

---

## Retrieve multiple blocks with additional info

Retrieves a json representations of **blocks** with block **account**, transaction **amount**, block **balance**, block **height** in account chain, block local modification **timstamp**, block **confirmation status** (since version 19.0), **subtype** (for state blocks since version 19.0).

Starting from version 19.0, **blocks_not_found** is an additional field in the response with the blocks that were not found locally. Before this version, an error would occur instead.

Request
```json
{
  "action": "blocks_info",
  "hashes": ["87434F8041869A01C8F6F263B87972D7BA443A72E0A97D7A3FD0CCC2358FD6F9", "0000000000000000000000000000000000000000000000000000000000000001"]
}
```
Response:
```json
{
  "blocks" : {
    "87434F8041869A01C8F6F263B87972D7BA443A72E0A97D7A3FD0CCC2358FD6F9": {
         "block_account": "xrb_1ipx847tk8o46pwxt5qjdbncjqcbwcc1rrmqnkztrfjy5k7z4imsrata9est",
         "amount": "30000000000000000000000000000000000",
         "balance": "5606157000000000000000000000000000000",
         "height": "58",
         "local_timestamp": "0",
         "confirmed": "false",
       "contents": "{\n
         \"type\": \"state\",\n
         \"account\": \"xrb_1ipx847tk8o46pwxt5qjdbncjqcbwcc1rrmqnkztrfjy5k7z4imsrata9est\",\n
         \"previous\": \"CE898C131AAEE25E05362F247760F8A3ACF34A9796A5AE0D9204E86B0637965E\",\n
         \"representative\": \"xrb_1stofnrxuz3cai7ze75o174bpm7scwj9jn3nxsn8ntzg784jf1gzn1jjdkou\",\n
         \"balance\": \"5606157000000000000000000000000000000\",\n
         \"link\": \"5D1AA8A45F8736519D707FCB375976A7F9AF795091021D7E9C7548D6F45DD8D5\",\n
         \"link_as_account\": \"xrb_1qato4k7z3spc8gq1zyd8xeqfbzsoxwo36a45ozbrxcatut7up8ohyardu1z\",\n
         \"signature\": \"82D41BC16F313E4B2243D14DFFA2FB04679C540C2095FEE7EAE0F2F26880AD56DD48D87A7CC5DD760C5B2D76EE2C205506AA557BF00B60D8DEE312EC7343A501\",\n
         \"work\": \"8a142e07a10996d5\"\n
      }\n",
      "subtype": "send"
     }
  },
  "blocks_not_found": [
    "0000000000000000000000000000000000000000000000000000000000000001"
  ]
}
```

(...)

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -4423,54 +4423,89 @@ TEST (rpc, blocks_info)
 	nano::ipc_rpc_processor ipc_rpc_processor (system.io_ctx, rpc_config);
 	nano::rpc rpc (system.io_ctx, rpc_config, ipc_rpc_processor);
 	rpc.start ();
+	auto check_blocks = [&](test_response & response) {
+		for (auto & blocks : response.json.get_child ("blocks"))
+		{
+			std::string hash_text (blocks.first);
+			ASSERT_EQ (system.nodes[0]->latest (nano::genesis_account).to_string (), hash_text);
+			std::string account_text (blocks.second.get<std::string> ("block_account"));
+			ASSERT_EQ (nano::test_genesis_key.pub.to_account (), account_text);
+			std::string amount_text (blocks.second.get<std::string> ("amount"));
+			ASSERT_EQ (nano::genesis_amount.convert_to<std::string> (), amount_text);
+			std::string blocks_text (blocks.second.get<std::string> ("contents"));
+			ASSERT_FALSE (blocks_text.empty ());
+			boost::optional<std::string> pending (blocks.second.get_optional<std::string> ("pending"));
+			ASSERT_FALSE (pending.is_initialized ());
+			boost::optional<std::string> source (blocks.second.get_optional<std::string> ("source_account"));
+			ASSERT_FALSE (source.is_initialized ());
+			std::string balance_text (blocks.second.get<std::string> ("balance"));
+			ASSERT_EQ (nano::genesis_amount.convert_to<std::string> (), balance_text);
+			ASSERT_TRUE (blocks.second.get<bool> ("confirmed")); // Genesis block is confirmed by default
+		}
+	};
 	boost::property_tree::ptree request;
 	request.put ("action", "blocks_info");
 	boost::property_tree::ptree entry;
-	boost::property_tree::ptree peers_l;
+	boost::property_tree::ptree hashes;
 	entry.put ("", system.nodes[0]->latest (nano::genesis_account).to_string ());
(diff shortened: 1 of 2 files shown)


guilhermelawless:
`blocks_not_found` is always included in the response even if there are no blocks there. Could also not include it if it's empty. I think any of the approaches are reasonable.

wezrule:
This is a breaking change if someone expects an error if any of the blocks can't be found; this is the pattern we have in quite a few places. There should probably be an optional `bool` argument to turn this behavior on/off so that it's backwards compatible. What do you think @zhyatt?

guilhermelawless:
It's a good point, though I doubt someone is using this to find out the blocks can't be found, as it doesn't offer any feedback regarding which. Might be worth breaking for added convenience.

guilhermelawless:
It's a good point. Could also include the `error` field if someone is relying on that, along with the rest of fields.

zhyatt:
I imagine the current behavior is more of an issue than the proposed fix would cause, but we don't want to do breaking changes now. I am not sure including `error` when there is a partially valid response makes sense, do we use that pattern anywhere else? I would lean toward an optional argument to enable this behavior of providing `blocks_not_found` in the response, and then when we do a bigger refactor RPC we can evaluate removing that and going to a standard `blocks_not_found` response because breaking changes for RPC will be happening then anyways.

guilhermelawless:
@wezrule Do you use the `bool` to change the type of response anywhere else? Would like to follow the naming convention.

wezrule:
> @wezrule Do you use the `bool` to change the type of response anywhere else? Would like to follow the naming convention.

Look like `output_raw` is an example inside `void nano::json_handler::account_history ()` (`json_handler.cpp`).

nano/rpc_test/rpc.cpp
```diff
@@ -4423,54 +4423,89 @@ TEST (rpc, blocks_info)
 	nano::ipc_rpc_processor ipc_rpc_processor (system.io_ctx, rpc_config);
 	nano::rpc rpc (system.io_ctx, rpc_config, ipc_rpc_processor);
 	rpc.start ();
+	auto check_blocks = [&](test_response & response) {
```
wezrule: Can you pass in exactly what you need in this lambda capture. Otherwise LGTM.

PR #1951, opened 2019-05-03
Allow long running read transactions to be broken up

wezrule:
This enables long running read transactions to be broken up and reuse the same transaction object instead of creating a new one each time periodically which is more efficient as per the official [documentation](http://www.lmdb.tech/doc/starting.html#optim). This will help with some of the forthcoming transaction breakup tasks.

As read/write now return different concrete objects (to make use of RAII for automatic cleanup) they can no longer be specified by using a `bool` argument (which is a good thing imo). This is what most of the changes done are for.

It is currently possible to pass a read only transaction object to something that expects a write transaction, this is only apparent at run-time. This can now be made a compile-time constraint by specifying a `write_transaction`. Widespread changes can be made to accommodate this but were not done here as to make only necessary changes during the RC.

### DIFF
--- a/nano/node/lmdb.cpp
+++ b/nano/node/lmdb.cpp
@@ -68,16 +68,74 @@ nano::mdb_env::operator MDB_env * () const
 	return environment;
 }
 
-nano::transaction nano::mdb_env::tx_begin (bool write_a) const
+nano::read_transaction nano::mdb_env::tx_begin_read () const
 {
-	return { std::make_unique<nano::mdb_txn> (*this, write_a) };
+	return nano::read_transaction{ std::make_unique<nano::read_mdb_txn> (*this) };
+}
+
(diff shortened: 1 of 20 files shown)


nano/node/lmdb.cpp
```diff
@@ -68,16 +68,74 @@ nano::mdb_env::operator MDB_env * () const
...
+
+nano::read_mdb_txn::read_mdb_txn (nano::mdb_env const & environment_a)
+{
+	auto status (mdb_txn_begin (environment_a, nullptr, MDB_RDONLY, &handle));
+	release_assert (status == 0);
+}
+
+nano::read_mdb_txn::~read_mdb_txn ()
+{
+	mdb_txn_abort (handle);
```
cryptocode: Do you have to worry about freeing cursors here?

PR #1953, opened 2019-05-04
Print info about network and path, when running daemon

guilhermelawless:
After multiple human mistakes, this is my proposal.

- ~~Adds a `active_network_label` method to `network_constants`~~ #1957
- Always log active network and the application path
- When using daemon, prints out the network, version and path

In CLI:
```
./nano_node --daemon --network test
Network: test, version: 20.0
Path: /home/USER/NanoTest
```

In log:
```
[2019-05-04 10:51:12.104364]: Active network: test
```


nano/qt/qt.cpp
```diff
@@ -74,14 +74,10 @@ balance_label (new QLabel),
 wallet (wallet_a)
 {
 	your_account_label->setStyleSheet ("font-weight: bold;");
-	std::string network = "Live";
-	if (wallet.node.network_params.network.is_beta_network ())
+	std::string network = wallet.node.network_params.network.get_current_network_as_string ();
+	if (!network.empty ())
```
wezrule: this is probably unnecessary (overly defensive) as network will always be populated, but its fine as well.

PR #1947, opened 2019-05-01
LMDB transaction time tracker

wezrule:
LMDB uses a multi-version concurrency control (MVCC) system which allows multiple threads to coordinate simultaneous access to a database. Each read is not affected by any writes (due to keeping multiple versions)` but it does mean that the memory cannot be reused until the read transaction has finished. LMDB database never gets smaller (unless copied and compacted), it only allows previous space to be reused. 

This is a debug tool to help investigate ledger bloat. There is a new RPC call `database_lock_tracker` which takes a `min_time` and displays any currently open mdb read/write transactions which have been open for or longer than `min_time`. Use 0 to show all currently open transactions. When a transaction is opened for longer than the minimum in the `txn_tracking` config section various information is written to the log file (false by default).

New RPC command (Debug/MSVC)
```
{
  "action": "database_txn_tracker",
  "min_read_time" : "1000"
  "min_write_time" : "0"
}
```
```
{
    "json": [
        {
            "thread": "Blck processing",  // What thread held the transaction
            "time_held_open": "2",          // Seconds the transaction has currently been held open for
            "write": "true",                       // If true it is a write lock, otherwise false.
            "stacktrace": [
                ...
                {
                    "name": "nano::mdb_store::tx_begin_write",
                    "address": "00007FF7142C5F86",
                    "source_file": "c:\\users\\wesley\\documents\\raiblocks\\nano\\node\\lmdb.cpp",
                    "source_line": "825"
                },
                {
                    "name": "nano::block_processor::process_batch",
                    "address": "00007FF714121EEA",
                    "source_file": "c:\\users\\wesley\\documents\\raiblocks\\nano\\node\\blockprocessor.cpp",
                    "source_line": "243"
                },
                {
                    "name": "nano::block_processor::process_blocks",
                    "address": "00007FF71411F8A6",
                    "source_file": "c:\\users\\wesley\\documents\\raiblocks\\nano\\node\\blockprocessor.cpp",
                    "source_line": "103"
                },
                {
                    "name": "<lambda_60e93c1383202c7d5b79d5d064c37649>::operator()",
                    "address": "00007FF71403BA2E",
                    "source_file": "c:\\users\\wesley\\documents\\raiblocks\\nano\\node\\node.cpp",
                    "source_line": "1048"
                },
                {
                    "name": "boost::detail::thread_data<<lambda_60e93c1383202c7d5b79d5d064c37649> >::run",
                    "address": "00007FF7140D3781",
[...]

PR #1954, opened 2019-05-04
Reduce length holding read transaction open for in confirmation height processor

wezrule:
Using #1947 it was found that the confirmation height processor can hold a read transaction open for several minutes when traversing very long account chains. This can have impacts on the database size as any writes cannot use the pages this is holding on to until the read has finished.

Added a generic test for reset/renew on the database as one didn't exist.

Made a slight optimization which mean that the sideband data was no longer needed when reading in the blocks on an account.


nano/core_test/block_store.cpp
```diff
@@ -1683,6 +1683,39 @@ TEST (block_store, incompatible_version)
...
+	read_transaction.reset ();
+
+	// Write the block
+	{
+		auto write_transaction (store.tx_begin_write ());
+		nano::block_sideband sideband (nano::block_type::open, 0, 0, 0, 0, 0);
+		store.block_put (write_transaction, hash1, block, sideband);
+	}
+
+	read_transaction.renew ();
```
cryptocode: The `reset` and `renew` terms are kind of similar. 

Would `suspend` and `resume` work in general? At least in this example, that would make it obvious what's going on. Alternatively `release` and `acquire` or some other pair.

nano/node/confirmation_height_processor.cpp
```diff
@@ -293,19 +298,27 @@ void nano::confirmation_height_processor::collect_unconfirmed_receive_and_source
...
+				next_height = block_height;
 			}
 
 			hash = block->previous ();
 		}
+
+		// We could be traversing a very large account so we don't want to have open read transactions for too long.
+		if (num_to_confirm % batch_read_size == 0)
+		{
+			transaction_a.refresh ();
```
cryptocode: What about calling this `restart`?

PR #1957, opened 2019-05-06
Pass command line arguments (network/data_path) to nano_rpc child process

wezrule:
When `nano_rpc` is used in-process it now gets passed network/data_path command line arguments so that it is in sync with `nano_node` should they not be the defaults.

Improve error message when `rpc_path` does not point to a valid file.

Removed unused `network_params` variables from some rpc code

### DIFF
--- a/nano/lib/config.hpp
+++ b/nano/lib/config.hpp
@@ -116,6 +116,22 @@ public:
 		return err;
 	}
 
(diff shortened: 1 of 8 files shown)


wezrule:
@cryptocode thanks. Didn't see that, but gonna steal how he has done it :)

PR #1959, opened 2019-05-06
Backup config files when upgrading

wezrule:
It's reasonable practice to create backups of files which the user is allowed to modify if you are going to modify them. There were also multiple times during the v19 RC1 period where having the original config file beta testers were using before an upgrade would have been useful.

During any upgrades of the config a new file will be created, for instance: `config_backup_15571460497935531.json`

Note: It might look initially strange during upgrades from v18 as the new `rpc_config.json` will have an initial backup because the migration only copies things from the original `config.json`, all the new options for the rpc are done in an upgrade. This can be added to documentation as expected behaviour.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -863,6 +863,57 @@ TEST (json, upgrade_from_existing)
 	ASSERT_EQ ("changed", object1.text);
 }
 
+/** Test that backups are made only when there is an upgrade */
+TEST (json, backup)
+{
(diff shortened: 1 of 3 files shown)


wezrule:
> LGTM but test would be nice :)

Yep good idea. It actually helped uncover a bug where the backup file is always created, even when creating a fresh one (not upgrading). Also got rid of an unnecessary template parameter declaration for the `read` function.

PR #1962, opened 2019-05-07
 Send live network messages over TCP

SergiySW:
- ~channel modified to include transport_type, last_packet_sent~ (https://github.com/nanocurrency/nano-node/pull/1971)
- merge_peer function modified to try TCP connection before UDP
- TCP channels ongoing_keepalive function used to wake up long not sending channels to prevent TCP disconnection from server
- Live TCP connection starts with node ID handshake
- After handshake node sends preferred peering ports for response channels (keepalive self)
- Received TCP live message is responded to sender with response channels
- Using https://github.com/nanocurrency/nano-node/pull/1938 multi-writer for TCP channel
- Multiple nodes system () tests using TCP connections by default
- Some tests modified to check 2 live network options: TCP connections & UDP connections
- ~Contains rpc.history & rpc.account_history tests fixes~ (https://github.com/nanocurrency/nano-node/pull/1966)
- ~RPC peers modified to return channel type with option "peer_details"~ (https://github.com/nanocurrency/nano-node/pull/1971)
- ~tcp_incoming_connections_max added, max established incoming TCP connections~ (https://github.com/nanocurrency/nano-node/pull/1976)
- bootstrap_server modified to accept live network messages
- ~confirm_req/confirm_ack headers improved to include hashes count in extensions for proper using of payload_length_bytes ()~ (https://github.com/nanocurrency/nano-node/pull/1972)


nano/node/transport/udp.hpp
```diff
@@ -69,14 +88,29 @@ namespace transport
 			node_id = node_id_a;
 		}
 
+		nano::transport::transport_type get_type () const
```
clemahieu: This converts channel_udp in to more of a channel_supporting_live class, I'd rather see tcp functionality remain in channel_tcp.

I think we should make a new class tcp_channels which holds all live-message tcp channels and is owned by nano::network.

nano/node/transport/udp.hpp
```diff
@@ -69,14 +88,29 @@ namespace transport
...
+		}
+
+		void set_type (nano::transport::transport_type type_a)
+		{
+			std::lock_guard<std::mutex> lk (channel_mutex);
+			type = type_a;
+		}
+
 		unsigned network_version{ nano::protocol_version };
+		std::shared_ptr<nano::socket> socket;
```
clemahieu: This is mixing TCP in to a UDP class and I think these should remain separate.

nano/node/node.cpp
```diff
@@ -253,6 +277,7 @@ void nano::network::flood_message (nano::message const & message_a)
 	auto list (node.network.udp_channels.list_fanout ());
```
clemahieu: This is where we should be additionally searching for channels in tcp_channels that are not set to bootstrap_only. In the future this will additionally search for channels in channel_quic or whatever.

nano/node/transport/udp.cpp
```diff
@@ -61,7 +70,16 @@ std::function<void(boost::system::error_code const &, size_t)> nano::transport::
...
 std::string nano::transport::channel_udp::to_string () const
 {
-	return boost::str (boost::format ("UDP: %1%") % endpoint);
+	return boost::str (boost::format ("%1%") % endpoint);
+}
+
+nano::transport::channel_udp::~channel_udp ()
+{
+	std::lock_guard<std::mutex> lk (channel_mutex);
+	if (type == nano::transport::transport_type::tcp && socket != nullptr)
```
clemahieu: Mixing TCP in to UDP is causing this cleanup complexity, this is automatically handled by channel_tcp.

nano/node/transport/udp.cpp
```diff
@@ -298,14 +317,26 @@ bool nano::transport::udp_channels::reserved_address (nano::endpoint const & end
 	return result;
 }
 
+nano::endpoint nano::transport::udp_channels::find_node_id (nano::account const & node_id_a)
```
clemahieu: I think this should return the channel instead of the endpoint, the caller can always query for the endpoint if that's what they want.

nano/node/transport/udp.cpp
```diff
@@ -390,12 +421,36 @@ void nano::transport::udp_channels::stop ()
...
+	else
+	{
+		boost::asio::dispatch (strand, [this] {
+			this->close_socket ();
+		});
+	}
 	// clang-format on
+
+	// Close all TCP sockets
+	for (auto i (channels.begin ()), j (channels.end ()); i != j; ++i)
```
clemahieu: This is more tcp-induced cleanup complexity. This should be inside a tcp_channels class that's dedicated to managing tcp-related things.

nano/node/transport/udp.hpp
```diff
@@ -213,7 +269,7 @@ namespace transport
 		attempts;
 		std::unordered_map<nano::endpoint, syn_cookie_info> syn_cookies;
 		std::unordered_map<boost::asio::ip::address, unsigned> syn_cookies_per_ip;
-		nano::node & node;
+		std::unordered_map<nano::tcp_endpoint, std::vector<nano::endpoint>> response_channels;
```
clemahieu: I think nano::network is responsible for managing response channels. Right now it's only udp_channels though we could create a new class tcp_channels which would also be inside nano::network and nano::network is responsible for live messages across all channel types.

nano::network should be able to handle the traffic in a channel-agnostic way.

nano/rpc_test/rpc.cpp
```diff
@@ -2159,11 +2159,11 @@ TEST (rpc, peers)
 	ASSERT_EQ (200, response.status);
 	auto & peers_node (response.json.get_child ("peers"));
 	ASSERT_EQ (2, peers_node.size ());
-	ASSERT_EQ (std::to_string (nano::protocol_version), peers_node.get<std::string> ("UDP: [::1]:24001"));
+	ASSERT_EQ (std::to_string (nano::protocol_version), peers_node.get<std::string> ("[::1]:24001"));
```
clemahieu: We can push this change in by itself to make the patch simpler.

nano/node/transport/udp.hpp
```diff
@@ -57,6 +64,18 @@ namespace transport
 			last_packet_received = time_a;
 		}
 
+		std::chrono::steady_clock::time_point get_last_packet_sent () const
+		{
+			std::lock_guard<std::mutex> lk (channel_mutex);
+			return last_packet_sent;
+		}
+
+		void set_last_packet_sent (std::chrono::steady_clock::time_point const time_a)
```
clemahieu: This can be pushed in by itself to simplify this patch.

nano/core_test/testutil.hpp
```diff
@@ -36,4 +40,83 @@ extern nano::uint256_union const & nano_test_account;
...
+			while (!success)
+			{
+				success = count >= required_count;
+				if (!success && timer.before_deadline (deadline_duration_a))
+				{
+					std::unique_lock<std::mutex> lock (mutex);
+					cv.wait_for (lock, std::chrono::milliseconds (1));
+				}
+			}
+			return timer.before_deadline (deadline_duration_a);
```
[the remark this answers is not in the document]
SergiySW: Seems like part of https://github.com/nanocurrency/nano-node/pull/1938

nano/core_test/testutil.hpp
```diff
@@ -36,4 +40,83 @@ extern nano::uint256_union const & nano_test_account;
...
+			auto val (count.fetch_add (1));
+			if (val >= required_count)
+			{
+				notify ();
+			}
+			return val;
+		}
+
+	private:
+		std::atomic<size_t> count{ 0 };
```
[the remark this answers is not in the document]
SergiySW: part of https://github.com/nanocurrency/nano-node/pull/1938

nano/lib/utility.hpp
```diff
@@ -118,8 +118,12 @@ class thread_runner final
 public:
 	thread_runner (boost::asio::io_context &, unsigned);
 	~thread_runner ();
-	void join ();
+	/** Tells the io context to stop processing events.*/
+	void stop_event_processing ();
+	/** Wait for IO threads to complete. If \p stop_event_processing is set, tell the io context to stop processing events.  */
+	void join (bool stop_event_processing = false);
```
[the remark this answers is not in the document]
SergiySW: https://github.com/nanocurrency/nano-node/pull/1938 ?

nano/rpc_test/rpc.cpp
SergiySW: You mean separate PR?

nano/core_test/testutil.hpp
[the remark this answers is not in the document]
cryptocode: Addressed in 1938

nano/core_test/testutil.hpp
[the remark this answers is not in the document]
cryptocode: Addressed in 1938

nano/lib/utility.hpp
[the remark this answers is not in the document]
cryptocode: Addressed in 1938

nano/node/common.cpp
```diff
@@ -650,6 +707,20 @@ void nano::confirm_ack::visit (nano::message_visitor & visitor_a) const
...
+size_t nano::confirm_ack::size (nano::block_type type_a, size_t count)
+{
+	size_t result (sizeof (nano::account) + sizeof (nano::signature) + sizeof (uint64_t));
+	if (type_a != nano::block_type::invalid && type_a != nano::block_type::not_a_block)
+	{
+		result += nano::block::size (type_a);
+	}
+	else if (type_a == nano::block_type::not_a_block)
+	{
+		result += count * sizeof (nano::block_hash);
```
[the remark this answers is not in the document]
SergiySW: Count should be uint8_t

nano/node/transport/udp.hpp
SergiySW: Now part of https://github.com/nanocurrency/nano-node/pull/1971

nano/rpc_test/rpc.cpp
SergiySW: Now part of https://github.com/nanocurrency/nano-node/pull/1971

nano/node/common.cpp
```diff
@@ -85,6 +85,16 @@ void nano::message_header::block_type_set (nano::block_type type_a)
...
 }
 
+size_t nano::message_header::count_get () const
+{
+	return ((extensions << 8) >> 8).to_ulong ();
+}
+
+void nano::message_header::count_set (size_t count)
+{
+	extensions |= std::bitset<16> (static_cast<unsigned long long> (count));
```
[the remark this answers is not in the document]
SergiySW: Now part of https://github.com/nanocurrency/nano-node/pull/1972 with test

nano/node/common.cpp
```diff
@@ -85,6 +85,16 @@ void nano::message_header::block_type_set (nano::block_type type_a)
 	extensions |= std::bitset<16> (static_cast<unsigned long long> (type_a) << 8);
 }
 
+size_t nano::message_header::count_get () const
+{
+	return ((extensions << 8) >> 8).to_ulong ();
```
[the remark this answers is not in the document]
SergiySW: Mask part of https://github.com/nanocurrency/nano-node/pull/1972

nano/node/common.cpp
```diff
@@ -124,6 +134,33 @@ size_t nano::message_header::payload_length_bytes () const
...
+		{
+			return nano::confirm_ack::size (block_type (), count_get ());
+		}
+		case nano::message_type::confirm_req:
+		{
+			return nano::confirm_req::size (block_type (), count_get ());
+		}
+		case nano::message_type::node_id_handshake:
+		{
+			if (extensions.test (nano::node_id_handshake::query_flag))
```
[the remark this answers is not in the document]
SergiySW: Part of https://github.com/nanocurrency/nano-node/pull/1972

nano/node/common.cpp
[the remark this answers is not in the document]
SergiySW: Part of https://github.com/nanocurrency/nano-node/pull/1972
Now limited to 4 bit (15 hashes seems enough for our small messages, if we want more in the future, then can use 8 bit part)

nano/node/transport/udp.cpp
SergiySW: Updated

nano/node/transport/udp.hpp
SergiySW: Updated with tcp_channels

nano/node/transport/udp.cpp
SergiySW: Updated

nano/node/transport/udp.cpp
SergiySW: Updated

nano/node/node.cpp
SergiySW: Updated

nano/node/transport/udp.hpp
SergiySW: Updated

cryptocode:
There's a bit of code duplication between udp_channels and tcp_channels - like for instance purge_syn_cookies. Should we have a common base class for these things? 

I guess this is a bit tricky because we don't have a common endpoint type, and maybe it's best to do it in a separate PR anyway. 

(The base class could take the endpoint type as a template parameter, or if we want an endpoint abstraction, there's one in https://github.com/nanocurrency/nano-node/pull/1721 as a starting point, but that would be a much bigger change)

nano/node/bootstrap.cpp
```diff
@@ -2312,9 +2385,58 @@ class request_response_visitor : public nano::message_visitor
...
+			connection->node->logger.try_log (boost::str (boost::format ("Received node_id_handshake message from %1%") % connection->remote_endpoint));
+		}
+		if (message_a.query)
+		{
+			boost::optional<std::pair<nano::account, nano::signature>> response (std::make_pair (connection->node->node_id.pub, nano::sign_message (connection->node->node_id.prv, connection->node->node_id.pub, *message_a.query)));
+			assert (!nano::validate_message (response->first, *message_a.query, response->second));
+			auto cookie (connection->node->network.tcp_channels.assign_syn_cookie (connection->remote_endpoint));
+			nano::node_id_handshake response_message (cookie, response);
+			auto bytes = response_message.to_bytes ();
+			connection->socket->async_write (bytes, [connection = connection](boost::system::error_code const & ec, size_t size_a) {
```
wezrule: `bytes` needs to be captured by the lambda so that it remains in scope for asio

nano/node/node.cpp
```diff
@@ -596,17 +617,152 @@ bool nano::network::reachout (nano::endpoint const & endpoint_a, bool allow_loca
...
+{
+	std::unordered_set<std::shared_ptr<nano::transport::channel>> result (tcp_channels.random_set (count_a));
+	std::unordered_set<std::shared_ptr<nano::transport::channel>> udp_random (udp_channels.random_set (count_a));
+	for (auto i (udp_random.begin ()), n (udp_random.end ()); i != n && result.size () < count_a * 1.5; ++i)
+	{
+		result.insert (*i);
+	}
+	while (result.size () > count_a)
+	{
+		result.erase (result.begin ());
```
wezrule: Can this use the ranged iterator version of `erase`?

nano/node/transport/tcp.cpp
```diff
@@ -63,3 +71,561 @@ std::string nano::transport::channel_tcp::to_string () const
...
+{
+	size_t channels_count = 0;
+	size_t attemps_count = 0;
+	size_t syn_cookies_count = 0;
+	size_t syn_cookies_per_ip_count = 0;
+	{
+		std::lock_guard<std::mutex> guard (mutex);
+		channels_count = channels.size ();
+		attemps_count = attempts.size ();
+		std::lock_guard<std::mutex> syn_cookie_guard (syn_cookie_mutex);
```
wezrule: The first guard can be put inside a new scope to release earlier (and prevent possible deadlock).

nano/node/transport/tcp.hpp
```diff
@@ -55,5 +59,137 @@ namespace transport
...
+		void modify (std::shared_ptr<nano::transport::channel_tcp>, std::function<void(std::shared_ptr<nano::transport::channel_tcp>)>);
+		void update (nano::tcp_endpoint const &);
+		// Connection start
+		void start_tcp (nano::endpoint const &, std::function<void(std::shared_ptr<nano::transport::channel>)> const & = nullptr);
+		void start_tcp_receive_node_id (std::shared_ptr<nano::transport::channel_tcp>, nano::endpoint const &, std::shared_ptr<std::vector<uint8_t>>, std::function<void(std::shared_ptr<nano::transport::channel>)> const &);
+		void udp_fallback (nano::endpoint const &, std::function<void(std::shared_ptr<nano::transport::channel>)> const &);
+		nano::node & node;
+
+	private:
+		void close_socket ();
```
wezrule: Is this function defined/used?

SergiySW: Nope, copied from udp_channels, thanks!

nano/node/transport/tcp.cpp
SergiySW: makes sense

nano/node/bootstrap.cpp
SergiySW: Good catch

nano/node/bootstrap.cpp
```diff
@@ -1888,13 +1888,27 @@ std::unique_ptr<seq_con_info_component> collect_seq_con_info (bootstrap_listener
...
 	}
-	std::lock_guard<std::mutex> lock (node->bootstrap.mutex);
-	node->bootstrap.connections.erase (this);
+	if (bootstrap_connection)
+	{
+		--node->bootstrap.bootstrap_count;
+	}
+	if (socket != nullptr)
+	{
+		socket->close ();
```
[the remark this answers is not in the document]
SergiySW: What is your suggestion for dropping connections with invalid node ID handshake?

nano/node/bootstrap.cpp
```diff
@@ -2071,7 +2114,15 @@ void nano::bootstrap_server::receive_frontier_req_action (boost::system::error_c
 			{
 				node->logger.try_log (boost::str (boost::format ("Received frontier request for %1% with age %2%") % request->start.to_string () % request->age));
 			}
-			add_request (std::unique_ptr<nano::message> (request.release ()));
+			if (!bootstrap_connection && !node->flags.disable_bootstrap_listener && node->bootstrap.bootstrap_count < node->config.bootstrap_connections_max)
+			{
+				++node->bootstrap.bootstrap_count;
+				bootstrap_connection = true;
+			}
```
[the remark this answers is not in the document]
SergiySW: makes sense

nano/node/node.cpp
SergiySW: Then it will be something like this I guess
```
if (result.size () > count_a)
{
	auto iterator (result.begin ());
	for (auto count (0); iterator != result.end () && count != count_a; ++iterator, ++count)
	{
	}
	result.erase (iterator, result.end ());
	assert (result.size () == count_a);
}
```

wezrule: does this not work?
```
if (result.size () > count_a)
{
result.erase (result.begin (), result.begin () + result.size () - count_a);
}
```
There can be a performance gain removing in bulk for some data structures, `unordered_set` probably not, so don't worry about it too much.

nano/node/bootstrap.cpp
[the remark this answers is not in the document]
wezrule: I just mean isn't it possible that `stop ()` can be called more than once now? (which can log multiple times etc..) Perhaps I'm just misunderstanding.

nano/node/bootstrap.cpp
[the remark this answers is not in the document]
SergiySW: Updated

nano/node/bootstrap.cpp
SergiySW: Updated

nano/node/transport/tcp.cpp
SergiySW: Updated

nano/node/transport/tcp.hpp
SergiySW: Updated

nano/node/node.cpp
SergiySW: Not working this way for unordered_set

nano/node/bootstrap.cpp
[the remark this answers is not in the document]
SergiySW: Is this better? https://github.com/nanocurrency/nano-node/pull/1962/commits/67efe40ce17ef05d3a30bbdd1599f5a9d4a2f941

nano/node/bootstrap.cpp
```diff
@@ -2252,45 +2330,40 @@ class request_response_visitor : public nano::message_visitor
...
 	}
-	void confirm_ack (nano::confirm_ack const &) override
+	void confirm_ack (nano::confirm_ack const & message_a) override
 	{
-		assert (false);
+		connection->finish_request_async ();
+		auto connection_l (connection->shared_from_this ());
+		connection->node->background ([connection_l, message_a]() {
+			connection_l->node->network.tcp_channels.process_message (message_a, connection_l->remote_endpoint, connection_l->remote_node_id);
+		});
```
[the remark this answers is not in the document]
SergiySW: There are some issues with capturing abstract type nano:message in lambda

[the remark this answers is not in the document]
wezrule: ah, yeh it's making a copy, so would have to be a function template then. Don't worry about it

PR #1972, opened 2019-05-10
Framing support for realtime network messages

SergiySW:
- Sizes addes for each message types & subtypes
- New count field added to message header extensions (4 bit, max size 15) for confirm_req/confirm_ack by hash
- Node ID handshake header flags functions moved to nano::message_header instead from nano::node_id_handshake
- Tests expanded to check count extensions field


SergiySW:
Breaking up #1962

cryptocode:
If we don't want to expend extension bits, we could maybe retrofit max version to be a count field, since it's not used and doesn't seem useful in the future either.

nano/node/bootstrap.cpp
```diff
@@ -2183,6 +2215,32 @@ void nano::bootstrap_server::receive_bulk_pull_account_action (boost::system::er
 	}
 }
 
+void nano::bootstrap_server::receive_frontier_req_action (boost::system::error_code const & ec, size_t size_a, nano::message_header const & header_a)
+{
+	if (!ec)
+	{
+		auto error (false);
+		nano::bufferstream stream (receive_buffer->data (), header_a.payload_length_bytes ());
```
clemahieu: I think these should be size_a instead of payload_length_bytes () since that's the number of bytes actually in the buffer.

nano/node/bootstrap.cpp
```diff
@@ -2326,15 +2437,15 @@ class request_response_visitor : public nano::message_visitor
 			}
 		});
 	}
-	void publish (nano::publish const &) override
+	void publish (nano::publish const & message_a) override
```
clemahieu: Since we don't use message_a this will give unused variable warnings, omitting the name suppresses the warning.

nano/node/bootstrap.cpp
SergiySW: Done

nano/node/bootstrap.cpp
SergiySW: Sure

PR #1973, opened 2019-05-10
Add severity logging levels, and send errors to syslog/Event Viewer (incl failed rollbacks for confirmed blocks)

wezrule:
When the rollback visitor tries to rollback a confirmed block, it currently just outputs to the log file. For greater visibility it now also gets sent to the system logger as priority `error` which should flag up any process monitoring applications.

Added `severity_level` to the logging, currently just with `normal` and `error`, where normal is the default unless otherwise specified. The sys log is filtered on equal or greater than the `severity_logging::error`, so that only crucial messages are sent to the sys log on Unix & Event log on Windows.

The Unix system logger output `/var/log/syslog` doesn't show priority by default. When testing I followed this [page](https://askubuntu.com/questions/79684/syslog-not-showing-log-levels-in-messages) to check that the priority was set correctly. Using the new `--debug_sys_logging` CLI command the follow is output:

`3,1,May 10 13:27:01,ubuntu-s-1vcpu-1gb-lon1-01,nano_node:, Error: 3: Testing system logger`

Where 3 is the `error` priority based on RFC5424.


wezrule:
So the event viewer was a bit more of a pain. From the boost [docs]( https://www.boost.org/doc/libs/1_67_0/libs/log/doc/html/log/detailed/sink_backends.html#log.detailed.sink_backends.event_log)

> The library must be built as a dynamic library in order to use this backend flawlessly. Otherwise event description resources are not linked into the executable, and the Event Viewer is not able to display events properly.

Which we don't do, it is a static library. There are also some permissions problems as it needs to create a registry entry which fails unless the node is run as an admin, which we don't enforce. However checking the Event viewer on Windows 10 it still has the correct entry marked as an error. For instance, using using the new cli command (`--debug_sys_logging`) the description field contains:

```
The description for Event ID 259 from source Nano cannot be found. Either the component that raises this event is not installed on your local computer or the installation is corrupted. You can install or repair the component on the local computer.

If the event originated on another computer, the display information had to be saved with the event.

The following information was included with the event: 

Error: 58 Testing system logger
```

Which I think is good enough, it still contains the relevant log message.

SergiySW:
Could it print to log file "Error: " as well?

PR #1976, opened 2019-05-11
Add config option tcp_incoming_connections_max

SergiySW:
- Add config option tcp_incoming_connections_max to limit incoming TCP connections instead of bootstrap_connections_max
- Use bootstrap_connections_max for bootstrap packets


nano/node/nodeconfig.hpp
```diff
@@ -64,6 +64,8 @@ class node_config
 	std::chrono::seconds tcp_client_timeout{ std::chrono::seconds (5) };
 	std::chrono::seconds tcp_server_timeout{ std::chrono::seconds (30) };
 	std::chrono::nanoseconds pow_sleep_interval{ 0 };
+	/** Default maximum incoming TCP connections, including realtime network & bootstrap */
+	unsigned tcp_incoming_connections_max{ 256 };
```
cryptocode: Since we lack an overlay network, is this too low?

SergiySW: Updated to 1024

SergiySW:
Merged as part of https://github.com/nanocurrency/nano-node/pull/1962

PR #1977, opened 2019-05-11
Add specific rpc process config options

wezrule:
It wasn't very clear that some of the config options were only applied in certain situations (i.e depending on how the rpc server was run; in-process, child process or out of process). I have made it a bit clearer. The `config.json` file for instance a new config node (child_process):
```
    "rpc": {
       ...
        "child_process": {
            "enable": "false",
            "rpc_path": "C:\\Users\\Wesley\\Documents\\raiblocks\\build\\Debug\\nano_rpc.exe"
        }
    }
```
This only lists things which are `child_process` specific. If enable is false (default) then it uses the in-process RPC.

`rpc_config.json` similarly has a new config node for options specific for an rpc server in a different process (child/out of process) rather than being in the main rpc config node:
```
    "process": {
        "ipc_port": "56000",
        "num_ipc_connections": "4"
    }
```

### DIFF
--- a/load-tester/src/launch_node_and_rpc.rs
+++ b/load-tester/src/launch_node_and_rpc.rs
@@ -48,8 +48,10 @@ pub fn launch_node_and_rpc(
         "version": "1",
         "enable_sign_hash": "false",
         "max_work_generate_difficulty": "ffffffffc0000000",
-        "rpc_in_process": "false"
-    },	
+        "child_process": {
+            "enable": "false"
+        }
+    },
     "node": {
         "version": "17",
         "peering_port": peering_port.to_string(),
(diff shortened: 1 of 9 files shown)


cryptocode:
The "rpc semantics change" label indicates a potentially breaking change for RPC clients. Maybe we should have a new "configuration" label and general purpose "breaking" and "nonbreaking" labels that can be combined with other labels. Thoughts on that organization @zhyatt ?

wezrule:
Right it's just config changes, and it's still a pre-release change, so will only affect people who have upgraded to the RC already. @cryptocode sometimes I just close my eyes and pick a label :D, I think they could do with a complete rework as a lot has changed since some of these were first introduced.

cryptocode:
Right, asking because we used to use "rpc semantics change" to collect things that could break RPC clients into release notes.

PR #1978, opened 2019-05-11
Prevent (optionally) block processor batched writes from being logged by the lmdb tracker

wezrule:
Most write log entries from the lmdb tracker are for the `Blck processing` thread. This is expected when bootstrapping as we process writes in batches according to the `block_processor_batch_max_time`. This can become quite noisy (~80000 lines are safe to ignore in a full bootstrap on my Windows/debug test). This is added to the `diagnostics` -> `txn_tracker` config so that it can be toggled (default is to ignore them).

I get  a lot of writes held longer than the "max", for instance using the default 5000ms `block_processor_batch_max_time` there are a few like so:
> [2019-05-11 16:33:54.757606]: 7547ms write lock held on thread Blck processing

So I have added a few seconds buffer onto it when checking whether it should be output to trim these away as well.


zhyatt:
@wezrule Does `ignore_writes_below_block_processor_max_time` need to be added as an option on the database_txn_tracker RPC call? It is unclear to me whether documentation changes are needed. https://docs.nano.org/commands/rpc-protocol/#database_txn_tracker

zhyatt:
Looks like I got this mixed up with that database_txn_tracker RPC call. I just saw the config entry for `ignore_writes_below_block_processor_max_time` on diagnostics https://docs.nano.org/running-a-node/configuration/#configjson. Let me know if any updates should be done, but we may be fine.

wezrule:
@zhyatt it's only in the config, so I think we are good 👍

PR #1981, opened 2019-05-13
Finish up difficulty/multiplier semantics in RPC work_generate and work_validate

guilhermelawless:
- Adds value + multiplier to `work_generate`
- Changes all "value" semantics to "difficulty", per [suggestion](https://github.com/nanocurrency/nano-node/pull/1981#issuecomment-492190033). This affects `work_generate` and `work_validate` documentations.
- Makes sure to have high precision (`double`) in all cases, which includes `active_difficulty`

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -838,7 +838,7 @@ void nano::json_handler::active_difficulty ()
 	response_l.put ("difficulty_threshold", nano::to_string_hex (node.network_params.network.publish_threshold));
 	auto difficulty_active = node.active.active_difficulty ();
 	response_l.put ("difficulty_active", nano::to_string_hex (difficulty_active));
-	float multiplier = nano::difficulty::to_multiplier (difficulty_active, node.network_params.network.publish_threshold);
+	auto multiplier = nano::difficulty::to_multiplier (difficulty_active, node.network_params.network.publish_threshold);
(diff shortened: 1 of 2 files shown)


zhyatt:
@guilhermelawless Is the use case for this item to allow systems doing pre-caching to be able to track generated work levels for better pro-active rework when network load is heavy?

guilhermelawless:
@zhyatt not only for precaching (an excellent use case I hadn't thought of yet), but also for services deciding to implement their own re-work will need to know the value of the work generated and compare against current `active_difficulty` or other measures.

This change avoids having to use RPC `work_validate`.

SergiySW:
"value" doesn't sound like clear name for difficulty

guilhermelawless:
Agreed, but "difficulty" would clash against the request field "difficulty", while "value" maintains coherence with other RPCs.

Maybe the request field should change to threshold so that the reply fields can change to difficulty.

clemahieu:
I think it's fine just being included every time, it doesn't need to be requested/optional.
Is it an issue if the names clash? The request says "give me something over this difficulty" and the response says "this is the actual difficulty".

guilhermelawless:
@clemahieu that makes sense! Made changes to reflect this, all changes documented in the first comment. Breaks RPC but only for v19 beta testers.

PR #1982, opened 2019-05-14
Prioritize frontiers for background confirmation

wezrule:
Updating the confirmation height for frontiers in the background (in the request loop) takes too long to get the cemented block count close to the total block count currently. The accounts are just searched from lowest account number onwards, which has now been improved by using the request loop wait period to scan through as many accounts as possible and prioritize the ones which have the most uncemented blocks. These are then voted on, which allows more efficient voting traffic as the larger unconfirmed accounts are more likely to confirm other accounts.

Some other parameters have also been modified to make confirming blocks more aggressive when the active root count is low.


nano/core_test/node.cpp
```diff
@@ -2562,6 +2564,75 @@ TEST (node, dont_write_lock_node)
 	finished_promise.set_value ();
 }
 
+namespace nano
```
SergiySW: Do we need namespace here?

wezrule: Unfortunately yes otherwise it doesn't compile:
https://github.com/google/googletest/blob/3306848f697568aacf4bcca330f6bdd5ce671899/googletest/include/gtest/gtest_prod.h#L55

This class is made a friend inside `nano::active_transactions`

nano/node/active_transactions.cpp
```diff
@@ -34,40 +34,63 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
...
+
 		// 4 times slower check if all frontiers were confirmed
-		int fully_confirmed_factor = (elections_count <= max_elections) ? 4 : 1;
+		int fully_confirmed_factor = (elections_count < max_elections) ? 4 : 1;
 		// Calculate next check time
-		next_frontier_check = steady_clock::now () + seconds ((representative_factor * fully_confirmed_factor) / test_network_factor);
+		next_frontier_check = steady_clock::now () + (representative_factor * fully_confirmed_factor / test_network_factor);
 		// Set next account to 0 if all frontiers were confirmed
-		next_frontier_account = (elections_count <= max_elections) ? 0 : next_frontier_account;
+		if (elections_count <= max_elections)
```
SergiySW: Seems not required now after https://github.com/nanocurrency/nano-node/pull/1982/files#diff-725dceebd1b7fd4fdf9ab179415dc61bR385

nano/node/active_transactions.cpp
```diff
@@ -310,15 +333,60 @@ void nano::active_transactions::request_loop ()
...
 		request_confirm (lock);
 		update_active_difficulty (lock);
+		const auto extra_delay (std::min (roots.size (), max_broadcast_queue) * node.network.broadcast_interval_ms * 2);
+		condition.wait_for (lock, std::chrono::milliseconds (node.network_params.network.request_interval_ms + extra_delay));
+	}
+}
 
-		// This prevents unnecessary waiting if stopped is set in-between the above check and now
-		if (stopped)
+void nano::active_transactions::prioritize_frontiers_for_confirmation (std::chrono::milliseconds time_a)
```
SergiySW: There is already confirm_frontiers transaction_a, may be use it?

nano/node/active_transactions.hpp
```diff
@@ -137,8 +150,18 @@ class active_transactions final
...
+		return lhs.blocks_uncemented < rhs.blocks_uncemented;
+	};
+	// clang-format on
+	void prioritize_frontiers_for_confirmation (std::chrono::milliseconds);
+	std::set<nano::cementable_account, decltype (comp), boost::fast_pool_allocator<nano::cementable_account>> priority_cementable_frontiers{ comp };
+	static size_t constexpr max_priority_cementable_frontiers{ 100000 };
+	static size_t constexpr confirmed_frontiers_max_pending_cut_off{ 1000 };
 	boost::thread thread;
+
+	friend class confirmation_height_prioritize_frontiers_Test;
```
SergiySW: Is it used somewhere?

nano/node/active_transactions.cpp
wezrule: Yep good spot

nano/node/active_transactions.cpp
wezrule: makes sense :)

nano/node/active_transactions.hpp
wezrule: The test calls the private method `prioritize_frontiers_for_confirmation ()` of this class, so either the test is a friend or that method is made public. But it is only used by the test so I thought it would be more encapsulated this way.

PR #1984, opened 2019-05-15
Work multiplier format and RPC work_validate fix for lower difficulties

guilhermelawless:
Reported by @cryptocode , thanks.

`{"action": "work_validate", "work":"4f4a0cfb4a25dc78", "hash":"0000000000000000000000000000000000000000000000000000000000000002", "difficulty":"000000000000001"}`

Before:
```json
{
"valid": "0",
"difficulty": "82ddbf2c0bd1ae98",
"multiplier": "3.0484946624529728e-08"
}
```

Now:
```json
{
"valid": "1",
"difficulty": "82ddbf2c0bd1ae98",
"multiplier": "0.000000030484947"
}
```

---

Note valid 1 (c6d7f18) and the format fix. Precision is 15 decimal cases, even with higher difficulties.

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -3,6 +3,8 @@
 #include <algorithm>
 #include <boost/array.hpp>
 #include <boost/bind.hpp>
+#include <boost/convert.hpp>
+#include <boost/convert/stream.hpp>
 #include <boost/endian/conversion.hpp>
 #include <boost/polymorphic_cast.hpp>
 #include <boost/property_tree/json_parser.hpp>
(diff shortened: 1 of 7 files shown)


nano/node/json_handler.cpp
```diff
@@ -839,7 +841,7 @@ void nano::json_handler::active_difficulty ()
 	auto difficulty_active = node.active.active_difficulty ();
 	response_l.put ("difficulty_active", nano::to_string_hex (difficulty_active));
 	auto multiplier = nano::difficulty::to_multiplier (difficulty_active, node.network_params.network.publish_threshold);
-	response_l.put ("multiplier", std::to_string (multiplier));
+	response_l.put ("multiplier", boost::convert<std::string> (multiplier, boost::cnv::cstream () (std::fixed) (std::setprecision (std::numeric_limits<double>::digits10))).value ());
```
cryptocode: This change fixes the issue.

It's a big expression for a simple conversion though, so I propose making a `nano::to_string` since it's used a few places. The first version can take a double and a precision parameter whose default value is `std::numeric_limits<double>::digits10`. If we want to support more types later, we can overload or make templates.

We already have `nano::to_string_hex` in lib/blocks.hpp, so it can probably live there for now. Maybe move both to number.hpp or a new conversions.hpp later on (there might be other conversion as well, but we can do that refactoring in a separate PR)

guilhermelawless: So the new functionality should be a encapsulated in a `nano::to_string (double, ...)`, and `to_string_hex` will still have that name since it's not a simple `to_string` conversion, but specifically in `hex` format.

I will go with `numbers.hpp` for now since there are also the difficulty to/from multipliers there, and a refactoring can change everything later.

Thanks!

cryptocode: Sounds like a plan!

PR #1990, opened 2019-05-16
Bounded active transaction

argakiig:
This PR serves to bound the active transactions container dynamically based on the number of transactions that are over the long_announcement threshold and the number of confirmed blocks per second. If the roots container is over 75,50,25% saturated with long_announcements it will drop the two lowest adjusted_difficulty(accounting for block ordering during confirmation and insertion) blocks that are not managed/monitored by the work_watcher that are also long_unconfirmed


nano/node/active_transactions.cpp
```diff
@@ -638,4 +739,27 @@ std::unique_ptr<seq_con_info_component> collect_seq_con_info (active_transaction
 	composite->add_component (std::make_unique<seq_con_info_leaf> (seq_con_info{ "confirmed", confirmed_count, sizeof (decltype (active_transactions.confirmed)::value_type) }));
 	return composite;
 }
+transaction_counter::transaction_counter () :
+trend_last (std::chrono::steady_clock::now ()),
+counter (0),
+rate (0)
```
wezrule: The order in the construct member initialization list is not the same as the order the members are declared in the class. Could just add the members to the class definition and remove the need for this constructor entirely, get 2 birds stoned at once :).

nano/node/active_transactions.cpp
```diff
@@ -579,6 +588,80 @@ void nano::active_transactions::erase (nano::block const & block_a)
...
+	}
+	else
+	{
+		minimum_size = counter.rate * 10;
+	}
+	if (roots.size () > minimum_size)
+	{
+		if (counter.rate <= 10)
+		{
+			if (static_cast<double> (roots.size () * .75) < long_unconfirmed_size ())
```
wezrule: The static cats shouldn't be needed as the `.75` will make the resulting expression a double anyway.

nano/node/active_transactions.cpp
```diff
@@ -579,6 +588,80 @@ void nano::active_transactions::erase (nano::block const & block_a)
...
+			}
+			else
+			{
+				++it;
+			}
+		}
+		else
+		{
+			//early return if 2 have been dropped
+			return result;
```
cryptocode: Can maybe just `break` here?

nano/node/active_transactions.cpp
```diff
@@ -579,6 +588,80 @@ void nano::active_transactions::erase (nano::block const & block_a)
...
+		}
+		else if (counter.rate <= 1000)
+		{
+			if (static_cast<double> (roots.size () * .25) < long_unconfirmed_size ())
+			{
+				result = true;
+			}
+		}
+	}
+	counter_lock.unlock ();
```
cryptocode: `counter_lock` being a `unique_lock`, unlocking will happen on return.

nano/node/active_transactions.cpp
```diff
@@ -579,6 +588,80 @@ void nano::active_transactions::erase (nano::block const & block_a)
...
 
+bool nano::active_transactions::should_flush ()
+{
+	bool result (false);
+	counter.trend_sample ();
+	size_t minimum_size (1);
+	std::unique_lock<std::mutex> counter_lock (counter.mutex);
+	if (counter.rate == 0)
+	{
+		minimum_size = 4;
```
SergiySW: Isn't default minimum too low? Why drop from 4 active elections on live/beta?

nano/node/active_transactions.cpp
```diff
@@ -591,6 +674,24 @@ size_t nano::active_transactions::size ()
 	return roots.size ();
 }
 
+size_t nano::active_transactions::long_unconfirmed_size ()
```
SergiySW: Have to iterate over all roots each time. Probably variable can stored from nano::active_transactions::request_confirm (), unconfirmed_count? And then updated in nano::election::confirm_once () if announcements > announcement_long

nano/node/wallet.cpp
```diff
@@ -1492,6 +1492,13 @@ void nano::work_watcher::add (std::shared_ptr<nano::block> block_a)
 	}
 }
 
+bool nano::work_watcher::is_watched (nano::qualified_root const root_a)
```
SergiySW: could be `const &` ?

nano/node/active_transactions.cpp
```diff
@@ -579,6 +588,80 @@ void nano::active_transactions::erase (nano::block const & block_a)
...
+	assert (!roots.empty ());
+	auto & sorted_roots = roots.get<1> ();
+	for (auto it = sorted_roots.rbegin (); it != sorted_roots.rend ();)
+	{
+		if (count != 2)
+		{
+			auto election = it->election;
+			if (election->announcements > announcement_long && !election->confirmed && !node.wallets.watcher.is_watched (it->root))
+			{
+				result = it->difficulty;
```
SergiySW: Should it be adjusted difficulty?

argakiig: this is an artifact from some debug testing I was doing, it doesn't actually matter one way or another, will probably void this function, there is nothing needing to be returned

nano/node/active_transactions.cpp
argakiig: minimum moved to the counter multiplier of 512, but added check for test_network to drop it down to 4 to make testing easier

SergiySW: Seems reasonable

nano/node/active_transactions.hpp
```diff
@@ -44,6 +44,23 @@ class election_status final
 	std::chrono::milliseconds election_duration;
 };
 
+class transaction_counter final
+{
+public:
+	transaction_counter ();
```
wezrule: No need to supply an empty default constructor here

nano/node/active_transactions.cpp
```diff
@@ -579,6 +589,71 @@ void nano::active_transactions::erase (nano::block const & block_a)
...
+	if (counter.rate == 0)
+	{
+		//set minimum size to 4 for test network
+		minimum_size = node.network_params.network.is_test_network () ? 4 : 512;
+	}
+	else
+	{
+		minimum_size = counter.rate * 512;
+	}
+	if (roots.size () > minimum_size)
```
SergiySW: Not sure if there can be some upper limit as well, i.e. 100k or something

argakiig: added early return if over 100k in roots

nano/node/active_transactions.hpp
argakiig: thanks, forgot i removed everything from this :)

PR #1992, opened 2019-05-16
Fix off by one error in frontier req server

orhanhenrik:
The current implementation will send 1 too many items, since count starts at 0.

### DIFF
--- a/nano/node/bootstrap.cpp
+++ b/nano/node/bootstrap.cpp
@@ -3018,7 +3018,7 @@ count (0)
 
 void nano::frontier_req_server::send_next ()
 {
-	if (!current.is_zero () && count <= request->count)
+	if (!current.is_zero () && count < request->count)
 	{
 		{
 			send_buffer->clear ();


orhanhenrik:
Any idea why the test is failing? I cannot see how it would be related to my code. The test `active_difficulty.recalculate_work` is failing on the macos build.

cryptocode:
`frontier_req_server::sent_action` increments the count before calling `send_next`. Doesn't that make the original code correct?

orhanhenrik:
If the wanted count is 1, and you run the code, the following happens:
1. send_next (count=0), 0<=1, ok
2. sent_action, increment count=1
3. send_next (count=1), 1<=1, ok
4. sent_action, increment count=2
5. send_next (count=2), 2<=1, not ok

So the server will send 2 results and then exit if I have understood the code correctly.  
  
There should probably be a test for this as well, but I'm not very good at C++ and I couldn't even get the project to build on my mac, so it's a bit hard to write and check a test..

orhanhenrik:
Tests are passing on travis here: https://travis-ci.org/orhanhenrik/nano-node, so I'm not sure what's wrong with this test run. Maybe a race condition in the code?

cryptocode:
I see, send_next is initially called from the visitor. I think clients currently always use max here (2^31-1), which might be why tests doesn't pick it up.

orhanhenrik:
Yeah, I noticed that max is always used. I’m writing a separate node implementation and was getting the wrong number of results, so that’s how I found this minor bug.

PR #2001, opened 2019-05-21
Websockets: prevent re-sub

guilhermelawless:
- Change variable and method names from "subscription" to "subscriber" in general
- (bug fix) Do not increase subscribers when re-subbing to the same topic, instead update the current one
- (fix) Do not output an error to log if the error code is EOF (session closed cleanly)
- Move increase/decrease_subscriber_count methods to private, and make `session` a friend of `listener`. I tried to make only the 2 functions that need to access them as friends, but one of them is private. Any alternative?
- Add a getter for `subscriber_count` , useful only in the new test for now but you never know.

### DIFF
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -71,7 +71,8 @@ boost::optional<std::string> websocket_test_call (std::string host, std::string 
 	{
 		boost::beast::error_code ec_ignored;
 		ws.async_close (boost::beast::websocket::close_code::normal, [](boost::beast::error_code const & ec) {
-			//
+			// A synchronous close usually hangs in tests when the server's io_context stops looping
+			// An async_close solves this problem
 		});
 	}
 	return ret;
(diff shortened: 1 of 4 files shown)


nano/core_test/websocket.cpp
```diff
@@ -75,6 +75,86 @@ boost::optional<std::string> websocket_test_call (std::string host, std::string
...
+		system.deadline_set (5s);
+		while (!ack_ready)
+		{
+			ASSERT_NO_ERROR (system.poll ());
+		}
+		ASSERT_EQ (0, node1->websocket_server->subscriber_count (nano::websocket::topic::confirmation));
+	}
+
+	// Second unsub, should acknowledge but not decrease subscriber count
+	std::this_thread::sleep_for (50ms);
```
guilhermelawless: @cryptocode this is an interesting place to test the issues we were having with websocket closing before being done, and crashing. Removing this line and running this test under Debug makes it crash every single time for me. Maybe you can get some insight?

cryptocode: Yeah the sessions are outliving the node in the tests. This issue will be fixed as part of a websocket PR I'm working on.

guilhermelawless: #2007

PR #2006, opened 2019-05-21
Set minimum difficulty for RPC "work_generate"

SergiySW:
to prevent possible OpenCL validation loop
https://github.com/nanocurrency/nano-node/blob/master/nano/node/openclwork.cpp#L698
work_validate is invalid for difficulty < threshold


guilhermelawless:
I don't know about non-breaking... I did use it once :)
I reported that some difficulties below threshold will work, but if it's several orders of magnitude below it hangs.

SergiySW:
@guilhermelawless yes, probably breaking

zhyatt:
I removed the "rpc semantics change" label as this is reserved for changes that break RPC between official releases. This change only breaks it between RC releases since submitting a work value for work_generate to cause this issue is new with V19.

PR #2014, opened 2019-05-22
Add threshold optional to RPC unopened and ledger

guilhermelawless:
RPCs [ledger](https://github.com/nanocurrency/nano-node/wiki/RPC-protocol#ledger) and [unopened
](https://github.com/nanocurrency/nano-node/wiki/RPC-protocol#unopened)

Suggested by @renesq , this is an alternative to pagination for lower powered nodes who want information about the ledger (and helps the scalability of this command in the future).

In both cases, the optional `count` is still honored.

Documentation update needed.

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -2248,6 +2248,7 @@ void nano::json_handler::key_expand ()
 void nano::json_handler::ledger ()
 {
 	auto count (count_optional_impl ());
+	auto threshold (threshold_optional_impl ());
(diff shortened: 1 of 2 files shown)


renesq:
Wondering if we should incorporate the pending balance into this filter's consideration when pending balances are requested. The decision tree would look a bit different then.

cryptocode:
After discussions with @guilhermelawless, we ended up deciding this PR should be updated to include @renesq's idea

PR #2021, opened 2019-05-23
[WIP] Websockets: vote received/replay/sent

guilhermelawless:
- Adds a `vote_code::sent` that is only used for the purpose of websocket broadcasting.
- Only `vote_code::vote` votes are processed in the first observer notify, to keep the same functionality.

Todo:
- [ ]  Validate the approach
- [ ]  Confirm all locations of vote processing are covered
- [ ]  Add websocket tests
- [ ]  Add filtering options by vote type
- [ ]  Add filtering tests
- [ ]  Documentation

@zhyatt I checked and for the v19RC1&2 tests on the beta network, only `received` votes were broadcasted through the websocket. No `replay` nor `sent` votes were there. This is because the `replay` votes don't enter the notify branch on [this switch](https://github.com/nanocurrency/nano-node/blob/V19.0RC2/nano/node/node.cpp#L856).

I updated my analysis script based on this PR as of commit 0b8ca62 and am attaching a sample. See the head of the file for a description. I guess that all recorded `sent` votes that are not from my representative is my own node rebroadcasting votes.

[vote_batching.zip](https://github.com/nanocurrency/nano-node/files/3211857/vote_batching.zip)

### DIFF
--- a/nano/node/bootstrap.cpp
+++ b/nano/node/bootstrap.cpp
@@ -2387,7 +2387,7 @@ public:
 			auto cookie (connection->node->network.tcp_channels.assign_syn_cookie (connection->remote_endpoint));
 			nano::node_id_handshake response_message (cookie, response);
 			auto bytes = response_message.to_bytes ();
-			connection->socket->async_write (bytes, [ bytes, connection = connection ](boost::system::error_code const & ec, size_t size_a) {
+			connection->socket->async_write (bytes, [bytes, connection = connection](boost::system::error_code const & ec, size_t size_a) {
 				if (ec)
 				{
 					if (connection->node->config.logging.network_node_id_handshake_logging ())
--- a/nano/node/bootstrap.cpp
+++ b/nano/node/bootstrap.cpp
@@ -2387,7 +2387,7 @@ public:
 			auto cookie (connection->node->network.tcp_channels.assign_syn_cookie (connection->remote_endpoint));
 			nano::node_id_handshake response_message (cookie, response);
 			auto bytes = response_message.to_bytes ();
-			connection->socket->async_write (bytes, [bytes, connection = connection](boost::system::error_code const & ec, size_t size_a) {
+			connection->socket->async_write (bytes, [ bytes, connection = connection ](boost::system::error_code const & ec, size_t size_a) {
 				if (ec)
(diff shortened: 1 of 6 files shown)


zhyatt:
@guilhermelawless In the sample vote_batching you provided I see that 12 "sent" votes were recorded for nano_3qkzggguxhcbfaqczybhhjc97ytudhaqke3jcjfe3fcey8i786fan78p6k51. As this isn't your Rep and it also isn't a Principal Rep, then no rebroadcasting should have been done for their votes. This appears to mean the assumption "I guess that all recorded `sent` votes that are not from my representative is my own node rebroadcasting votes" must be wrong. We should keep this in mind and make sure we understand what the definition of `send` votes will be in the websocket.

guilhermelawless:
@zhyatt Thank you for validating it. I am struggling to connect all the pieces here to be absolutely sure, but my hope is that the PR in its current form can help do exactly that.

I had missed spots where `vote_processor.vote ()` is called. There are 3 of them:

1. In [confirm_ack](https://github.com/nanocurrency/nano-node/blob/471c587afd3d4d555ae9bfedc3349392a37de55d/nano/node/node.cpp#L547)
2. In [compute_rep_votes](https://github.com/nanocurrency/nano-node/blob/471c587afd3d4d555ae9bfedc3349392a37de55d/nano/node/node.cpp#L2798)
3. In [vote_generator::send](https://github.com/nanocurrency/nano-node/blob/471c587afd3d4d555ae9bfedc3349392a37de55d/nano/node/voting.cpp#L55)

I'm not sure about what happens in 1 and 2. It may be even better to have subtypes for `sent` blocks, such as `self generated`, `rebroadcast`, (etc ?)

guilhermelawless:
```bash
grep -r 'compute_rep_votes ('
nano/core_test/node.cpp:                election->compute_rep_votes (transaction);
nano/node/node.hpp:     void compute_rep_votes (nano::transaction const &);
nano/node/node.cpp:void nano::election::compute_rep_votes (nano::transaction const & transaction_a)
```

Is this unused except for tests?

guilhermelawless:
I've noted several things wrong in the approach, will make another PR in the future. Closing this one.

PR #2024, opened 2019-05-23
Factor out network

SergiySW:
Part of #1112


nano/node/transport/transport.hpp
```diff
@@ -1,10 +1,19 @@
...
 
+#include <unordered_set>
+
+#include <boost/asio/buffer.hpp>
+#include <boost/multi_index/hashed_index.hpp>
+#include <boost/multi_index/mem_fun.hpp>
+#include <boost/multi_index/member.hpp>
+#include <boost/multi_index/ordered_index.hpp>
+#include <boost/multi_index/random_access_index.hpp>
+#include <boost/multi_index_container.hpp>
```
wezrule: why are these headers included? It doesn't look like a multi_index container is used in this header file.

nano/node/network.hpp
```diff
@@ -0,0 +1,139 @@
+#pragma once
+
+#include <nano/node/common.hpp>
+#include <nano/node/stats.hpp>
```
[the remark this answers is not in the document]
SergiySW: Sure

nano/node/transport/transport.hpp
SergiySW: It's used in tcp.hpp & udp.hpp. Can be moved to this files

SergiySW: Updated

PR #2026, opened 2019-05-24
Remove unnecessary copies when using lmdb cursors

wezrule:
The cursors used when iterating tables in lmdb already contain the objects in question as a pair (key -> value). A lot of places wrap `first`/`second` to make it more readable, but inadvertently end up copying the objects. I've replaced them with references.

Also found a few other places where a copy is made unnecessarily so have used references there as well.


guilhermelawless:
@wezrule i suppose the copy can't be deleted? Will be a hard habbit to break.

PR #2035, opened 2019-05-27
Bandwidth limiting

argakiig:
The purpose of this PR is to help ease resource usage by allowing a limit to be set on outbound bandwidth sent through transport::channel::send()
new config option `bandwidth_limit` with default of 1572864 bytes per sec
setting `bandwidth_limit` to 0 will leave it unbounded which is the current behavior
add bool not_dropable default false to nano::transport::channel::send allowing calls to channel::send allow for dropping of traffic types if desired
collect rate for at least 50ms, sample into circular buffer trending over 20 samples to ramp up and down rate


nano/lib/stats.cpp
```diff
@@ -545,6 +545,173 @@ std::string nano::stat::detail_to_string (uint32_t key)
 	return res;
 }
 
+std::string nano::stat::detail_raw_to_string (nano::stat::detail detail)
```
cryptocode: This duplicates everything in `detail_to_string`. Maybe convert detail to key and forward to `detail_to_string`.

nano/node/transport/transport.cpp
```diff
@@ -78,8 +79,20 @@ void nano::transport::channel::send (nano::message const & message_a, std::funct
...
-	node.stats.inc (nano::stat::type::message, detail, nano::stat::dir::out);
+	if (!limiter.should_drop (detail, buffer->size ()))
+	{
+		send_buffer (buffer, detail, callback_a);
+		node.stats.inc (nano::stat::type::message, detail, nano::stat::dir::out);
+	}
+	else
+	{
+		node.stats.inc (nano::stat::type::drop, detail, nano::stat::dir::out);
+		if (node.config.logging.network_packet_logging () == true)
```
SergiySW: Can be just `if (node.config.logging.network_packet_logging ())`

nano/node/transport/transport.hpp
```diff
@@ -8,6 +8,25 @@
...
+	// initialize with rate 0 = unbounded
+	bandwidth_limiter (const size_t);
+	bool should_drop (const nano::stat::detail &, const size_t &);
+	size_t get_rate ();
+
+private:
+	//last time rate was adjusted
+	std::chrono::steady_clock::time_point last_poll;
+	//array of message_types that could be dropped
+	std::array<nano::stat::detail, 3> could_drop{ { nano::stat::detail::confirm_req, nano::stat::detail::confirm_ack, nano::stat::detail::publish } };
```
wezrule: Can this be made static constexpr?

nano/node/transport/transport.hpp
```diff
@@ -8,6 +8,27 @@
...
+	bool should_drop (const size_t &);
+	size_t get_rate ();
+
+private:
+	//last time rate was adjusted
+	std::chrono::steady_clock::time_point last_poll;
+	//trend rate over 20 poll periods
+	boost::circular_buffer<size_t> rate_buffer{ 20, 0 };
+	//limit bandwidth to
+	size_t limit;
```
wezrule: This can be const as it's not modified outside of the constructor member initializer list

nano/node/transport/transport.cpp
```diff
@@ -73,13 +75,13 @@ node (node_a)
 {
 }
 
-void nano::transport::channel::send (nano::message const & message_a, std::function<void(boost::system::error_code const &, size_t)> const & callback_a)
+void nano::transport::channel::send (nano::message const & message_a, std::function<void(boost::system::error_code const &, size_t)> const & callback_a, bool const & is_dropable)
```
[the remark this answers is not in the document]
clemahieu: I think we should flip the logic here. By default things should be droppable so we should be true if they must be sent.

nano/node/transport/transport.cpp
```diff
@@ -188,3 +203,46 @@ bool nano::transport::reserved_address (nano::endpoint const & endpoint_a, bool
...
+{
+	using namespace std::chrono_literals;
+	bool result (false);
+	if (limit == 0) //never drop if limit is 0
+	{
+		return result;
+	}
+	std::lock_guard<std::mutex> lock (mutex);
+
+	if (last_poll + 50ms < std::chrono::steady_clock::now ())
```
clemahieu: This seems like it would miss the case where a bunch of traffic has come in within the last 50ms interval.

Can this be implemented simply by dividing the limit/sec by (50ms / 1000ms) and then seeing if rate is greater?  If it is, drop, if it isn't, add message_size to rate and let it pass.

nano/node/active_transactions.cpp
```diff
@@ -72,14 +72,14 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
 
 			if (info.block_count > info.confirmation_height && !node.pending_confirmation_height.is_processing_block (info.head))
 			{
-				auto block (node.store.block_get (transaction_a, info.head));
-				if (!start (block))
+				auto block (this->node.store.block_get (transaction_a, info.head));
```
SergiySW: Is `this` required?

nano/node/active_transactions.cpp
```diff
@@ -333,12 +333,6 @@ void nano::active_transactions::request_loop ()
 	{
 		request_confirm (lock);
 		update_active_difficulty (lock);
-
-		// This prevents unnecessary waiting if stopped is set in-between the above check and now
-		if (stopped)
```
SergiySW: Should it be removed?

nano/node/transport/transport.cpp
```diff
@@ -68,18 +70,31 @@ nano::tcp_endpoint nano::transport::map_endpoint_to_tcp (nano::endpoint const &
...
-void nano::transport::channel::send (nano::message const & message_a, std::function<void(boost::system::error_code const &, size_t)> const & callback_a)
+void nano::transport::channel::send (nano::message const & message_a, std::function<void(boost::system::error_code const &, size_t)> const & callback_a, bool const & is_dropable)
 {
 	callback_visitor visitor;
 	message_a.visit (visitor);
 	auto buffer (message_a.to_bytes ());
 	auto detail (visitor.result);
-	send_buffer (buffer, detail, callback_a);
-	node.stats.inc (nano::stat::type::message, detail, nano::stat::dir::out);
+	if (is_dropable || !limiter.should_drop (buffer->size ()))
```
SergiySW: `!is_droppable` ?

argakiig: flipped

nano/node/active_transactions.cpp
argakiig: no it shouldn't was a missed merge commit

nano/node/active_transactions.cpp
argakiig: missed merge commit :/

PR #2036, opened 2019-05-27
Update broadcast_confirm_req to use confirm_req by hash

SergiySW:
for non-live networks


nano/node/network.cpp
```diff
@@ -278,9 +278,19 @@ void nano::network::broadcast_confirm_req_base (std::shared_ptr<nano::block> blo
 	auto count (0);
 	while (!endpoints_a->empty () && count < max_reps)
 	{
-		nano::confirm_req req (block_a);
 		auto channel (endpoints_a->back ());
-		channel->send (req);
+		// Confirmation request with hash + root
+		if (!node.network_params.network.is_live_network ()
```
cryptocode: maybe flip this so the normal case (live network) is first

SergiySW: Makes sense

PR #2043, opened 2019-05-28
Fix deadlock in tests

wezrule:
On Windows (at least) when running the tests it would often crash on `node.fork_bootstrap_flip` among other tests due to a mutex being locked which is already locked higher up in the stack. It is caused by `bootstrap_initiator.stop ()` being called while the mutex is locked before the `attempt = nullptr` call. The `attempt` is holding the last reference to the `node` shared_ptr which may already have been stopped by the `system::~system` destructor which explicitly calls `node->stop ()`. It then eventually calls `bootstrap_initiator.stop ()` which leaks to a deadlock. There's probably a few ways to solve this, but this could be a problem with other node members (I have seen similar errors before) so I decided to add a check in `node::stop` to see if it has been stopped already before trying to stop again

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -1120,25 +1120,31 @@ void nano::node::start ()
 
 void nano::node::stop ()
 {
-	logger.always_log ("Node stopping");
-	block_processor.stop ();
-	if (block_processor_thread.joinable ())
+	std::unique_lock<std::mutex> lk (stopped_mutex);
+	if (!stopped)
 	{
-		block_processor_thread.join ();
+		stopped = true;
+		lk.unlock ();
+		logger.always_log ("Node stopping");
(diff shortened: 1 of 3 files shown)


nano/node/node.cpp
```diff
@@ -1120,25 +1120,28 @@ void nano::node::start ()
 
 void nano::node::stop ()
 {
-	logger.always_log ("Node stopping");
-	block_processor.stop ();
-	if (block_processor_thread.joinable ())
+	if (!stopped.exchange (true))
```
SergiySW: I guess same can be used in `void nano::bootstrap_server::stop ()` (to not create PR with 1 line)

wezrule: Sure :)

PR #2045, opened 2019-05-29
Attempt to start TCP connections to known UDP peers

SergiySW:


nano/node/transport/tcp.cpp
```diff
@@ -464,6 +464,20 @@ void nano::transport::tcp_channels::ongoing_keepalive ()
...
 	}
+	// Attempt to start TCP connections to known UDP peers
+	if (!node.network_params.network.is_test_network ())
+	{
+		size_t random_count (std::min (static_cast<size_t> (6), static_cast<size_t> (std::ceil (std::sqrt (node.network.udp_channels.size ())))));
+		auto random_udp_list (node.network.udp_channels.random_set (random_count));
+		for (auto udp_channel : random_udp_list)
+		{
+			auto endpoint (udp_channel->get_endpoint ());
+			if (find_channel (nano::transport::map_endpoint_to_tcp (endpoint)) == nullptr)
```
cryptocode: I don't think `nano::transport::map_endpoint_to_tcp` ever returns nullptr

zhyatt:
@SergiySW As discussed we will want network version limits on this and also explore options for cool down periods to avoid hitting repeatedly hitting any peers not properly answering.

SergiySW:
@zhyatt So there is network version limit now with https://github.com/nanocurrency/nano-node/pull/2045/files#diff-7ad572fe8ea10cad777d4f06175f03f2R239
As for cool down periods, last_bootstrap_attempt_tag is used (similar to bootstrap attemps). Do we need extra cooldowns?

zhyatt:
@SergiySW I think this cool down method works well.

PR #2051, opened 2019-06-01
active_transaction.roots bounding by config instead of confirmation rate

argakiig:
remove transaction counter and associated tests
only flush if roots.size() >= active_elections_size
active_elections_size added to config and re-factor simplified should_flush
update config v17 and upgrade path tests
ensure active_elections_size is 250 or greater


nano/node/nodeconfig.hpp
```diff
@@ -66,6 +66,7 @@ class node_config
 	/** Timeout for initiated async operations */
 	std::chrono::seconds tcp_io_timeout{ (network_params.network.is_test_network () && !is_sanitizer_build) ? std::chrono::seconds (5) : std::chrono::seconds (15) };
 	std::chrono::nanoseconds pow_sleep_interval{ 0 };
+	size_t active_elections_size{ 4000 };
```
wezrule: This should be added to the node config upgrade test

PR #2058, opened 2019-06-03
Add active container observer

SergiySW:


cryptocode:
websocket subs will be added to this PR

nano/node/node.cpp
```diff
@@ -343,6 +345,9 @@ startup_time (std::chrono::steady_clock::now ())
 					}
 				}
 			});
+			observers.active_stopped.add ([this](nano::block_hash const & hash_a) {
+				// Dummy
```
cryptocode: this can be added in the websocket PR

PR #2062, opened 2019-06-05
Limit socket queue_size

SergiySW:
to 128 items


cryptocode:
I think we wanted a socket_drop stat for this

PR #2063, opened 2019-06-05
Add launch flag --disable_udp

SergiySW:


nano/node/transport/tcp.cpp
```diff
@@ -242,7 +242,7 @@ void nano::transport::tcp_channels::process_message (nano::message const & messa
 			{
 				node.network.process_message (message_a, channel);
 			}
-			else
+			else if (!node.flags.disable_udp || (message_a.header.type != nano::message_type::confirm_req && message_a.header.type != nano::message_type::confirm_req))
```
cryptocode: Same type checked twice

SergiySW: Thanks

SergiySW: Replaced

PR #2066, opened 2019-06-07
Websocket/extended confirmation support

cryptocode:
Part 2 of #2058.

Includes stopped-election subscription, and option to get confirmations without full block info.

A documentation PR remains.

This could probably use some more unit tests. If we want this in quickly for testing confirmation callbacks, I could add those to a separate PR, otherwise I'll mark as incomplete.

### DIFF
--- a/nano/node/websocket.cpp
+++ b/nano/node/websocket.cpp
@@ -88,11 +88,11 @@ bool nano::websocket::confirmation_options::should_filter (nano::websocket::mess
 	{
 		should_filter_conf_type_l = false;
 	}
-	if (type_text_l == "active_confirmation_height" && confirmation_types & type_active_confirmation_height)
+	else if (type_text_l == "active_confirmation_height" && confirmation_types & type_active_confirmation_height)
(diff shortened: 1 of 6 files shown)


nano/node/websocket.cpp
```diff
@@ -36,7 +81,23 @@ all_local_accounts (options_a.get<bool> ("all_local_accounts", false))
...
 {
-	bool should_filter_l (true);
+	bool should_filter_conf_type_l (true);
+
+	auto type_text_l (message_a.contents.get<std::string> ("confirmation_type"));
+	if (type_text_l == "active_quorum" && confirmation_types & type_active_quorum)
+	{
+		should_filter_conf_type_l = false;
+	}
+	if (type_text_l == "active_confirmation_height" && confirmation_types & type_active_confirmation_height)
```
SergiySW: may be it & next one can be `else if`

cryptocode: yeah it can

PR #2069, opened 2019-06-08
Use correct index when iterating prioritized frontiers

wezrule:
#2053 changed the collection used for prioritizing frontiers to be a `boost::multi_index` container, but was iterating over the default (wrong) index afterwards. There's not really a way to test this as the active roots has no "insertion" order, I do however check that the `roots` does contain the expected frontier hashes though.

I've also modified the `prioritize_frontiers_for_confirmation` so that instead of stopping iterating when the container is full, it continues and replaces the frontier with the lowest amount of uncemented blocks with higher ones.

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -42,7 +42,9 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
 	int test_network_factor = is_test_network ? 1000 : 1;
 	auto roots_size = size ();
 	auto max_elections = (max_broadcast_queue / 4);
+	std::unique_lock<std::mutex> lk (mutex);
 	auto check_time_exceeded = std::chrono::steady_clock::now () >= next_frontier_check;
+	lk.unlock ();
 	auto low_active_elections = roots_size < max_elections;
 	if (check_time_exceeded || (!is_test_network && low_active_elections))
 	{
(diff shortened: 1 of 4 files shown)


nano/node/active_transactions.cpp
```diff
@@ -56,33 +58,34 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
...
 
 			if (info.block_count > info.confirmation_height && !this->node.pending_confirmation_height.is_processing_block (info.head))
 			{
-				auto block (this->node.store.block_get (transaction_a, info.head));
-				if (!this->start (block))
+				auto block (node.store.block_get (transaction_a, info.head));
+				lk.lock ();
+				auto root_exists = roots.find (block->qualified_root ()) != roots.end ();
+				lk.unlock ();
+				if (!root_exists && !start (block))
```
SergiySW: I don't think we should check root existence here because it's in start () / add () function

nano/node/active_transactions.cpp
```diff
@@ -56,33 +58,34 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
...
 		size_t elections_count (0);
-		std::unique_lock<std::mutex> lock (mutex);
+		lk.lock ();
 		while (!priority_cementable_frontiers.empty () && !stopped && elections_count < max_elections)
 		{
-			auto cementable_account = *priority_cementable_frontiers.begin ();
-			priority_cementable_frontiers.erase (priority_cementable_frontiers.begin ());
-			lock.unlock ();
+			auto cementable_account = *priority_cementable_frontiers.get<1> ().begin ();
+			priority_cementable_frontiers.get<1> ().erase (priority_cementable_frontiers.get<1> ().begin ());
```
SergiySW: 2 times get same info `priority_cementable_frontiers.get<1> ().begin ()`

wezrule: Made it a separate variable

nano/node/active_transactions.cpp
wezrule: ok removed

PR #2070, opened 2019-06-08
Factor out online_reps and gap_cache

cryptocode:
Part of #1112 
Moves out corresponding classes and adds an erase function to gap_cache to improve encapsulation (though more work should be done here in the future) + adds a bit of doxygen docs.


nano/node/online_reps.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+#include <nano/lib/numbers.hpp>
+#include <nano/lib/utility.hpp>
+
+#include <memory>
```
wezrule: Is `<memory>` required?

nano/node/gap_cache.hpp
```diff
@@ -0,0 +1,54 @@
+#pragma once
+
+#include <nano/lib/numbers.hpp>
+#include <nano/lib/utility.hpp>
+#include <nano/secure/common.hpp>
```
wezrule: Is the secure include needed in the header?

nano/node/online_reps.cpp
```diff
@@ -0,0 +1,99 @@
+#include <nano/node/node.hpp>
+#include <nano/node/online_reps.hpp>
+
+#include <cassert>
+
+nano::online_reps::online_reps (nano::node & node_a, nano::uint128_t minimum_a) :
```
wezrule: I wonder if during these refactorings we could change what we pass to these classes rather than `node` so it's clearer what dependencies it has (as we have quite a few initialization issues with `node`). For instance this class just needs `ledger` and `max_weight_samples`. Doesn't have to be done now, but something to consider in the future.

nano/node/gap_cache.hpp
cryptocode: Yep

nano/node/online_reps.cpp
cryptocode: Yeah we can do a separate PR that goes through these injections

nano/node/online_reps.hpp
cryptocode: Because of unique_ptr

PR #2077, opened 2019-06-12
Use the MDB_NOMEMINIT flag with the ledger db environment

wezrule:
I researched various options that we might be able to use to improve some performance, and this seemed quite useful. From `lmdb.h` for MDB_NOMEMINIT:
>Don't initialize malloc'd memory before writing to unused spaces in the data file. By default, memory for pages written to the data file is obtained using malloc. While these pages may be reused in subsequent transactions, freshly malloc'd pages will be initialized to zeroes before use. This avoids persisting leftover data from other code (that used the heap and subsequently freed the memory) into the data file. Note that many other system libraries may allocate and free memory from the heap for arbitrary uses. E.g., stdio may use the heap for file I/O buffers. **This initialization step has a modest performance cost so some applications may want to disable it using this flag. This option can be a problem for applications which handle sensitive data like passwords, and it makes memory checkers like Valgrind noisy.**

I've highlighted the last 2 sentences as they are probably the most important. I am now setting this flag for the ledger (data.ldb) but not for (wallets.ldb) as it deals with sensitive information. I have not done any profiling to see how much of an improvement this will give, I don't imagine it will be that impressive but seems worth doing anyway.

I did an ASAN run on Mac which turned up nothing but Valgrind on Ubuntu did show some warnings (as expected). We have a suppressions file but I couldn't get this to work correctly with the necessary wildcards. I found that the header file valgrind.h can be used by clients which provides a handy MACRO `RUNNING_ON_VALGRIND` which we can use to tell if we are launched under valgrind. As per the [documentation](http://valgrind.org/docs/manual/manual-core-adv.html#manual-core-adv.clientreq)
> The Valgrind headers, unlike most of the rest of the code, are under a BSD-style license so you may include them without worrying about license incompatibility.

I downloaded the latest files from the valgrind website but it gave some errors when actually trying to compile on all platforms. The first result when googling `valgrind.h` on google is: https://github.com/adobe/chromium/blob/master/base/third_party/valgrind/valgrind.h which is a few years old but does work, and this file is under BSD license still so should be no problem using it.

### DIFF
--- a/nano/node/lmdb.cpp
+++ b/nano/node/lmdb.cpp
@@ -26,11 +26,11 @@ nano::mdb_env::mdb_env (bool & error_a, boost::filesystem::path const & path_a, 
 			release_assert (status2 == 0);
 			auto running_within_valgrind = (RUNNING_ON_VALGRIND > 0);
 			auto map_size = map_size_a;
-			auto half_gig = 512 * 1024 * 1024;
-			if (running_within_valgrind && map_size_a > half_gig)
+			auto max_valgrind_map_size = 16 * 1024 * 1024;
+			if (running_within_valgrind && map_size_a > max_valgrind_map_size)
 			{
 				// In order to run LMDB under Valgrind, the maximum map size must be smaller than half your available RAM
-				map_size = half_gig;
+				map_size = max_valgrind_map_size;
 			}
 			auto status3 (mdb_env_set_mapsize (environment, map_size));
 			release_assert (status3 == 0);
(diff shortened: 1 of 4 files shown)


SergiySW:
We could probably use RUNNING_ON_VALGRIND to reduce map_size in lmdb.hpp to several MB instead of 128GB as well

wezrule:
> We could probably use RUNNING_ON_VALGRIND to reduce map_size in lmdb.hpp to several MB instead of 128GB as well
Was going to do that in a different PR in case we didn't want this lmdb flag, have added a check in the source file to check  if size is large instead though, I think that may be more robust.

PR #2087, opened 2019-06-15
Bidirectional TCP sockets

SergiySW:


nano/node/bootstrap.hpp
```diff
@@ -289,6 +289,13 @@ class bootstrap_listener final
 std::unique_ptr<seq_con_info_component> collect_seq_con_info (bootstrap_listener & bootstrap_listener, const std::string & name);
 
 class message;
+enum class bootstrap_server_type
```
cryptocode: Maybe a separate PR, but probably wanna rename boostrap_server to tcp_server at this point (and move to separate file?), and bootstrap_server_type - > tcp_server_type.

SergiySW: Good idea! But probably better separate PR with just renames

nano/core_test/node.cpp
```diff
@@ -2617,7 +2617,12 @@ TEST (node, peers)
 	{
 		ASSERT_NO_ERROR (system.poll ());
 	}
-
+	// Wait to finish  TCP node ID handshakes
```
wezrule: Minor whitespace issue, double space after `finish`

nano/core_test/node.cpp
```diff
@@ -2764,6 +2769,75 @@ TEST (node, dont_write_lock_node)
...
+		ASSERT_NO_ERROR (system.poll ());
+	}
+	// Test block confirmation from node 1
+	system.wallet (0)->insert_adhoc (nano::test_genesis_key.prv);
+	bool confirmed (false);
+	system.deadline_set (10s);
+	while (!confirmed)
+	{
+		auto transaction1 (node1->store.tx_begin_read ());
+		auto transaction2 (node1->store.tx_begin_read ());
```
wezrule: should probably be `node2`? (same issue later in the test as well)

SergiySW: Good find!

nano/node/transport/udp.cpp
```diff
@@ -215,6 +215,21 @@ std::shared_ptr<nano::transport::channel_udp> nano::transport::udp_channels::fin
...
+	std::lock_guard<std::mutex> lock (mutex);
+	std::vector<nano::endpoint> endpoints;
+	auto existing (channels.get<node_id_tag> ().equal_range (node_id_a));
+	for (auto & record : boost::make_iterator_range (existing))
+	{
+		endpoints.push_back (record.endpoint ());
+	}
+	for (auto & endpoint : endpoints)
+	{
+		channels.get<endpoint_tag> ().erase (endpoint);
```
clemahieu: I believe multiindex has a range erase method that can use .first and .second of equal_range directly.

nano/node/transport/tcp.cpp
```diff
@@ -514,6 +566,12 @@ void nano::transport::tcp_channels::start_tcp_receive_node_id (std::shared_ptr<n
 										{
 											callback_a (channel_a);
 										}
+										// Listen for possible responses
+										channel_a->response_server = std::make_shared<nano::bootstrap_server> (channel_a->socket, node_l);
+										channel_a->response_server->keepalive_first = false;
+										channel_a->response_server->type = nano::bootstrap_server_type::realtime_response_server;
+										channel_a->response_server->remote_node_id = channel_a->get_node_id ().get ();
```
clemahieu: It seems like bootstrap_server should be holding a channel_tcp directly instead of the socket + tcp_endpoint.   Then we can eliminate duplicate data like the node_id, tcp_endpoint, node shared_ptr etc.

This can be a separate PR though.

SergiySW: Replaced partially. remote_endpoint remains to remove response_channel on closing until we are 100% sure that socket returning correct endpoint if remote peer closed connection. Also removing node shared pointer & replacing with channel node reference cause some random segfaults as it is

PR #2088, opened 2019-06-15
Fix duplicate iterating in conf height processor with self and circular sends

wezrule:
There were reports of duplicates found in the callbacks, and this was cause by self sends and sends to accounts who sent back to the originating account. While this didn't compromise the actual confirmation heights being set it did extra processing for no reason and in the observer multiple callbacks for the same inactive blocks could be sent. This has been rectified by having a separate "iterated" height in addition to a pending confirmation height and checking this. I've updated all confirmation_height tests to check callback counts and a new test to check for the case where a block was being confirmed and it iterates over a block which was voted on previously and added to pending_confirmation_height. So the callback is called after the election, then it is seen as inactive when it gets iterated over by a successor and another callback is made for that.

Also removed the block arrival check in the websockets as it didn't report all inactive blocks, found during a discussion with @cryptocode & @SergiySW.

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -1497,6 +1497,16 @@ TEST (bootstrap, tcp_node_id_handshake)
 	}
 }
 
+namespace
+{
+void add_callback_stats (nano::node & node)
+{
+	node.observers.blocks.add ([& stats = node.stats](nano::election_status const & status_a, nano::account const &, nano::amount const &, bool) {
+		stats.inc (nano::stat::type::http_callback, nano::stat::detail::http_callback, nano::stat::dir::out);
+	});
+}
+}
+
 TEST (confirmation_height, single)
 {
 	auto amount (std::numeric_limits<nano::uint128_t>::max ());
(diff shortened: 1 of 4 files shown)


nano/core_test/network.cpp
```diff
@@ -2134,9 +2164,62 @@ TEST (confirmation_height, observers)
...
 	ASSERT_EQ (1, node1->stats.count (nano::stat::type::http_callback, nano::stat::detail::http_callback, nano::stat::dir::out));
 }
 
+namespace nano
+{
+TEST (confirmation_height, pending_observer_callbacks)
+{
+	nano::system system;
+	nano::node_flags node_flags;
+	node_flags.delay_frontier_confirmation_height_updating = true;
```
cryptocode: This has been added to quite a few tests. For another PR, but I wonder if this should be default true if testnet, then set it to false in `node.bootstrap_confirm_frontiers`

wezrule: Yeh probably :)

Issue #2083, opened 2019-06-13
Add optional argument to block_count RPC for including cemented blocks


zhyatt:
Cement block counts are currently only available via the [`--debug_cemented_block_count`](https://docs.nano.org/commands/command-line-interface/#-debug_cemented_block_count) CLI command but an RPC option is desired. This was originally kept as CLI command due to the lengthy calculation process for cemented counts (10-15s has been seen) but making it optional via RPC makes sense.

Add optional "include_cemented" boolean parameter to [`block_count`](https://docs.nano.org/commands/rpc-protocol/#block_count) RPC to return `cemented` count in addition to `count`(total) and `unchecked`.

wezrule:
Done in #2089. Requires `enable_control` to be set in `rpc_config.json`

PR #2089, opened 2019-06-15
Add optional cemented count to block_count RPC command

wezrule:
This solves #2083


cryptocode:
@wezrule https://docs.nano.org/commands/rpc-protocol/#block_count needs a doc update

Issue #2080, opened 2019-06-13
Create websocket subscription for active_difficulty changes


zhyatt:
Currently the [`active_difficulty`](https://docs.nano.org/commands/rpc-protocol/#active_difficulty) RPC must be polled to monitor changes in work difficulty seen on the network. To assist integrations managing work generation efforts external to the node, add a new websocket subscription for `active_difficulty`.

This requires creation of a new observer set that would monitor and fire notifications with each cycle of difficulty changes (16-36sec depending on network conditions). These notifications should include the same details as found in the [`active_difficulty`](https://docs.nano.org/commands/rpc-protocol/#active_difficulty) RPC - `network_minimum`, `network_current` and `multiplier`.

A single subscription without additional options for this should work.

zhyatt:
@guilhermelawless

zhyatt:
Thanks for offering help on this @chrislinegar.

chrislinegar:
Hi @guilhermelawless, I've submitted a PR with a first pass at this: https://github.com/nanocurrency/nano-node/pull/2091

I couldn't add a reviewer on the PR though, is this something you'd be able to have a look at?

Thanks in advance!

zhyatt:
Solved with #2091 and #2109

PR #2091, opened 2019-06-16
Web socket subscription for active difficulty

chrislinegar:
This implements #2080 - monitors changes in difficulty and notifies via web socket if the active difficulty changes

### DIFF
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -220,7 +220,7 @@ TEST (websocket, subscribe_active_difficulty)
 	nano::from_string_hex (message_contents.get<std::string> ("network_current"), network_current);
 	ASSERT_EQ (network_current, node1->active.active_difficulty ());
 
-	double multiplier = std::stod (message_contents.get<std::string> ("multiplier"));
+	double multiplier = message_contents.get<double> ("multiplier");
 	ASSERT_NEAR (multiplier, nano::difficulty::to_multiplier (node1->active.active_difficulty (), node1->network_params.network.publish_threshold), 1e-6);
 
 	node1->stop ();
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -161,7 +161,7 @@ TEST (websocket, subscription_edge)
 }
 
 // Test client subscribing to changes in active_difficulty
-TEST (websocket, subscribe_active_difficulty)
+TEST (websocket, active_difficulty)
 {
 	nano::system system (24000, 1);
 	nano::node_init init1;
@@ -182,7 +182,7 @@ TEST (websocket, subscribe_active_difficulty)
 		auto response = websocket_test_call ("::1", "24078", R"json({"action": "subscribe", "topic": "active_difficulty", "ack": true})json", true, true);
 		return response;
 	});
-	auto client_future = std::async (client_task);
+	auto client_future = std::async (std::launch::async, client_task);
 
 	// Wait for acknowledge
 	system.deadline_set (5s);
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -160,6 +160,40 @@ TEST (websocket, subscription_edge)
 	node1->stop ();
 }
 
+/** Tests clients subscribing multiple times or unsubscribing without a subscription */
+TEST (websocket, subscribe_active_difficulty)
+{
+  nano::system system (24000, 1);
+  nano::node_init init1;
+  nano::node_config config;
+  nano::node_flags node_flags;
+  config.websocket_config.enabled = true;
+  config.websocket_config.port = 24078;
+
+  auto node1 (std::make_shared<nano::node> (init1, system.io_ctx, nano::unique_path (), system.alarm, config, system.work, node_flags));
+  node1->start ();
+  system.nodes.push_back (node1);
+
+  ASSERT_EQ (0, node1->websocket_server->subscriber_count (nano::websocket::topic::active_difficulty));
+
+  // First subscription
+  {
+    ack_ready = false;
+    std::thread subscription_thread ([]() {
+      websocket_test_call ("::1", "24078", R"json({"action": "subscribe", "topic": "active_difficulty", "ack": true})json", true, false);
+    });
+    system.deadline_set (5s);
+    while (!ack_ready)
+    {
+      ASSERT_NO_ERROR (system.poll ());
+    }
+    subscription_thread.join ();
+    ASSERT_EQ (1, node1->websocket_server->subscriber_count (nano::websocket::topic::active_difficulty));
+  }
+
+  node1->stop ();
+}
+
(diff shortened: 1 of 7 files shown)


nano/node/active_transactions.cpp
```diff
@@ -639,7 +639,12 @@ void nano::active_transactions::update_active_difficulty (std::unique_lock<std::
 	auto sum (std::accumulate (multipliers_cb.begin (), multipliers_cb.end (), double(0)));
 	auto difficulty = nano::difficulty::from_multiplier (sum / multipliers_cb.size (), node.network_params.network.publish_threshold);
 	assert (difficulty >= node.network_params.network.publish_threshold);
+	
+	bool notify_change = trended_active_difficulty != difficulty;
```
chrislinegar: Currently implemented to only notify if the active difficulty changes - is this correct or should it notify on every iteration even if active difficulty remains the same?

cryptocode:
Thanks @chrislinegar. Before reviewing this, could you fix the formatting issues reported by CI? There's a `clang-format-all` script under the ci directory.

nano/core_test/websocket.cpp
```diff
@@ -160,6 +160,72 @@ TEST (websocket, subscription_edge)
...
+	auto message_contents = event.get_child ("message");
+	uint64_t network_minimum;
+	nano::from_string_hex (message_contents.get<std::string> ("network_minimum"), network_minimum);
+	ASSERT_EQ (network_minimum, node1->network_params.network.publish_threshold);
+
+	uint64_t network_current;
+	nano::from_string_hex (message_contents.get<std::string> ("network_current"), network_current);
+	ASSERT_EQ (network_current, node1->active.active_difficulty ());
+
+	double multiplier = std::stod (message_contents.get<std::string> ("multiplier"));
```
guilhermelawless: `boost::property_tree` has a specialized `get` for **double**

chrislinegar: good point, done!

chrislinegar:
> Thanks @chrislinegar. Before reviewing this, could you fix the formatting issues reported by CI? There's a `clang-format-all` script under the ci directory.

thanks, done :)

chrislinegar:
`continuous-integration/appveyor/pr ` was failing on last commit - will wait for it to run again and investigate

nano/core_test/websocket.cpp
```diff
@@ -160,6 +160,72 @@ TEST (websocket, subscription_edge)
...
+
+	ASSERT_EQ (0, node1->websocket_server->subscriber_count (nano::websocket::topic::active_difficulty));
+
+	// Subscribe to active_difficulty and wait for response asynchronously
+	ack_ready = false;
+	auto client_task = ([&node1]() -> boost::optional<std::string> {
+		auto response = websocket_test_call ("::1", "24078", R"json({"action": "subscribe", "topic": "active_difficulty", "ack": true})json", true, true);
+		return response;
+	});
+	auto client_future = std::async (client_task);
```
guilhermelawless: Seems like the problem with travis might be here, can you try forcing the policy? Would be something like:
```cpp
auto client_future = std::async (std::launch::async, client_task);
```

Without forcing, it could fallback to `defer` which makes the test fail. However, forcing the policy can [throw](https://en.cppreference.com/w/cpp/thread/async). If that doesn't work, could do similarly to other tests where a flag is passed by reference to a `std::thread` directly.

nano/core_test/websocket.cpp
```diff
@@ -160,6 +160,72 @@ TEST (websocket, subscription_edge)
 	node1->stop ();
 }
 
+// Test client subscribing to changes in active_difficulty
+TEST (websocket, subscribe_active_difficulty)
```
guilhermelawless: Suggeset keeping the naming convention of the tests: `websocket.active_difficulty`

nano/core_test/websocket.cpp
chrislinegar: thanks! fixed now, tests passing

nano/node/active_transactions.cpp
```diff
@@ -639,7 +639,13 @@ void nano::active_transactions::update_active_difficulty (std::unique_lock<std::
 	auto sum (std::accumulate (multipliers_cb.begin (), multipliers_cb.end (), double(0)));
 	auto difficulty = nano::difficulty::from_multiplier (sum / multipliers_cb.size (), node.network_params.network.publish_threshold);
 	assert (difficulty >= node.network_params.network.publish_threshold);
+
+	bool notify_change = trended_active_difficulty != difficulty;
 	trended_active_difficulty = difficulty;
+	if (notify_change)
+	{
+		difficulty_observer (trended_active_difficulty);
```
cryptocode: To simplify a bit, can the `difficulty_observer` std::function be removed, and then call `node.observers.difficulty.notify (trended_active_difficulty);` directly? I believe that's the general pattern we use elsewhere.

nano/node/active_transactions.cpp
cryptocode: Only when it changes I think

guilhermelawless:
Current failure you might be seeing is outlined in #2101

guilhermelawless:
Thank you @chrislinegar . Only needs another pass of `clang-format-all`

nano/node/active_transactions.cpp
zhyatt: If the call is sufficiently lightweight, I could see value in triggering it every cycle, specifically to ensure consistent data points for any tracking/visualizations, and also as a mechanism that allows watching for that consistency. If gaps are seen, this allows polling for difficulty separately through RPC as fallback or alerting of a possible issue (if these notifications aren't seen for a time period). Any additional thoughts @cryptocode or @guilhermelawless ?

guilhermelawless: I think that's a good idea, it's very light data especially considering the ~20 second period.

cryptocode: Sounds good to me if it simplifies clients and is low frequency.

chrislinegar: The `request_loop` runs faster in testing - will publishing on every iteration have an impact on other tests if running faster?:

`request_interval_ms = is_test_network () ? (is_sanitizer_build ? 100 : 20) : 16000;`

chrislinegar:
> Current failure you might be seeing is outlined in #2101

Just to confirm - is there any action needed on this PR for this issue?

guilhermelawless:
> Just to confirm - is there any action needed on this PR for this issue?

No, it's resolved.

nano/node/active_transactions.cpp
guilhermelawless: Not a problem I believe, there is much higher frequency data being passed around. Tests should pass.

zhyatt:
@chrislinegar Are you considering this PR wrapped up on your end (all comments resolved)?

nano/node/active_transactions.cpp
zhyatt: We agree with @guilhermelawless, sending it every cycle should be fine.

PR #2092, opened 2019-06-17
Restrict early vote generator wakeup

guilhermelawless:
- Simplifies vote_generator::run() and adds a check for spurious wakeups.
- Adding a hash only notifies for complete vote by hash packets (>=12
hashes ready). This is the improvement.

Under similar testing conditions, saw an increase in the number of hashes packed per vote, both for a 300 tps test and a 80 tps test. Under low tps a 50ms delay always occurs.

imagem
*Note* - not scaled to the number of hashes per vote.

------

@cryptocode For previous tests I had always locally pushed blocks but turns out this is the best case scenario for the vote generator constant delay and results are very different. These ones and any future vote tests from me will be done with blocks being pushed by another node. (Thank you @Joohansson).

### DIFF
--- a/nano/node/voting.cpp
+++ b/nano/node/voting.cpp
@@ -18,11 +18,13 @@ thread ([this]() { run (); })
 
 void nano::vote_generator::add (nano::block_hash const & hash_a)
 {
+	std::unique_lock<std::mutex> lock (mutex);
+	hashes.push_back (hash_a);
+	if (hashes.size () >= 12)
 	{
-		std::lock_guard<std::mutex> lock (mutex);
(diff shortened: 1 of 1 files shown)


guilhermelawless:
With this PR it also becomes easier to setup an algorithm to automatically adjust the vote generator delay, with a floor and a ceiling, based on the rate of votes arriving. Current idea is to take a look at the difference (derivative) in the `hashes` queue between sleeps and incrementally increase or decrease the delay based on that.

guilhermelawless:
Closing, came up with something better

Issue #2101, opened 2019-06-20
Test bug introduced by #2090


guilhermelawless:
#2090 

https://github.com/nanocurrency/nano-node/blob/a54f2925d405a25c59fbc88bb09a011dcc1248a9/nano/rpc_test/rpc.cpp#L6400-L6401

https://github.com/nanocurrency/nano-node/blob/a54f2925d405a25c59fbc88bb09a011dcc1248a9/nano/node/active_transactions.cpp#L619

These two blocks are incompatible (I probably forgot to test in Debug, sorry). Would send a fix but not sure how to escape this without removing the `assert`

PR #2102, opened 2019-06-21
Fix locking issue and intermittent failure in rpc.active_difficulty

cryptocode:
Closes https://github.com/nanocurrency/nano-node/issues/2101
Thanks to @guilhermelawless for validating the trend-sequence fix

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -6397,10 +6397,10 @@ TEST (rpc, active_difficulty)
 	request.put ("action", "active_difficulty");
 	node->active.multipliers_cb.push_front (1.5);
 	node->active.multipliers_cb.push_front (4.2);
-	std::mutex mutex;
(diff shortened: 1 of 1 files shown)


nano/rpc_test/rpc.cpp
```diff
@@ -6397,10 +6397,10 @@ TEST (rpc, active_difficulty)
 	request.put ("action", "active_difficulty");
 	node->active.multipliers_cb.push_front (1.5);
 	node->active.multipliers_cb.push_front (4.2);
-	std::mutex mutex;
-	std::unique_lock<std::mutex> lock (mutex);
+	std::unique_lock<std::mutex> lock (node->active.mutex);
```
guilhermelawless: Best lock before pushing the values. Also fixes an intermittent issue where a `1` is placed in between them and the test fails.

cryptocode: Thanks, moved the lock

PR #2103, opened 2019-06-21
Flag to disable TCP realtime network

SergiySW:
Similar to disabling UDP


wezrule:
Is it possible to add a test for this?
Also what if a user disables both udp and tcp, perhaps this combination should be prevented?

PR #2105, opened 2019-06-23
Improve batching multiple blocks pending confirmation height processing

wezrule:
This PR does a few things:
1 - Introduces a minimum time when there are blocks pending confirmation height to batch writes, currently set to 50ms but can be change via a new config option `conf_height_processor_batch_min_time`
2 - Checks whether there are other txn writes pending (currently just `process_batch`). If there is and the confirmation height processor has other blocks pending confirmation height then it will continue processing them until it runs out or the `process_batch` write txn lock is finished. 
3 - (Unrelated) Removing the assert in `nano::active_transactions::add`. We have multiple checks for elections but I still think implicit confirmations can cause issues here (@cryptocode had run into the `assert`) and I think the following can happen:
We check if block is confirmed (no)
Background conf height processor implicitly makes it confirmed
We check it is not confirmed and assert.
4 - (Unrelated) A test didn't exist in the case where blocks differed inbetween reads/writes. Due to the `assert`s which will always fire as the block doesn't exist, I now check whether we are in the `testnet`

Note for beta testers:
Using the following RPC command, check that the `pending_confirmation_height` -> `pending` entry is kept low during any stress tests.
`curl -d '{"action":"stats", "type":"objects"}'`
 Those that have close cemented and total block counts before any such tests, should also see them remain close through any spam as well.
`curl -d '{"action":"block_count", "include_cemented":"true"}'`
Both RPC commands require `enable_control` to be set in `rpc_config.json`

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -7,12 +7,13 @@
 
 std::chrono::milliseconds constexpr nano::block_processor::confirmation_request_delay;
 
-nano::block_processor::block_processor (nano::node & node_a) :
+nano::block_processor::block_processor (nano::node & node_a, nano::write_database_queue & write_database_queue_a) :
 generator (node_a),
 stopped (false),
 active (false),
 next_log (std::chrono::steady_clock::now ()),
-node (node_a)
+node (node_a),
+write_database_queue (write_database_queue_a)
 {
 }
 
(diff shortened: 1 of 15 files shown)


nano/node/blockprocessor.cpp
```diff
@@ -241,6 +242,7 @@ void nano::block_processor::process_batch (std::unique_lock<std::mutex> & lock_a
 		}
 	}
 	lock_a.unlock ();
+	auto write_guard = write_database_queue.wait (nano::writer::process_batch);
```
cryptocode: None of the `auto write_guard` variables in the PR appears to be used - can they be removed?

wezrule: It is similar to something like `lock_guard`, which isn't directly called but will pop the writer off the front of the queue when it goes out of scope.

cryptocode: Ah yes, missed that in the surface scan.

Could we name it `scoped_write_guard` ?

PR #2106, opened 2019-06-24
Fuzzer

cryptocode:
Based on llvm libfuzzer and requires a very recent clang version. Tested with clang-9 on Ubuntu. Combine with a ASAN and other sanitizers. This is minimal at this stage but it still obtains decent coverage.

(Incomplete, pushing now to get automated test integration going)


cryptocode:
Closing in favour of a different approach that's being researched. This version pumped data through UDP, but now we're mainly on TCP even for live traffic so a different approach will be needed for sufficient coverage (a challenge is that currently the tcp server will indirectly disconnect on invalid data through the shared_from_this design, and reconnecting is way too slow. There are other issues with the fuzzer not making progress when using tcp, possibly due to the effect of input mutations being on async handlers on different threads (?), timeouts when the fuzzer is sending insufficient data amounts, and possibly tcp flow control; the fuzzer client only sends and doesn't read.

PR #2110, opened 2019-06-28
Vote generator max latency

guilhermelawless:
Some things only show up during beta net testing due to the variety of hardware. I am seeing some nodes pack many votes even at 1 tps - which should not happen, they should be sent after 50ms with a single hash.

Likeliest explanation for me is that these are lower core count nodes where threads are fighting each other for time. Votes get queued up and hit `vote_generator::add` in quick succession, going over `vote_generator_threshold` and so the node continues to wait for more.

I'm adding a check that does not change current logic besides adding a maximum latency limit.

This change prevents slower nodes from packing too many hashes at low load, something that increases confirmation time (although bounded to around 500ms). A config option `vote_generator_maximum_latency` is added, defaulting to 150ms. This is the absolute maximum delay (counting from the first hash being inserted), after which the node will immediately send the vote.

This change only matters for low-medium load (up to `12/.150 = 80`tps), and it ensures we should continue to see sub-second confirmation while still enjoying the benefit of VBH at any load.

A discussion point would be which maximum latency to go for.


guilhermelawless:
Replacing with simpler approach.

PR #2116, opened 2019-07-02
remove else and always flush if over config.json active_elections_size

argakiig:


clemahieu:
Can we get a test that will trigger the incorrect behavior.

PR #2132, opened 2019-07-05
Cleanup inactive channels from rep crawler

SergiySW:


nano/node/repcrawler.cpp
```diff
@@ -205,6 +198,45 @@ void nano::rep_crawler::on_rep_request (std::shared_ptr<nano::transport::channel
...
+		}
+	}
+	// Remove reps with inactive channels
+	for (auto i : channels)
+	{
+		bool equal (false);
+		if (i->get_type () == nano::transport::transport_type::tcp)
+		{
+			auto find_channel (node.network.tcp_channels.find_channel (i->get_tcp_endpoint ()));
+			if (find_channel != nullptr && *find_channel == *dynamic_cast<nano::transport::channel_tcp *> (i.get ()))
```
wezrule: As you don't check if it's null (and this class doesn't use virtual inheritance) this can just be a `static_cast`

nano/node/repcrawler.cpp
```diff
@@ -205,6 +198,45 @@ void nano::rep_crawler::on_rep_request (std::shared_ptr<nano::transport::channel
...
+		}
+		else if (i->get_type () == nano::transport::transport_type::udp)
+		{
+			auto find_channel (node.network.udp_channels.channel (i->get_endpoint ()));
+			if (find_channel != nullptr && *find_channel == *static_cast<nano::transport::channel_udp *> (i.get ()))
+			{
+				equal = true;
+			}
+		}
+		if (!equal)
```
cryptocode: Could the if/else above be deleted and this line be replaced with 

`if (!node.network.find_channel(i->get_endpoint ()))` 

The tcp channel's get_endpoint converts to tcp_endpoint. Would simplify the code a lot of it works.

PR #2144, opened 2019-07-11
Set genesis account to have a confirmation height of 1 after initial upgrade

wezrule:
We want the genesis account to have a minimum cemented count of 1 at all times. This is done elsewhere but not during the initial upgrade.


cryptocode:
Should resetting conf' height do the same?

wezrule:
I guess it depends where we think the logic for it should go. It does get checked in the cli where's it's used (`reset_confirmation_heights` in the cli.cpp). Actually that's a lie, it doesn't do it in the instance of a per account clear, it probably should... I will update that. Thanks

PR #2152, opened 2019-07-14
Fix node_id assign for bootstrap_server

SergiySW:


nano/node/bootstrap.cpp
```diff
@@ -2436,10 +2436,10 @@ class request_response_visitor : public nano::message_visitor
 		}
 		else if (message_a.response)
 		{
-			auto node_id (message_a.response->first);
-			connection->remote_node_id = node_id;
+			nano::account node_id (message_a.response->first);
```
wezrule: This existed already, but seeing as you are changing the variable to have an explicit type it could be made a const reference to remove a copy.

SergiySW: Good idea

wezrule: Can probably change the other one a bit lower as well then

PR #2154, opened 2019-07-15
Prioritize node wallet frontiers during background confirmations

wezrule:
Currently we prioritize based on the number of uncemented blocks on all accounts in the ledger. This PR still does that, but gives a greater priority to all node wallet accounts first (again sorted by uncemented blocks).

(Unrelated) nano/node/wallet.hpp I noticed some function delcarations had no definitions so removed them.

TSAN has been run on core_tests and showed no warnings.


nano/node/active_transactions.cpp
```diff
@@ -346,73 +350,128 @@ void nano::active_transactions::request_loop ()
...
+							// If it exists in normal priority collection delete from there.
+							auto it = priority_cementable_frontiers.find (account);
+							if (it != priority_cementable_frontiers.end ())
+							{
+								priority_cementable_frontiers.erase (it);
+							}
+
+							prioritize_account_for_confirmation (priority_local_cementable_frontiers, priority_local_cementable_frontiers_size, account, info);
+
+							if (timer.since_start () >= time_a)
```
SergiySW: Not sure if wallets mutex should be locked for 2 seconds

nano/node/active_transactions.cpp
```diff
@@ -346,73 +350,128 @@ void nano::active_transactions::request_loop ()
...
+							}
 						}
+						next_local_frontier_account = account.number () + 1;
 					}
-					else
+					// Go back to the beginning when we have reached the end of the local accounts and make sure we skip local accounts
+					if (i == n)
 					{
-						priority_cementable_frontiers.emplace (account, num_uncemented);
+						next_local_frontier_account = 0;
```
SergiySW: By default should be nano::wallet_store::special_count according to nano::wallet_store::begin () to not iterate over tech fields. Same for default value in header

nano/node/active_transactions.cpp
```diff
@@ -346,73 +350,128 @@ void nano::active_transactions::request_loop ()
...
+							}
+
+							prioritize_account_for_confirmation (priority_local_cementable_frontiers, priority_local_cementable_frontiers_size, account, info);
+
+							if (timer.since_start () >= time_a)
+							{
+								break;
+							}
 						}
+						next_local_frontier_account = account.number () + 1;
```
SergiySW: Also not sure if this value is required. One for all wallets. Only if store last checked wallet or something like that

nano/node/active_transactions.cpp
wezrule: Good point, I have lowered it to 50ms

nano/node/active_transactions.cpp
wezrule: Updated, thanks

nano/node/active_transactions.cpp
wezrule: As the time spent iterating has been lowered and some services may have many accounts in a single wallet it may need to resume iterating the same wallet in the next request loop. I have added another collection so that we can skip wallets which have already been iterated (and so get straight to ones which haven't), as well as having a "next frontier account" per wallet as well in case there are many accounts in this wallet.

PR #2155, opened 2019-07-16
Improve votes caching for same account & increase max cache size

SergiySW:


nano/node/voting.cpp
```diff
@@ -114,6 +114,16 @@ void nano::votes_cache::add (std::shared_ptr<nano::vote> const & vote_a)
...
 			cache.get<1> ().modify (existing, [vote_a](nano::cached_votes & cache_a) {
+				// Erase old vote for same representative & hash
+				for (auto i (cache_a.votes.begin ()), n (cache_a.votes.end ()); i != n; ++i)
+				{
+					if ((*i)->account == vote_a->account)
+					{
+						cache_a.votes.erase (i);
+						break;
+					}
+				}
```
wezrule: Does the order of `cache_a.votes` matter? If not then an optimization can be made in the erase/add case by changing the vote in-place rather than removing one first, which requires shifting all elements after it to the left by 1.

Alternatively if order must be kept then you could use the `std::find` algorithm (or remove-erase idiom but that would iterate all elements, maybe that's ok?) instead of a raw loop, e.g:
```
auto it = std::find (cache_a.votes.begin (), cache_a.votes.end (), vote_a->account);
if (it != cache_a.votes.end())
{
    cache_a.votes.erase (it);
}
```

SergiySW: Order is not important

wezrule: Looks good. You could still use the `find` stl algorithm if you wanted, but up to you (it basically does the same loop).

PR #2156, opened 2019-07-16
Remove querying pending confirmation height in various operations

wezrule:
During the development of the confirmation height processor (at that time) it was quite slow, so anyone calling `block_confirm` would have seen inconsistent results from `block_info` which would state that a block was unconfirmed. It is decided that any block pending confirmation or currently undergoing it would show up as confirmed as well. But this leads to other inconsistencies as blocks below it may not be seen as confirmed but this was seen as acceptable at the time. Now after many performance improvements since then, and that in v19 most nodes upgrading to v20 will have synced cemented counts anyway this restriction has been removed so a block will only show as confirmed in RPCs if all blocks below it are also marked as confirmed.


zhyatt:
As discussed in our meeting we are holding off on merging this until closer to RC period.

PR #2158, opened 2019-07-16
Fix compile errors in release mode after warning->error change

cryptocode:
Using the (void)var idiom used elsewhere in the code base for unused vars (can move to [[maybe_unused]] in c++17). There's also a default-in-switch-related uninitialized error fixed.


wezrule:
LGTM. gcc 8.2 and clang 6.0 on Ubuntu both currently compile on master in release mode. I've requested @argakiig to add a mac/release build to travis to catch future errors. I suspect we will likely come across more of these warning -> errors in the future with different version of the same compiler as well.

PR #2159, opened 2019-07-17
Cache recently arrived votes

SergiySW:
without active elections


nano/node/election.cpp
```diff
@@ -298,3 +298,22 @@ void nano::election::clear_blocks ()
...
+void nano::election::insert_inactive_votes_cache ()
+{
+	auto winner_hash (status.winner->hash ());
+	auto reps (node.active.find_inactive_votes_cache (winner_hash));
+	for (auto & rep : reps)
+	{
+		auto last_vote_it (last_votes.find (rep));
+		if (last_vote_it == last_votes.end ())
+		{
+			last_votes[rep] = { std::chrono::steady_clock::time_point::min (), 0, winner_hash };
```
clemahieu: Let's get a test for this, showing we wouldn't clobber an existing vote as this is important.

wezrule: Does this code need to be performant? If so you may consider using lower_bound/insert instead of operator[] (Effective STL - Item 24, Scott Meyers)

SergiySW: Is it true for unordered_map as well?

wezrule: Ah no sorry thought I checked that, it only works with map. Actually because we don't care about updating (only inserting) the whole thing can be replace with an `emplace`, something like:
`last_votes.emplace (rep, vote_info{std::chrono::steady_clock::time_point::min (), 0, winner_hash});`
This only inserts if the key doesn't already exist:
https://en.cppreference.com/w/cpp/container/unordered_map/emplace

SergiySW: Replaced

nano/node/gap_cache.cpp
```diff
@@ -54,48 +54,53 @@ void nano::gap_cache::vote (std::shared_ptr<nano::vote> vote_a)
...
+	{
+		if (tally >= node.config.online_weight_minimum.number ())
+		{
+			start_bootstrap = true;
+		}
+	}
+	else if (!node.flags.disable_legacy_bootstrap && tally > bootstrap_threshold (transaction_a))
+	{
+		start_bootstrap = true;
+	}
```
cryptocode: Identical if/else in `gap_cache::vote`, so maybe extract this into `should_bootstrap` and call that from both places.

PR #2160, opened 2019-07-18
Update entry script

argakiig:
Use ENTRYPOINT CMD instead of CMD to allow for greater flexibility
ex. `docker logs <container>` after running with `nano_node --wallet_list`
```
EXECUTING: nano_node --wallet_list

Wallet ID: 4633783D98EFE13AC7884FB8299E39032358575C4DA696B9800C6EB0AD1634A4
nano_1womoper6rkhq1atc9y5rsqr6jzatia9ch4mezmtxdixhip68aibxtuefadz
```
If vacuum limit is set and db size is greater than limit vacuum before launching daemon
container stops if daemon dies allowing docker to restart if configured
This is a change from the how things currently run as the size is not checked and the daemon restarted if over size.
```
Usage:
   /entry.sh nano_node [daemon] [cli_options] [-l] [-v size]
     daemon
       start as daemon
     cli_options
       nano_node cli options <see nano_node --help>
     -l
       log to console <use docker logs {container}>
     -v<size>
       vacuum database if over size GB on startup
   /entry.sh bash [other]
     other
       bash pass through
   /entry.sh [*]
     *
       usage
 default:
   /entry.sh nano_node daemon -l
```


argakiig:
moving https://github.com/nanocurrency/nano-node/commit/555c555aab91f8fb11b867a5fbc3a7cd30a992ad to a separate PR to allow for further discussion

argakiig:
doc updates at https://github.com/nanocurrency/nano-docs/pull/76

PR #2161, opened 2019-07-18
Filter recent elections in update_active_difficulty

guilhermelawless:
If an election was inserted within the current loop, don't consider it for the median calculation.

This should solve the effect during low load where active difficulty gets quite high while sampling only one or two very recent elections.

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -623,9 +623,11 @@ void nano::active_transactions::update_active_difficulty (std::unique_lock<std::
 	{
 		std::vector<uint64_t> active_root_difficulties;
 		active_root_difficulties.reserve (roots.size ());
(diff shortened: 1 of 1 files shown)


nano/node/active_transactions.cpp
```diff
@@ -623,9 +623,11 @@ void nano::active_transactions::update_active_difficulty (std::unique_lock<std::
 	{
 		std::vector<uint64_t> active_root_difficulties;
 		active_root_difficulties.reserve (roots.size ());
+		auto min_election_time (std::chrono::milliseconds (node.network_params.network.is_test_network () ? 0 : 2000));
```
clemahieu: I think we should wait the amount in nano::active_transactions::request_loop, basically the average of any blocks that are stuck around since the last cycle.

PR #2162, opened 2019-07-20
Allow the RPC server to be run on a different system

wezrule:
Currently the child/out of process RPC server uses the `address` config option in `rpc_config.json` when communicating to the node with the ipc port. This is incorrect as this address is for filtering out unauthorised RPC access not the ipc address (location of the node). It will still work ok with the default address (node/rpc on the same computer) or an address where the node is located on, but will fail for everything else.

I've added `ipc_address` to the `process` config section, as well as `version` as it is now an upgrade.


cryptocode:
looks good pending formatting

PR #2163, opened 2019-07-20
Assert for disabled bootstrap modes

SergiySW:


SergiySW:
@cryptocode indeed, thanks!

PR #2168, opened 2019-07-22
RPC publish optional work_watch flag to add to work_watcher

argakiig:
Adds flag work_watch, default true, to rpc publish allowing for blocks processed via RPC call to not be eligible for dropping from active ensuring the block is not dropped from active_transactions until confirmed. Messages will continue to originate from the block as long as the block is not confirmed

With this change publishes originating from the local node will not be dropped ensuring your blocks are always broadcast and can propagate the network


nano/node/network.cpp
```diff
@@ -215,15 +215,21 @@ bool nano::network::send_votes_cache (std::shared_ptr<nano::transport::channel>
 	return result;
 }
 
-void nano::network::flood_message (nano::message const & message_a)
+void nano::network::flood_message (nano::message const & message_a, bool const & is_dropable)
```
wezrule: Minor typo `is_droppable`. Passing a bool by ref won't be faster, primitives are generally passed by value. Also it should probably have the `_a` suffix to be consistent with the existing parameter. Same in `flood_block`

nano/node/network.cpp
```diff
@@ -215,15 +215,21 @@ bool nano::network::send_votes_cache (std::shared_ptr<nano::transport::channel>
...
 {
 	auto list (list_fanout ());
 	for (auto i (list.begin ()), n (list.end ()); i != n; ++i)
 	{
-		(*i)->send (message_a);
+		(*i)->send (message_a, nullptr, is_dropable);
 	}
 }
 
+void nano::network::flood_block (std::shared_ptr<nano::block> block_a, bool const & is_dropable)
```
SergiySW: Probably still can be in network.hpp

argakiig: was moved when I added something to the function, good catch, moved it back

nano/node/network.cpp
argakiig: thanks :)

nano/node/blockprocessor.cpp
```diff
@@ -382,7 +387,7 @@ void nano::block_processor::process_live (nano::block_hash const & hash_a, std::
 	});
 }
 
-nano::process_return nano::block_processor::process_one (nano::transaction const & transaction_a, nano::unchecked_info info_a)
+nano::process_return nano::block_processor::process_one (nano::transaction const & transaction_a, nano::unchecked_info info_a, const bool work_watcher_a)
```
guilhermelawless: Shouldn't it also be called `watch_work_a` here?

argakiig: yup, thank you I thought I got them all but good catch

nano/node/json_handler.cpp
```diff
@@ -2818,6 +2818,7 @@ void nano::json_handler::payment_wait ()
 void nano::json_handler::process ()
 {
 	const bool json_block_l = request.get<bool> ("json_block", false);
+	const bool watch_work_l = request.get<bool> ("watch_work", true);
```
SergiySW: Probably this option requires protection with enable_control? Same as work_generate

argakiig: handled same way as process force

PR #2170, opened 2019-07-23
add git commit hash and compiler info to logging and version calls

argakiig:
include build info in version calls
`Build Info <git hash> "<compiler> version " "<compiler version string>" "BOOST <boost version>" BUILT "<build date>"`


argakiig:
Updated build info string
```
Version 20.0
Build Info 3d66a9a "Clang version " "10.0.1 (clang-1001.0.46.4)" "BOOST 107000" BUILT "Jul 23 2019"
```

nano/nano_node/daemon.cpp
```diff
@@ -63,7 +63,8 @@ void nano_daemon::daemon::run (boost::filesystem::path const & data_path, nano::
 				auto network_label = node->network_params.network.get_current_network_as_string ();
 				auto version = (NANO_VERSION_PATCH == 0) ? NANO_MAJOR_MINOR_VERSION : NANO_MAJOR_MINOR_RC_VERSION;
 				std::cout << "Network: " << network_label << ", version: " << version << std::endl
-				          << "Path: " << node->application_path.string () << std::endl;
+				          << "Path: " << node->application_path.string () << std::endl
+				          << "Build Info: " << BUILD_INFO << std::endl;
```
wezrule: Would be better to use `\n` instead of `std::endl` multiple times otherwise you are doing multiple flushes. But it's only called once so not too bothered.

argakiig: done and changed thanks :)

PR #2175, opened 2019-07-27
Add frontiers confirmation modes

SergiySW:
always - always algressive confirm frontiers
automatic - confirm frontiers in always mode if node contains representative with at least 50% of principal weight, confirm frontiers less often otherwise
disabled - do not confirm frontiers


nano/node/nodeconfig.cpp
```diff
@@ -448,6 +459,47 @@ nano::error nano::node_config::deserialize_json (bool & upgraded_a, nano::jsonco
...
+	}
+	else if (string_a == "auto")
+	{
+		return nano::frontiers_confirmation_mode::automatic;
+	}
+	else if (string_a == "lazy")
+	{
+		return nano::frontiers_confirmation_mode::lazy;
+	}
+	if (string_a == "disabled")
```
cryptocode: Was an else if meant here? Not that it matters functionally in this case.

nano/node/wallet.cpp
```diff
@@ -808,12 +808,13 @@ nano::public_key nano::wallet::deterministic_insert (nano::transaction const & t
 		{
 			work_ensure (key, key);
 		}
+		auto supply (wallets.node.online_reps.online_stake ());
+		auto half_principal_weight ((supply / 1000) / 2);
```
cryptocode: Now that https://github.com/nanocurrency/nano-node/pull/2182 is merged, this should probably use `principal_weight_factor`

nano/node/wallet.cpp
```diff
@@ -852,11 +853,12 @@ nano::public_key nano::wallet::insert_adhoc (nano::transaction const & transacti
 		{
 			work_ensure (key, wallets.node.ledger.latest_root (block_transaction, key));
 		}
-		if (wallets.node.ledger.weight (block_transaction, key) >= wallets.node.config.vote_minimum.number ())
+		auto supply (wallets.node.online_reps.online_stake ());
+		auto half_principal_weight ((supply / 1000) / 2);
```
cryptocode: Same as above (principal_weight_factor)

nano/node/wallet.cpp
```diff
@@ -1826,10 +1828,29 @@ void nano::wallets::clear_send_ids (nano::transaction const & transaction_a)
...
+	return result;
+}
+
 void nano::wallets::compute_reps ()
 {
 	std::lock_guard<std::mutex> lock (mutex);
 	reps_count = 0;
+	half_principal_reps_count = 0;
+	auto supply (node.online_reps.online_stake ());
+	auto half_principal_weight ((supply / 1000) / 2);
```
cryptocode: Same as above (principal_weight_factor)

nano/node/nodeconfig.cpp
cryptocode: Was an else-if meant here? Not that it matters functionally in this case.

nano/node/active_transactions.cpp
```diff
@@ -49,7 +51,7 @@ void nano::active_transactions::confirm_frontiers (nano::transaction const & tra
 	// To minimise dropping real-time transactions, set the maximum number of elections
 	// for cementing frontiers to half the total active election maximum.
 	const auto max_active = node.config.active_elections_size / 2;
-	if (roots_size <= max_active && (check_time_exceeded || (!is_test_network && low_active_elections)))
+	if (roots_size <= max_active && (check_time_exceeded || (!is_test_network && low_active_elections && !lazy_mode)))
 	{
 		// When the number of active elections is low increase max number of elections for setting confirmation height.
 		if (max_active > roots_size + max_elections)
```
wezrule: So as I understand it:
- Lazy mode, every 3 minutes confirm up to half of max active elections (25k) 
- Auto (not enough weight) - Don't confirm 
- Auto (with enough weight) - Confirm every 15 minutes (or if there is a very low number of active elections) up to half max active elections.
- Always - Same as above.
Is this correct?

wezrule:
So as I understand it:
- Lazy mode, every 3 minutes confirm up to half of max active elections (25k) 
- Auto (not enough weight) - Don't confirm 
- Auto (with enough weight) - Confirm every 15 minutes (or if there is a very low number of active elections) up to half max active elections.
- Always - Same as above.

Is this correct?

nano/node/nodeconfig.cpp
SergiySW: Thanks

SergiySW:
@wezrule should be 3 minutes for always, 15 minutes for lazy. Corrected

nano/core_test/node.cpp
```diff
@@ -3153,6 +3217,7 @@ TEST (confirmation_height, prioritize_frontiers)
 	transaction.refresh ();
 	node->active.prioritize_frontiers_for_confirmation (transaction, std::chrono::seconds (1), std::chrono::seconds (1));
 	ASSERT_TRUE (priority_orders_match (node->active.priority_wallet_cementable_frontiers, std::array<nano::account, num_accounts>{ key3.pub, nano::genesis_account, key4.pub, key1.pub, key2.pub }));
+	node->active.confirm_frontiers (transaction);
```
wezrule: I don't think you need the scoped block below anymore (`node->active.next_frontier_check = ...`). It was overriding the frontier check so that in the next request loop it would call `confirm_frontiers` (which you are now calling manually) which wouldn't do anything because the `frontier_confirmation` is set to `disabled`.

cryptocode:
This now needs to (de)serialize toml instead (no json or upgrade)

PR #2183, opened 2019-07-30
Fix lazy bootstrap balances insert

SergiySW:
Adding lazy balances check was unreachable
`if (total_blocks == 0)`

### DIFF
--- a/nano/node/bootstrap.cpp
+++ b/nano/node/bootstrap.cpp
@@ -282,7 +282,7 @@ nano::bulk_pull_client::bulk_pull_client (std::shared_ptr<nano::bootstrap_client
 connection (connection_a),
 known_account (0),
 pull (pull_a),
-total_blocks (0),
+pull_blocks (0),
 unexpected_count (0)
 {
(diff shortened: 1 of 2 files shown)


nano/node/bootstrap.cpp
```diff
@@ -470,8 +470,8 @@ void nano::bulk_pull_client::received_block (boost::system::error_code const & e
 				connection->start_time = std::chrono::steady_clock::now ();
 			}
 			connection->attempt->total_blocks++;
-			total_blocks++;
 			bool stop_pull (connection->attempt->process_block (block, known_account, total_blocks, block_expected));
+			total_blocks++;
```
cryptocode: Maybe a good opportunity to do a related improvement: The parameter `total_blocks` in `bootstrap_attempt::process_block` shadows a field with the same name, so maybe rename the parameter to `total_blocks_a`

The parameter and field are incremented at the same time, but the field is used by `populate_connections` as well, so I assume they mean different things.

Other than that, LGTM

SergiySW: Good idea

PR #2185, opened 2019-07-30
boost 1.70 requires log_setup component specifically updated CMAKE 1.…

argakiig:
…67 and 1.69 dont require explicit but will use

update boostrap_boost.sh to mirror https://docs.nano.org/integration-guides/build-options/


EvertonMelo:
@argakiig 
if necessary see.

https://security-tracker.debian.org/tracker/CVE-2016-10642 
about x86 exponse with Cmake. 

x86 support study will be required.

PR #2186, opened 2019-07-30
Beta reset with lower work threshold

guilhermelawless:
- [x] 16x lower threshold + test
- [x] New genesis
- [x] Fix to allow missing rep_weights_beta.bin
- [x] Preconfigured representatives
- [x] Rebase onto master when all necessary commits are in


cryptocode:
For completeness, do we wanna update TEST (difficulty, multipliers) with a test for the beta threshold?

Issue #2181, opened 2019-07-29
Update node ID to use node_ instead of nano_ prefix


zhyatt:
To avoid the potential for confusing the node ID value with a Nano address, since they currently have the same format, @BitDesert suggested in https://github.com/nanocurrency/nano-docs/pull/59 that we use `node_` as a prefix instead of `nano_`.

zhyatt:
@SergiySW Is there any reason for us to not use `node_` as the node ID prefix?

SergiySW:
We can make it with replace

arranHarty123:
this caught me out as well

zhyatt:
Resolved in #2191

zhyatt:
Reopening until release

PR #2191, opened 2019-08-03
Use node_ prefix for node ID

SergiySW:
https://github.com/nanocurrency/nano-node/issues/2181


nano/node/json_handler.cpp
```diff
@@ -2502,6 +2502,7 @@ void nano::json_handler::node_id ()
 		response_l.put ("private", node.node_id.prv.data.to_string ());
 		response_l.put ("public", node.node_id.pub.to_string ());
 		response_l.put ("as_account", node.node_id.pub.to_account ());
+		response_l.put ("as_node_id", node.node_id.pub.to_node_id ());
```
cryptocode: The "peers" RPC returns this value as "node_id" and think this should too, and maybe even deprecate "as_account" in the docs?

SergiySW: Seems right. Same for docs

Issue #2190, opened 2019-08-03
confirmation stats in socket message


schenkty:
It would be extremely helpful if confirmation stats such as duration and tally were added to the socket message when a block is confirmed. I have found the socket to be the most reliable option to capture block activity under heavy stress than really anything else. confirmation_history works but will slow down under stress due to the fact that you need to increase the confirmation_history size to capture everything needed. Great work on the socket, would love if it could be expanded to include those values. 

Thanks,
Ty

clemahieu:
Makes sense since we’re trying to deprecate the confirmation_history method since it’s polled. 

We should make the additional data configurable since it’ll increase the web socket bandwidth usage.

PR #2195, opened 2019-08-04
Add support for election info in websocket confirmations

cryptocode:
Closes https://github.com/nanocurrency/nano-node/issues/2190

The election info fields matches those of `confirmation_history`


cryptocode:
Docs: https://github.com/nanocurrency/nano-docs/pull/69

Issue #2194, opened 2019-08-04
Wallet actions can take a long time due to flush() being called


cryptocode:
A few people have reported this, and I've noticed it under load on a nearly-synced node. The reason is that wallet actions currently attempts to flush the block queue, which can take a long time.

SergiySW:
Closed via https://github.com/nanocurrency/nano-node/pull/2196

PR #2196, opened 2019-08-04
Process wallet blocks outside of block processor queue

SergiySW:
https://github.com/nanocurrency/nano-node/issues/2194

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -608,6 +608,20 @@ nano::process_return nano::node::process (nano::block const & block_a)
 	return result;
 }
 
+nano::process_return nano::node::process_local (std::shared_ptr<nano::block> block_a)
+{
+	// Add block hash as recently arrived to trigger automatic rebroadcast and election
+	block_arrival.add (block_a->hash ());
+	// Set current time to trigger automatic rebroadcast and election
+	nano::unchecked_info info (block_a, block_a->account (), nano::seconds_since_epoch (), nano::signature_verification::unknown);
+	// Notify block processor to release write lock
+	block_processor.wait_write ();
+	// Process block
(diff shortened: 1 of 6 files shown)


nano/node/wallet.cpp
```diff
@@ -1136,12 +1144,16 @@ std::shared_ptr<nano::block> nano::wallet::send_action (nano::account const & so
 			wallets.node.work_generate_blocking (*block, wallets.node.active.active_difficulty ());
 		}
 		wallets.watcher.add (block);
-		wallets.node.process_active (block);
-		wallets.node.block_processor.flush ();
-		if (generate_work_a)
+		error = (wallets.node.process_local (block).code != nano::process_result::progress);
```
cryptocode: Can drop the parenthesis here

nano/node/node.cpp
```diff
@@ -608,6 +608,20 @@ nano::process_return nano::node::process (nano::block const & block_a)
...
+	block_arrival.add (block_a->hash ());
+	// Set current time to trigger automatic rebroadcast and election
+	nano::unchecked_info info (block_a, block_a->account (), nano::seconds_since_epoch (), nano::signature_verification::unknown);
+	// Notify block processor to release write lock
+	block_processor.wait_write ();
+	// Process block
+	auto transaction (store.tx_begin_write ());
+	auto result (block_processor.process_one (transaction, info));
+	return result;
+}
```
cryptocode: Maybe just `return block_processor.process_one(...)` as the result local isn't used.

nano/node/wallet.cpp
SergiySW: I guess it's easier to read with it?

Issue #1690, opened 2019-02-03
Provide optional automatic ledger/wallet backups before an upgrade 


guilhermelawless:
QOL enhancement.

If upgrading the ledger, do a backup (controlled on/off via config, default off, and maybe change location via a launch flag). If upgrading the wallet, do the same.

guilhermelawless:
Config backup was done here: https://github.com/nanocurrency/nano-node/pull/1959

zhyatt:
Closed by #1959

guilhermelawless:
@zhyatt The issue was regarding ledger (data.ldb) and wallet (wallet.ldb) backups, before things like the cementing or sideband upgrade. But these should be less frequent over time I guess (maybe again for pruning?).

That said, an automatic config backup is a great feature.

zhyatt:
Wow, did I ever read that wrong. :) Reopening

guilhermelawless:
Closed by #2198

PR #2198, opened 2019-08-05
Provide optional automatic ledger/wallet backups before an upgrade

SergiySW:
https://github.com/nanocurrency/nano-node/issues/1690


nano/node/lmdb/lmdb.cpp
```diff
@@ -60,9 +60,15 @@ txn_tracking_enabled (txn_tracking_config_a.enable)
...
 		// open inactive nodes which can otherwise be locked here if there is a long write
-		// (can be a few minutes with the --fastbootstrap flag for instance)
+		// (can be a few minutes with the --fast_bootstrap flag for instance)
 		if (!is_fully_upgraded)
 		{
+			// Backup database before upgrade
+			if (backup_before_upgrade)
+			{
+				auto backup_path = path_a.parent_path () / "backup.upgrade.ldb";
+				mdb_env_copy (env, backup_path.string ().c_str ());
```
wezrule: If they wanted a backup we should probably inform them and exit the node if the backup fails (lack of disk space etc), so they can rectify it and try again.

SergiySW:
@guilhermelawless imagine sevweral data.ldb backups size

guilhermelawless:
That's a good point. Can we then follow some other convention, for example `data.ldb.bak` and `wallets.ldb.bak`?

SergiySW:
I guess we can use config_backup_15571460497935531.json style for wallets (usually small size) & other for data.ldb

nano/node/lmdb/lmdb.cpp
SergiySW: Good point

schenkty:
Why wouldn't this happen by default? Many of us service operators have run into countless issues where something is corrupt after an upgrade and we have to start from scratch. Would be awesome if it could backup automatically then I suppose delete the backup once everything is successful and just make it optional to retain the backup post upgrade

guilhermelawless:
@schenkty All it takes is changing a config flag then you're set for every future upgrade

cryptocode:
Adding toml label as `backup_before_upgrade` upgrade change will need to be reverted and placed into toml base file.

guilhermelawless:
This will have to be changed as #2203 was merged which creates the new config upgrade path.

PR #2199, opened 2019-08-05
Only use -Werror by default with ci

wezrule:
To prevent inevitable warnings from different compilers/systems which are not caught by the versions of gcc/clang we use, this is now only turned on by default for our ci. It can still be turned on by setting the `NANO_WARN_TO_ERR` CMake flag though.


zhyatt:
@wezrule Can you set up either an issue or submit doc changes via PR for this? https://github.com/nanocurrency/nano-docs/issues

PR #2201, opened 2019-08-05
Deprecate nano- and xrb- address support

SergiySW:
Deprecate prefixes with dashes as almost not used


nano/node/json_handler.cpp
```diff
@@ -3816,9 +3780,14 @@ void nano::json_handler::version ()
 void nano::json_handler::validate_account_number ()
```
guilhermelawless: As mentioned this could go through `account_impl ()`

PR #2203, opened 2019-08-06
Simplify vote generator logic with a dual sleep

guilhermelawless:
Also adds the RPC upgrade path for v18 config.

This is a simpler approach than before:
1. Sleep 100ms, only interruptible to send 12 hashes
2. If 3+ hashes arrived, sleep 100ms more
3. Send anything at the end
4. Repeat

This re-uses previous RPC config options. The current set of choices means that, on average:
- Votes will have 12 hashes at ~60 tps  (`12 / (2*0.100)`)
- Up to ~30 tps (`3 / 0.100`) the delay is 100ms
- Above ~30 tps, the delay is 200ms
- The min delay is increase from 50ms to 100ms

These targets are slightly more conservative (favor packaging more than latency) than currently.

### DIFF
--- a/nano/core_test/wallets.cpp
+++ b/nano/core_test/wallets.cpp
@@ -87,13 +87,15 @@ TEST (wallets, upgrade)
 		nano::node_init init1;
 		auto node1 (std::make_shared<nano::node> (init1, system.io_ctx, 24001, path, system.alarm, system.logging, system.work));
 		ASSERT_FALSE (init1.error ());
-		node1->wallets.create (id.pub);
+		bool error (false);
+		nano::wallets wallets (error, *node1);
+		wallets.create (id.pub);
 		auto transaction_source (node1->wallets.env.tx_begin_write ());
 		auto tx_source = static_cast<MDB_txn *> (transaction_source.get_handle ());
 		auto & mdb_store (dynamic_cast<nano::mdb_store &> (node1->store));
 		auto transaction_destination (mdb_store.tx_begin_write ());
 		auto tx_destination = static_cast<MDB_txn *> (transaction_destination.get_handle ());
-		node1->wallets.move_table (id.pub.to_string (), tx_source, tx_destination);
+		wallets.move_table (id.pub.to_string (), tx_source, tx_destination);
 		node1->store.version_put (transaction_destination, 11);
 
 		nano::account_info info;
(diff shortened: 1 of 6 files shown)


cryptocode:
Adding new `toml` label as new config options should not be in json upgrade paths but in the new toml base file. Not sure how best to deal with this - we can maybe make new PRs to revert the added upgrade paths.

nano/node/voting.cpp
```diff
@@ -79,14 +76,14 @@ void nano::vote_generator::run ()
...
-			if (!condition.wait_for (lock, node.config.vote_generator_delay, [this]() { return this->wakeup; }))
+			condition.wait_for (lock, node.config.vote_generator_delay, [this]() { return this->hashes.size () >= 12; });
+			if (hashes.size () >= node.config.vote_generator_threshold && hashes.size () < 12)
 			{
-				// Did not wake up early. Likely not under high load, ok to send lower number of hashes
-				if (!hashes.empty ())
-				{
-					send (lock);
-				}
+				condition.wait_for (lock, node.config.vote_generator_delay, [this]() { return this->hashes.size () >= 12; });
```
cryptocode: Would it make sense to compute the additional wait as something like `(delay/12)*(12-hashes.size())` so it's inversely proportional to current hash count? Otherwise we'll wait an extra 100ms whether 1 or 9 hashes are missing for a full VBH. 

With default settings, for 3 hashes total wait would be 175ms ... 6 -> 150ms ... 11 -> ~110ms. IOW, extra wait is between 10 and 75 ms.

guilhermelawless: That could be done, although `vote_generator::add` will wake it up from the sleep early if it reaches 12:
https://github.com/nanocurrency/nano-node/pull/2203/files#diff-6892810124346aa7a6e440e1e922d28eL26

nano/core_test/wallets.cpp
```diff
@@ -87,13 +87,15 @@ TEST (wallets, upgrade)
 		nano::node_init init1;
 		auto node1 (std::make_shared<nano::node> (init1, system.io_ctx, 24001, path, system.alarm, system.logging, system.work));
 		ASSERT_FALSE (init1.error ());
-		node1->wallets.create (id.pub);
```
guilhermelawless: The changes in timing with this PR revealed that by not creating the wallets with `nano::wallets wallets (error, *node1);` it was not creating the environment correctly, therefore it was hitting a failure in mdb_iterator in an unrelated function that acessed the wallets (prioritize frontiers).

Please check that the changes don't invalidate what this test is trying to do.

PR #2206, opened 2019-08-07
Add test for retrieving non-existing account

guilhermelawless:
Also fixes `release_assert` for the confirmation height database.

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -4426,19 +4426,8 @@ TEST (rpc, account_info)
 	nano::system system (24000, 1);
 	nano::keypair key;
 	nano::genesis genesis;
-	system.wallet (0)->insert_adhoc (nano::test_genesis_key.prv);
-	system.wallet (0)->insert_adhoc (key.prv);
+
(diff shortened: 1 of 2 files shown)


nano/rpc_test/rpc.cpp
```diff
@@ -4426,73 +4426,96 @@ TEST (rpc, account_info)
...
+		test_response response (request, rpc.config.port, system.io_ctx);
+		system.deadline_set (5s);
+		while (response.status == 0)
+		{
+			ASSERT_NO_ERROR (system.poll ());
+		}
+
+		auto error (response.json.get_optional<std::string> ("error"));
+		ASSERT_TRUE (error.is_initialized ());
+		ASSERT_EQ (error.get (), "Account not found");
```
wezrule: You should be able to use the error code rather than hardcode the string, e.g:
```
std::error_code ec (nano::error_common::account_not_found);
ASSERT_EQ (*error, ec.message ());
```

guilhermelawless: Thanks, went ahead and changed other cases of hardcoded strings in `rpc.cpp`

PR #2210, opened 2019-08-09
Ignoring unreachable work peers

guilhermelawless:
In my network, when a work peer is unavailable, the connection hangs and takes a very long time to timeout, which makes it unusable as the node writes to all peers before falling back to local work generation, so the node itself hangs.

The deadline of 5 seconds, on expiry, closes the socket connection with status 125 (Operation cancelled), and adds the peer to a blacklist in memory only. On resolving addresses, those that error are also added to the list, so that they don't need to be resolved again.

If the failure is temporary due to connection failure (write and read are excluded), it will require a restart of the node, re-adding via RPC `work_peer_add` or using the new RPC `work_peers_reset` (note: not the same as `work_peers_clear`).

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -4561,7 +4561,7 @@ void nano::json_handler::work_peer_add ()
 	{
 		auto peer (std::make_pair (address_text, port));
 		node.config.work_peers.insert (peer);
-		node.blacklisted_work_peers.erase (peer);
+		node.excluded_work_peers.erase (peer);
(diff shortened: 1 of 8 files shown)


zhyatt:
@guilhermelawless What part of this is a breaking RPC change?

guilhermelawless:
In the sense that peers can get excluded so `work_generate` will not reach them afterwards

guilhermelawless:
We don't want people to lose access to work_peers, so another approach will follow.

Issue #806, opened 2018-04-16
Logging output can spread multiple lines, which makes it difficult to export


meltingice:
I'm currently sending logs to Graylog for monitoring and analysis by tailing the logs files and sending each line as a new message, but there are a few logging commands that are outputting data over multiple lines and makes it difficult to parse. One example of this is when you [log the full contents of a block](https://github.com/nanocurrency/raiblocks/blob/6527617ebffb1d250d7c25873ad3584143b3a8ef/rai/node/bootstrap.cpp#L635), it pretty prints the JSON instead of dumping it all on a single line.

Would love to see it either dump on a single line by default, or have a config option that lets you control pretty printing of the logs.

cryptocode:
There are many valid reasons to log things verbatim. filebeat supports multiple lines, or feed it through logstash.

cryptocode:
Sounds like a config flag to turn on single-line is what we want, as to not break any existing log parsing scripts.

zhyatt:
Resolved by #2214

PR #2214, opened 2019-08-13
Config option for single line logging records

SergiySW:
for block content and for election tally
https://github.com/nanocurrency/nano-node/issues/806


cryptocode:
toml note: revert the "single_line_record" change.

PR #2216, opened 2019-08-15
Throttled received block for bootstrap

SergiySW:
Bootstrap can now fill up to half of the block processor size. If half full, tries to continue later with a delayed receive.

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -48,6 +48,12 @@ bool nano::block_processor::full ()
 	return (blocks.size () + state_blocks.size ()) > node.flags.block_processor_full_size;
 }
 
+bool nano::block_processor::half_full ()
+{
+	std::unique_lock<std::mutex> lock (mutex);
+	return (blocks.size () + state_blocks.size ()) > node.flags.block_processor_full_size / 2;
+}
+
 void nano::block_processor::add (std::shared_ptr<nano::block> block_a, uint64_t origination)
 {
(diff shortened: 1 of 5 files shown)


nano/node/blockprocessor.cpp
```diff
@@ -48,6 +48,12 @@ bool nano::block_processor::full ()
 	return (blocks.size () + state_blocks.size ()) > node.flags.block_processor_full_size;
 }
 
+bool nano::block_processor::half_full ()
+{
+	std::unique_lock<std::mutex> lock (mutex);
+	return (blocks.size () + state_blocks.size ()) > node.flags.block_processor_full_size / 2;
```
guilhermelawless: Could extract `(blocks.size () + state_blocks.size ())`  into a `size ()` method since it's repeated in `full ()`

guilhermelawless:
LGTM pending clang-format

nano/node/blockprocessor.cpp
```diff
@@ -42,10 +42,20 @@ void nano::block_processor::flush ()
...
-	return (blocks.size () + state_blocks.size ()) > node.flags.block_processor_full_size;
+	return (blocks.size () + state_blocks.size () + forced.size ());
+}
+
+bool nano::block_processor::full ()
+{
+	return size () > node.flags.block_processor_full_size;
+}
+
+bool nano::block_processor::half_full ()
```
cryptocode: I think call sites would be easier to read if this was called `above_half_full`, I had to check the relational operator a couple of times on my first read-through.

PR #2217, opened 2019-08-15
Cache rep weights

wezrule:
Construct rep weights during node start from the frontiers and keep this in memory for any further rep weight operations. This does mean that they might be slightly out of sync with frontiers in the ledger because the operations are no longer atomic, but I don't think it should be problematic. Locks could be used to keep it in sync but that seems to defeat the purpose of this being done for performance.

A database upgrade removes the representation table.

Appropriate tests have been modified

A lot of the changes are removing `nano::node_init`. I needed to pass this to the `ledger` constructor to check if there were any errors, because the database may have failed to load correctly before the `node` constructor has exited out with an appropriate message. This needs to be done in `online_rep` as well. These constructors are passed the `store` already so I thought the store could tracked if it had an error itself, it doesn't add any dependencies because it's just a `bool` and saves it being passed in everywhere.

TSAN clean on Ubuntu/Clang (for anything related to this at least)


nano/secure/ledger.cpp
```diff
@@ -675,6 +684,18 @@ check_bootstrap_weights (true),
...
+	if (!store.init_error ())
+	{
+		auto transaction = store.tx_begin_read ();
+		for (auto i (store.latest_begin (transaction)), n (store.latest_end ()); i != n; ++i)
+		{
+			nano::account_info const & info (i->second);
+			nano::block_hash rep_block (representative_calculated (transaction, info.head));
+			auto source_block (store.block_get (transaction, rep_block));
+			auto source_rep (source_block->representative ());
+			rep_weights.representation_add (source_rep, info.balance.number ());
```
guilhermelawless: I think here we can go with
```cpp
auto block (store.block_get (transaction, info.rep_block));
assert (block != nullptr);
rep_weights.representation_add (block->account (), info.balance.number ());
```

It should be faster than the visitor used by `representative_calculated`

nano/secure/ledger.hpp
```diff
@@ -47,11 +48,15 @@ class ledger final
 	void dump_account_chain (nano::account const &);
 	bool could_fit (nano::transaction const &, nano::block const &);
 	bool is_epoch_link (nano::uint256_union const &);
+	void set_bootstrap_weight (nano::account const & account, nano::uint128_t const & weight);
+	//	std::unordered_map<nano::account, nano::uint128_t> construct_reps ();
```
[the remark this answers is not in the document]
wezrule: Yes whoops, removed. Thanks

nano/secure/ledger.cpp
wezrule: Yes good point. I had copied that from `--debug_dump_representatives`, `info.rep_block` has already cached it so can be used.

nano/nano_node/entry.cpp
```diff
@@ -266,26 +266,12 @@ int main (int argc, char * const * argv)
 			nano::inactive_node node (data_path);
```
SergiySW: I guess inactive_node needs special launch flag to not perform reps weights calculation each time. Expect for debug_dump_representatives

nano/lib/rep_weights.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+	std::lock_guard<std::mutex> guard (mutex);
+	auto source_previous (get (source_rep));
+	put (source_rep, source_previous + amount_a);
+}
+
+void nano::rep_weights::representation_put (nano::account const & account_a, nano::uint128_union const & representation_a)
+{
+	std::lock_guard<std::mutex> guard (mutex);
+	return put (account_a, representation_a);
```
SergiySW: return for void

nano/lib/rep_weights.cpp
```diff
@@ -0,0 +1,69 @@
...
+nano::uint128_t nano::rep_weights::get (nano::account const & account_a)
+{
+	auto it = rep_amounts.find (account_a);
+	if (it != rep_amounts.end ())
+	{
+		return it->second;
+	}
+	else
+	{
+		return nano::uint128_t{};
```
SergiySW: Is default value 0?

wezrule: Yeh from https://www.boost.org/doc/libs/1_67_0/libs/multiprecision/doc/html/boost_multiprecision/tut/ints/cpp_int.html "Default constructed cpp_int_backends have the value zero." I have added 0 anyway now to remove all doubts :)

nano/lib/rep_weights.cpp
wezrule: good spot.

nano/nano_node/entry.cpp
wezrule: I've added a flag in the `inactive_node` constructor as that seemed easiest, also added it to the `node_flags` which is constructed in the `inactive_node` constructor as well. Possibly `inactive_node` should take `node_flags` as well, but each one CLI call would then have to set it to false. Something to think about in a future PR maybe :)

PR #2221, opened 2019-08-16
 TOML config file support and migration

cryptocode:
This PR adds TOML support and migration of json config files on startup.

User facing changes:

* The node and rpc binaries can operate without a config file as every setting has a default. The purpose of the config file is to allow these defaults to be overridden. The user only need to add the settings that are required instead of a complete file.
* Config options can also be specified using a new `--config` option, which can be repeated multiple times. These options will take precedence over any entries in the config file (which doesn't need to exist). Example: `nano_node --daemon --config rpc.enable=true`
* After migration, config.json is migrated into config-node.toml. Only non-default values are written to keep the config file small. Comments are added. rpc_config.json  is migrated in a similar fashion.
* The same `config-node.toml` file can be used for both the node and the Qt wallet.
* After migration from json, config-node.toml and config-rpc.toml are never written to.
* If migrating a Qt wallet config, config-qtwallet.toml will be created, containing wallet id and account. The account is updated if changed via the UI. If running the Qt wallet on a node directory, this file will be created.
* Complete config files can be generated using `--generate_config node|rpc`. The user can use these directly, or pick what they need. Comments with schema information is added to facilitate future tooling/editor support. Sample output: https://gist.github.com/cryptocode/4d05774390518ddaadab61f838288d20
* I've used the opportunity to clean up some settings during the migration. For instance, rpc_enable and opencl_enable are moved to `enable` under their respective config sections, to keep it consistent with other sections. The HTTP callback settings are grouped under a separate section.

Developer facing changes:

* A git submodule update required for building. I'm maintaining a fork of cpptoml that fixes bugs and improves formatting, as well as adds features like key sorting and programmatically adding documentation. None of the upstream parsers supporting TOML 0.5 suffice.
* JSON serialization is kept and there's a final upgrade path before migrating to TOML. This means the upgraded v20 JSON files on beta will work in the migration. After v20, no more upgrade paths are added and we simply add fields and defaults (as well as any validation code we want on the values.) At some point we can remove the JSON config code and require the user to upgrade via v2x first if they have json configs they wanna upgrade.

Should be ready for review, though some updates will be added:

- [x] Documentation PR/issue: https://github.com/nanocurrency/nano-docs/issues/94
- [x] Add full coverage of the generated TOML in the daemon_config_deserialize unit test
- [x] Additional/improved TOML comments

We might extract more settings into separate toml files in upcoming PRs - the work generation part has been discussed, details remain.

### DIFF
--- a/nano/lib/tomlconfig.hpp
+++ b/nano/lib/tomlconfig.hpp
@@ -338,7 +338,7 @@ public:
 
 protected:
 	template <typename T, typename = std::enable_if_t<nano::is_lexical_castable<T>::value>>
-	tomlconfig & get_config (bool optional, std::string key, T & target, T default_value = T ())
+	tomlconfig & get_config (bool optional, std::string const & key, T & target, T default_value = T ())
 	{
 		try
 		{
@@ -369,7 +369,7 @@ protected:
 
 	// boost's lexical cast doesn't handle (u)int8_t
 	template <typename T, typename = std::enable_if_t<std::is_same<T, uint8_t>::value>>
-	tomlconfig & get_config (bool optional, std::string key, uint8_t & target, uint8_t default_value = T ())
+	tomlconfig & get_config (bool optional, std::string const & key, uint8_t & target, uint8_t default_value = T ())
 	{
 		try
 		{
@@ -404,7 +404,7 @@ protected:
 	}
 
 	template <typename T, typename = std::enable_if_t<std::is_same<T, bool>::value>>
-	tomlconfig & get_config (bool optional, std::string key, bool & target, bool default_value = false)
+	tomlconfig & get_config (bool optional, std::string const & key, bool & target, bool default_value = false)
 	{
 		auto bool_conv = [this, &target, &key, optional](std::string val) {
 			if (val == "true")
--- a/nano/lib/tomlconfig.hpp
+++ b/nano/lib/tomlconfig.hpp
@@ -47,13 +47,19 @@ public:
 	 */
 	nano::error & read (boost::filesystem::path const & path_a)
 	{
+		std::stringstream stream_override_empty;
+		return read (stream_override_empty, path_a);
+	}
+
+	nano::error & read (std::istream & stream_overrides, boost::filesystem::path const & path_a)
+	{
 		std::fstream stream;
 		open_or_create (stream, path_a.string ());
 		if (!stream.fail ())
 		{
 			try
 			{
-				read (stream);
+				read (stream_overrides, stream);
 			}
 			catch (std::runtime_error const & ex)
 			{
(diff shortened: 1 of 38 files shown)


nano/lib/tomlconfig.hpp
```diff
@@ -0,0 +1,540 @@
...
+#include <cpptoml.h>
+
+namespace nano
+{
+/** Manages a table in a toml configuration table hierarchy */
+class tomlconfig : public nano::configbase
+{
+public:
+	tomlconfig () :
+	tree (cpptoml::make_table ())
```
wezrule: Can the non-template method definitions of this class go into a separate `.cpp` file? (same with `walletconfig.hpp`)

nano/lib/tomlconfig.hpp
```diff
@@ -0,0 +1,540 @@
...
+	std::shared_ptr<cpptoml::table> tree;
+
+	/** Compare two stringified configs, remove keys where values are equal */
+	void erase_defaults (std::shared_ptr<cpptoml::table> base, std::shared_ptr<cpptoml::table> other, std::shared_ptr<cpptoml::table> update_target)
+	{
+		std::vector<std::string> erased;
+		assert (other != nullptr);
+		for (auto & item : *other)
+		{
+			std::string key = item.first;
```
wezrule: This can be a const ref

nano/node/nodeconfig.cpp
```diff
@@ -62,6 +63,272 @@ logging (logging_a)
...
+				bool result = port_position == -1;
+				if (!result)
+				{
+					auto port_str (entry.substr (port_position + 1));
+					uint16_t port;
+					result |= parse_port (port_str, port);
+					if (!result)
+					{
+						auto address (entry.substr (0, port_position));
+						this->work_peers.push_back (std::make_pair (address, port));
```
wezrule: can use `emplace_back` and pass the arguments directly (saves a copy/move)

nano/node/nodeconfig.cpp
```diff
@@ -62,6 +63,272 @@ logging (logging_a)
...
+		if (toml.has_key ("statistics"))
+		{
+			auto stat_config_l (toml.get_required_child ("statistics"));
+			stat_config.deserialize_toml (stat_config_l);
+		}
+
+		if (toml.has_key ("work_peers"))
+		{
+			work_peers.clear ();
+			toml.array_entries_required<std::string> ("work_peers", [this](std::string entry) {
```
wezrule: entry can be const ref

nano/lib/tomlconfig.hpp
cryptocode: Will add a cpp file for walletconfig. I'll keep the tomlconfig as-is for now since it mirrors jsonconfig. I can refactor both in a separate pr.

nano/node/node_rpc_config.cpp
```diff
@@ -17,6 +18,39 @@ nano::error nano::node_rpc_config::serialize_json (nano::jsonconfig & json) cons
...
 }
 
+nano::error nano::node_rpc_config::serialize_toml (nano::tomlconfig & toml) const
+{
+	toml.put ("enable_sign_hash", enable_sign_hash, "Allow or disallow signing of hashes\ntype:bool");
+	toml.put ("max_work_generate_difficulty", nano::to_string_hex (max_work_generate_difficulty), "Maximum allowed difficulty request for work generation\ntype:string,hex");
+
+	nano::tomlconfig child_process_l;
+	child_process_l.put ("enable", child_process.enable, "Enable or disable RPC child process. If false, an in-process RPC server is used.\ntype:bool");
+	child_process_l.put ("rpc_path", child_process.rpc_path, "Path to the nano_rpc executable. Must be set if child process is enabled.\ntype:bool");
```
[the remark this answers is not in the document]
cryptocode: Good catch, thanks

PR #2222, opened 2019-08-16
Configurable work watcher period

guilhermelawless:
New option `work_watcher_period` defaulting to 5 seconds.

Also moved the `condition.wait` to the end of the loop so that `stopped` is checked before going through blocks.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -854,10 +854,12 @@ TEST (node_config, v17_v18_upgrade)
 	config.logging.init (path);
 	// These config options should not be present
 	ASSERT_FALSE (tree.get_optional_child ("backup_before_upgrade"));
+	ASSERT_FALSE (tree.get_optional_child ("work_watcher_period"));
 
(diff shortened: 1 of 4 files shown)


SergiySW:
Are all timings in config currently in ms?

guilhermelawless:
I believe so, but documentation covers it. Do we need that much control?

guilhermelawless:
@SergiySW there are 2 instances of times in seconds: `tcp_io_timeout` and IPC transport `timeout`.

SergiySW:
I guess we need check for it to not be 0?

cryptocode:
@guilhermelawless for the toml PR: could you attach a comment here with a suitable one-liner description of the new config option? No need to mention data type or default value as the toml file will contain those.

guilhermelawless:
@cryptocode it's best to refer to the nano-docs [PR](https://github.com/nanocurrency/nano-docs/pull/91) in case it changes. That's a one liner in the config documentation.

PR #2230, opened 2019-08-20
Distibuted work fixes

guilhermelawless:
Distributed work was having some issues:
- Work cancels were not being sent at all - bad file descriptor. The socket was not specified with the required host and port
- Never using local work generation if any work peer is set. This is a problem if the work peers are unresponsive. To keep this behavior now, set work threads to 0 and don't use OpenCL
- Work requests hanging and not being handled correctly when destroying the distributed work object
- (general) work pool was not being stopped by `node::stop()` until work finished

So those are all fixed, and added a startup log if there is no local work generation set and no work peers defined.


nano/node/node.cpp
```diff
@@ -1193,15 +1244,22 @@ class distributed_work : public std::enable_shared_from_this<distributed_work>
...
 	unsigned int backoff; // in seconds
 	std::shared_ptr<nano::node> node;
 	nano::block_hash root;
 	std::mutex mutex;
 	std::map<boost::asio::ip::address, uint16_t> outstanding;
+	std::vector<std::weak_ptr<work_request>> connections;
 	std::vector<std::pair<std::string, uint16_t>> need_resolve;
 	std::atomic_flag completed;
 	uint64_t difficulty;
+	std::atomic_bool stopped{ false };
```
cryptocode: We use `std::atomic<bool>` everywhere (mainly consistency now, the type used to be a problem on some compilers)

nano/node/node.cpp
```diff
@@ -1085,19 +1111,57 @@ class distributed_work : public std::enable_shared_from_this<distributed_work>
...
+					}
 				});
-			});
+			}
+		});
+	}
+	void stop (bool const local_stop)
+	{
+		std::lock_guard<std::mutex> lock (mutex);
+		if (!stopped)
```
cryptocode: Perhaps use `if (!stopped.exchange (true))` like in node::stopped - can also move the lock inside the scope as it's an atomic.

guilhermelawless:
@SergiySW please see last commit and its commit message. Local generation is now only started as a last resort and, for future requests, only as long as peers are unresponsive.

PR #2233, opened 2019-08-21
Fire callback & add to history after confirmation height is set

wezrule:
After a block has won an election, the roots are removed but the blocks/dependents are not removed until confirmation height is now set, same with dependent elections.

Moved existing confirmation height tests to their own file, as they are getting quite large (1000+ lines). Added 2 new tests which check that the callback/confirmation history is set correctly for inactive/quorum and dependent elections observer stats.

Added `pause/unpause` functionality to the confirmation height processor, so that I can more easily prevent it confirming any blocks until desired.

Should be used in conjunction with https://github.com/nanocurrency/nano-node/pull/2156 to be completely consistent.

Note for reviewers:
Consider the following situation, a simple chain with 2 un-cemented send blocks:
1 - The frontier is being confirmed with an active election, it gets added to the conf height processor and then iterates over both blocks and it about to write the confirmation height of the account to include both blocks.
2 - `block_confirm` is called on the dependent block and is not marked as confirmed yet so an election is started.
3 - The confirmation height processor in the first step now sets the confirmation height on the frontier and the dependent block, callbacks and confirmation history are updated for both blocks.
4 - Election in 2 finishes, what should happen?

In master this will fire off the callback and add to the confirmation history due to the finished election, but now we only want to do this after setting the confirmation height. The confirmation height for the block is already set and will only fire a duplicate callback in `json_handler::block_confirm` if the block is confirmed at that point in time. Is this an issue? Should the confirmation height processor detect the already confirmed block and if from a confirmed election always fire the callback as well? (for consistent duplicates at least?).


nano/node/node.cpp
```diff
@@ -1187,22 +1187,9 @@ void nano::node::process_confirmed (nano::election_status const & status_a, uint
...
-
-		receive_confirmed (transaction, block_a, hash);
-		nano::account account (0);
-		nano::uint128_t amount (0);
-		bool is_state_send (false);
-		nano::account pending_account (0);
-		process_confirmed_data (transaction, block_a, hash, sideband, account, amount, is_state_send, pending_account);
-		observers.blocks.notify (status_a, account, amount, is_state_send);
-		if (amount > 0)
+		if (status_a.type == nano::election_status_type::active_confirmed_quorum)
```
SergiySW: Probably this can be before read transaction. Also if `type != nano::election_status_type::active_confirmed_quorum` then `else if (iteration < 20)` part seems not required

nano/node/election.cpp
```diff
@@ -50,11 +50,7 @@ void nano::election::confirm_once (nano::election_status_type type_a)
 		{
 			--node.active.long_unconfirmed_size;
 		}
-		auto root (status.winner->qualified_root ());
-		node.active.add_confirmed (status, root);
-		clear_blocks ();
```
SergiySW: Not sure why clear_blocks (); & clear_dependent (); were removed here. Because after removing root we ideally should immediately clean blocks (for correct difficulty adjustment) & adjust difficulty

nano/node/node.cpp
wezrule: sounds good :)

nano/node/election.cpp
wezrule: I needed the blocks at a later time to get the election information, but am just making a new collection with the necessary ones now.

PR #2236, opened 2019-08-22
Start frontiers confirmation after initial bootstrap

SergiySW:


nano/node/active_transactions.cpp
```diff
@@ -115,9 +115,9 @@ void nano::active_transactions::request_confirm (std::unique_lock<std::mutex> &
 	std::deque<std::shared_ptr<nano::block>> rebroadcast_bundle;
 	std::deque<std::pair<std::shared_ptr<nano::block>, std::shared_ptr<std::vector<std::shared_ptr<nano::transport::channel>>>>> confirm_req_bundle;
 
-	// Confirm frontiers when there aren't many confirmations already pending
+	// Confirm frontiers when there aren't many confirmations already pending and node finished initial boostrap
```
wezrule: minor type `boostrap`

SergiySW: Thanks!

PR #2240, opened 2019-08-23
Wrap boost::asio::async_write to ensure lifetime of buffers

wezrule:
Introduce a new class `nano::shared_const_buffer` which @cryptocode found an example for here: https://www.boost.org/doc/libs/1_69_0/doc/html/boost_asio/example/cpp11/buffers/reference_counted.cpp This should alleviate any buffer memory not being captured or having sufficient lifetime during the `boost::asio::async_write` call. When building travis I check if `boost::asio::async_write` is called anywhere other than the new file which wraps it `nano/lib/asio.hpp` and error out with a message if it is.

I think any minor performance hits we have by always using a `shared_ptr` are outweighed by keeping the program correct, as this has bitten us in the past and I found yet more places (nano::rpc_request_processor::try_reconnect_and_execute_request) where this wasn't done properly even though we actively try make sure it is.


cryptocode:
For another PR, but would something similar work for reads? I.e. a shared_mutable_buffer class (using boost::asio::mutable_buffer)?

nano/lib/asio.hpp
```diff
@@ -0,0 +1,34 @@
...
+	const boost::asio::const_buffer * begin () const;
+	const boost::asio::const_buffer * end () const;
+
+	size_t size () const;
+
+private:
+	std::shared_ptr<std::vector<uint8_t>> m_data;
+	boost::asio::const_buffer m_buffer;
+};
+
```
cryptocode: Maybe add a static assert here to ensure it's compliant to asio's ConstBufferSequence concept, should it change in the future. My notes says something this should work:

`static_assert(boost::asio::is_const_buffer_sequence<shared_const_buffer<uint8_t>>::value, "Not ConstBufferSequence compliant");`

wezrule:
The MutableBufferSequence has similar requirements to the ConstBufferSequence so I think it should be fine to add something similar for `async_read`
https://www.boost.org/doc/libs/1_67_0/doc/html/boost_asio/reference/MutableBufferSequence.html

nano/lib/asio.hpp
wezrule: good idea

PR #2244, opened 2019-08-24
Fixes account_history RPC to include receive blocks when filtering accounts.

Srayman:
Previously receive blocks were filtered out as it was comparing link hash instead of link as account.

Also added test case for receive blocks in addition to existing send block test case.


guilhermelawless:
Added documentation tag, @Srayman  could you add an accompanying PR to nano-docs, with a link to this one?

PR #2253, opened 2019-08-26
Fix distributed work not triggering callback on work cancel

guilhermelawless:
Callback was never triggered due to the `if (work_a)` check.

Switched the callback functions to boost optionals so that `json_handler::work_generate ()` can properly handle as "Cancelled" the results of distributed work. Implicit conversion from boost::optional ...

For now, `work_generate_blocking` continues to assume correct generation and returns a non-optional. Leaving that for future work, going through all cases and handling bad/cancelled work generation adequately. I assume that might be why we once ran into a `work_validate` assert.


guilhermelawless:
Replacing with a refactor which will handle work cancellingcorrectly. As it stands there is no way to connect work request cancelling (which happens in the work pool) to distributed work requests with peers. A new observer for work cancels will accomplish this.

PR #2255, opened 2019-08-27
Distributed work refactor

guilhermelawless:
By introducing:
- A distributed work factory that also saves distributed work objects (as weak_ptr)
- A work cancel observer

Work can now be properly cancelled across local generation and work peers, simply by using the observer.

`work_generate` now takes an optional so the caller may choose to handle work cancellation (work not set once callback was fired)


guilhermelawless:
Most of the apparent changes are simply extracting `distributed_work` from node.cpp, the actual changes are `distributed_work_factory` and the new observer.

nano/node/distributed_work.cpp
```diff
@@ -0,0 +1,367 @@
...
+				boost::property_tree::write_json (ostream, request);
+				request_string = ostream.str ();
+			}
+			auto request (std::make_shared<boost::beast::http::request<boost::beast::http::string_body>> ());
+			request->method (boost::beast::http::verb::post);
+			request->set (boost::beast::http::field::content_type, "application/json");
+			request->target ("/");
+			request->version (11);
+			request->body () = request_string;
+			request->prepare_payload ();
```
cryptocode: These seven lines are duplicated in distributed_work::start_work (). Maybe use this refactoring as an opportunity to extract this into a `prepare_request` function or something similar.

nano/node/distributed_work.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+	auto last (remove (address));
+	handle_failure (last);
+}
+
+void nano::distributed_work::handle_failure (bool const last)
+{
+	if (last)
+	{
+		if (!completed)
```
cryptocode: This is from the old code as well, what do you think about doing `if (last && !completed)` instead to improve readability and save an indentation?

nano/node/distributed_work.cpp
guilhermelawless: Definitely, fits well into the `work_peer_request` class imo, despite not needing anything from it.

nano/node/distributed_work.cpp
guilhermelawless: Agreed, thanks!

nano/core_test/distributed_work.cpp
```diff
@@ -0,0 +1,127 @@
...
+	// now using observer
+	done = false;
+	node->distributed_work.make (hash, callback_to_cancel, nano::difficulty::from_multiplier (1000000, node->network_params.network.publish_threshold));
+	system.deadline_set (1s);
+	while (node->distributed_work.work.empty ())
+	{
+		ASSERT_NO_ERROR (system.poll ());
+	}
+	node->observers.work_cancel.notify (hash);
+	while (!done)
```
wezrule: Should have a `system.deadline_set` here

nano/core_test/distributed_work.cpp
```diff
@@ -0,0 +1,127 @@
...
+	nano::system system (24000, 1);
+	auto node (system.nodes[0]);
+	nano::block_hash hash;
+	bool done{ false };
+	auto callback_to_cancel = [&done](boost::optional<uint64_t> work_a) {
+		ASSERT_FALSE (work_a.is_initialized ());
+		done = true;
+	};
+	node->distributed_work.make (hash, callback_to_cancel, nano::difficulty::from_multiplier (1000000, node->network_params.network.publish_threshold));
+	system.deadline_set (1s);
```
wezrule: Is 1 second enough in the various places used? Sometimes we run on slow hardware and TSAN/ASAN/Valgrind can make tests run a lot slower.

nano/node/distributed_work.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+void nano::distributed_work_factory::make (unsigned int backoff_a, nano::block_hash const & root_a, std::function<void(boost::optional<uint64_t>)> const & callback_a, uint64_t difficulty_a)
+{
+	auto distributed (std::make_shared<nano::distributed_work> (backoff_a, node, root_a, callback_a, difficulty_a));
+	{
+		std::lock_guard<std::mutex> guard (mutex);
+		work[root_a].emplace_back (distributed);
+	}
+	distributed->start ();
+	cleanup ();
```
wezrule: Would it make sense to call `cleanup` first, so you don't waste time iterating over the newly created distributed work? (although it's probably negligible) Also we generally use the term `cleanup ()` to indicate it's basically annihilating the whole object. Perhaps add a suffix or use a different term to indicate it's only a partial cleanup, `cleanup_expired`, `purge_finished_work` or something.

nano/node/CMakeLists.txt
```diff
@@ -37,6 +37,8 @@ add_library (node
 	confirmation_height_processor.cpp
 	daemonconfig.hpp
 	daemonconfig.cpp
+	distributed_work.cpp
+	distributed_work.hpp
```
wezrule: We put the header files first

nano/core_test/distributed_work.cpp
guilhermelawless: Thanks, waiting isn't even necessary as the function emplaces before returning.

nano/core_test/distributed_work.cpp
guilhermelawless: Good catch, increasing to 5s

nano/node/distributed_work.cpp
guilhermelawless: Makes sense! Changing to `cleanup_finished` 👍

nano/core_test/distributed_work.cpp
wezrule: There's another `1s` is the `no_peers_cancel` test

PR #2260, opened 2019-08-28
Prevent adjusted difficulty overflow for extremly high or low difficulty

SergiySW:


nano/core_test/active_transactions.cpp
```diff
@@ -114,6 +124,123 @@ TEST (active_transactions, adjusted_difficulty_priority)
...
+		});
+		node1.active.roots.modify (open1_root, [](nano::conflict_info & info_a) {
+			info_a.difficulty = std::numeric_limits<std::uint64_t>::min () + 1;
+		});
+		node1.active.roots.modify (open2_root, [](nano::conflict_info & info_a) {
+			info_a.difficulty = std::numeric_limits<std::uint64_t>::min () + 1;
+		});
+		node1.active.roots.modify (send3_root, [](nano::conflict_info & info_a) {
+			info_a.difficulty = std::numeric_limits<std::uint64_t>::min () + 1;
+		});
```
wezrule: These `modify` calls are very similar, could reduce some duplication by wrapping the whole thing in a local lambda like:
```
auto modify_difficulty = [&roots = node1.active.root](auto & root) {
	roots.modify (root, [](nano::conflict_info & info_a) {
		info_a.difficulty = std::numeric_limits<std::uint64_t>::min () + 1;
	});
};

modify_difficulty (send1_root);
....
```
Or put it in a function, or perhaps just extract the callback lambda in modify function call (and make a local variable to `roots`) would be fine too :).

Same can be done in the other newly added test.

SergiySW: Updated

PR #2263, opened 2019-08-30
Prevent supurious wakeup in active transactions request loop

guilhermelawless:
Some supurious wakeup prevention:
- Critical: `active_transactions::request_loop ()` has a long wait
- Non-critical: could try to stop attempts more than once in bootstrapping, and increase stats more than once in some places
- Non-functional: writing `while(!predicate) wait(lock)` as `wait(lock, predicate)` 

Also moved the lock creation in `nano::worker::run ()` so it doesn't get re-created every loop

### DIFF
--- a/nano/lib/utility.cpp
+++ b/nano/lib/utility.cpp
@@ -238,9 +238,9 @@ thread ([this]() {
 
 void nano::worker::run ()
 {
-	std::unique_lock<std::mutex> lk (mutex);
 	while (!stopped)
 	{
+		std::unique_lock<std::mutex> lk (mutex);
 		if (!queue.empty ())
 		{
 			auto func = queue.front ();
@@ -250,7 +250,6 @@ void nano::worker::run ()
 			// So that we reduce locking for anything being pushed as that will
 			// most likely be on an io-thread
 			std::this_thread::yield ();
-			lk.lock ();
 		}
 		else
 		{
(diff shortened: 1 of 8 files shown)


nano/node/active_transactions.cpp
```diff
@@ -20,10 +20,7 @@ thread ([this]() {
 })
 {
 	std::unique_lock<std::mutex> lock (mutex);
-	while (!started)
-	{
-		condition.wait (lock);
-	}
+	condition.wait (lock, [this] { return this->started; });
```
wezrule: Try and only capture what you need to in lambdas otherwise it has a greater dependency than it needs, so in this case `&started = started` would be enough. And same in other places

nano/lib/utility.cpp
```diff
@@ -238,9 +238,9 @@ thread ([this]() {
 
 void nano::worker::run ()
 {
+	std::unique_lock<std::mutex> lk (mutex);
```
wezrule: Not sure how I feel about this one, because it's now going to be protecting `stopped` which it doesn't need to. If you are going to do this then can you make `stopped` non-atomic and protect it in `stop`. It's cheap to make a `unique_lock` fyi, it's just a pointer to a mutex and a `bool`. The `lk.lock ()` in the loop is superfluous so probably shouldn't have been there to begin with.

guilhermelawless: Thanks, did not know about this. Did your other comment disappear somehow? I applied both

wezrule: Yeh, had a talk with @cryptocode and didn't want to force you to do it :P

PR #2281, opened 2019-09-04
Handle work generation failure and cancel

guilhermelawless:
Adds handling of work generation failures in critical paths (everything that is not a test). Failure can be a simple work cancellation which can happen for many reasons - even manually.

- Extract `nano::wallet::action_complete` with work validation + generation + block processing. Was repeated in all `*_action`
- Move work generation logging to distributed work, which makes logging now work for 99% of cases rather than only for precaching
- Fix a minor issue in distributed work where work cancellation would make it start local generation immediately the next time

### DIFF
--- a/nano/node/distributed_work.cpp
+++ b/nano/node/distributed_work.cpp
@@ -19,7 +19,8 @@ backoff (backoff_a),
 node (node_a),
 root (root_a),
 need_resolve (node.config.work_peers),
-difficulty (difficulty_a)
+difficulty (difficulty_a),
+elapsed (nano::timer_state::started, "distributed work generation timer")
 {
 	assert (!completed);
 }
@@ -257,6 +258,11 @@ void nano::distributed_work::set_once (boost::optional<uint64_t> work_a)
 	if (!completed.exchange (true))
 	{
 		callback (work_a);
+		if (node.config.logging.work_generation_time ())
+		{
+			node.logger.try_log (boost::str (boost::format ("Work generation for %1%, with a threshold difficulty of %2% (multiplier %3%x) complete: %4% ms")
+			% root.to_string () % nano::to_string_hex (difficulty) % nano::to_string (nano::difficulty::to_multiplier (difficulty, node.network_params.network.publish_threshold), 2) % elapsed.stop ().count ()));
+		}
 	}
 }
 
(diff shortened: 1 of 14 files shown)


nano/node/wallet.cpp
```diff
@@ -1138,24 +1116,36 @@ std::shared_ptr<nano::block> nano::wallet::send_action (nano::account const & so
...
+			block = nullptr;
 		}
-		wallets.watcher->add (block);
-		error = (wallets.node.process_local (block).code != nano::process_result::progress);
-		if (!error && generate_work_a)
+	}
+	return block;
+}
+
+bool nano::wallet::action_complete (std::shared_ptr<nano::block> block_a, nano::account const & account_a, bool const generate_work_a)
```
wezrule: `block_a` can be passed by const ref &

nano/node/distributed_work.cpp
```diff
@@ -257,6 +258,11 @@ void nano::distributed_work::set_once (boost::optional<uint64_t> work_a)
 	if (!completed.exchange (true))
 	{
 		callback (work_a);
+		if (node.config.logging.work_generation_time ())
+		{
+			node.logger.try_log (boost::str (boost::format ("Work generation for %1%, with a threshold difficulty of %2% (multiplier %3%x) complete: %4% ms")
+			% root.to_string () % nano::to_string_hex (difficulty) % nano::to_string (nano::difficulty::to_multiplier (difficulty, node.network_params.network.publish_threshold), 2) % elapsed.stop ().count ()));
```
wezrule: This statement is really long, can `"Work generation for %1%, with a threshold difficulty of %2% (multiplier %3%x) complete: %4% ms"` and `nano::to_string (nano::difficulty::to_multiplier (difficulty, node.network_params.network.publish_threshold), 2)` be put into local variables? that should shorten it a lot.

nano/node/wallet.cpp
guilhermelawless: Need to pass as shared_ptr to  `wallets.watcher->add` and `wallets.node.process_local`

wezrule: That's fine, those functions will copy (and extend lifetime) of the `shared_ptr` if necessary, but you don't need to copy it here.

nano/node/distributed_work.cpp
guilhermelawless: Done, makes sense!

PR #2287, opened 2019-09-04
Add --config option to rpc process

cryptocode:
Same option as for the node, allowing toml options to be passed to the RPC daemon via arguments. Includes a var name change for consistency.

Example: `nano_rpc --daemon --config port=9999`

Not adding a doc tag as there's already https://github.com/nanocurrency/nano-docs/issues/94

### DIFF
--- a/nano/lib/rpcconfig.cpp
+++ b/nano/lib/rpcconfig.cpp
@@ -187,7 +187,7 @@ nano::error nano::rpc_config::deserialize_toml (nano::tomlconfig & toml)
 
 namespace nano
 {
-nano::error read_rpc_config_toml (boost::filesystem::path const & data_path_a, nano::rpc_config & config_a)
+nano::error read_rpc_config_toml (boost::filesystem::path const & data_path_a, nano::rpc_config & config_a, std::vector<std::string> const & config_overrides)
 {
 	nano::error error;
 	auto json_config_path = nano::get_rpc_config_path (data_path_a);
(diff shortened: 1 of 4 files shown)


nano/lib/rpcconfig.cpp
```diff
@@ -233,10 +233,21 @@ nano::error read_rpc_config_toml (boost::filesystem::path const & data_path_a, n
 	// Parse and deserialize
 	nano::tomlconfig toml;
 
+	std::stringstream config_overrides_stream;
+	for (auto const & entry : config_overrides)
+	{
+		config_overrides_stream << entry << std::endl;
```
wezrule: Could use `"\n"` instead of flushing each time with `std::endl`

nano/lib/rpcconfig.cpp
```diff
@@ -233,10 +233,21 @@ nano::error read_rpc_config_toml (boost::filesystem::path const & data_path_a, n
...
+
 	// Make sure we don't create an empty toml file if it doesn't exist. Running without a toml file is the default.
 	if (!error && boost::filesystem::exists (toml_config_path))
 	{
-		error = toml.read (toml_config_path);
+		error = toml.read (config_overrides_stream, toml_config_path);
+	}
+	else if (!error)
+	{
+		toml.read (config_overrides_stream);
```
wezrule: I wonder if this should be moved into a single block, so `!error` isn't repeated, i.e:
```
if (!error)
{
    if (boost::filesystem::exists (toml_config_path))
    {
        error = toml.read (config_overrides_stream, toml_config_path);
    }
    else
    {
        toml.read (config_overrides_stream);
    }
}
```
But is up to you.

nano/lib/rpcconfig.cpp
cryptocode: What does flushing mean for a stringstream?

wezrule: The same for any output stream https://en.cppreference.com/w/cpp/io/basic_ostream/flush:
```
Writes uncommitted changes to the underlying output sequence.
If rdbuf() is a null pointer, does nothing
Otherwise, behaves as an UnformattedOutputFunction (since C++11). After constructing and checking the sentry object, calls rdbuf()->pubsync(). If the call returns -1, calls setstate(badbit).
```

I did some benchmarking to see if there was any meaningful difference:
http://quick-bench.com/BGpm2DCcIS8xk65lUvmHMzQiGLA

Strangely with libstdc++ the `std::endl` is a bit quicker! but libc++ is a lot slower. I doubt it matters here anyway given there won't be many. My comment about using endl after is irrelevant here sorry.

cryptocode: Right, probably fine as this is only on startup. But that benchmark is good to know about, esp in performance sensitive code on libc++ 😅

PR #2290, opened 2019-09-06
Add block count cache

SergiySW:
Estimation used for comparision with bootstrap_weight_max_blocks


nano/secure/ledger.cpp
```diff
@@ -711,6 +711,9 @@ check_bootstrap_weights (true)
 				cemented_count += i->second;
 			}
 		}
+
+		// Cache block count. 1 for not yet initialized ledger
+		block_count_cache = std::max (static_cast<size_t> (1), store.block_count (transaction).sum ());
```
wezrule: The rep weight counts and cemented count is passed to and updated in `block_store_partial::initialize`. To be consistent this new block cache should probably be there too because that's where the actual `block_put` call is made. As we are getting more and more of these, I wonder if it makes sense to have a new class `ledger_cache` which holds these so only 1 object needs to be passed around. Or move the setting of those to here, although easier I'm not sure if is the most correct thing to do, but I'll let you ponder that one :)

PR #2293, opened 2019-09-08
Add account_state class

clemahieu:
Add the account_state class which is an abstraction for ledger-users to decouple from the underlying representation of the state of an account in the block store.

With state blocks, a lot of the fields are available in the head block itself and don't need to be duplicated in the account_info table. This change normalizes this information and pulls it from the head block itself.


nano/node/lmdb/lmdb.cpp
```diff
@@ -473,25 +473,24 @@ void nano::mdb_store::upgrade_v14_to_v15 (nano::write_transaction const & transa
...
 		release_assert (rep_block != nullptr);
-		account_infos.emplace_back (i->first, nano::account_info{ account_info_v14.head, rep_block->representative (), account_info_v14.open_block, account_info_v14.balance, account_info_v14.modified, account_info_v14.block_count, account_info_v14.epoch });
+		account_infos.emplace_back (i->first, nano::account_info{ account_info_v14.head, rep_block->representative (), account_info_v14.open_block }, account_info_v14.epoch);
 		confirmation_height_put (transaction_a, i->first, i->second.confirmation_height);
 	}
 
 	for (auto const & account_info : account_infos)
 	{
-		account_put (transaction_a, account_info.first, account_info.second);
+		account_put (transaction_a, std::get<0> (account_info), std::get<1> (account_info), std::get<2> (account_info));
```
cryptocode: Typed extractors, e.g. `std::get<nano::account>`, could maybe help readability here

nano/lib/blocks.cpp
```diff
@@ -991,7 +1008,8 @@ hashables (error_a, stream_a)
 }
 
 nano::state_block::state_block (bool & error_a, boost::property_tree::ptree const & tree_a) :
-hashables (error_a, tree_a)
+hashables (error_a, tree_a),
+epoch_m (nano::epoch::unspecified)
```
wezrule: Rather than repeat this default value in all constructors (and risk missing one in the future) you could add it as an in-class member initializer.

nano/secure/ledger.hpp
```diff
@@ -15,6 +15,27 @@ class shared_ptr_block_hash final
...
+{
+public:
+	account_state () = default;
+	account_state (nano::account_info const &, std::shared_ptr<nano::block>, nano::block_sideband const &);
+	account_state (nano::transaction const &, nano::block_store &, nano::account const &);
+	account_state (nano::transaction const &, nano::block_store &, nano::account_info const &);
+	nano::uint256_union head () const;
+	nano::uint256_union rep () const;
+	nano::uint256_union open () const;
+	nano::amount balance () const;
```
wezrule: `head`, `rep`, `open` can all be `const &`, also could use the appropriate aliases we have currently, i.e `nano::block_hash`, `nano::account`; I am in the process of updating them all anyway though.

nano/node/lmdb/lmdb.cpp
wezrule: or a new struct/class with these as members (can even be private to this function)

nano/node/json_handler.cpp
```diff
@@ -1950,11 +1950,11 @@ void nano::json_handler::delegators ()
 		auto transaction (node.store.tx_begin_read ());
 		for (auto i (node.store.latest_begin (transaction)), n (node.store.latest_end ()); i != n; ++i)
 		{
-			nano::account_info const & info (i->second);
-			if (info.representative == account)
+			auto state (node.ledger.account_state (transaction, nano::account_info (i->second)));
```
wezrule: This is going to make the delegators RPC slow again. The 1.1M+ point reads for blocks can take a couple minutes on some systems. I assumed we would be combining the `head` block into the `account_info` and removing duplicate fields so that we could get info about both in 1 I/O read (and in this case still scan a range quickly (100-150ms).

nano/secure/ledger.cpp
```diff
@@ -687,6 +669,106 @@ bool nano::shared_ptr_block_hash::operator() (std::shared_ptr<nano::block> const
 	return lhs->hash () == rhs->hash ();
 }
 
+nano::account_state::account_state (nano::account_info const & info_a, std::shared_ptr<nano::block> block_a, nano::block_sideband const & sideband_a) :
```
wezrule: `block_a` can be passed by `const &`

nano/lib/blocks.cpp
```diff
@@ -357,6 +357,11 @@ bool nano::send_block::valid_predecessor (nano::block const & block_a) const
 	return result;
 }
 
+nano::epoch nano::send_block::epoch () const
```
SergiySW: Because it's similar for send/open/change/receive types, then probably it can be just common `nano::epoch nano::block::epoch () const` with override for state_block (like nano::block::link (), account (), representative ())

nano/node/bootstrap/bootstrap_frontier.cpp
```diff
@@ -320,11 +320,11 @@ void nano::frontier_req_server::next ()
 		auto transaction (connection->node->store.tx_begin_read ());
 		for (auto i (connection->node->store.latest_begin (transaction, current.number () + 1)), n (connection->node->store.latest_end ()); i != n && accounts.size () != max_size; ++i)
 		{
-			nano::account_info const & info (i->second);
-			if (!skip_old || (now - info.modified) <= request->age)
+			auto state (connection->node->ledger.account_state (transaction, i->second));
```
SergiySW: This request seems like extra i/o if `request->age == std::numeric_limits<decltype (request.age)>::max ()` (default). Then `i->second.head` can be used

SergiySW: Actually it can be like
```
bool send_frontier (true);
if (skip_old)
{
	auto state (connection->node->ledger.account_state (transaction, i->second));
	send_frontier = (now - state.modified ()) <= request->age;
}
if (send_frontier)
{
	nano::account const & account (i->first);
	nano::account_info const & info (i->second);
	accounts.emplace_back (account, info.head);
}
```

clemahieu:
Closing for now since the performance implications are significant.

PR #2296, opened 2019-09-09
Prefer requesting confirmation via active elections

guilhermelawless:
**WIP!** Description is outdated

Single hash confirmation in `blockprocessor::process_live ()` is now disabled once 1000 roots are in active elections. At that point, confirmation times are severely increased anyway, and from tests [on beta], single-hash confirm_req+ack are severely overloading the nodes, possibly inhibiting prioritization.

Therefore, with this PR the focus is on confirmation via active elections. `request_interval_ms` and `max_broadcast_queue` must be experimented with as that will be the confirmation rate limit. The loop occurs every `request_interval_ms` (set to 8000 here), and 7*`max_broadcast_queue` (7*1000 without #2137 ) confirmations are requested in `max_broadcast_queue` packets of 7 hashes each.

In my view, the loop would instead be shorter (lower number of roots) and much more frequent, but for that prioritization must be flawless.

### DIFF
--- a/nano/node/blockprocessor.hpp
+++ b/nano/node/blockprocessor.hpp
@@ -59,6 +59,7 @@ private:
 	void verify_state_blocks (nano::unique_lock<std::mutex> &, size_t = std::numeric_limits<size_t>::max ());
 	void process_batch (nano::unique_lock<std::mutex> &);
 	void process_live (nano::block_hash const &, std::shared_ptr<nano::block>, const bool = false);
+	void requeue_invalid (nano::block_hash const &);
 	bool stopped;
 	bool active;
 	bool awaiting_write{ false };
--- a/nano/node/blockprocessor.hpp
+++ b/nano/node/blockprocessor.hpp
@@ -53,8 +53,6 @@ public:
 	nano::vote_generator generator;
 	// Delay required for average network propagartion before requesting confirmation
 	static std::chrono::milliseconds constexpr confirmation_request_delay{ 1500 };
-	// Threshold for not requesting confirmation when there are many active elections
-	static size_t constexpr confirmation_request_cutoff{ 50 };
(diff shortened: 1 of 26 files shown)


nano/node/blockprocessor.hpp
```diff
@@ -53,6 +53,8 @@ class block_processor final
 	nano::vote_generator generator;
 	// Delay required for average network propagartion before requesting confirmation
 	static std::chrono::milliseconds constexpr confirmation_request_delay{ 1500 };
+	// Threshold for not requesting confirmation when there are too many active elections
+	static size_t constexpr confirmation_request_cutoff{ 1000 };
```
cryptocode: Is this duplication of `max_broadcast_queue` or is the same value coincidental ? Asking since  max_broadcast_queue is used in computation of the extra delay, removed by this PR.

guilhermelawless: Coincidence, but good catch! We can still tweak `max_broadcast_queue` if we see fit

Srayman:
I really like these changes however I think confirmation_request_cutoff is too high at 1,000.  It might be suitable for the beta network with only 30 peers, but only live with 325+ it has 10x the traffic going on.  The image below is from August 10th, 2019 where a test on live of 40 CPS was performed.  You can see it handled 40 CPS for a while, but when active confirmations hit about 1,000+ it started to break down quickly.  This is identical to what we see on beta when active confirmations fill up and confirm_ack requests spike.  

If batching of confirm_req can happen efficiently and quickly then I think the lowest it can go while not negatively impacting regular confirmation times would be ideal.  The more batching that can happen the more efficient the network traffic.  For live in this example I think in the range of 500 would be better or maybe a bit lower if possible.

image

PR #2299, opened 2019-09-11
Verify pubkey against special keys in wallet

cryptocode:
Marking as a breaking RPC change, since wallet_add_watch may now fail after pubkey validation. Contains some unrelated formatting fixes on master.


cryptocode:
That's what clang-format 8.0.0 generates for me. I'll have to guard it with clang-format off

zhyatt:
@cryptocode Any documentation considerations needed for this?

cryptocode:
@zhyatt I couldn't think of any doc needs, I believe making a label in the release notes is sufficient, nobody can rely on the current behaviour I think.

PR #2303, opened 2019-09-15
Update broadcast_confirm_req_batch to handle PoW priority

Srayman:
Changing requests_bundle to use deque instead of vector and then pop_front instead of pop_back.  This resolves the inverse PoW priority issues seen during active transaction processing.  Live traffic is still unaffected as this only takes effect once active transaction processing starts.

There are some other version difference between the images below, but PoW priority is shown.

Bootstrapping V20.0 Beta Before the change ...
image

Compared to bootstrapping V20.0 Beta after the change ...
image


guilhermelawless:
Thanks! To be clear prioritization is working correctly across the active elections (roots) container, but once hashes go into the `requests_bundle`, they are later inserted in the wrong order (up to 7k hashes).

Srayman:
Correct!  And that explains the strange coloring in the first chart where the blue seems to invert mid way through bootstrapping.  During that phase of bootstrapping the active elections (roots) container had fewer blocks than the requests bundle size so it sent all in reverse order.  Whereas you can see the priority happening on the other ends when it active elections count was higher.

PR #2304, opened 2019-09-15
RPC epoch_upgrade

SergiySW:
This is an internal/diagnostic RPC, do not rely on its interface being stable

Updated with https://github.com/nanocurrency/nano-node/pull/2310


nano/node/json_handler.cpp
```diff
@@ -2011,6 +2011,185 @@ void nano::json_handler::deterministic_key ()
...
 }
 
+void epoch_upgrader (std::shared_ptr<nano::node> node_a, nano::uint256_union const & prv_a, nano::epoch epoch_a)
+{
+	nano::block_builder builder;
+	auto link (node_a->ledger.link (epoch_a));
+	nano::raw_key raw_key;
+	raw_key.data = prv_a;
+	auto signer (nano::pub_key (prv_a));
+	assert (signer == node_a->ledger.signer (link));
```
cryptocode: Somewhat unrelated to PR, but do you think we could rename `ledger::link` to `epoch_link` and `ledger::signer` to `epoch_signer` to help readability? Would make it more consistent with is_epoch_link too.

SergiySW: I think yes

PR #2306, opened 2019-09-18
Active elections loop tuning

guilhermelawless:
Closes #2137 , closes #2139  (alternative to both)

**Todo**
- [x] Adjust frontier confirmation mode timings with the decreased elections size and loop time
- [x] Allow a certain time for low-difficulty elections before flushing
- [x] Split into smaller, contained PRs #2327 

**Documentation**
- [x] active_difficulty updates are now done every ~500ms (give or take), and not 16-36 seconds

**PR Description**
This PR has 2 main changes:
- Disables per-block confirm_req in the block processor
- Changes the confirmation loop such that it has a period of 0.5s instead of 16-36 seconds.

Other things had to be tuned consequently, such as frontier prioritization timings. Data collected on the beta network shows there are large improvements to bandwidth usage under high load, much more agile PoW prioritization, and reduced tcp write drops.

The optimized bandwidth allows for doing a larger number of batch confirm_req directly to representatives in the same amount of time. The PoW prioritization being more agile means even under saturation the network should correctly prioritize blocks with the highest PoW such that those are confirmed fastest.

The active elections container is reduced to 10k soft maximum. As blocks arrive that would immediately be dropped due to a low work difficulty, the election is allowed to live for up to 10 seconds, after which it is dropped.

There is a minimum time of 1 second until an election is considered for confirm_req, and the same time for being consider for active difficulty.

**Future improvements**
- Using a passive elections container to store a larger amount of elections that would otherwise be dropped. If the size of a passive election is small and not many operations are performed on the container, it can possibly be in the order of 100k in memory.

- Consider not rebroadcasting votes if the node is saturating, at least for blocks that would go into passive elections. This means all blocks would be confirmed via confirm_req. A better alternative would be #1006 . Currently, even under saturation, there's a change that blocks are confirmed via live elections even if they have a low PoW difficulty. 

**Acknowledgments**
Thanks to @Srayman for the enormous help in tuning and testing, and @Joohansson for the blocks.

### DIFF
--- a/nano/node/nodeconfig.cpp
+++ b/nano/node/nodeconfig.cpp
@@ -577,7 +577,8 @@ bool nano::node_config::upgrade_json (unsigned version_a, nano::jsonconfig & jso
 		}
 		case 17:
 		{
-			json.put ("vote_generator_delay", vote_generator_delay.count ()); // Update value
+			json.put ("active_elections_size", 10000); // Update value
+			json.put ("vote_generator_delay", 100); // Update value
 			json.put ("backup_before_upgrade", backup_before_upgrade);
 			json.put ("work_watcher_period", work_watcher_period.count ());
 		}
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -22,15 +22,15 @@ TEST (active_transactions, bounded_active_elections)
 	{
 		node.process_active (send);
 		node.active.start (send);
+		ASSERT_NO_ERROR (system.poll ());
 		ASSERT_FALSE (node.active.empty ());
 		ASSERT_LE (node.active.size (), node.config.active_elections_size);
+		++count;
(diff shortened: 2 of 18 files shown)


nano/node/nodeconfig.cpp
```diff
@@ -577,7 +577,8 @@ bool nano::node_config::upgrade_json (unsigned version_a, nano::jsonconfig & jso
 		}
 		case 17:
 		{
-			json.put ("vote_generator_delay", vote_generator_delay.count ()); // Update value
+			json.put ("active_elections_size", 10000); // Update value
```
SergiySW: Do we need it? If right now it's toml

guilhermelawless: Otherwise will remain 50k during the v19->v20 upgrade, this forces the change one last time.

PR #2319, opened 2019-09-29
Periodic rep crawler weights update

SergiySW:
Also update probable rep with response only if channel was changed


cryptocode:
lgtm, pending some msvc compile issue

PR #2320, opened 2019-09-29
Add confirmation request count to election_status

guilhermelawless:
Used in the block confirmation websocket to keep track of confirmation request counts outside the node.


guilhermelawless:
@cryptocode done, I think we should since that's [marked](https://docs.nano.org/commands/rpc-protocol/#confirmation_history) as a debug RPC anyway.

Fixed a bug where the "warning: empty filter" log would come up when it shouldn't.

Docs PR coming up.

PR #2322, opened 2019-10-01
Profile work validation

guilhermelawless:
```bash
./nano_node --debug_profile_validate
Starting validation profile
Average validation time: 184 ns (5414818 validations/s)
```


guilhermelawless:
Noticed we already had a verify command, but it's non-intuitive to use, so this replaces that one.

PR #2324, opened 2019-10-03
Add new state2 block type to hold nano pow proofs

wezrule:
Marking as incomplete and just putting this up to make sure it is the direction it should be going; it still needs some cleaning up. The work_generate/validation stuff hasn't been touched as requested (**Edit: As no one is working on this stuff now, I'm going to create some skeletons around it which should simplify some of the changes I've made**), this however made it somewhat difficult in some areas because it always returned a `legacy_pow` when a `nano_pow` would have been expected, so some temporary hacks have been added to get around this.

The epoch version field is currently part of the sideband information for state blocks (originally from the v14 to v15 upgrade), I have left it there, unsure if it makes sense to strip it all out and put into the `state_block` class itself. The work field is currently the determining factor in knowing whether this is a `state` or `state2` block type. It uses a new class called `proof_of_work` which stores a variant which holds either a `legacy_pow` or a `nano_pow`, this makes it easier to know which one is actually stored rather than passing a `nano::epoch` version everywhere with it, due to the inheritance involved with with blocks a common type was needed for the `block_work/set` functions.

A new block type `state2` has been added and the work field has been extended to 12 bytes. However it was noted that the work generation functions may return 16 bytes I wasn't entirely how I should be interacting with this (**Edit: Going to make `nano_pow` hold the number of bytes which it will serialize instead of 16, should simplify some things**), so in the new `nano_pow` class it has space for 16 bytes, but is only currently serializing 12 bytes (stored in big endian) like the legacy pow does.

On the wire the changes are only to use a block type of `state2` and the work field widened to 12 from 8, no extra fields are added nor is the signing changed.


wezrule:
This is now out of date and mostly irrelevant

PR #2327, opened 2019-10-07
Basic confirmation request loop enhancements

guilhermelawless:
This is split 1/? (a safe one) from #2306 

- Rename variables with suffix `_l` for local variables in `request_confirm` as it's a large function
- Rename some variables so their purpose is obvious
- Add callbacks to allow throttling broadcasting. It could happen that more broadcasts would be queued while the previous ones hadn't finished, especially under load
- Add condition `node.ledger.block_count_cache > node.ledger.cemented_count + roots.size () + pending_confirmation_height_size` before checking frontiers as that can read the disk for 2 seconds when not necessary. It's still not perfect, per the in-file comment, as e.g., 10 roots can be added and they're transformed into a single frontier by the confirmation height processor. Open to ideas on making this condition more consistent.

**Discussion needed** - I got rid of the `unordered_set` of channels. Not sure why this was in there; if it was for randomizing the channels, we should probably shuffle the vector instead.

### DIFF
--- a/nano/node/network.cpp
+++ b/nano/node/network.cpp
@@ -224,22 +224,26 @@ void nano::network::flood_message (nano::message const & message_a, bool const i
 	}
 }
 
-void nano::network::flood_block_batch (std::deque<std::shared_ptr<nano::block>> blocks_a, unsigned delay_a)
+void nano::network::flood_block_many (std::deque<std::shared_ptr<nano::block>> blocks_a, std::function<void()> callback_a, unsigned delay_a)
 {
-	auto block (blocks_a.front ());
+	auto block_l (blocks_a.front ());
 	blocks_a.pop_front ();
-	flood_block (block);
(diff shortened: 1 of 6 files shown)


nano/node/network.cpp
```diff
@@ -319,29 +323,29 @@ void nano::network::broadcast_confirm_req_base (std::shared_ptr<nano::block> blo
...
-	while (!request_bundle_a.empty () && count < max_reps)
+	auto count_l (0);
+	while (!request_bundle_a.empty () && count_l < max_reps)
 	{
 		auto j (request_bundle_a.begin ());
 		while (j != request_bundle_a.end ())
 		{
-			count++;
-			std::vector<std::pair<nano::block_hash, nano::root>> roots_hashes;
+			count_l++;
```
SergiySW: Can be ++count_l I guess

PR #2330, opened 2019-10-07
Secondary list of peers for generating work

guilhermelawless:
This secondary list of peers can be used only via RPC work_generate by setting the option "secondary_work_peers" to "true" along with "use_peers"

- Adds secondary_work_peers under a new "experimental" config
- Changes distributed_work to receive a list of peers
- Some fixes to distributed_work where it wouldn't call the failure callback in edge scenarios
- Collect object stats for distributed_work

### DIFF
--- a/nano/node/nodeconfig.cpp
+++ b/nano/node/nodeconfig.cpp
@@ -105,12 +105,6 @@ nano::error nano::node_config::serialize_toml (nano::tomlconfig & toml) const
 		work_peers_l->push_back (boost::str (boost::format ("%1%:%2%") % i->first % i->second));
 	}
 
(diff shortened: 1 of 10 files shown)


nano/node/nodeconfig.cpp
```diff
@@ -217,6 +223,26 @@ nano::error nano::node_config::deserialize_toml (nano::tomlconfig & toml)
...
+				{
+					auto port_str (entry.substr (port_position + 1));
+					uint16_t port;
+					result |= parse_port (port_str, port);
+					if (!result)
+					{
+						auto address (entry.substr (0, port_position));
+						this->new_work_peers.emplace_back (address, port);
+					}
+				}
```
cryptocode: This is duplicated now, would be nice to get it extracted into a helper function, passing container and `entry`.

PR #2333, opened 2019-10-10
include nano_pow_server as a submodule and build into ci framework for tagged releases

argakiig:
nano-pow-server included as submodule, doesnt define NANO_POW_STANDALONE so as to not install files in the wrong location on OSx
'-DNANO_POW_SERVER=ON` is used to indicate if the standalone server should be built and bundled
binary is added to docker as well


argakiig:
Windows currently failing if not tagged, will fix in the AM

CMakeLists.txt
```diff
@@ -346,6 +350,7 @@ add_subdirectory(nano/nano_node)
 add_subdirectory(nano/rpc)
 add_subdirectory(nano/nano_rpc)
 
+
```
wezrule: Probably don't need an extra space here

PR #2335, opened 2019-10-10
Restart dropped elections for unconfirmed blocks with higher PoW

guilhermelawless:
With this and #2306 prioritization should be fully guaranteed.

Requires:
- block get to check for work
- account get + confirmation height get to check for confirmation

A 60s buffer of dropped elections is added which sets the maximum time another node has to send the upgraded block before it will only be confirmed via frontier confirmation. This was added to prevent high DB usage for any old block, instead it is only done if the root has been recently dropped. This grants protection from bad actors as they do not know which elections we dropped.

However:
- It is possible to infer with a certain probability which blocks each node is dropping if they use a default config
- It is not ideal to have a maximum time to upgrade a block

Therefore, another approach will follow in the future using a large container of passive elections.


guilhermelawless:
Marked as incomplete, will add a cache of recently dropped elections to reduce DB I/O in this path.

guilhermelawless:
Finished, up to 32k dropped elections (less than 100 bytes each, 3MB) and only considered for checking the DB if the upgraded PoW arrived within 60 seconds (can increase this value).

nano/node/active_transactions.hpp
```diff
@@ -76,6 +77,13 @@ class confirmed_set_info final
 	nano::qualified_root root;
 };
 
+class dropped_election final
```
cryptocode: This has the same definition as `confirmed_set_info`. An option is to rename that to `election_timepoint` or something like that and have only one class (and split into two classes if the types need to diverge)

This would allow the two multi_index_container members to be replaced with a type alias.

guilhermelawless: Great suggestion, thanks!  29cd26b

nano/node/active_transactions.hpp
```diff
@@ -71,13 +72,20 @@ class cementable_account final
...
 
-class confirmed_set_info final
+class election_timepoint final
 {
 public:
 	std::chrono::steady_clock::time_point time;
 	nano::qualified_root root;
 };
 
+typedef boost::multi_index_container<
```
wezrule: `typedef` is legacy c++, aliases are preferred, i.e `using ordered_elections_timepoint = ... `

nano/node/active_transactions.cpp
```diff
@@ -742,26 +743,58 @@ bool nano::active_transactions::active (nano::block const & block_a)
...
+			nano::block_sideband existing_sideband;
+			auto hash (block_a->hash ());
+			auto existing_block (node.store.block_get (*opt_transaction_a, hash, &existing_sideband));
+			release_assert (existing_block != nullptr);
+			uint64_t confirmation_height;
+			release_assert (!node.store.confirmation_height_get (*opt_transaction_a, node.store.block_account (*opt_transaction_a, hash), confirmation_height));
+			bool confirmed = (confirmation_height >= existing_sideband.height);
+			if (!confirmed && existing_block->block_work () != block_a->block_work ())
+			{
+				uint64_t existing_difficulty, new_difficulty;
```
wezrule: I've not seen many (if any) areas where we declare multiple variables on the same line, so to be consistent they should probably be on separate lines.

nano/node/active_transactions.hpp
guilhermelawless: Thanks, changed and moved to inside the class as it's not needed elsewhere.

PR #2343, opened 2019-10-15
script to build qtbase on mac from src git, use for mac artifacts

argakiig:
update qt to use artifacts generated from https://github.com/nanocurrency/travis-cache uploaded to S3


argakiig:
closing for now, persuing another method will make a new PR

PR #2349, opened 2019-10-17
Confirm frontiers for bootstrap with multiple failures

SergiySW:


nano/node/bootstrap/bootstrap.cpp
```diff
@@ -573,6 +592,158 @@ void nano::bootstrap_attempt::add_bulk_push_target (nano::block_hash const & hea
...
 
+void nano::bootstrap_attempt::attempt_restart_check (nano::unique_lock<std::mutex> & lock_a)
+{
+	if (!confirmed_frontiers && requeued_pulls > (!node->network_params.network.is_test_network () ? nano::bootstrap_limits::requeued_pulls_limit : nano::bootstrap_limits::requeued_pulls_limit_test))
+	{
+		confirm_frontiers (lock_a);
+		if (!confirmed_frontiers)
+		{
+			node->stats.inc (nano::stat::type::bootstrap, nano::stat::detail::frontier_confirmation_failed, nano::stat::dir::in);
+			node->bootstrap_initiator.blacklist.add (endpoint_frontier_request);
```
cryptocode: Would be nice if the blacklisting (with IP/port) was always_log`ed.

SergiySW: Done

nano/node/json_handler.cpp
```diff
@@ -1679,6 +1680,8 @@ void nano::json_handler::bootstrap_status ()
 		response_l.put ("target_connections", std::to_string (attempt->target_connections (attempt->pulls.size ())));
 		response_l.put ("total_blocks", std::to_string (attempt->total_blocks));
 		response_l.put ("runs_count", std::to_string (attempt->runs_count));
+		response_l.put ("requeued_pulls", std::to_string (attempt->requeued_pulls));
+		response_l.put ("confirmed_frontiers", std::to_string (attempt->confirmed_frontiers));
```
guilhermelawless: Renaming this to `frontiers_confirmed` makes it more clear it's a boolean

nano/node/json_handler.cpp
```diff
@@ -1601,6 +1601,7 @@ void nano::json_handler::bootstrap ()
 {
 	std::string address_text = request.get<std::string> ("address");
 	std::string port_text = request.get<std::string> ("port");
+	const bool confirmed_frontiers = request.get<bool> ("confirmed_frontiers", false);
```
guilhermelawless: I think `bypass_frontier_confirmation` conveys the intention better (default false as well)

nano/node/bootstrap/bootstrap.cpp
```diff
@@ -573,6 +592,159 @@ void nano::bootstrap_attempt::add_bulk_push_target (nano::block_hash const & hea
...
+	for (auto i (pulls.begin ()), end (pulls.end ()); i != end && frontiers.size () != nano::bootstrap_limits::bootstrap_max_confirm_frontiers; ++i)
+	{
+		if (!i->head.is_zero ())
+		{
+			frontiers.push_back (i->head);
+		}
+	}
+	for (auto i (recent_pulls_head.begin ()), end (recent_pulls_head.end ()); i != end && frontiers.size () != nano::bootstrap_limits::bootstrap_max_confirm_frontiers; ++i)
+	{
+		if (!(*i).is_zero ())
```
guilhermelawless: Any reason to not use `i->is_zero ()` ?

nano/node/bootstrap/bootstrap.cpp
```diff
@@ -573,6 +592,159 @@ void nano::bootstrap_attempt::add_bulk_push_target (nano::block_hash const & hea
...
+							{
+								rep_request->second.emplace_back (*ii, *ii);
+							}
+						}
+					}
+					++ii;
+				}
+			}
+		}
+		if (frontiers.size () < frontiers_count * 0.1) // 90% of frontiers confirmed
```
guilhermelawless: Maybe `0.1` can be defined somewhere. Suggest `required_frontier_confirmation_ratio = 0.9` and place it on the left size of condition)

nano/node/bootstrap/bootstrap.cpp
```diff
@@ -573,6 +592,159 @@ void nano::bootstrap_attempt::add_bulk_push_target (nano::block_hash const & hea
...
+	auto reps_weight (node->rep_crawler.total_weight ());
+	auto representatives_copy (representatives);
+	nano::uint128_t total_weight (0);
+	// Select random peers from bottom 50% of principal representatives
+	if (!representatives.empty ())
+	{
+		if (representatives.size () > 1)
+		{
+			std::reverse (representatives.begin (), representatives.end ());
+			representatives.resize (representatives.size () / 2, nano::representative{ 0, 0, nullptr });
```
wezrule: I think it may be simpler to add a default `= default` constructor in `class representative` instead of adding the second argument to all the `resize` calls. The container is only getting smaller in these cases so shouldn't really need them, but I guess the compiler complains. It would reduce some noise, however only it if you think having a default constructor on `representative` is correct.

I did wonder if there could be some optimization on the `reverse` as you only really need to reverse the top half of the representatives as the rest is discarded, but I don't think it's worth the hassle as the container will be relatively small.

nano/node/bootstrap/bootstrap.cpp
```diff
@@ -573,6 +592,159 @@ void nano::bootstrap_attempt::add_bulk_push_target (nano::block_hash const & hea
...
+	auto frontiers_count (frontiers.size ());
+	const size_t reps_limit = 20;
+	auto representatives (node->rep_crawler.representatives ());
+	auto reps_weight (node->rep_crawler.total_weight ());
+	auto representatives_copy (representatives);
+	nano::uint128_t total_weight (0);
+	// Select random peers from bottom 50% of principal representatives
+	if (!representatives.empty ())
+	{
+		if (representatives.size () > 1)
```
wezrule: I think you can probably get rid of the `representatives.empty ()` outer if block, both blocks inside are fine if it is empty, and saves a nested (indented) scope.

nano/node/json_handler.cpp

nano/node/json_handler.cpp

nano/node/bootstrap/bootstrap.cpp
SergiySW: Changed. Is it more clear now?

nano/node/bootstrap/bootstrap.cpp
SergiySW: Updated

nano/node/bootstrap/bootstrap.cpp
guilhermelawless: Perfect

nano/node/bootstrap/bootstrap.cpp
SergiySW: Updated

nano/node/bootstrap/bootstrap.cpp
SergiySW: Updated

guilhermelawless:
Added documentation label so the corresponding RPC changes are tracked.

PR #2351, opened 2019-10-17
TSAN error with block work

wezrule:
The following TSAN error is generated by the `active_transactions.update_difficulty` test with Clang on Ubuntu 18.04:
https://gist.github.com/wezrule/febc782e9d8c1863760be2055b1e728e

It is caused by calling `block->block_work_set` while reading from it in another thread, in this case serializing the block during publishing. `nano::wallet::action_complete` seems to be the only place that modifies the work in the wild. ~~I'm not sure if this is just a test issue and I'm being pessimistic here, but it seems plausible that the work on a block might be changed while something else is reading from it. The simplest solution which mimics existing behaviour is to just make the work an atomic variable.~~ @cryptocode pointed out that blocks shouldn't be shared between threads so the root of the issue is probably elsewhere, and @SergiySW pointed to the work watcher which modifies work for blocks. I have somewhat followed what is done in `nano::work_watcher::watching` and limited changes to the test itself.

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -523,8 +523,8 @@ TEST (active_transactions, update_difficulty)
 		ASSERT_NO_ERROR (system.poll ());
 	}
 	// Update work with higher difficulty
-	auto work1 = node1.work_generate_blocking (genesis.hash (), difficulty1 + 1, boost::none);
-	auto work2 = node1.work_generate_blocking (send1->hash (), difficulty2 + 1, boost::none);
+	auto work1 = node1.work_generate_blocking (send1->root (), difficulty1 + 1, boost::none);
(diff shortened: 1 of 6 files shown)


nano/core_test/active_transactions.cpp
```diff
@@ -523,8 +523,32 @@ TEST (active_transactions, update_difficulty)
 		ASSERT_NO_ERROR (system.poll ());
 	}
 	// Update work with higher difficulty
-	node1.work_generate_blocking (*send1, difficulty1 + 1);
-	node1.work_generate_blocking (*send2, difficulty2 + 1);
+	auto work1 = node1.work_generate_blocking (genesis.hash (), difficulty1 + 1, boost::none);
```
SergiySW: probably `send1->root ()` is easier to undershand here? Same for below

wezrule: Done

PR #2352, opened 2019-10-17
Add support for launch flags passed to nano_wallet

argakiig:
show help in messagebox linking docs launch flags with detail text being the help output
support launch flags being passed through nano_wallet like nano_node does

### DIFF
--- a/nano/nano_wallet/entry.cpp
+++ b/nano/nano_wallet/entry.cpp
@@ -315,7 +315,7 @@ int main (int argc, char * const * argv)
 						data_path = nano::working_path ();
 					}
 					nano::node_flags flags;
-					nano::update_flags(flags, vm);
+					nano::update_flags (flags, vm);
 					auto config (vm.find ("config"));
 					if (config != vm.end ())
 					{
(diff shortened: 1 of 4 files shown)


nano/nano_wallet/entry.cpp
```diff
@@ -239,21 +289,45 @@ int main (int argc, char * const * argv)
...
+			("disable_udp", "Disables UDP realtime network")
+			("disable_unchecked_cleanup", "Disables periodic cleanup of old records from unchecked table")
+			("disable_unchecked_drop", "Disables drop of unchecked table at startup")
+			("fast_bootstrap", "Increase bootstrap speed for high end nodes with higher limits")
+			("batch_size",boost::program_options::value<std::size_t> (), "Increase sideband batch size, default 512")
+			("block_processor_batch_size",boost::program_options::value<std::size_t> (), "Increase block processor transaction batch write size, default 0 (limited by config block_processor_batch_max_time), 256k for fast_bootstrap")
+			("block_processor_full_size",boost::program_options::value<std::size_t> (), "Increase block processor allowed blocks queue size before dropping live network packets and holding bootstrap download, default 65536, 1 million for fast_bootstrap")
+			("block_processor_verification_size",boost::program_options::value<std::size_t> (), "Increase batch signature verification size in block processor, default 0 (limited by config signature_checker_threads), unlimited for fast_bootstrap");
```
cryptocode: Looks like this duplicates the code in node's entry.cpp. Any way to share the options? Otherwise we'll have to update two places as we maintain options and descriptions.

argakiig: yeah, this should be better

PR #2368, opened 2019-10-28
Fix non-restarting timer in vote processor log

guilhermelawless:


nano/node/vote_processor.cpp
```diff
@@ -18,7 +18,7 @@ thread ([this]() {
 
 void nano::vote_processor::process_loop ()
 {
-	nano::timer<std::chrono::milliseconds> elapsed;
+	nano::timer<std::chrono::milliseconds> elapsed (nano::timer_state::stopped);
```
cryptocode: A nano timer is initially stopped so passing the stopped state shouldn't be necessary. Otherwise LGTM.

PR #2369, opened 2019-10-29
Improve distributed_work stopping with ongoing worker tasks

guilhermelawless:
Found while running RPC `epoch_upgrade` which places the task in `nano::worker`. When preemptively stopping the node (with SIGINT, not RPC "stop") the I/O threads are destroyed and the background `work_generate_blocking` task can't complete if using work peers.

This adds manual canceling of all ongoing work before attempting to destroy the `worker`.

Also fixes an issue where canceled work would turn into zero-filled work, which was only a problem when stopping the node.

### DIFF
--- a/nano/node/distributed_work.cpp
+++ b/nano/node/distributed_work.cpp
@@ -456,9 +456,8 @@ void nano::distributed_work_factory::cleanup_finished ()
 
 void nano::distributed_work_factory::stop ()
 {
-	if (!stopped)
+	if (!stopped.exchange (true))
 	{
-		stopped = true;
 		// Cancel any ongoing work
 		std::unordered_set<nano::root> roots_l;
 		nano::unique_lock<std::mutex> lock_l (mutex);
--- a/nano/node/distributed_work.cpp
+++ b/nano/node/distributed_work.cpp
@@ -457,7 +457,7 @@ void nano::distributed_work_factory::stop ()
 	std::unordered_set<nano::root> roots_l;
 	{
 		nano::lock_guard<std::mutex> guard_l (mutex);
-		for (auto & item_l : items)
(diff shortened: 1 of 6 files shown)


nano/node/distributed_work.cpp
```diff
@@ -448,6 +451,23 @@ void nano::distributed_work_factory::cleanup_finished ()
 	}
 }
 
+void nano::distributed_work_factory::stop ()
+{
```
wezrule: Should `items` be cleared at the end of this function? Otherwise in the destructor they can be cancelled again

guilhermelawless: Each vector of work is erased in `cancel()` , but I've added a clear to make sure, and a `stopped` flag to avoid calling `stop()` twice or adding new work in `make ()` after stopping

nano/node/distributed_work.cpp
```diff
@@ -448,6 +454,27 @@ void nano::distributed_work_factory::cleanup_finished ()
...
+			roots_l.insert (item_l.first);
+		}
+		lock_l.unlock ();
+		for (auto const & root_l : roots_l)
+		{
+			cancel (root_l, true);
+		}
+		lock_l.lock ();
+		items.clear ();
+	}
```
cryptocode: I wonder if this can be simplified using `swap` (we use that pattern a few places). The first loop and the clear would be gone. Or do you need `items` to be alive during the cancellation?

guilhermelawless: Yeah it needs to persist, otherwise `cancel ()` won't find the root in items. An alternative here would be to replicate `cancel ()` in `stop ()`  directly, but that has code redundancy. WDYT?

PR #2378, opened 2019-11-01
[ASAN] memory leak between tcp channels and sockets

wezrule:
This was reproducible on Ubuntu with Clang, not on MacOS (thanks @cryptocode for testing it and assistance during this), Valgrind also caught this. It was most reproducible with the `node.send_single_many_peers` test, however may need to be run multiple times before seeing it. These leak stacktraces were shown: https://gist.github.com/wezrule/02f5a8d922c13b4bc277581b9e919f64

From the stacktraces it is not apparent what the cause of the issue is, and many different things were tried. The main issue stems from a circular relationship between `channel_tcp`, `bootstrap_server` and `socket`. `channel_tcp` stores a `shared_ptr` to both `bootstrap_server` and `socket`. `socket` stores callbacks in `send_queue`, most of these contain `channel_tcp` shared_ptrs,  so `socket` and `channel_tcp` can reference each other. Breaking this relationship doesn't fully fix the issue as `channel_tcp` also has a `bootstrap_server` member which has a `socket` member so it implicitly still ends up with a circular reference, so this needed removing too.

Deciding what should be responsible for owning the lifetime was non-obvious due to the multi-step nature of a node handshake. What I ended up doing was temporarily storing the `socket` in a collection during the node id handshake and passing it off to the `bootstrap_server` (and subsequently removing it from this collection). Was not sure if this was sufficient in case of handshake failures, so am also purging sockets in the existing `tcp_channels::purge ()` function.

### DIFF
--- a/nano/node/bootstrap/bootstrap_bulk_pull.cpp
+++ b/nano/node/bootstrap/bootstrap_bulk_pull.cpp
@@ -121,62 +121,81 @@ void nano::bulk_pull_client::throttled_receive_block ()
 void nano::bulk_pull_client::receive_block ()
 {
 	auto this_l (shared_from_this ());
-	connection->channel->socket->async_read (connection->receive_buffer, 1, [this_l](boost::system::error_code const & ec, size_t size_a) {
-		if (!ec)
-		{
-			this_l->received_type ();
-		}
-		else
-		{
-			if (this_l->connection->node->config.logging.bulk_pull_logging ())
+	if (auto socket_l = connection->channel->socket.lock ())
+	{
+		socket_l->async_read (connection->receive_buffer, 1, [this_l](boost::system::error_code const & ec, size_t size_a) {
+			if (!ec)
 			{
-				this_l->connection->node->logger.try_log (boost::str (boost::format ("Error receiving block type: %1%") % ec.message ()));
+				this_l->received_type ();
 			}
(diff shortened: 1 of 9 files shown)


nano/node/bootstrap/bootstrap_bulk_pull.cpp
```diff
@@ -121,62 +121,81 @@ void nano::bulk_pull_client::throttled_receive_block ()
...
 	nano::block_type type (static_cast<nano::block_type> (connection->receive_buffer->data ()[0]));
+	auto socket_l = connection->channel->socket.lock ();
 	switch (type)
 	{
 		case nano::block_type::send:
 		{
-			connection->channel->socket->async_read (connection->receive_buffer, nano::send_block::size, [this_l, type](boost::system::error_code const & ec, size_t size_a) {
-				this_l->received_block (ec, size_a, type);
-			});
+			if (socket_l)
```
SergiySW: Why check if here for each block type & not before for all?

SergiySW:
Tested on Ubuntu with GCC 9, no asan in `node.send_single_many_peers`

wezrule:
I did a bootstrap overnight with ASAN. I'm up to 17M checked blocks without any issues. 192 tcp channels and "temp_node_id_handshake_socket_container" was 0 too. Realised that name is wrong, updated to "node_id_handshake_socket".

PR #2386, opened 2019-11-04
Prevent concurrent bootstrap populate_connections ()

SergiySW:
From run (), lazy_run () etc. if there are several runs in single bootstrap attempt

### DIFF
--- a/nano/node/bootstrap/bootstrap.cpp
+++ b/nano/node/bootstrap/bootstrap.cpp
@@ -231,7 +231,7 @@ void nano::bootstrap_attempt::run_start (nano::unique_lock<std::mutex> & lock_a)
 void nano::bootstrap_attempt::run ()
 {
 	assert (!node->flags.disable_legacy_bootstrap);
-	populate_connections ();
+	start_populate_connections ();
 	nano::unique_lock<std::mutex> lock (mutex);
 	run_start (lock);
 	while (still_pulling ())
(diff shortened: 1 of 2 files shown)


nano/node/bootstrap/bootstrap.cpp
```diff
@@ -879,7 +883,10 @@ void nano::bootstrap_attempt::lazy_clear ()
 void nano::bootstrap_attempt::lazy_run ()
 {
 	assert (!node->flags.disable_lazy_bootstrap);
-	populate_connections ();
+	if (!populate_connections_started)
```
cryptocode: I think we want `if (populate_connections_started.exchange(true)` everywhere, otherwise there's a race (and remove the assignment in populate_connections). Maybe factor that out into 

```
void try_populate_connections()
{
    if (populate_connections_started.exchange(true)
    {
            populate_connections ();
    }
}
```
?

nano/node/bootstrap/bootstrap.cpp
```diff
@@ -455,6 +452,14 @@ void nano::bootstrap_attempt::populate_connections ()
 	}
 }
 
+void nano::bootstrap_attempt::start_populate_connections ()
+{
+	if (populate_connections_started.exchange (true))
```
guilhermelawless: The condition should be negated I think

> Return value
    The value of the atomic variable before the call. 

https://en.cppreference.com/w/cpp/atomic/atomic/exchange

cryptocode: yep, i gave a bad snippet in the review 😬

PR #2387, opened 2019-11-05
Fix early return preventing mutex lock in attempt_restart_check

guilhermelawless:
Best reviewed without whitespace changes.
Thanks to sev for the detailed report.

### DIFF
--- a/nano/node/bootstrap/bootstrap.cpp
+++ b/nano/node/bootstrap/bootstrap.cpp
@@ -617,7 +617,7 @@ void nano::bootstrap_attempt::attempt_restart_check (nano::unique_lock<std::mute
 	if (!frontiers_confirmed && (requeued_pulls > (!node->network_params.network.is_test_network () ? nano::bootstrap_limits::requeued_pulls_limit : nano::bootstrap_limits::requeued_pulls_limit_test) || total_blocks > nano::bootstrap_limits::frontier_confirmation_blocks_limit))
 	{
 		confirm_frontiers (lock_a);
-		assert (!lock_a.try_lock ());
+		assert (lock_a.owns_lock ());
 		if (!frontiers_confirmed)
 		{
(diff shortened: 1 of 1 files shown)


nano/node/bootstrap/bootstrap.cpp
```diff
@@ -617,6 +617,7 @@ void nano::bootstrap_attempt::attempt_restart_check (nano::unique_lock<std::mute
 	if (!frontiers_confirmed && (requeued_pulls > (!node->network_params.network.is_test_network () ? nano::bootstrap_limits::requeued_pulls_limit : nano::bootstrap_limits::requeued_pulls_limit_test) || total_blocks > nano::bootstrap_limits::frontier_confirmation_blocks_limit))
 	{
 		confirm_frontiers (lock_a);
+		assert (!lock_a.try_lock ());
```
wezrule: "If try_lock is called by a thread that already owns the mutex, the behavior is undefined." https://en.cppreference.com/w/cpp/thread/mutex/try_lock
Can that happen here?

guilhermelawless: Interesting. No such mention in https://en.cppreference.com/w/cpp/thread/unique_lock/try_lock (for unique locks), although we use this idiom at least 10 times throughout the code directly with mutexes.

wezrule: It also says it effectively just calls `mutex.try_lock`. I guess they are all surrounded in `assert`s so no user would see anything anyway, so it's probably fine.

guilhermelawless: Do you think it's better to move it outside the assert and use `owns_lock()` ?

wezrule: `owns_lock` is probably better yes

PR #2389, opened 2019-11-05
[Valgrind] Various uninitialized data accesses

wezrule:
1. In `bulk_pull` class https://gist.github.com/wezrule/387084bf4189f5dd85fb044abd917d1c
2. In `websocket.work` test https://gist.github.com/wezrule/174edc5e141143fb5656dbe7ad9a5522
3. In `rpc.wallet_create_seed` test https://gist.github.com/wezrule/05b1a84d0582b40cc3f7c16932f3367c (same in `rpc.wallet_change_seed`)
4. In `rpc.wallet_create_fail` https://gist.github.com/wezrule/014686344f0ff43470143d1fa25a7a55

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -977,6 +977,7 @@ TEST (rpc, wallet_create_seed)
 	nano::system system (24000, 1);
 	scoped_io_thread_name_change scoped_thread_name_io;
 	nano::raw_key seed;
+	seed.data = 1;
 	auto prv = nano::deterministic_key (seed, 0);
 	auto pub (nano::pub_key (prv));
 	auto node = system.nodes.front ();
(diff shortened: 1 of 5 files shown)


nano/rpc_test/rpc.cpp
```diff
@@ -977,6 +977,7 @@ TEST (rpc, wallet_create_seed)
 	nano::system system (24000, 1);
 	scoped_io_thread_name_change scoped_thread_name_io;
 	nano::raw_key seed;
+	seed.data = 1;
```
SergiySW: May be better use nano::keypair that generates ramdom value?

wezrule: Good point, calling `nano::random_pool::generate_block` to generate random data now.

PR #2394, opened 2019-11-06
Improve bootstrap frontiers confirmation

SergiySW:
* start new bootstrap attempt after failed confirmation
* prevent duplicates in frontiers vector
* increase chances of starting first bootstrap connection to requested peer
* disable rep crawler for specific tests (it can start elections)
* return confirmation status in confirm_frontiers () function
* add `force` option to `bootstrap_any` RPC

### DIFF
--- a/nano/node/bootstrap/bootstrap.cpp
+++ b/nano/node/bootstrap/bootstrap.cpp
@@ -101,10 +101,10 @@ bool nano::bootstrap_attempt::should_log ()
 	return result;
 }
 
-bool nano::bootstrap_attempt::request_frontier (nano::unique_lock<std::mutex> & lock_a)
+bool nano::bootstrap_attempt::request_frontier (nano::unique_lock<std::mutex> & lock_a, bool first_attempt)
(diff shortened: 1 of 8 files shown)


nano/node/bootstrap/bootstrap.cpp
```diff
@@ -1298,9 +1309,19 @@ nano::bootstrap_initiator::~bootstrap_initiator ()
 	stop ();
 }
 
-void nano::bootstrap_initiator::bootstrap ()
+void nano::bootstrap_initiator::bootstrap (bool force)
 {
 	nano::unique_lock<std::mutex> lock (mutex);
+	if (force)
+	{
+		if (attempt != nullptr)
```
wezrule: minor, but could probably combine the `if (attempt != nullptr)` statement into the one above and have `if (force && attempt != nullptr)` to save a scope & indentation.

SergiySW: Sure

PR #2399, opened 2019-11-09
Add HOST header to distributed work request

SergiySW:


nano/node/distributed_work.cpp
```diff
@@ -7,6 +7,13 @@ std::shared_ptr<request_type> nano::work_peer_request::get_prepared_json_request
 	auto request (std::make_shared<boost::beast::http::request<boost::beast::http::string_body>> ());
 	request->method (boost::beast::http::verb::post);
 	request->set (boost::beast::http::field::content_type, "application/json");
+	auto address_string (address.to_string ());
+	auto pos (address_string.find ("::ffff:"));
+	if (pos != std::string::npos)
+	{
+		address_string.replace (pos, 7, "");
+	}
```
cryptocode: This could be simplified (if including <boost/algorithm/string/erase.hpp> is fine):

`auto address_string = boost::algorithm::erase_first_copy (address.to_string (), "::ffff:");`

SergiySW: Updated

PR #2403, opened 2019-11-12
The start of CLI tests

wezrule:
Our CLI code currently has no test coverage. This is the beginning of it, any new CLI commands should now have tests added. The hope is that existing CLI commands will have tests added over time too, but some may need some restructuring to make them more easily testable though.


nano/core_test/cli.cpp
```diff
@@ -0,0 +1,50 @@
...
+	vm.emplace ("key_create", boost::program_options::variable_value ());
+
+	std::stringstream ss;
+	nano::cout_redirect redirect (ss.rdbuf ());
+
+	// Execute CLI command. This populates the stringstream with a string like: "Private: 123\n Public: 456\n Account: nano_123"
+	auto ec = nano::handle_node_options (vm);
+	ASSERT_FALSE (ec);
+
+	// Extract the private, public and account values. The regular expression extracts anything between the semi-colon and new line.
```
SergiySW: Could this be some common function for other tests here?

nano/core_test/cli.cpp
```diff
@@ -0,0 +1,50 @@
...
+		search_start = matches.suffix ().first;
+	}
+
+	// Get the contents of the private key and check that the public key and account are successfully derived from the private key
+	auto private_key_str = vals.front ();
+	nano::private_key private_key;
+	private_key.decode_hex (private_key_str);
+
+	nano::public_key public_key;
+	ed25519_publickey (private_key.bytes.data (), public_key.bytes.data ());
```
SergiySW: I guess can use `nano::pub_key (...)` without direct inluding ed25519 header

PR #2406, opened 2019-11-15
[MSAN] remove zero_key from ledger_constants

wezrule:
This was found when trying MSAN. This is the first error encountered, it complains about generating the keypair from the `"0"` string, inside `nano::decode_hex`. `zero_key` isn't used so just removing it.


wezrule:
oh, incoming travis failure then :D, I didn't try Gui build

wezrule:
Closing this because it's still used, another way will need to be done to remove the MSAN warning.

PR #2407, opened 2019-11-16
Improve indentation formatting with clang-format

cryptocode:
This sets an indent for continuations, and makes constructor initializer lists line up. This is subjective, but I think it improves the readability of:

- Large constructors (like node::node)
- Templates declarations (like anything boost multi-index related)
- When lambdas are function arguments
- Fluent calls (like stream piping and block builder calls)

These formatting rules were applied in nano-pow-server, which seemed to work well.


wezrule:
LGTM, I too prefer the constructor member initializer syntax here as each new/deletion only changes 1 line in the diff. While we are at it I have a couple other personal suggestions. I like to use 2 space tabs in C++ because it's verbose so long lines are common, every little helps to reduce that. And what about limiting column length to something like 96? Might promote more refactoring of variables etc when it ends up looking crappy by the formatter.

cryptocode:
@wezrule right, second commit uses tabs for continuation indents so your personal tab size preference should work.

nano/node/network.cpp
```diff
@@ -4,13 +4,14 @@
...
-buffer_container (node_a.stats, nano::network::buffer_size, 4096), // 2Mb receive buffer
-resolver (node_a.io_ctx),
-node (node_a),
-udp_channels (node_a, port_a),
-tcp_channels (node_a),
-disconnect_observer ([]() {})
+nano::network::network (nano::node & node_a, uint16_t port_a)
+	: buffer_container (node_a.stats, nano::network::buffer_size, 4096)
+	, // 2Mb receive buffer
+	resolver (node_a.io_ctx)
```
wezrule: The placement of the comment looks a bit strange, perhaps just remove it (it seems inaccurate anyway?)

nano/node/write_database_queue.cpp
```diff
@@ -15,8 +15,8 @@ nano::write_guard::~write_guard ()
 	cv.notify_all ();
 }
 
-nano::write_database_queue::write_database_queue () :
-// clang-format off
+nano::write_database_queue::write_database_queue ()
+	: // clang-format off
 guard_finish_callback ([&queue = queue, &mutex = mutex]() {
```
wezrule: Again placement of comment looks strange

cryptocode: I don't think that can be controlled, but in this case the format directive can enclose the entire constructor, which is empty.

nano/node/network.cpp
cryptocode: Looks like it's accurate (512*4096) but seems prone to get out of sync, so can probably be removed.

nano/node/write_database_queue.cpp
cryptocode: Updated

nano/node/network.cpp
cryptocode: Updated

zhyatt:
@theohax Include this in your evaluation of clang-format items

theohax:
Superseded by https://github.com/nanocurrency/nano-node/pull/3506

PR #2408, opened 2019-11-17
message_queue websocket for recording tcp write drops

Srayman:
Similar to confirmation_quorum peer_details and peers that show relevant information related to peers and vote websocket that can be used to measure vote performance, this websocket sends a message with the endpoint whenever the message queue reaches queue_size_max (eg 128) for a particular endpoint.  Tests so far have shown this could be a way to measure node performance but further tests are needed to confirm and ultimately sampled and averaged across multiple nodes to get usable data.

Core test case only tests the websocket subscription, I'm not sure the best way to automate testing a full message queue, but bootstrapping 600k blocks on beta resulted in 536 tcp_write_drop count which matches the messages returned by the websocket.

### DIFF
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -778,6 +778,41 @@ TEST (websocket, work)
 	ASSERT_EQ (contents.get<std::string> ("reason"), "");
 }
 
+/** Subscribes to message_queue */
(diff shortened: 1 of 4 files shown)


guilhermelawless:
Closing this out, an approach will follow likely for V22, but outside of this hot code path.

PR #2411, opened 2019-11-22
Fix active_transactions not doing any confirmation request in tests

guilhermelawless:
Some tests expect an increasing confirmation_request_count even when confirm_req fails (e.g. no reps available), so I made it always increase for the test network.

Without that, note that `bootstrap_processor.lazy_unclear_state_link_not_existing` fails, because active_transactions would not escalate to broadcasting the block.

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -6,6 +6,43 @@
 
 using namespace std::chrono_literals;
 
+TEST (active_transactions, confirm_one)
+{
+	nano::system system;
+	nano::node_config node_config (24000, system.logging);
+	auto & node1 = *system.add_node (node_config);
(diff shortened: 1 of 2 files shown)


nano/core_test/active_transactions.cpp
```diff
@@ -6,6 +6,43 @@
 
 using namespace std::chrono_literals;
 
+TEST (active_transactions, confirm_one)
+{
+	nano::system system;
+	nano::node_config node_config (24000, system.logging);
+	auto & node1 = *system.add_node (node_config);
+	nano::genesis genesis;
```
SergiySW: Not required here it seems

guilhermelawless: Thanks, leftover from testing

PR #2413, opened 2019-11-24
Fix distributed_work segfault when local work generation fails

guilhermelawless:
In start_work () :
`else if (!this_l->cancelled && !this_l->completed)`

Was not setting any of these atomics to true, so the same callback would possibly be called in `cancel_once`. Not normally an issue, but can happen and came up with Serg's epoch storm test. Instead of making a new atomic, joined both into a single `finished` atomic, and added a `work_generation_status` enum.

Also added a check `if (callback)` on every call.

### DIFF
--- a/nano/node/distributed_work.hpp
+++ b/nano/node/distributed_work.hpp
@@ -32,6 +32,14 @@ public:
 	boost::asio::ip::tcp::socket socket;
 };
 
+enum class work_generation_status
+{
+	ongoing,
+	success,
+	cancelled,
+	failure
+};
+
 /**
  * distributed_work cancels local and peer work requests when going out of scope
  */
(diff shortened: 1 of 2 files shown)


nano/node/distributed_work.hpp
```diff
@@ -32,6 +32,14 @@ class work_peer_request final
 	boost::asio::ip::tcp::socket socket;
 };
 
+enum class work_generation_status
+{
+	ongoing,
+	success,
+	cancelled,
+	failure
+};
```
cryptocode: Minor: Maybe move this inside class `distributed_work` which is the only customer, as to not pollute the `nano` namespace.

guilhermelawless: Good idea; thanks!

PR #2414, opened 2019-11-25
Add block and voters count in election information

guilhermelawless:
Considered adding more details such as the number of votes for each block, but would need to tally up each election.

Similar to a previous PR, uses `assert (!node.active.mutex.try_lock ());` which can have [undefined behavior](https://en.cppreference.com/w/cpp/thread/mutex/try_lock), but only used in Debug. Don't see an easy alternative in this case, and it was already used in `election.cpp`.

### DIFF
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -468,6 +468,8 @@ TEST (websocket, confirmation_options)
 			// Duration and request count may be zero on testnet, so we only check that they're present
 			ASSERT_EQ (1, election_info.count ("duration"));
 			ASSERT_EQ (1, election_info.count ("request_count"));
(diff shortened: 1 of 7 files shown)


cryptocode:
Most likely for a separate PR: I know it's already done a couple of places, but `election` referencing `node.active.mutex` seems wrong to me. Not sure if we should add more of those. There are many data types that need external synchronization, and they generally don't reach out to an external mutex to see if it's locked. Maybe the cleanest solution is to put a mutex on election and make sure its properly encapsulated.

guilhermelawless:
Agree 100%, see issue 2415 for tracking this. One glaring location is `vote_processor::vote_blocking`

PR #2416, opened 2019-11-25
Disable vote rebroadcasting while cemented count is below hardcoded bootstrap block count

### LINKED ISSUE #2410: Remove vote rebroadcasting for bootstrapping nodes
During bootstrapping from scratch, I've noticed that vote rebroadcasting seems to be fairly high.  At least going by confirm_ack out stats.  If votes received during frontier confirmations are rebroadcasted while bootstrapping could this be removed?  I would think there would be little to no value in rebroadcasting votes for old blocks/frontiers which adds extra network traffic and processing time for the nodes that receive them.

Also this would reduce the network overhead for nodes bootstrapping from scratch.  For example, I manually reduced the fanout from sqrt(peers) to sqrt(peers)/4 while bootstrapping and noticed a significant reduction in upload bandwidth.  During frontier confirmation after the change I was receiving about 4.1 Mbps down but 25 Mbps upload which would line up with the reduction in fanout (eg sqrt(290)/4 is a fanout of 5 peers for each vote received and 4.1 * 5 plus the additional confirm_req is close to 25 Mbps)

guilhermelawless:
Closes #2410 . It's an heuristic since blocks that don't go towards the bootstrap block count can be cemented, but close enough. Other heuristics can follow in the future to understand the likelihood of a vote being useful to the network.

One thing to take into account is nodes with disabled frontier confirmation ( #2175 ); it's likely they also reach this cemented count through the confirmation height processor trickling down through millions of blocks. Disabling vote rebroadcasting completely for these nodes would be another option.

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -692,7 +692,7 @@ bool nano::active_transactions::vote (std::shared_ptr<nano::vote> vote_a, bool s
 		{
 			lock = nano::unique_lock<std::mutex> (mutex);
 		}
-		for (auto vote_block : vote_a->blocks)
+		for (auto & vote_block : vote_a->blocks)
 		{
 			nano::election_vote_result result;
 			if (vote_block.which ())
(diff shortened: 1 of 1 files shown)


zhyatt:
We are planning to hold off on merging this for now in favor of looking into limiting rebroadcasting of votes from confirm_req more generally.

PR #2417, opened 2019-11-25
Consistent checks for recently confirmed blocks

guilhermelawless:
It is currently possible for an election to be finished (confirmed) and under processing by the confirmation_height_processor, but not signalled as such by `pending_confirmation_height::is_processing_block`.

`active_transactions` has the `pending_conf_height` set, where blocks are added when the election is confirmed, and only removed after the confirmation height is set. So I added `active::is_confirming_block` in place of the previous `is_processing_block`.

Symptom of tests failing was the following. `active_transactions::search_frontiers` polls `pending_confirmation_height::is_processing_block` to know if it should add a frontier. When added, RPCs pending and wallet_pending would go through `json_handler::block_confirmed`. This checks the active elections, which would contain the wrongly re-added frontier.

Unrelated, but since this was discovered with the rpc.pending test, adjusted `json_handler::pending` to have a similar loop to `ledger::account_pending`.


guilhermelawless:
With the short 500ms active_transactions loop, and since it already goes over finished elections anyway, I changed so that `active.erase ()` is not called from the election itself.

guilhermelawless:
Last commit makes it so that roots are only removed by dropping (within active_transactions , which is the parent of elections) or when elections are finished, but only triggered from the election itself to ensure it only happens after sending for confirmation height processing. This ensures the election can always be found, in the following order:

1. In active.roots / active.blocks
2. The currently processing hash in confirmation_height
3. Pending in confirmation_height
4. Writing by confirmation_height
5. In the ledger

guilhermelawless:
This requires more testing, closing this one and opening another PR in the near future.

PR #2419, opened 2019-11-27
Set wallet representative with RPC receive on unopened accounts

### LINKED ISSUE #2418: Wallet RPC action "receive" sets the wrong representative/ignores the wallet representative
This issue has been present since V14 or so, it is the same as #1019 - but was missing an important detail which is why it was closed as un-reproducible

**Description of bug:**

When you use the rpc `receive` action on a node/development wallet then the representative is set to the account itself, rather than the wallet representative.

Example, if you have an account in the node wallet:

`nano_1hmefcfq35td5f6rkh15hbpr4bkkhyyhmfhm7511jaka811bfp17xhkboyxo`

And you do an RPC `receive` for any block, the representative for that account will become `nano_1hmefcfq35td5f6rkh15hbpr4bkkhyyhmfhm7511jaka811bfp17xhkboyxo`

This issue **does not** occur when transactions are auto-received, only when they are received using the RPC receive action.

1. Create a wallet via `wallet_create` with auto receive disabled
2. Create an account on the wallet via `account_create`
3. Use rpc `receive` to pocket the funds from the account
4. The representative for that account is now set to itself

Accounts in the node/developer wallet always set the representative field to themselves when using rpc `receive`

Accounts use the wallet representative

Issue happens consistently and is repeatable.

**Environment**:

- Ubuntu 18.04
- Nodes V14, V15, V16, V17, V18, V19, V20

guilhermelawless:
Closes #2418 

Will update documentation, should state that the wallet representative will be used from V21, and the own account is used in V20 and before, for unopened accounts.

Adds a suite of tests for RPC receive.

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -3373,10 +3373,14 @@ void nano::json_handler::receive ()
 					}
 					if (!ec)
 					{
+						// Representative is only used by receive_action when opening accounts
+						// Set a wallet default representative for new accounts
+						nano::account representative (wallet->store.representative (node.wallets.tx_begin_read ()));
+
 						bool generate_work (work == 0); // Disable work generation if "work" option is provided
 						auto response_a (response);
 						// clang-format off
-						wallet->receive_async(std::move(block), account, node.network_params.ledger.genesis_amount, [response_a](std::shared_ptr<nano::block> block_a) {
+						wallet->receive_async(std::move(block), representative, node.network_params.ledger.genesis_amount, [response_a](std::shared_ptr<nano::block> block_a) {
 							if (block_a != nullptr)
 							{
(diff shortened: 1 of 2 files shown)


nano/node/json_handler.cpp
```diff
@@ -3373,10 +3373,14 @@ void nano::json_handler::receive ()
 					}
 					if (!ec)
 					{
+						// Representative is only used by receive_action when opening accounts
+						// Set a wallet default representative for new accounts
+						nano::account representative (wallet->store.representative (node.wallets.tx_begin_read ()));
```
cryptocode: json_handler::receive () has already started a read tx that can be used here right?

guilhermelawless: Yes it does, missed that. Thanks!

PR #2420, opened 2019-11-28
Add database vendor string to RPC version

guilhermelawless:
`"store_vendor": "RocksDB 6.6.0"` or `"LMDB 0.9.23"` in my case.

Also logs the same string on node startup, currently it's only printing to stdcout.


wezrule:
I wonder if the node startup standard output should include this too, as we include other version information there too.

guilhermelawless:
Your wish == my commit

PR #2425, opened 2019-12-04
Improve build times

wezrule:
Added new file `random_pool_shuffle.hpp` because I wanted to remove the cryptopp `orng.h`  header (which is quite large) from `random_pool.hpp` which is included in a lot of places, and only a couple of source files were actually making use of shuffling.

Removed the top level nano/boost/asio.hpp, beast.hpp and process.hpp header files, because these are quite large (asio is 200k LOC for instance), and we only use a subset of it, so this prevent it being used. As a consequence I had to add a lot of the other child headers to our `nano/boost` directory, so that warnings for these headers are not displayed on Windows.

Created out-of-class function definitions in source files for template classes where the types of template arguments are known in advance (and within our dependency scope). This means we can provide explicit instantiations for these templates once rather than for every source file which includes them. This was done for `blockbuilders.hpp/cpp`, `timer.hpp/cpp`,. For `block_store_partial` this could not be done because lmdb/rocksdb are in a higher level, so extern templates were used (thanks @cryptocode).

Config files (`websocket_config`, `rpc_config`, `node_config`): Removed `boost::asio::ip::address_v6` members which required a boost header file and replaced with `std::string` members which are much smaller. This did mean that when a `boost::asio::ip::address_v6` object was needed it had to be created on the fly but it didn't seem to be that expensive an operation.

Created source files for `jsonconfig.hpp` and `tomlconfig.hpp` to move some of the dependencies into the source file. Moved more inline definitions from `stats.hpp` into the source file too.

Create new `nano/lib/threading.hpp` file, because the threading stuff was in a generic `nano/lib/utility.hpp` file which has dependencies the threading part doesn't care about.

Replaced `boost::thread` with `std::thread` members where appropriate (the `<thread>` header is smaller).

Created forward declarations in header files where possible (i.e the type is just a pointer/reference, essentially anywhere `sizeof` does not need to be called on it).

Separated out `distributed_work_factory` into its own file so that it could be included without other dependencies and similar for various other files.

Used ReSharper extension for Visual studio to remove unused includes, this however broke a lot of the other platforms on travis, so stitched them back together gradually.

The remaining issue in compile times is the `node.hpp` header file. Removing the node dependency in the various files we have in `/node/*` will help substantially, there is a related ticket for this already. However as part of this I removed some node dependencies for some of the classes which were too much effort/had low regression risk. This was done for `online_reps`, `websocket`, `voting`, `vote_processor` and `online_reps`. Related to this though is the `core_test` files which include `node.hpp`, which is most of them. However they do legitimately use the node, I think an easy win would be to combine some of the smaller test files dependent on the node, this would mean the `node.hpp` header file doesn't need to be parsed as often, but I will leave this as an after thought.

Using `time make -j 4` on my linux subsystem for Windows I got the following results (full build):

This branch:
real    6m15.598s
user    16m30.063s
sys     5m25.875s

develop branch:
real    9m6.802s
user    24m37.578s
sys     7m28.969s

Seems to be around 30% build improvement.

My MSVC builds started failing when using boost 1.71, complaining that it required our source to be built with `/EHsc`, I don't know if this is something I screwed up on my system or not, but added here as it seems benign enough.

I haven't done any build optimisation for the QT wallet.

Unrelated: A few files had some unused variables causing warnings on mac which I've removed.
Unrelated: `boost::asio::ip::address_v6::from_string` is deprecated, using `boost::asio::ip::make_address_v6` instead.


SergiySW:
Clean recompile changed from 1:28 to 1:07 for me

PR #2432, opened 2019-12-09
Generated votes cache size increase and upper bound

guilhermelawless:
Memory footprint is insignificant compared to the cache for votes from other representatives.


SergiySW:
I guess it can use number of voting account in node as well to prevent overflow with current settings

guilhermelawless:
@SergiySW updated, had to make a copy of representatives in `wallet::foreach_representative` to avoid a deadlock.

New test added and a weird clang-format change.

PR #2439, opened 2019-12-14
[TSAN] More data races solved in distributed_work

guilhermelawless:
Use an atomic in `rpc.work_peer_many` and check if the request wasn't `stopped` before proceeding.

Placing the local work generation part at the end of `start_work ()` solves an issue where it would stop and remove the request, resulting in a `heap-after-free` on the `outstanding` object. This only happens in tests.

Still have one data race to be solved between `async_write` and `socket.close` , will need more investigation though. Unless you have any idea @cryptocode ? Does this need a strand?

### DIFF
--- a/nano/node/distributed_work.cpp
+++ b/nano/node/distributed_work.cpp
@@ -94,6 +94,81 @@ void nano::distributed_work::start_work ()
 {
 	auto this_l (shared_from_this ());
 
+	if (!outstanding.empty ())
+	{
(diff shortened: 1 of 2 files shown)


cryptocode:
@guilhermelawless yeah, strands would be best. For closing the socket, this means wrapping the close call in asio's post or dispatch functions.

guilhermelawless:
New PR will replace this

PR #2440, opened 2019-12-14
WIP: Passive elections

guilhermelawless:
Adds the concept of passive elections, with both local changes to the active_transactions loop and non-breaking rule changes to answering requests for votes.

The goal is to better align the network on the highest priority transactions. One of the current observed issues is a lot of dropped messages under saturation. It is likely that a lot of these messages are blocks that won't be seen by every voting node consistently, resulting in delayed elections. Most changes here affect the behavior of nodes only under saturation.

The PR is split into three parts, described below.

1. Adding the concept of passive elections to active_transactions
    - Instead of dropping elections into a cache (V20 addition), elections are now completely dropped once the `roots` container is full. However, the size of this container is greatly increased (to 50k from 10k, and could likely be larger - tests needed). By default, only 2k elections are active, which the request loop actively goes through and requests confirmations for. The other [up to] 48k elections are passive.
    - Dropped elections can later be confirmed via frontier confirmation, since the dropped cache has been removed for simplicity. Nodes should actively target active difficulty even before pushing a block, or they may see their transactions dropped by representatives. This may need some changes to the node wallet algorithm.
    - Active difficulty is calculated from active elections only. The result would be a higher difficulty than before under saturation. I changed to use the 10th percentile of active difficulties instead of the median, which I think makes more sense here given that we're trying to give users the minimum expected difficulty to have good prioritization. This needs testing.
    - Since there is no random access to the `roots` container, an approximation is made by saving the difficulty of the last active election every loop when under saturation. With this approach, when a new election is started, we know immediately, with the possibility of some error, if the inserted election should be active or passive.
    - I decided against adding a second container of passive roots for 2 reasons: it's faster to sort through a large container than 2 smaller ones (with the same combined size), and it would imply moving memory around.

2. Local actions performed for inserted passive elections
    - The node does not immediately vote for an election inserted as passive. This is the main contributor to hopefully lower overall bandwidth under saturation, thus focusing network resources on the current set of prioritized elections.
    - When an election goes from passive to active, votes are created and pushed via the vote generator for the winner via the vote generator.
    - Since votes for any election (even if passive) are republished, this is only effective network-wise when most nodes are using these changes. While we could have nodes not republish votes for passive elections (this was the case before the second commit here), they would have to be delayed votes to account for timing differences, which creates its own set of problems.

3. Restrictions on replying to vote requests
    - The above changes are a choice of each node, and they cannot be enforced by representatives unless a change is made to how we handle confirm_req. Currently, the following checks are done: 1) votes cache, 2) block exists, 3) root successor exists. This PR adds another case before checking the ledger. If there is an active election containing the hash, then the vote will be for the **election winner**, and if the election is passive, the **request is ignored**. This change forces other nodes to respect the rules laid above. Otherwise, they could bypass them and request confirmation for a low difficulty block.

Finally, there is a change in elections when a new block is the winner - the votes_cache is invalidated and votes are generated for the new winner. This is unrelated but included since changes were made to handling requests.

-----

Notes for reviews:
- The optional return in `block_account_height` is not my favorite approach, but I couldn't think of anything better and easy to use.
- Made a few TSAN runs with no issues and this was run on a stress test on beta. The node does not perform any worse than before, with bandwidth improvements. Once more nodes and especially reps use this PR we may see the other benefits outlined above.

-----

To-do

- [ ] Decide on `active_elections_size` and `passive_elections_size`
- [ ] Decide on percentile for `active_difficulty`
- [ ] Documentation for RPC, stats and active difficulty changes
- [ ] Include "active" state in confirmation callbacks

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -7,21 +7,26 @@
 
 using namespace std::chrono_literals;
 
-TEST (active_transactions, confirm_one)
+TEST (active_transactions, confirm_active)
 {
 	nano::system system;
-	nano::node_config node_config (24000, system.logging);
+	nano::node_config node_config (nano::get_available_port (), system.logging);
 	auto & node1 = *system.add_node (node_config);
 	// Send and vote for a block before peering with node2
 	system.wallet (0)->insert_adhoc (nano::test_genesis_key.prv);
 	auto send (system.wallet (0)->send_action (nano::test_genesis_key.pub, nano::public_key (), node_config.receive_minimum.number ()));
 	system.deadline_set (5s);
-	while (!node1.active.empty () && !node1.block_confirmed_or_being_confirmed (node1.store.tx_begin_read (), send->hash ()))
+	while (node1.ledger.cemented_count != 2)
 	{
 		ASSERT_NO_ERROR (system.poll ());
 	}
-	node_config.peering_port = 24001;
+	node_config.peering_port = nano::get_available_port ();
 	auto & node2 = *system.add_node (node_config);
+	std::atomic<bool> confirmed{ false };
+	// Attach to node2 confirmation observer
+	node2.observers.blocks.add ([&confirmed](const nano::election_status & status_a, const nano::account &, const nano::uint128_t &, bool) {
+		confirmed = true;
+	});
 	system.deadline_set (5s);
 	// Let node2 know about the block
 	while (node2.active.empty ())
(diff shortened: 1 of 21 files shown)


guilhermelawless:
This is being replaced by a few separate PRs:

- #2641 increases active elections to 50k allowing more flexibility in managing active elections
- Local voting behavior changes for inserted elections with PoW difficulty below a certain dynamic threshold - so called passive elections
- Restrictions on replying to confirm_req for passive elections
- (maybe) changes to the active_difficulty algorithm

PR #2441, opened 2019-12-16
WIP: Ensure the winner of an election is always in confirmation or confirmed

guilhermelawless:
Currently, a second election may be started for a confirming block, which is not intended but also not critical. This adds two checks to be sure of the block confirmation state of a block originating from an election (most cases by far after bootstrapping):
- The block is not being committed to the ledger ( `writing` in confirmation height processor )
- The election has finished but the block not yet queued to the confirmation height processor (seems to be usually due to a rollback, otherwise would not have to be done in the background - @SergiySW  ?)

Also cleans up active_transactions tests and, unrelated, RPC pending loop non-functional change, only because this issue was discovered through that RPC.

An intermittent failure in confirmation_height.pending_observer_callbacks is fixed by waiting for cementing.

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -9,18 +9,18 @@ using namespace std::chrono_literals;
 
 TEST (active_transactions, confirm_one)
 {
-	nano::system system;
-	nano::node_config node_config (24000, system.logging);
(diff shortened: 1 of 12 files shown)


guilhermelawless:
Updated changes at https://github.com/nanocurrency/nano-node/pull/2625

PR #2444, opened 2019-12-17
Indeterminate vote status and enhanced websocket vote sub

### LINKED ISSUE #2421: replay and valid vote counts do not align with expectations
**Description of bug:**
While analyzing votes from the votes websocket I have noticed that votes from the same rep for the same block hash are often duplicated multiple times.  This is expected from the fanout, however it appears as though the vote_valid and vote_replay counts are not exactly accurate from stats->counters.  This also leads to the websocket producing extra votes when it should be skipping replay votes.

After running the V20.0 node for about 6 hours it has confirmed 13,815 blocks through active quorum.  It is reporting a total of 16,575,829 votes (confirm_ack in) of which 14,365,862 are "vote_valid" and 2,209,967 are "vote_replay".

With 92 online representatives I would expect a max of 1,270,980 valid votes for 13,815 blocks and the remaining 15,304,849 as replay votes or votes for inactive blocks.  

In addition to the counters being off the websocket performance is poor at high load.  It would be nice to be able to filter out the actual replay or inactive block votes from the real votes that are used for active transactions.  I imagine it would be difficult to differentiate between old blocks that have already been confirmed and cases where the block has not arrived yet and the vote is stored in the vote cache.  Ideally the vote cache examples would be counted as valid once the block arrived but those that are duplicate or late votes after a block has been confirmed are counted as inactive.

guilhermelawless:
Important note: if a vote arrives before the corresponding block, it's added to inactive_votes_cache and the vote code is `indeterminate`. When the vote is seen again it will be a `replay` if the election has since started. That means these votes are never broadcasted as unique votes. This is important to leave clear in the documentation for anyone using this websocket subscription.

------

Fixes #2421 by not broadcasting votes with unknown status to the websocket unless specified - which was the initial intention.

When elections are over, since we clear the votes immediately, we cannot determine if an arriving vote is a replay or not.

Late votes are now no longer marked as non-replays incorrectly; however, the indeterminate votes still follow the same path as unique votes, through the gap cache and representative crawler. This behavior can be changed but is left untouched in this PR.

Part of #1901 , now the vote messages include a type `{vote, replay, indeterminate}` and the vote websocket can optionally include replays and indeterminate votes. One related breaking change - if the resulting filter of accounts is empty or only has invalid accounts, the result is no account filter (current behavior is everything is filtered). This websocket is marked as [debug](https://docs.nano.org/integration-guides/advanced/#websocket-support).

### DIFF
--- a/nano/node/vote_processor.cpp
+++ b/nano/node/vote_processor.cpp
@@ -202,8 +202,7 @@ nano::vote_code nano::vote_processor::vote_blocking (nano::transaction const & t
 	auto result (nano::vote_code::invalid);
 	if (validated || !vote_a->validate ())
 	{
-		auto is_replay (active.vote (vote_a, true));
-		result = boost::logic::indeterminate (is_replay) ? nano::vote_code::indeterminate : is_replay ? nano::vote_code::replay : nano::vote_code::vote;
+		result = active.vote (vote_a, true);
 		observers.vote.notify (vote_a, channel_a, result);
 		// This tries to assist rep nodes that have lost track of their highest sequence number by replaying our highest known vote back to them
 		// Only do this if the sequence number is significantly different to account for network reordering
--- a/nano/node/vote_processor.cpp
+++ b/nano/node/vote_processor.cpp
@@ -202,29 +202,17 @@ nano::vote_code nano::vote_processor::vote_blocking (nano::transaction const & t
 	auto result (nano::vote_code::invalid);
 	if (validated || !vote_a->validate ())
 	{
+		auto is_replay (active.vote (vote_a, true));
+		result = boost::logic::indeterminate (is_replay) ? nano::vote_code::indeterminate : is_replay ? nano::vote_code::replay : nano::vote_code::vote;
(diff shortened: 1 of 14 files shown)


nano/node/vote_processor.cpp
```diff
@@ -202,29 +202,17 @@ nano::vote_code nano::vote_processor::vote_blocking (nano::transaction const & t
 	auto result (nano::vote_code::invalid);
 	if (validated || !vote_a->validate ())
 	{
+		auto is_replay (active.vote (vote_a, true));
+		result = boost::logic::indeterminate (is_replay) ? nano::vote_code::indeterminate : is_replay ? nano::vote_code::replay : nano::vote_code::vote;
```
cryptocode: Could `vote` return `nano::vote_code` directly?

guilhermelawless: Definitely, thanks for noticing. Made `active::vote` much more evident in what the return is; tribool is useful but not in this case where we want `indetermined || false` to be false.

Issue #2225, opened 2019-08-17
Node telemetry


zhyatt:
Provide methods for nodes to send various metrics between each other for better monitoring of network status and upgrades progress. Some metrics being considered:

- Block count
- Cemented block count
- Set bandwidth cap
- Protocol version number
- Node vendor version
- Peer counts

Are there any other metrics to consider for inclusion?

wezrule:
Perhaps number of accounts

schenkty:
@wezrule I assume you are referring to number of accounts in the node wallet? If so, I think it is a very slippery slope to include any wallet data. It may be too precautionary but anything along the lines of sending wallet data to peers could have a mistake at some point where private wallet data is leaked to a peer which concerns me.

wezrule:
I am referring to the number of the accounts in the ledger

schenkty:
@wezrule oh thank you. I was worried 😂

arranHarty123:
Would bootstrap settings be useful? e.g. my node is not bootstrapping anyone, but will establish x bootstrap connections to catch up.

zhyatt:
Related to #1308

bryanwag:
It is crucial to have a reliable and objective way to measure node performance because the scalability of the network depends heavily on node hardware, yet we currently have to rely on rep's self-report and stress tests to even get a basic idea on that metric. This method is neither quantitative nor reliable. Thus, it's easy to game and hard to educate users on how to choose a rep that is beneficial to the network. 

If we allow each peer to report vote latency of other peers (time(Rep A vote arrives) - time(quorum reached)) and aggregate this result, with appropriate statistical modeling we can quantitatively and reliably measure the performance of reps.

I'm not a dev and not sure whether vote latency can/should be reported by peers within the node software. If this is the wrong place please let me know. Thanks!

zhyatt:
Completed by #2446 and related tickets. Above suggestions captured in new issues: #2657 and #2658

PR #2446, opened 2019-12-20
Node telemetry

wezrule:
There are 10 pieces of telemetry data which can now be accessible, 6 of which were originally mentioned in the description of https://github.com/nanocurrency/nano-node/issues/2225, an additional 4 have been added (sizes noted as well):

1. block count (uint64)
2. cemented count (uint64)
3. node vendor version (uint8)
4. protocol version (uint8)
5. peer count (uint32)
6. bandwidth cap (uint64)
7. account count (uint64)
8. unchecked count (uint64)
9. uptime (from @argakiig) (uint64)
10. genesis block hash (uint256)

**Total: 86 bytes**

Started having a separate message for each of these, but it became quite messy and I think unnecessary because we can fit these easily (as well as a lot more) into the 508 UDP payload without fragmenting so I saw no reason not to pack them all into a single message named `telemetry_ack`.

2 modes of operation
1 - Random selection of peers
The way it works is by sending a `telemetry_req` to square root random peers and waiting for responses. Only peers we sent `telemetry_req` messages to are accepted, others are discarded. If a response (telemetry_ack) is not seen within 2 seconds, it is dropped from the "wait" list.
If we send another telemetry_req to the same peer, and happen to receive an out of date telemetry_ack from a previous request it is ignored. There is a sequence number associated with each request (locally on the client side only).

There's also a way to "consolidate data" which works as follows:
- Takes the **mean** of the block count, cemented count, peer count, account count, unchecked count and uptime.
- For vendor version, protocol version and genesis_block, the **mode** is returned, and if there is no metric with more than 1 result in common (unlikely), it returns the first one.
- For bandwidth cap, the **mode** is taken otherwise an **average** is used, but **0** is not taken into account as it has a special meaning of unlimited.

2 - Single request
Similar to above except only a single peer is requested data (no consolidation is necessary)

There are async & blocking variants.

Works over both udp and tcp real-time connection.

Protocol number has been increased.

Many tests added.

Added rpc "node_telemetry", which has a optional boolean argument "raw", which displays as an array all metrics from all nodes, instead of just a consolidated view. Can also specify address/port to request metrics from a specific peer.

Added TOML config option for logging telemetry requests/receives.

**Of note for reviewer:** When getting a **random_set** of channels, I had issues when the temporary channel was being returned, I couldn't see a usecase where we would want this, so I am excluding it from the list of channels to return. However this is done for all callers, I can limit the scope for the new changes if desired.

TSAN/ASAN was run over the new tests.


nano/node/network.cpp
```diff
@@ -508,6 +509,42 @@ class network_message_visitor : public nano::message_visitor
...
+		telemetry_data.protocol_version_number = node.network_params.protocol.protocol_version;
+		telemetry_data.vendor_version = nano::get_major_node_version ();
+		telemetry_data.uptime = std::chrono::duration_cast<std::chrono::seconds> (std::chrono::steady_clock::now () - node.startup_time).count ();
+		telemetry_data.unchecked_count = node.ledger.cache.unchecked_count;
+		telemetry_data.genesis_block = nano::genesis ().hash ();
+
+		{
+			auto transaction = node.store.tx_begin_read ();
+			telemetry_data.account_count = node.store.account_count (transaction);
+			telemetry_data.peer_count = node.store.peer_count (transaction);
```
SergiySW: May be use network::size () ?

Srayman:
I would suggest median average instead of mean for block count and cemented counts.  If you query a node that is bootstrapping and has a really low block count the mean average will be very skewed.  If still using mean average then might want to consider removing outliers from the average.

wezrule:
> I would suggest median average instead of mean for block count and cemented counts. If you query a node that is bootstrapping and has a really low block count the mean average will be very skewed. If still using mean average then might want to consider removing outliers from the average.

This crossed my mind. I think it has pros/cons especially with a random selection of nodes, because the outlier could actually be the only honest node, so we'd be throwing away the only valid data, but I guess we are treating all data as potentially invalid it would make sense to go with the majority rather than skewing valid cases too, so I will bring it up. Thanks!

wezrule:
Merging this in so that there is a baseline similar to DB4 (and the PR is quite large already). An additional PR will be made with community suggestions

PR #2448, opened 2019-12-20
Erase inactive votes cache after election stop / finish

SergiySW:
... and for forks elections

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -1128,10 +1128,10 @@ nano::gap_information nano::active_transactions::find_inactive_votes_cache (nano
 
 void nano::active_transactions::erase_inactive_votes_cache (nano::block_hash const & hash_a)
 {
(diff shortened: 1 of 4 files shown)


nano/node/active_transactions.cpp
```diff
@@ -1126,6 +1126,15 @@ nano::gap_information nano::active_transactions::find_inactive_votes_cache (nano
 	}
 }
 
+void nano::active_transactions::erase_inactive_votes_cache (nano::block_hash const & hash_a)
+{
+	auto existing (inactive_votes_cache.get<1> ().find (hash_a));
+	if (existing != inactive_votes_cache.get<1> ().end ())
+	{
+		inactive_votes_cache.get<1> ().erase (existing);
+	}
```
guilhermelawless: Could use `nano::gap_cache::tag_hash` here, similar to other uses of inactive_votes_cache

SergiySW: Updated

PR #2452, opened 2019-12-29
Confirmation solicitor

clemahieu:


nano/node/confirmation_solicitor.cpp
```diff
@@ -0,0 +1,55 @@
...
+		}
+		else if (single_count++ < max_confirm_req)
+		{
+			node.network.broadcast_confirm_req (i->election->status.winner);
+			++i;
+		}
+	}
+	requests.clear ();
+	rebroadcasted = 0;
+	representatives = node.rep_crawler.representatives ();
```
guilhermelawless: Would this fit better in a `prepare ()` method called right before we start adding new requests?

nano/node/active_transactions.cpp
```diff
@@ -363,74 +274,17 @@ void nano::active_transactions::request_confirm (nano::unique_lock<std::mutex> &
...
-					increment_counter_l = false;
-				}
-			}
-			if (increment_counter_l || node.network_params.network.is_test_network ())
-			{
-				++election_l->confirmation_request_count;
-			}
 		}
 	}
 	ongoing_broadcasts = !blocks_bundle_l.empty () + !batched_confirm_req_bundle_l.empty () + !single_confirm_req_bundle_l.empty ();
```
guilhermelawless: This is not getting decremented anymore. With the change to start broadcasting / sending to channels within the solicitor, it seems this might not be necessary. The request loop waits on these here: https://github.com/nanocurrency/nano-node/blob/42e5a27545e73606993dcfc520ea8b69d6e75e0f/nano/node/active_transactions.cpp#L324-L327

guilhermelawless: Any concern if requests pile up in slow nodes, with more getting added before the previous ones go through?

nano/node/active_transactions.cpp
```diff
@@ -344,12 +246,17 @@ void nano::active_transactions::request_confirm (nano::unique_lock<std::mutex> &
 	 * Elections extending the soft config.active_elections_size limit are flushed after a certain time-to-live cutoff
 	 * Flushed elections are later re-activated via frontier confirmation
 	 */
+	solicitor.prepare ();
```
cryptocode: Your concern about missed calls was mentioned in a comment. One thing I've found useful in these types of situations is to have a `state` enum field in the class so future refactorings don't break the logic. So #add and #flush would assert `state == prepared` and #prepare would assert `state == flushed` (the initial state), or something to that effect. Just an idea (used in `nano::timer`)

PR #2456, opened 2020-01-01
Reduce memory used by callback data in confirmation height processor

wezrule:
#2233 made the confirmation height processor use a lot more memory as it stored the whole block & sideband for each block it traversed. This has been changed to only store the block hash. An additional `store.block_get ()` is then required later to retrieve the contents but this didn't have any measurable performance hit from benchmarking and gave considerable memory reduction gains, as shown below:
image

It shows for the `confirmation_height.long_chains` test in the `slow_tests` a reduction from 82MB to 12MB (86% less memory) for the affected containers. The memory is still unbounded (as well as for `receive_source_pairs`) but it has been significantly reduced and a proper bounded solution is scheduled for a future release.

I increased the number of blocks used in the slow_test.


wezrule:
No longer needed thanks to https://github.com/nanocurrency/nano-node/pull/2531

PR #2459, opened 2020-01-03
Publish cleanup

clemahieu:
Performing fork republishing after block processing and removing validate_block_by_previous which was performing validity checking that's already performed in the ledger block processor.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -3032,12 +3032,17 @@ TEST (node, fork_election_invalid_block_signature)
 			election = existing->second;
 		}
 	}
+	nano::unique_lock<std::mutex> lock (node1.active.mutex);
 	ASSERT_EQ (1, election->blocks.size ());
+	lock.unlock ();
 	node1.network.process_message (nano::publish (send3), channel1);
 	node1.network.process_message (nano::publish (send2), channel1);
+	lock.lock ();
 	while (election->blocks.size () == 1)
 	{
+		lock.unlock ();
 		ASSERT_NO_ERROR (system.poll ());
+		lock.lock ();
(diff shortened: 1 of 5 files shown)


nano/core_test/node.cpp
```diff
@@ -2724,41 +2724,60 @@ TEST (node, fork_invalid_block_signature)
...
 		ASSERT_NO_ERROR (system.poll ());
+		nano::lock_guard<std::mutex> lock (system.nodes [0]->active.mutex);
+		auto existing = system.nodes [0]->active.blocks.find (send1->hash ());
+		if (existing != system.nodes [0]->active.blocks.end ())
+		{
+			election = existing->second;
+		}
 	}
-	while (!system.nodes[0]->block (send2->hash ()))
+	ASSERT_EQ (1, election->blocks.size ());
```
SergiySW: Do we need active mutex lock to prevent TSAN issues here and later for election access?

clemahieu: Fixed.

nano/node/election.cpp
```diff
@@ -211,21 +211,16 @@ bool nano::election::publish (std::shared_ptr<nano::block> block_a)
...
-				blocks.emplace (block_a->hash (), block_a);
-				insert_inactive_votes_cache (block_a->hash ());
-				confirm_if_quorum ();
-				node.network.flood_block (block_a, false);
-			}
-			else
-			{
-				result = true;
-			}
+			blocks.emplace (std::make_pair (block_a->hash (), block_a));
```
wezrule: This didn't create a pair before, was there a reason for needing to do this? Not a big issue, so I'll approve regardless

PR #2460, opened 2020-01-03
WIP: Consistent election difficulty updates

guilhermelawless:
Ensures difficulty is updated correctly for old blocks and forks, as well as work watcher propagation of updated blocks. It's important to ensure these cases are handled correctly, for which tests have been added:
- New forks
- Updates for forks
- Updates for ledger blocks
- Work watcher block propagation

Summary of the changes:
- Extract active_transactions::restart for dropped elections
- Erase from dropped cache after restarting
- Split into update_difficulty and update_difficulty_impl
- Run update_difficulty for forks by running update_difficulty_impl in active_transactions::publish
- Check node::block_confirmed_or_being_confirmed in active_transactions::restart rather than confirmation height only
- Simplify work watcher and allow stopping even if hashes don't match (as long as the root does). #2518 removed flooding from the work watcher but 0e52159 shows it's necessary or else the block doesn't propagate to other nodes. Furthermore, difficulty updates aren't republished, so #2549 will change this to be a flood to all PRs

Best reviewed ignoring whitespaces.


nano/node/active_transactions.cpp
```diff
@@ -869,7 +872,7 @@ size_t nano::active_transactions::size ()
 
 bool nano::active_transactions::publish (std::shared_ptr<nano::block> block_a)
 {
-	nano::lock_guard<std::mutex> lock (mutex);
+	nano::unique_lock<std::mutex> lock (mutex);
```
SergiySW: Can remain as `lock_guard`, right?

nano/node/wallet.cpp
```diff
@@ -1471,11 +1471,13 @@ void nano::work_watcher::watching (nano::qualified_root const & root_a, std::sha
 											current->second = block;
 										}
 									}
-									watcher_l->node.network.flood_block (block, false);
-									assert (!watcher_l->node.active.update_difficulty (block));
-									watcher_l->update (root_a, block);
+									if (!watcher_l->node.active.update_difficulty (block))
```
clemahieu: Does this need a test since it's a new control flow?

guilhermelawless: Hard to test this since the block would have to be confirmed after a successful work generation but before the difficulty update. But I can perhaps change tests to ensure blocks are no longer under watch once confirmed

guilhermelawless: Added tests to ensure blocks are removed when winning an election, but also on losing (this was not the case before, changed here).

Simplified this method and toned down the number of indentation levels.

guilhermelawless:
#2518 removed flooding from the work watcher but 0e52159 shows it's necessary or else the block doesn't propagate to other nodes. Furthermore, difficulty updates aren't republished, so #2549 will change this to be a flood to all PRs

PR #2464, opened 2020-01-09
Don't bind to UDP socket with --disable_udp flag

SergiySW:
* don't bind to UDP socket with --disable_udp flag
* don't start port mapping with both UDP & TCP sockets disabled
* use predefined port in `nano::network::endpoint ()` function rather than call to UDP socket
* update TCP (bootstrap) server `endpoint ()` function to return empty endpoint if server is not started
* update several tests to run twice: with UDP only and with TCP only
* attepmt to start realtime TCP connection to RPC bootstrap peer if UDP is disabled


nano/node/transport/udp.hpp
```diff
@@ -172,7 +172,7 @@ namespace transport
 		attempts;
 		// clang-format on
 		boost::asio::strand<boost::asio::io_context::executor_type> strand;
-		boost::asio::ip::udp::socket socket;
+		std::shared_ptr<boost::asio::ip::udp::socket> socket;
```
wezrule: This could probably be a `unique_ptr` as I don't think anything else shares ownership of it (unless I missed it)

SergiySW: Updated

PR #2468, opened 2020-01-09
Publish prototype

clemahieu:
This is a prototype of changing how votes are relayed with respect to principal representatives. In order to reduce network traffic on PRs, they will no longer relay votes, non-PR nodes will no longer relay votes to PRs, and PRs will announce their votes to all other PRs.


nano/node/network.cpp
```diff
@@ -236,6 +236,28 @@ void nano::network::flood_message (nano::message const & message_a, bool const i
...
+			(*i)->send (nano::confirm_ack (vote_a));
+		}
+	}
+}
+
+void nano::network::flood_vote_pr (std::shared_ptr<nano::vote> vote_a)
+{
+	nano::confirm_ack message (vote_a);
+	auto list = node.rep_crawler.principal_representatives ();
+	for (auto const & i: list)
```
SergiySW: I guess for consistency usually we have extra space between `i` and  `:`

nano/node/wallet.hpp
```diff
@@ -185,6 +185,10 @@ class wallet_representative_counts
 public:
 	uint64_t voting{ 0 }; // Representatives with at least the configured minimum voting weight
 	uint64_t half_principal{ 0 }; // Representatives with at least 50% of principal representative requirements
+	bool have_half_rep () const
```
guilhermelawless: This code seems to not be used

nano/node/network.cpp
```diff
@@ -236,6 +236,28 @@ void nano::network::flood_message (nano::message const & message_a, bool const i
 	}
 }
 
+void nano::network::flood_vote (std::shared_ptr<nano::vote> vote_a, float scale)
+{
+	auto list (list_fanout (scale));
+	for (auto i (list.begin ()), n (list.end ()); i != n; ++i)
+	{
+		if (!node.rep_crawler.is_pr (**i))
```
guilhermelawless: Do we risk not meeting the desired fanout because half of the list could be PRs? Not a problem when nodes >>PRs but for now?

nano/node/network.cpp
```diff
@@ -678,9 +700,9 @@ std::deque<std::shared_ptr<nano::transport::channel>> nano::network::list (size_
 }
 
 // Simulating with sqrt_broadcast_simulate shows we only need to broadcast to sqrt(total_peers) random peers in order to successfully publish to everyone with high probability
-std::deque<std::shared_ptr<nano::transport::channel>> nano::network::list_fanout ()
+std::deque<std::shared_ptr<nano::transport::channel>> nano::network::list_fanout (float scale)
 {
-	auto result (list (size_sqrt ()));
+	auto result (list (static_cast<size_t> (std::ceil (scale * size_sqrt ()))));
```
guilhermelawless: `size_sqrt()` returns the ceiling already, maybe it should return a float so that only one rounding up is done?

nano/node/wallet.hpp
cryptocode: It was used when I looked through earlier, got dropped in the conflict resolution merge it seems.

nano/node/network.cpp
clemahieu: Yes, we can do this in list. I don’t usually like to pass booleans to inject control in the target function. We could have a similar function to list that is list_non_pr which, after shuffle, iterates from the beginning removing prs until it reaches the end of list or count and then resizes the deque

nano/node/network.cpp
clemahieu: Yes

nano/node/network.cpp
clemahieu: I think I ran in to an issue while testing when the peer count was 1, possibly because sqrt(1.0) over floats yields close, but less than one which would round to zero. Returning a float would work but then all call sites need to cast. If it’s easy though we can do that.

nano/node/active_transactions.cpp
```diff
@@ -582,7 +582,7 @@ nano::vote_code nano::active_transactions::vote (std::shared_ptr<nano::vote> vot
 	{
 		if (processed)
```
clemahieu: This is supposed to include && !is_half_pr()

nano/node/network.cpp
guilhermelawless: Added, specifically in https://github.com/nanocurrency/nano-node/pull/2468/commits/60b2fe2b2a3adb7e7d5de0d0cdf046d2d4c398e4 . `fanout()` was extracted to avoid having both `list_fanout ()` and `list_non_pr_fanout ()`. The number of caller sites was sufficiently small that it seemed like a good idea, please take a look.

You mentioned a more complicated loop that would include PRs if the working list fell below count, is that still the intention? As it stands now it can't include PRs even if the count falls below the desired.

nano/node/network.cpp
guilhermelawless: There were 2 call sites, one was modified to call `fanout()` directly (see above). Extended test cases in `peer_container.list_fanout` to ensure there are no rounding errors for 1, 2, 1000.

nano/node/active_transactions.cpp
guilhermelawless: Assuming you meant `!node.wallets.rep_counts ().have_half_rep ()` this is now handled

nano/node/network.cpp
```diff
@@ -214,10 +214,27 @@ bool nano::network::send_votes_cache (std::shared_ptr<nano::transport::channel>
...
+void nano::network::flood_vote (std::shared_ptr<nano::vote> vote_a, float scale)
+{
+	nano::confirm_ack message (vote_a);
+	for (auto & i : list_non_pr (fanout (scale)))
+	{
+		i->send (message, nullptr);
+	}
+}
+
+void nano::network::flood_vote_pr (std::shared_ptr<nano::vote> vote_a)
```
wezrule: can use `const &` with this and `flood_vote` above

nano/node/network.cpp
```diff
@@ -264,7 +281,7 @@ void nano::network::broadcast_confirm_req (std::shared_ptr<nano::block> block_a)
 	if (list->empty () || node.rep_crawler.total_weight () < node.config.online_weight_minimum.number ())
 	{
 		// broadcast request to all peers (with max limit 2 * sqrt (peers count))
-		auto peers (node.network.list (std::min (static_cast<size_t> (100), 2 * node.network.size_sqrt ())));
+		auto peers (node.network.list (std::min<size_t> (100, node.network.fanout (2.F))));
```
wezrule: You use lowercase `F` in other places, should be consistent. Can also just remove it as int -> float should be allowed implicitly.

clemahieu:
This is commit is based on a contribution from @Srayman on a recommendation to reduce traffic to principle representatives. The substantive changes are:
- Principle representatives do not relay votes
- Principle representatives are not the target of vote relaying
- Each principle representative broadcasts its votes to all other principle representative and additional non-principle representatives.

An analysis of this methodology as well as other suggestions were posted https://medium.com/nanocurrency/proposal-for-nano-node-network-optimizations-21003e79cdba . 

Testing on beta shows upwards of a 50% reduction in traffic to principle representatives.

PR #2469, opened 2020-01-09
WIP: Conditional work validation

guilhermelawless:
Splits work validation into `work_validate_entry` - before ledger processing -, and `work_validate` - which is done during or after ledger processing, when details such as epoch version and block subtype can be retrieved

Some changes were made leading to less work validation overall:
- Took the validation step out of the node blockprocessor entry, as this is already done when deserializing from the network. Equivalent checks were added wherever a bootstrap process wants to add blocks to the processor
- Added work version and difficulty to ledger process_return, so as to pass the result to the corresponding next step: process_live, update_difficulty and process_fork for progress, old and fork respectively

Work validation is now done during ledger processing as it may require block details in the future. It only adds about 3 lines for each block processing method, so I recommend reviewing while ignoring whitespace changes.

work_version is defined within lib/work.hpp, and the method to transform (in the future) block details to work_version is defined within secure/work.hpp. For now, it simply returns work_1 as the only version available. When outside of ledger processing, this can be queried via ledger::work_version.

Example:
```cpp
nano::ledger::work_version (sideband.block_details);
nano::work_version_get (epoch, is_send, is_receive, is_epoch);
```

Added minimal non-breaking changes to RPCs work_validate and work_generate so they can accept a work_version field. For websockets, the only change was to output the work version via the work websocket.

Work validation has a default version to avoid changing every work_validate call in tests, but it asserts that it's a test network.

Work generation has an implicit default work version to avoid touching work generation as much as possible.


guilhermelawless:
This has been replaced by a few PRs, and more are coming as preparation for work difficulty updates based on each block subtype: https://medium.com/nanocurrency/development-update-v21-pow-difficulty-increases-362b5d052c8e

PR #2472, opened 2020-01-10
Confirmation solicitor revamp

guilhermelawless:
Being a revamp this is best reviewed in split mode. The changes are:

- Fix for getting stuck in the loop
- Requests ordering is now respected to ensure prioritization
- Batch count was global and not per-channel
- **confirmation_request_count no longer used to alternate between elections, instead now doing it time-based**. Using a counter + modulo was not great.
- Deleted all uses of single confirm_req since there is more than enough weight on TCP now. This can however be reverted.
- Added a flag to disable active_transactions to allow for testing the solicitor
- Added a basic test for the solicitor

Eventually, it seemed more efficient to get rid of the `request` class and do most of the logic in `::add()` and have `::flush()` only batch and dispatch the messages. Overall I think the solicitor has still simplified a lot of the logic, but I could not escape making it more complicated while still making sure the behavior is the same as before.

Rate-limiting logic is left in `active_transactions`

### DIFF
--- a/nano/node/confirmation_solicitor.cpp
+++ b/nano/node/confirmation_solicitor.cpp
@@ -1,62 +1,90 @@
 #include <nano/node/confirmation_solicitor.hpp>
-
 #include <nano/node/election.hpp>
 #include <nano/node/node.hpp>
 
 nano::confirmation_solicitor::confirmation_solicitor (nano::node & node_a) :
-node (node_a)
+node (node_a),
+max_confirm_req_batches (node_a.network_params.network.is_test_network () ? 1 : 20),
+max_block_broadcasts (node_a.network_params.network.is_test_network () ? 4 : 30),
+soliciting_alternating_factor (node_a.network_params.network.is_test_network () ? 2 : 4),
+block_flooding_alternating_factor (8)
 {
 }
 
-void nano::confirmation_solicitor::prepare ()
+void nano::confirmation_solicitor::prepare (std::vector<nano::representative> const & representatives_a)
 {
 	assert (!prepared);
 	requests.clear ();
 	rebroadcasted = 0;
-	representatives = node.rep_crawler.representatives ();
+	// Only representatives ready to receive batched confirm_req
+	representatives = representatives_a;
 	prepared = true;
 }
 
-void nano::confirmation_solicitor::add (std::shared_ptr<nano::election> election_a)
+bool nano::confirmation_solicitor::add (std::shared_ptr<nano::election> election_a)
 {
 	assert (prepared);
-	if (election_a->confirmation_request_count % 8 == 1 && rebroadcasted++ < max_block_broadcasts)
+	if (((election_a->confirmation_request_count > block_flooding_alternating_factor && election_a->confirmation_request_count % block_flooding_alternating_factor == 1) || node.network_params.network.is_test_network ()) && rebroadcasted++ < max_block_broadcasts)
 	{
 		node.network.flood_block (election_a->status.winner);
 	}
-	for (auto const & rep : representatives)
+	static auto max_channel_requests (max_confirm_req_batches * nano::network::confirm_req_hashes_max);
+	bool any_bundle{ false };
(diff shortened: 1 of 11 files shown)


nano/node/confirmation_solicitor.cpp
```diff
@@ -1,62 +1,94 @@
...
 void nano::confirmation_solicitor::flush ()
 {
 	assert (prepared);
-	size_t batch_count = 0;
-	size_t single_count = 0;
-	for (auto i = requests.begin (), n (requests.end ()); i != n;)
+	for (auto const & request_queue : requests)
 	{
-		if (batch_count++ < max_confirm_req_batches && i->channel->get_network_version () >= node.network_params.protocol.tcp_realtime_protocol_version_min)
+		auto channel (request_queue.first);
```
wezrule: This could be `const &` to save increment/decrementing the reference count

nano/node/election.hpp
```diff
@@ -79,6 +79,8 @@ class election final : public std::enable_shared_from_this<nano::election>
 	bool stopped;
 	std::unordered_map<nano::block_hash, nano::uint128_t> last_tally;
 	unsigned confirmation_request_count{ 0 };
+	std::chrono::steady_clock::time_point last_broadcast {};
+	std::chrono::steady_clock::time_point last_request {};
```
wezrule: Does value initializing `{}` these have a benefit?

nano/node/confirmation_solicitor.cpp
```diff
@@ -1,62 +1,94 @@
...
 {
 }
 
-void nano::confirmation_solicitor::prepare ()
+void nano::confirmation_solicitor::prepare (std::vector<nano::representative> const & representatives_a)
 {
 	assert (!prepared);
 	requests.clear ();
 	rebroadcasted = 0;
-	representatives = node.rep_crawler.representatives ();
```
wezrule: It looks like this class only needs `network` now (`network_params` could be made a member), so could remove the  `node` dependency on this class entirely. I guess not essential (only if bored) there's a generic ticket somewhere for me to do this :D

nano/node/election.hpp
guilhermelawless: As you noted, default constructor does the same (setting to zero)!

nano/node/confirmation_solicitor.cpp
guilhermelawless: Thanks, very good point and I should try addressing this more as I go (might try to for distributed_work when I head in there again).  `network_constants` are retrieved in the constructor itself and added a `network` reference as a member.

nano/node/nodeconfig.hpp
```diff
@@ -110,6 +110,7 @@ class node_flags final
 {
 public:
 	std::vector<std::string> config_overrides;
+	bool disable_active_transactions{ false };
```
SergiySW: Isn't this more like disable[_active_transaction]_request_loop? Not disabling elections completely

guilhermelawless: Good point, changed.

guilhermelawless:
Last commits bring the rate-limiting logic back to active_transactions and elections are now given as const ref to the solicitor.

nano/node/confirmation_solicitor.cpp
```diff
@@ -1,61 +1,77 @@
...
 #include <nano/node/election.hpp>
-#include <nano/node/node.hpp>
 
-nano::confirmation_solicitor::confirmation_solicitor (nano::node & node_a) :
-node (node_a)
+using namespace std::chrono_literals;
+
+nano::confirmation_solicitor::confirmation_solicitor (nano::network & network_a, nano::network_constants const & params_a) :
+max_confirm_req_batches (params_a.is_test_network () ? 1 : 20),
+max_block_broadcasts (params_a.is_test_network () ? 4 : 30),
```
wezrule: Could remove `network_constants` dependency and pass in `bool is_test_network` instead, but I don't think it's that important. Haven't tested it, but otherwise LGTM

PR #2474, opened 2020-01-11
Use a strand in distributed_work

guilhermelawless:
Wrap all distributed_work socket handling in a strand. Could use one strand for each peer [I think], but with the very low amount of messages (2 sent and 1 received at max) this seems unnecessary

Other fixes and enhancements:

- Fix a rare issue where generation would finish before `start_work()` is done (by moving the local generation code to the end of the function)
- TSAN race fix in rpc.work_peer_many
- Wrap request parameters in a work_request class
- No logging when ongoing operations are canceled
- Make `backoff` a chrono object
- Move the work_peer_request class into distributed_work and rename it

### DIFF
--- a/nano/node/distributed_work.cpp
+++ b/nano/node/distributed_work.cpp
@@ -291,11 +291,11 @@ void nano::distributed_work::stop_once (bool const local_stop_a)
 		{
 			if (auto connection_l = connection_w.lock ())
 			{
-				if (connection_l->socket.is_open ())
-				{
-					auto this_l (shared_from_this ());
-					boost::asio::post (strand, boost::asio::bind_executor (strand, [this_l, connection_l] {
-						boost::system::error_code ec;
+				auto this_l (shared_from_this ());
+				boost::asio::post (strand, boost::asio::bind_executor (strand, [this_l, connection_l] {
+					boost::system::error_code ec;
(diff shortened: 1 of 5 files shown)


nano/node/distributed_work.cpp
```diff
@@ -278,25 +285,32 @@ void nano::distributed_work::stop_once (bool const local_stop_a)
...
 			if (auto connection_l = connection_w.lock ())
 			{
-				boost::system::error_code ec;
-				connection_l->socket.cancel (ec);
-				if (ec)
-				{
-					node.logger.try_log (boost::str (boost::format ("Error cancelling operation with work_peer %1% %2%: %3%") % connection_l->address % connection_l->port % ec.message () % ec.value ()));
-				}
-				try
+				if (connection_l->socket.is_open ())
```
cryptocode: I think the `boost::asio::post (strand ...` below should enclose the whole thing up to there (if I'm reading the diff right), as no socket operations are thread safe.

guilhermelawless: Oh right, thanks for spotting that. Fixed!

cryptocode: I think even `is_open` will have to be included.

PR #2475, opened 2020-01-11
Remove unnecessary lock requirement in active_transactions::lock.

clemahieu:
This also removes some code-smell around passing in a bool to determine if the lock should actually be acquired.

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -637,11 +637,13 @@ TEST (active_transactions, restart_dropped)
 		}
 		ASSERT_NO_ERROR (system.poll ());
 	}
-	// Verify the block was updated in the ledger
+	std::shared_ptr<nano::block> block;
+	while (block == nullptr)
 	{
-		auto block (node.store.block_get (node.store.tx_begin_write (), send1->hash ()));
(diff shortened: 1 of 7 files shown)


nano/core_test/active_transactions.cpp
```diff
@@ -662,7 +662,7 @@ TEST (active_transactions, restart_dropped)
 	}
 	// Verify the block was updated in the ledger
 	{
-		auto block (node.store.block_get (node.store.tx_begin_read (), send1->hash ()));
+		auto block (node.store.block_get (node.store.tx_begin_write (), send1->hash ()));
```
wezrule: Was changing this to a write transaction required?

clemahieu: I experienced a race condition some times where the transaction for the block write had not completed before this code was reached.  This was attempting to block until the write had completed.

wezrule: ok, I wonder if it should have a comment then to prevent someone unknowingly changing it in the future, or perhaps call `node.store.block_exists` in a loop with a suitable timeout

clemahieu: Good point, that's a better idiom.

PR #2481, opened 2020-01-14
Upgrade confirmation height table to include cemented frontier

wezrule:
This is a preparation step for a confirmation height redesign I am doing. I'm modifying the confirmation height table to also include the hash at that height (cemented frontier). Currently the confirmation height processor starts from a hash in a chain and works downwards until it hits the first cemented block, because there's no other way to reach it. Then it starts traversing the bottom and cementing those blocks. This works fine when storing the receive blocks as we iterate, but this can lead to a lot of memory being used if long chains are traversed and it doesn't scale. It also doesn't play nicely with a bounded memory algorithm, because we would be wasting a lot of read IO traversing accounts many times from the top. I want to start from the bottom (cemented frontier) and work upwards, this limits waste and also cements the bottom most blocks quicker. To do this I need to know what the block hash is at that height though, which is what I am now storing.

Updated some conf heights tests which used multiple nodes, to just use 1 which makes debugging simpler.

Upgrade, this was from a v14 ledger (**Edit: Outdated as we now use v17**) (I left the later portion of the logs from v14 -> v15 at the top):
```
[2020-Jan-15 11:27:57.090806]: Database epoch merge upgrade 22 million state blocks upgraded
[2020-Jan-15 11:28:15.294133]: Epoch merge upgrade: Finished state blocks, now doing pending blocks
[2020-Jan-15 11:29:20.603369]: Finished epoch merge upgrade. Preparing vacuum...
[2020-Jan-15 11:29:21.295422]: Preparing v15 to v16 upgrade...
[2020-Jan-15 11:30:13.381820]: Confirmation height frontier set for 200k accounts
[2020-Jan-15 11:31:02.586240]: Confirmation height frontier set for 400k accounts
[2020-Jan-15 11:32:20.382162]: Confirmation height frontier set for 600k accounts
[2020-Jan-15 11:32:57.744189]: Confirmation height frontier set for 800k accounts
[2020-Jan-15 11:33:31.488086]: Confirmation height frontier set for 1000k accounts
[2020-Jan-15 11:34:01.524724]: Finished upgrading confirmation height frontiers
[2020-Jan-15 11:36:36.047500]: Vacuum succeeded.
```

### DIFF
--- a/nano/node/lmdb/lmdb.cpp
+++ b/nano/node/lmdb/lmdb.cpp
@@ -578,7 +578,7 @@ void nano::mdb_store::upgrade_v14_to_v15 (nano::write_transaction & transaction_
 		release_assert (rep_block != nullptr);
 		account_infos.emplace_back (account, nano::account_info{ account_info_v14.head, rep_block->representative (), account_info_v14.open_block, account_info_v14.balance, account_info_v14.modified, account_info_v14.block_count, i_account.from_first_database ? nano::epoch::epoch_0 : nano::epoch::epoch_1 });
 		// Move confirmation height from account_info database to its own table
-		confirmation_height_put (transaction_a, account, account_info_v14.confirmation_height);
+		mdb_put (env.tx (transaction_a), confirmation_height, nano::mdb_val (account), nano::mdb_val (account_info_v14.confirmation_height), MDB_APPEND);
 		i_account.from_first_database ? ++account_counters.after_v0 : ++account_counters.after_v1;
 	}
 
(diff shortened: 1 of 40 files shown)


nano/node/lmdb/lmdb.cpp
```diff
@@ -687,6 +687,71 @@ void nano::mdb_store::upgrade_v15_to_v16 (nano::write_transaction & transaction_
 		representation = 0;
 	}
 
+	auto account_info_i = latest_begin (transaction_a);
+	auto account_info_n = latest_end ();
+
+	// Set the confirmed frontier for each account in the confirmation height table
+	std::vector<std::pair<nano::account, nano::confirmation_height_info>> confirmation_height_infos;
+	for (nano::mdb_iterator<nano::account, uint64_t> i (transaction_a, confirmation_height), n (nano::mdb_iterator<nano::account, uint64_t>{}); i != n; ++i, ++account_info_i)
```
cryptocode: As this upgrade takes a few minutes, maybe do start/stop and periodic logging like is done in `upgrade_v14_to_v15`

wezrule: Good point, I have done that and updated the description of what logs I got from an old upgrade (1.1M accounts) from a v14 ledger. I noticed the vacuum log message doesn't carry forward well with future upgrades, so have updated that too. Also updated the test to make sure it hits all control paths as per private discussion.

PR #2482, opened 2020-01-14
Rep crawler cleanup

clemahieu:
Move processing effort on to rep_crawler and off of the vote processor.
- The response function has been converted to a lightweight function that simply queues votes to be checked on the next craw interval.
- Logic around updating reps has been moved out of the callback lambda and into the rep processor itself.
- The rep processor no longer accepts arbitrary values and directly computes weight from a given vote so tests have been updated accordingly.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -2037,7 +2037,7 @@ TEST (node, rep_weight)
 	node.rep_crawler.response (channel1, vote2);
 	while (node.rep_crawler.representative_count () != 2)
 	{
-		system.poll ();
+		ASSERT_NO_ERROR (system.poll ());
 	}
 	// Make sure we get the rep with the most weight first
 	auto reps (node.rep_crawler.representatives (1));
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1614,7 +1614,7 @@ TEST (node, DISABLED_fork_stale)
 	auto & node1 (*system1.nodes[0]);
 	auto & node2 (*system2.nodes[0]);
 	node2.bootstrap_initiator.bootstrap (node1.network.endpoint ());
-	auto channel (std::make_shared<nano::transport::channel_udp> (node2.network.udp_channels, node1.network.endpoint (), node2.network_params.protocol.protocol_version));
+	std::shared_ptr<nano::transport::channel> channel (std::make_shared<nano::transport::channel_udp> (node2.network.udp_channels, node1.network.endpoint (), node2.network_params.protocol.protocol_version));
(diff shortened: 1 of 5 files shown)


nano/core_test/node.cpp
```diff
@@ -1999,62 +2027,131 @@ TEST (node, rep_weight)
...
-	node.rep_crawler.response (channel1, keypair2.pub, amount50);
-	ASSERT_EQ (2, node.rep_crawler.representative_count ());
+	auto vote1 = std::make_shared<nano::vote> (nano::test_genesis_key.pub, nano::test_genesis_key.prv, 0, genesis.open);
+	auto vote2 = std::make_shared<nano::vote> (keypair1.pub, keypair1.prv, 0, genesis.open);
+	node.rep_crawler.add (genesis.open->hash ());
+	node.rep_crawler.response (channel0, vote1);
+	node.rep_crawler.response (channel1, vote2);
+	while (node.rep_crawler.representative_count () != 2)
+	{
+		system.poll ();
```
wezrule: We generally wrap `system.poll` with `ASSERT_NO_ERROR ()` and have a deadline timer. Might also want to consider resetting the timer in the `rep.remove` test before each polling session as TSAN runs can sometimes trip these up as it makes the tests run a lot slower.

nano/node/repcrawler.cpp
```diff
@@ -119,32 +176,17 @@ void nano::rep_crawler::query (std::shared_ptr<nano::transport::channel> channel
...
-	}
-	else
+	nano::lock_guard<std::mutex> lock (active_mutex);
+	for (auto i = vote_a->begin (), n = vote_a->end (); i != n; ++i)
 	{
-		probable_reps.emplace (nano::representative (rep_account_a, weight_a, channel_a));
-		updated_or_inserted = true;
+		if (active.count (*i) != 0)
+		{
+			responses.push_back (std::make_pair (channel_a, vote_a));
```
wezrule: Could use `emplace_back (channel_a, vote_a)` to save creating a temporary pair

nano/node/repcrawler.hpp
```diff
@@ -99,7 +96,7 @@ class rep_crawler
 	 * with high probability that the endpoint is a representative node.
 	 * @return True if the rep entry was updated with new information due to increase in weight.
 	 */
-	bool response (std::shared_ptr<nano::transport::channel> channel_a, nano::account const & rep_account_a, nano::amount const & weight_a);
+	void response (std::shared_ptr<nano::transport::channel> channel_a, std::shared_ptr<nano::vote>);
```
wezrule: The `vote` can be passed by reference (`channel` can do although this is existing code). Saves an inc/decrement of the atomic reference count.

PR #2485, opened 2020-01-15
Request aggregator

guilhermelawless:
Adds a class that runs in a new thread to pool confirmation requests by endpoint, before replying with cached votes and generating new votes as necessary.

This allows a reduction of bandwidth, vote generation, and moves some vote generation out of the I/O threads.

Another advantage is using 12-hash votes rather than only up to the number of individual requestt hashes (max 7)

Includes stats for how many votes were cached, generated, or ignored.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -2424,7 +2424,7 @@ TEST (node, local_votes_cache_batch)
 	auto channel (node.network.udp_channels.create (node.network.endpoint ()));
 	// Generates and sends one vote for both hashes which is then cached
 	node.network.process_message (message, channel);
-	system.deadline_set (5s);
+	system.deadline_set (3s);
 	while (node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out) < 1)
 	{
 		ASSERT_NO_ERROR (system.poll ());
@@ -2434,6 +2434,7 @@ TEST (node, local_votes_cache_batch)
 	ASSERT_FALSE (node.votes_cache.find (send2->hash ()).empty ());
 	// Only one confirm_ack should be sent if all hashes are part of the same vote
 	node.network.process_message (message, channel);
+	system.deadline_set (3s);
 	while (node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out) < 2)
 	{
 		ASSERT_NO_ERROR (system.poll ());
@@ -2443,12 +2444,14 @@ TEST (node, local_votes_cache_batch)
 	node.votes_cache.remove (send1->hash ());
 	node.votes_cache.remove (send2->hash ());
 	node.network.process_message (nano::confirm_req (send1->hash (), send1->root ()), channel);
+	system.deadline_set (3s);
 	while (node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out) < 3)
 	{
 		ASSERT_NO_ERROR (system.poll ());
 	}
 	ASSERT_EQ (3, node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out));
 	node.network.process_message (nano::confirm_req (send2->hash (), send2->root ()), channel);
+	system.deadline_set (3s);
 	while (node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out) < 4)
 	{
 		ASSERT_NO_ERROR (system.poll ());
@@ -2456,6 +2459,7 @@ TEST (node, local_votes_cache_batch)
 	ASSERT_EQ (4, node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out));
 	// There are two different votes, so both should be sent in response
 	node.network.process_message (message, channel);
+	system.deadline_set (3s);
 	while (node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out) < 6)
 	{
 		ASSERT_NO_ERROR (system.poll ());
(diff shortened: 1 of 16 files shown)


nano/node/request_aggregator.cpp
```diff
@@ -0,0 +1,225 @@
...
+		}
+		else if (!hash_root.first.is_zero () && store.block_exists (transaction_a, hash_root.first))
+		{
+			to_generate.push_back (hash_root.first);
+		}
+		else if (!hash_root.second.is_zero ())
+		{
+			nano::block_hash successor (0);
+			// Search for block root
+			successor = store.block_successor (transaction_a, hash_root.second);
```
SergiySW: Can be just `auto successor (store.block_successor ...`

nano/node/request_aggregator.cpp
```diff
@@ -0,0 +1,225 @@
...
+			else
+			{
+				stats.inc (nano::stat::type::requests, nano::stat::detail::requests_ignored);
+			}
+		}
+	}
+	// Unique votes
+	std::sort (cached_votes.begin (), cached_votes.end ());
+	cached_votes.erase (std::unique (cached_votes.begin (), cached_votes.end ()), cached_votes.end ());
+	for (auto const & vote : cached_votes)
```
SergiySW: Do we need some limiter to prevent situation with 12 cached votes for 12 requests?

nano/node/request_aggregator.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+void nano::request_aggregator::generate (nano::transaction & transaction_a, std::vector<nano::block_hash> hashes_a, std::shared_ptr<nano::transport::channel> & channel_a) const
+{
+	size_t generated_l = 0;
+	auto i (hashes_a.begin ());
+	auto n (hashes_a.end ());
+	while (i != n)
+	{
+		std::vector<nano::block_hash> hashes_l;
+		for (; i != n && hashes_l.size () < 12; ++i)
```
SergiySW: I think 12 should be some sort of constant here and in voting.cpp as well

nano/node/request_aggregator.hpp
```diff
@@ -0,0 +1,105 @@
...
+	size_t size ();
+	bool empty ();
+
+	const std::chrono::milliseconds max_delay;
+	const std::chrono::milliseconds small_delay;
+
+private:
+	void run ();
+	/** Aggregate and send cached votes for a pool, returning the leftovers that were not found in cached votes **/
+	std::vector<nano::block_hash> aggregate (nano::transaction &, channel_pool &) const;
```
SergiySW: Any specific reason to not use common in sourcs `nano::transaction const & `? Isn't much of a diffirence, but...

nano/node/request_aggregator.cpp
guilhermelawless: Good call, will be done in a future PR.

nano/node/request_aggregator.cpp
guilhermelawless: Done as `nano::network::confirm_ack_hashes_max` . Only place that cannot see this is `nano/secure/common.cpp` -> `nano::vote::vote` -> `assert (blocks_a.size () <= 12);`

nano/core_test/node.cpp
```diff
@@ -2424,21 +2424,42 @@ TEST (node, local_votes_cache_batch)
...
+	while (node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out) < 1)
+	{
+		ASSERT_NO_ERROR (system.poll ());
+	}
 	ASSERT_EQ (1, node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out));
 	ASSERT_FALSE (node.votes_cache.find (send1->hash ()).empty ());
 	ASSERT_FALSE (node.votes_cache.find (send2->hash ()).empty ());
 	// Only one confirm_ack should be sent if all hashes are part of the same vote
 	node.network.process_message (message, channel);
+	while (node.stats.count (nano::stat::type::message, nano::stat::detail::confirm_ack, nano::stat::dir::out) < 2)
```
wezrule: Should reset the deadline timer before each `system.poll` block. There's a few missing in this test and a test below, as well as in the new request_aggregator test

nano/node/request_aggregator.hpp
guilhermelawless: No reason, updated 👍

nano/node/request_aggregator.hpp
```diff
@@ -0,0 +1,105 @@
...
+		start (std::chrono::steady_clock::now ()),
+		channel (channel_a),
+		endpoint (nano::transport::map_endpoint_to_v6 (channel_a->get_endpoint ()))
+		{
+		}
+		std::vector<std::pair<nano::block_hash, nano::root>> hashes_roots;
+		std::chrono::steady_clock::time_point start;
+		std::chrono::steady_clock::time_point deadline;
+		nano::endpoint endpoint;
+		std::shared_ptr<nano::transport::channel> channel;
```
wezrule: channel and endpoint should be initialized in the order of the class members. You could also initialize some of them in-class.

nano/node/request_aggregator.cpp
```diff
@@ -0,0 +1,225 @@
...
+	}
+}
+
+void nano::request_aggregator::stop ()
+{
+	nano::unique_lock<std::mutex> lock (mutex);
+	stopped = true;
+
+	lock.unlock ();
+	condition.notify_all ();
```
wezrule: You can limit the lock scope and make use of the RAII pattern we use elsewhere i.e
```
{
  nano::lock_guard<std::mutex> guard (mutex);
  stopped = true;
}
condition.notify_all ();
```

nano/core_test/node.cpp
guilhermelawless: Indeed, thanks for taking a look at the tests 💯  addressed other comments too

PR #2487, opened 2020-01-15
IPC 2.0

cryptocode:
This PR proposes "IPC 2.0", on which RPC 2.0 and other things can be built (https://forum.nano.org/t/rpc-2-0-refactor/110)

Preliminary docs:
   - HTTP gateway: https://github.com/cryptocode/notes/wiki/IPC-2.0---RPC-gateway
   - Authorization: https://github.com/cryptocode/notes/wiki/IPC-Authorization
If anyone wanna test some the sample endpoints/test client, probably easiest to get started with no security (see the first config example)

Communication:
  - RPC 1.0 works as before, RPC 2.0 is on a new endpoint, /api/v2
  - IPC client and server supports full-duplex and can thus be used for subscriptions/callbacks, as well as request/response style APIs.
    A client can use both models through multiple connections.
  - A subscription system is in place (this should cover our callback need). In the future, "WebSockets 2.0" can be implemented in terms of this system.
  - Support for a new binary format based on Flatbuffers, a very efficient serialization format- and library.
  - Proposes a message format which can be used in both RPCs and messages oriented architectures (such as mqtt and websockets)
  - An HTTP gateway which automatically translates between Flatbuffers and JSON. This expects and delivers proper JSON instead of the ptree variant.
  - While RPC 2.0 is out of scope for this PR. a few tracer bullet RPCs and callbacks to verify the design have been added (such as AccountWeight and TopicConfirmation)

Security (enable_control replacement):
  - Rationale for this change is that the enable_control setting is too course grained, and that we'd like a common access control mechanism for both IPC and RPC.
  - Authentication through credentials. API keys are supported now, but the structure should be flexible enough to support stuff like JWT in the future.
  - Key concepts are Users, Roles and Permissions, where the code checks permissions. This is similar to Apache Shiro.
  - There is a default user to which we'll assign permissions we consider safe.
  - Roles and users (and default permission overrides) can be declared in `config-access.toml`
  - The access config file can be reloaded at runtime using a NOHUP signal

Services (using the IPC API):
  - We want to move more things out-of-process, but this gets unwieldy if done manually.
  - This is step one: A set of messages for IPC clients to register and stop/restart services.
    The next step (for a different PR) is to move the process startup from the daemon into a service manager

Internals:
  - A lock monitor is added to simplify threaded member access with finer granularity; this should be useful other places. Long-term it can be replaced by `libguarded` or similar.
  - For RPC 2.0, most of the meat will be in action_handler.* and ipc_broker.*

For RPC 2.0, we might consider writing a Flatbuffers transpiler for OpenAPI/Swagger.

In addition to generated docs, guides for programmers and node operators is needed.

@argakiig: as discussed for packaging, the Flatbuffers schema file in api/flatbuffers must be part of the redistribution.


argakiig:
Added in docker image installation and install targets for other os bundles

PR #2488, opened 2020-01-17
Unchecked cache not handling duplicate entries or unchecked_clear

wezrule:
The `unchecked_count` cache was not being updated when the unchecked table was cleared, which happened in 2 places. There was no `unchecked_clear` RPC test, so I've added one, the other code block is called on node start-up and isn't entered during tests, so couldn't make a test for it.

Processing blocks also just `unchecked_put` without any check for existence already, this is now checked. Added checks for tests which were confirmed to hit this. The extra `unchecked_exists` call only took 0.08% of CPU time during a snapshot on a bootstrapping attempt (gathering unchecked) on beta with LMDB. Untested on RocksDB, but `unchecked_get` is already an issue with it during bootstrapping, so any negative effects should be sorted for free alongside that.

Did a beta bootstrap and had 0 unchecked blocks and all checked blocks.

### DIFF
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -605,6 +605,9 @@ TEST (bootstrap_processor, lazy_max_pull_count)
 	{
 		ASSERT_NO_ERROR (system.poll ());
 	}
+
+	auto transaction = node1->store.tx_begin_read ();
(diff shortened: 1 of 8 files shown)


guilhermelawless:
Somewhat seems like it should be the ledger's job to handle this though, not the caller. Maybe in some backends it's always possible to know if data was already there.

PR #2497, opened 2020-01-20
Vote replay in crawler

clemahieu:
This change moves the responsibility for informing reps of very-behind sequence numbers off of the vote processor an on to the rep crawler. 
- Moves infrequently used codepath out of hot code section
- Removes database transactions from vote processor entirely.
- Eliminate unneeded deque construction of validated votes.


nano/node/repcrawler.cpp
```diff
@@ -38,6 +38,7 @@ void nano::rep_crawler::validate ()
 		nano::lock_guard<std::mutex> lock (active_mutex);
 		responses_l.swap (responses);
 	}
+	auto transaction (node.store.tx_begin_read ());
```
SergiySW: Could be moved to line 89 to prevent long tx holding

SergiySW:
Is informing reps of very-behind sequence numbers supposed to work only for crawler targets then?

clemahieu:
Yes, this is just to catch an actual rep up to speed. After the rep is updated they will continue to increase their sequence numbers as usual and everyone else will receive the new votes.

PR #2498, opened 2020-01-20
Emplacing wherever possible and other misc enhancements

guilhermelawless:
- Using emplace/emplace_back where possible. Had to make explicit constructors for [tcp_]endpoint_attempt and channel wrappers
- Looping with range for-loops and using const refs
- Avoiding one hash and lookup in ::reachout
- Using insert/emplace to insert multiple elements in some places rather than a loop + single insert/emplace

### DIFF
--- a/nano/node/transport/tcp.hpp
+++ b/nano/node/transport/tcp.hpp
@@ -137,7 +137,7 @@ namespace transport
 			std::shared_ptr<nano::transport::channel_tcp> channel;
 			std::shared_ptr<nano::socket> socket;
 			std::shared_ptr<nano::bootstrap_server> response_server;
-			channel_tcp_wrapper (std::shared_ptr<nano::transport::channel_tcp> & channel_a, std::shared_ptr<nano::socket> & socket_a, std::shared_ptr<nano::bootstrap_server> & server_a) :
+			channel_tcp_wrapper (std::shared_ptr<nano::transport::channel_tcp> const & channel_a, std::shared_ptr<nano::socket> const & socket_a, std::shared_ptr<nano::bootstrap_server> const & server_a) :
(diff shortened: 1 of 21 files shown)


cryptocode:
I know we don't use exceptions much (yet), but I seem to recall there being issues with emplace on smart pointer containers and exception safety.

guilhermelawless:
@cryptocode 
The only issue I have been able to find references to is emplacing a new pointer leading to a memory leak if it can't emplace due to lack of memory. In any case I've removed several `emplace` with smart pointers since they can't be made in-place.

The only location left is https://github.com/nanocurrency/nano-node/pull/2498/files#diff-a5e317b08279542bba01035dfeee7990R216 but what is constructed in-place is the pair

cryptocode:
@guilhermelawless I believe there is an exception safety case. Reading Scott Meyers item 42 now, but seems like it's related to custom deleters, which we've used only in one non-merged PR. So probably not an issue, just something to be aware of.

nano/node/transport/tcp.hpp
```diff
@@ -137,6 +137,10 @@ namespace transport
 			std::shared_ptr<nano::transport::channel_tcp> channel;
 			std::shared_ptr<nano::socket> socket;
 			std::shared_ptr<nano::bootstrap_server> response_server;
+			channel_tcp_wrapper (std::shared_ptr<nano::transport::channel_tcp> & channel_a, std::shared_ptr<nano::socket> & socket_a, std::shared_ptr<nano::bootstrap_server> & server_a) :
```
cryptocode: minor: I think we pass by const ref when passing a shared_ptr by reference (same below)

guilhermelawless: Good catch, did the same for the endpoint ref. Thanks!

PR #2499, opened 2020-01-20
Bootstrap attempts and connections/pulls separation

SergiySW:
- New class `nano::bootstrap_connections` to manage client connections & bulk pulls
- Separate source file for connections
- Restored bootstrap clients list (to close connections with `stop` command)
- Parent class `nano::bootstrap_attempt` & child classes for legacy, lazy & wallet bootstraps
- Separate source files for bootstrap attempts & lazy attempts
- Allowing several concurrent bootstrap attempts (currently for different bootstrap modes, in the future can be easily modified to allow same concurrent modes)
- Separate `nano::bootstrap_attempts` class for fast attempt search with incremental ID
- Bulk pull info is modified to include bootstrap attempt incremental ID
- `force` option in bootstrap RPCs & RPC "bootstrap" are currenlty designed to close all concurrent attempts
- Config field `bootstrap_initiator_threads` to manage bootstrap concurrency. Default 2 for multithreaded systems, 1 for singlethreaded & test network
- RPC "bootstrap_status" is modified to show all bootstrap attempts & new connections class
- Reduced lazy bootstrap attempts memory consumption: processed blocks unordered map is modified to store only 64 bit hash of block instead of full 256 bit union
- Fixed `websocket.bootstrap_exited` test TSAN warnings


guilhermelawless:
LGTM, tested a sync on beta net
```
{'reason': 'exited', 'id': 'FD73167472FD8EE570A21AC772C2979F', 'mode': 'legacy', 'total_blocks': '7027808', 'duration': '5657'}
```

nano/node/bootstrap/bootstrap.hpp
```diff
@@ -21,170 +21,27 @@ namespace mi = boost::multi_index;
 
 namespace nano
 {
-class bootstrap_attempt;
-class bootstrap_client;
 class node;
+
+class bootstrap_attempt_legacy;
+class bootstrap_attempt_lazy;
+class bootstrap_attempt_wallet;
```
wezrule: These seem unused

nano/node/bootstrap/bootstrap_lazy.cpp
```diff
@@ -0,0 +1,598 @@
...
+			lazy_pulls.pop_front ();
+			// Recheck if block was already processed
+			if (!lazy_blocks_processed (pull_start.first) && !node->store.block_exists (transaction, pull_start.first))
+			{
+				lock_a.unlock ();
+				node->bootstrap_initiator.connections->add_pull (nano::pull_info (pull_start.first, pull_start.first, nano::block_hash (0), incremental_id, batch_count, pull_start.second));
+				++pulling;
+				lock_a.lock ();
+				++count;
+			}
```
wezrule: Can move `lock_a.lock ();` below `++count;`

nano/node/bootstrap/bootstrap_lazy.cpp
```diff
@@ -0,0 +1,598 @@
...
+	}
+}
+
+void nano::bootstrap_attempt_lazy::lazy_blocks_erase (nano::block_hash const & hash_a)
+{
+	assert (!mutex.try_lock ());
+	auto erased (lazy_blocks.erase (std::hash<::nano::block_hash> () (hash_a)));
+	if (erased)
+	{
+		--lazy_blocks_count;
```
wezrule: Worth having an `assert` than it is greater than 0?

nano/node/bootstrap/bootstrap_lazy.hpp
```diff
@@ -0,0 +1,113 @@
...
+#include <boost/multi_index/hashed_index.hpp>
+#include <boost/multi_index/member.hpp>
+#include <boost/multi_index/ordered_index.hpp>
+#include <boost/multi_index_container.hpp>
+#include <boost/thread/thread.hpp>
+
+#include <atomic>
+#include <future>
+#include <queue>
+#include <unordered_set>
```
wezrule: Not sure all includes are needed: `<future>`, `<boost/thread/thread.hpp>` & `<nano/secure/ledger.hpp>`. Maybe others

nano/node/bootstrap/bootstrap_lazy.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+namespace nano
+{
+class bootstrap_attempt;
+class bootstrap_client;
+class node;
+namespace transport
+{
+	class channel_tcp;
+}
```
wezrule: `bootstrap_attempt`, `bootstrap_client` & `channel_tcp` seem unused in this file

nano/node/bootstrap/bootstrap.cpp
```diff
@@ -1,1547 +1,277 @@
...
-		}
-		if (!excluded_peers.check (nano::transport::map_endpoint_to_tcp (endpoint_a)))
-		{
-			attempt->add_connection (endpoint_a);
+			return i;
 		}
-		attempt->frontiers_confirmed = frontiers_confirmed;
-		condition.notify_all ();
 	}
+	return nullptr;
```
wezrule: Could use `std::find_if` algorithm

nano/node/bootstrap/bootstrap_attempt.hpp
```diff
@@ -0,0 +1,97 @@
...
+	class channel_tcp;
+}
+
+class frontier_req_client;
+class bulk_push_client;
+class bootstrap_attempt : public std::enable_shared_from_this<bootstrap_attempt>
+{
+public:
+	explicit bootstrap_attempt (std::shared_ptr<nano::node> node_a, nano::bootstrap_mode mode_a, uint64_t incremental_id_a, std::string id_a);
+	~bootstrap_attempt ();
```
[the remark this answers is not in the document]
SergiySW: Thanks!

nano/node/bootstrap/bootstrap.hpp
SergiySW: Updated

nano/node/bootstrap/bootstrap_lazy.hpp
SergiySW: Updated

nano/node/bootstrap/bootstrap_lazy.hpp
SergiySW: Updated

nano/node/bootstrap/bootstrap_lazy.cpp
SergiySW: It can become 0 so I guess we can check if it's not equal to `std::numeric_limits<size_t>::max ()` (less than 0 for uint)

wezrule: I meant 0 before deletion, but that works too :)

nano/node/bootstrap/bootstrap.cpp
SergiySW: Is this more symbols? Like
```
auto existing (std::find_if (attempts_list.begin (), attempts_list.end (), [&mode_a](auto const & item_a) { return item_a->mode == mode_a; }));
if (existing != attempts_list.end ()
{
	return *existing;
}
return nullptr;
```

wezrule: Was thinking this::
```
auto existing (std::find_if (attempts_list.cbegin (), attempts_list.cend (), [&mode_a](auto const & item_a) {
    return item_a->mode == mode_a;
}));
return (existing != attempts_list.cend ()) ? *existing : nullptr;
```

I prefer the named algorithm in this case than a raw for loop, but perhaps this usage of it isn't as great. Should have said it was optional :P

Srayman:
I noticed bootstrap connections are being limited to 1 per peer now compared to earlier where I could initiate 500-2000+ connections on the beta network.  Seems like it was always intended to be 1 per node/peer but maybe there was a bug?   Talking to Sergiy seems related to maybe the clients container.  While I agree with limiting the number of bootstrap connections are a critical area to prevent abuse could this be disabled on beta/test networks?  It's sometimes helpful to push the block limits more to test things downstream.

SergiySW:
@Srayman allowing duplicate connections for beta/test will be in different PR

SergiySW:
https://github.com/nanocurrency/nano-node/pull/2606

PR #2501, opened 2020-01-20
Bounded request aggregator

guilhermelawless:
With vote generation being out of I/O threads it's possible this new thread creates requests faster than they are replied to, needing some kind of upper bound.

- Limit the number of requests per channel to 512 by default (512 over 300ms is a large margin)
- Stop accepting requests if processing is slower than new requests are being queued by checking if the next deadline in the queue is too old

New stat `requests_dropped` is added.

Ideas for the future:
- Don't queue the request if the channel buffer is almost full
- Ban peers overloading with requests


SergiySW:
Exact limit can be beta tested, otherwise LGTM

PR #2504, opened 2020-01-21
Reinstate election request limit on the confirmation solicitor

guilhermelawless:


nano/node/confirmation_solicitor.hpp
```diff
@@ -22,15 +22,17 @@ class confirmation_solicitor final
...
 	/** Dispatch bundled requests to each channel*/
 	void flush ();
-	/** The maximum amount of confirmation requests (batches) to be sent to each channel */
+	/** Maximum amount of confirmation requests (batches) to be sent to each channel */
 	size_t const max_confirm_req_batches;
-	/** The global maximum amount of block broadcasts */
+	/** Global maximum amount of block broadcasts */
 	size_t const max_block_broadcasts;
+	/** Maximum amount of requests to be sent per election */
+	size_t const max_election_requests;
```
wezrule: Could this be a private member?

guilhermelawless: This is information that could be used by other parts of the node, similar to node config fields but static. But yes it could be private.

wezrule: Alright

PR #2505, opened 2020-01-22
Directed block broadcasting for long elections

guilhermelawless:
If we're not getting votes from representatives after some time, it makes more sense to send them the election winner rather than a random flood, but the flood is still kept at half scale. The duplicate filter prevents this from being a large burden on representatives.

Therefore, with this PR, the behavior is: **for up to 30 elections per loop, send to top sqrt(peers)/2 representatives that haven't voted for that election, and to sqrt(peers)/2 random peers.**

A second copy of `representatives` is made in the solicitor since some representatives could be erased from the list afterr #2649 .

Note: the change on the `update_difficulty` test is not necessary but makes it faster since it does not wait for the election broadcasting (outside of tests, the node broadcasts a block with updated difficulty)

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -590,8 +590,7 @@ TEST (active_transactions, update_difficulty)
 	node1.process_active (send1);
 	node1.process_active (send2);
 	node1.block_processor.flush ();
-	auto ec = system.poll_until_true (10s, [&node1, &node2] { return node1.active.size () == 2 && node2.active.size () == 2; });
-	ASSERT_NO_ERROR (ec);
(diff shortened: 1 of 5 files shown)


zhyatt:
@guilhermelawless Was a random flood previously sending the election winner to a random set of sqrt(peers)? And this instead would select reps to specifically target with the election winner, based on their not voting and prioritized by voting weight?

guilhermelawless:
@zhyatt that's correct, the idea is to target the reps that we'll be requesting confirmation later.

@SergiySW suggested keeping a lower random fanout to ensure some propagation (although reps will republish the block as well). Scaled fanout is available in #2468

PR #2510, opened 2020-01-24
Fix bootstrap_exited segfault on macos

cryptocode:


guilhermelawless:
Ok after 500 iterations under TSAN. Merging!

PR #2512, opened 2020-01-24
Cache genesis block hash in ledger constants

wezrule:
Introduced a new member in `ledger_constants` called `genesis_hash`. The only way to access it currently is through creating a new `class genesis` object and then calling a method which parses all of the `genesis_account` string each time, generates a block then gets the hash from that. This is what the node telemetry currently does, which can be a relatively expensive operation. Also tests use the genesis hash quite frequently so a helper has been added for them too.

I replaced all occurrences of `nano::genesis ().hash ()` with `nano::genesis_hash` in tests, and also started replacing any `nano::genesis genesis;` declarations which were only used with `genesis.hash` and `genesis.open->hash ()` but there was quite a few so haven't done all.

Unrelated - there was an [[no_discard]] warning in memory_pool.cpp on Windows which I have fixed with (void) cast.


SergiySW:
Also there is reference in json_handler.cpp `nano::json_handler::version ()`

PR #2514, opened 2020-01-27
macro guards for unstable tests

argakiig:
On actions node.fork_invalid_block_signature and node.bidirectional_tcp
intermittently fail for windows. 
#if !(defined(_WIN32) && CI) is used to guard these when running in actions if windows


argakiig:
closing for now

PR #2518, opened 2020-01-28
Active_transactions updates blocks when updating difficulty.

clemahieu:


nano/node/active_transactions.cpp
```diff
@@ -632,8 +632,13 @@ void nano::active_transactions::update_difficulty (std::shared_ptr<nano::block>
 			{
 				node.logger.try_log (boost::str (boost::format ("Block %1% was updated from difficulty %2% to %3%") % block_a->hash ().to_string () % nano::to_string_hex (existing_election->difficulty) % nano::to_string_hex (difficulty)));
 			}
-			roots.get<tag_root> ().modify (existing_election, [difficulty](nano::conflict_info & info_a) {
+			roots.get<tag_root> ().modify (existing_election, [election = existing_election->election, &block_a, difficulty](nano::conflict_info & info_a) {
```
guilhermelawless: Could use election from `info_a.election`, or is this deliberate?

guilhermelawless:
This code seems redundant now, could be removed:

https://github.com/nanocurrency/nano-node/blob/945c8ca3a12b331fbada1217e52ff8b7f0eb13bb/nano/node/wallet.cpp#L1458-L1473

nano/node/wallet.cpp
```diff
@@ -1455,23 +1455,6 @@ void nano::work_watcher::watching (nano::qualified_root const & root_a, std::sha
...
-											if (election->status.winner->hash () == hash)
-											{
-												election->status.winner = block;
-											}
-											auto current (election->blocks.find (hash));
-											assert (current != election->blocks.end ());
-											current->second = block;
-										}
-									}
-									watcher_l->node.network.flood_block (block, false);
```
guilhermelawless: Do we need flood_block forced here anyway since a different work value leads to the same hash?

clemahieu: Block flooding is handled by election which is informed of the update via active.update_difficulty

guilhermelawless: But in election.publish it will never enter the code that floods the block because the hash already existed from the block with the initial work.

But seems like this is a bug in the current implementation, updated blocks reaching other nodes are never republished, should we make it another PR? Perhaps checking if the work is updated [here](https://github.com/nanocurrency/nano-node/pull/2518/files#diff-2e3381130bf5f753e3cb73294268986eR222-R230) ?

guilhermelawless: Alternatively could flood the block to _all_ PRs from the work watcher with no need to be republished.

guilhermelawless: Placing the burden on the initial flood makes sense, in another PR.

PR #2521, opened 2020-01-29
Poll all nodes, remove outliers, ddos protection & amend RPC response with endpoint

wezrule:
This provides a set of enhancements for node telemetry:
1 - Remove outliers from consolidated data. 10% of the lower and upper bounds are removed
2 - Get metrics for all nodes instead of a random selection.
3 - Cache for 60 seconds on live, 15 seconds on beta. Nodes will reject `telemetry_req` messages if they are received from the same peer within this timeframe (+ the alarm cutoff tolerance).
4 - "cached" removed from RPC command
5 - "address" & "port" added to the RPC response when using "raw" to be able to match the metrics.
6 - "timestamp" added when using "raw" or single requests so that the data can be used accurately by services.

Other PRs will follow for some other requests


nano/node/json_handler.cpp
```diff
@@ -3910,6 +3910,12 @@ void nano::json_handler::telemetry ()
 			uint16_t port;
 			if (!nano::parse_port (*port_text, port))
 			{
+				if (!address_text->empty () && address_text->front () == '[' && address_text->back () == ']')
```
guilhermelawless: Strange that it's the first time running into this, should we extract a function?
@SergiySW bootstrap RPC also uses make_address_v6, should we do the same there?

nano/node/json_handler.cpp
```diff
@@ -3975,14 +3981,18 @@ void nano::json_handler::telemetry ()
 		// setting "raw" to true returns metrics from all nodes requested.
 		auto raw = request.get_optional<bool> ("raw");
 		auto output_raw = raw.value_or (false);
-		node.telemetry.get_metrics_random_peers_async ([rpc_l, output_raw](auto const & batched_telemetry_metrics_a) {
+		node.telemetry.get_metrics_peers_async ([rpc_l, output_raw](auto const & batched_telemetry_metrics_a) {
 			if (output_raw)
 			{
+				std::unordered_map<nano::endpoint, telemetry_data_time_pair> telemetry_data_time_pairs;
```
guilhermelawless: Appears to not be used

wezrule: Removed

nano/node/common.cpp
```diff
@@ -1550,6 +1431,12 @@ bool nano::parse_tcp_endpoint (std::string const & string, nano::tcp_endpoint &
 	return result;
 }
 
+std::chrono::seconds nano::telemetry_cache_cutoffs::network_to_time (network_constants const & network_constants)
+{
+	// 15s for beta to allow for quicker diagnostics, 60s for live
+	return std::chrono::seconds{ network_constants.is_live_network () ? 60 : network_constants.is_beta_network () ? 15 : 2 };
```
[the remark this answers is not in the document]
wezrule: Yep! good spot

nano/node/json_handler.cpp
wezrule: We had `parse_port` and `parse_port_and_address` in common so I've moved it there.
I think this is because in most places we don't use `[]`

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -239,10 +239,24 @@ void nano::bootstrap_server::receive_header_action (boost::system::error_code co
 				}
 				case nano::message_type::telemetry_req:
 				{
-					node->stats.inc (nano::stat::type::bootstrap, nano::stat::detail::telemetry_req, nano::stat::dir::in);
 					if (is_realtime_connection ())
 					{
-						add_request (std::make_unique<nano::telemetry_req> (header));
+						auto find_channel (node->network.tcp_channels.find_channel (remote_endpoint));
```
SergiySW: I think better make this check in another thread i.e. in request_response_visitor

SergiySW: Or `last_telemetry_req `can be variable in boostrap_server instead

wezrule: Moved to bootstrap_server, thanks

PR #2529, opened 2020-01-31
State blocks sideband upgrade adding is_send/receive/epoch

guilhermelawless:
This changes how the epoch byte in the state block sideband is [de]serialized, to also store these flags in the 3 most significant bits: `[is_send | is_receive | is_epoch | epoch (5 bits)]`

The upgrade is done in-place but takes a long time due to having to a random read to retrieve the previous block. A vacuum is performed at the end.

The upgrade does the following:
- Iterate through the state_blocks table
- For each block grab the previous block to determine the subtype of the current block
- Serialize the new sideband
- `mdb_cursor_put` with `MDB_CURRENT`

I was expecting no size growth due to using [MDB_CURRENT](http://www.lmdb.tech/doc/group__mdb__put.html#ga92f7832a496213db0f03105e6fa9afae), but no such luck. [mdb_cursor_put](http://www.lmdb.tech/doc/group__mdb.html#ga1f83ccb40011837ff37cc32be01ad91e) docs say:
> MDB_CURRENT - replace the item at the current cursor position. The key parameter must still be provided, and must match it. If using sorted duplicates (MDB_DUPSORT) the data item must still sort into the same place. This is intended to be used when the new data is the same size as the old. Otherwise it will simply perform a delete of the old record followed by an insert.

Since the new sideband deserialization discards the MSBs when reading the epoch, the upgrade can safely be stopped and restarted (from the beginning), and no additional versioning was necessary. This behavior is tested in the upgrade test.

However, it was necessary to move up the `block_sideband` class definition in order to have `block_w_sideband` declared in the same file (declaring in `versioning.hpp` would require saving a copy of the `block_sideband` class in `versioning.cpp`).

Also added a validation step of these new fields in the `--debug_validate_blocks` CLI.

-----

**Best case scenario test** (on a SATA III SSD)
Tested with @SergiySW 's daily ledger (which is rebuilt with https://github.com/nanocurrency/nano-node/pull/2435):
- Initial size 17GB
- Initial v16->v17 upgrade takes 2 minutes, basically no growth
- v17->v18 upgrade takes 45 minutes, growing to 29GB
- Post-upgrade vacuum brings it back to 17GB

This means the total space necessary for the best-case scenario was 29+17 = 46GB.

### DIFF
--- a/nano/nano_node/entry.cpp
+++ b/nano/nano_node/entry.cpp
@@ -1050,6 +1050,45 @@ int main (int argc, char * const * argv)
 							std::cerr << boost::str (boost::format ("Invalid signature for block %1%\n") % hash.to_string ());
 						}
 					}
+					// Validate block details set in the sideband
+					bool block_details_error = false;
+					if (block->type () != nano::block_type::state)
+					{
+						// Not state
+						block_details_error = sideband.details.is_send || sideband.details.is_receive || sideband.details.is_epoch;
+					}
+					else
+					{
+						auto prev_balance (node.node->ledger.balance (transaction, block->previous ()));
+						if (block->balance () < prev_balance)
+						{
+							// State send
+							block_details_error = !sideband.details.is_send || sideband.details.is_receive || sideband.details.is_epoch;
+						}
+						else
+						{
+							if (block->link ().is_zero ())
+							{
+								// State change
+								block_details_error = sideband.details.is_send || sideband.details.is_receive || sideband.details.is_epoch;
+							}
+							else if (block->balance () == prev_balance && node.node->ledger.is_epoch_link (block->link ()))
+							{
+								// State epoch
+								block_details_error = !sideband.details.is_epoch || sideband.details.is_send || sideband.details.is_receive;
+							}
+							else
+							{
+								// State receive
+								block_details_error = !sideband.details.is_receive || sideband.details.is_send || sideband.details.is_epoch;
+								block_details_error |= !node.node->store.source_exists (transaction, block->link ());
+							}
+						}
+					}
+					if (block_details_error)
+					{
+						std::cerr << boost::str (boost::format ("Incorrect sideband block details for block %1%\n") % hash.to_string ());
+					}
 					// Check if block work value is correct
 					if (nano::work_validate (*block.get ()))
 					{
--- a/nano/node/lmdb/lmdb.cpp
+++ b/nano/node/lmdb/lmdb.cpp
@@ -249,11 +249,14 @@ bool nano::mdb_store::do_upgrades (nano::write_transaction & transaction_a, bool
 			upgrade_v14_to_v15 (transaction_a);
 			needs_vacuuming = true;
 		case 15:
-			// Upgrades to v16 & v17 are both part of the v21 node release
+			// Upgrades to v16, v17 & v18 are all part of the v21 node release
 			upgrade_v15_to_v16 (transaction_a);
 		case 16:
 			upgrade_v16_to_v17 (transaction_a);
 		case 17:
+			upgrade_v17_to_v18 (transaction_a);
+			needs_vacuuming = true;
+		case 18:
 			break;
 		default:
 			logger.always_log (boost::str (boost::format ("The version of the ledger (%1%) is too high for this node") % version_l));
@@ -625,7 +628,7 @@ void nano::mdb_store::upgrade_v14_to_v15 (nano::write_transaction & transaction_
 		nano::state_block_w_sideband_v14 state_block_w_sideband_v14 (i_state->second);
 		auto & sideband_v14 = state_block_w_sideband_v14.sideband;
 
-		nano::block_sideband sideband{ sideband_v14.type, sideband_v14.account, sideband_v14.successor, sideband_v14.balance, sideband_v14.height, sideband_v14.timestamp, i_state.from_first_database ? nano::epoch::epoch_0 : nano::epoch::epoch_1 };
+		nano::block_sideband sideband (sideband_v14.type, sideband_v14.account, sideband_v14.successor, sideband_v14.balance, sideband_v14.height, sideband_v14.timestamp, i_state.from_first_database ? nano::epoch::epoch_0 : nano::epoch::epoch_1, false, false, false);
 
 		// Write these out
 		std::vector<uint8_t> data;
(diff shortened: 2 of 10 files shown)


guilhermelawless:
Log example:

```
[2020-Jan-31 12:51:39.787604]: Preparing v16 to v17 database upgrade...
[2020-Jan-31 12:52:05.966315]: Confirmation height frontier set for 200k accounts
[2020-Jan-31 12:52:27.812587]: Confirmation height frontier set for 400k accounts
[2020-Jan-31 12:52:48.697944]: Confirmation height frontier set for 600k accounts
[2020-Jan-31 12:53:09.644635]: Confirmation height frontier set for 800k accounts
[2020-Jan-31 12:53:29.040246]: Confirmation height frontier set for 1000k accounts
[2020-Jan-31 12:53:43.478123]: Finished upgrading confirmation height frontiers
[2020-Jan-31 12:53:43.480816]: Preparing v17 to v18 database upgrade...
[2020-Jan-31 12:55:01.512548]: Database sideband upgrade 1 million state blocks upgraded (out of 39024411)
[2020-Jan-31 12:56:13.118961]: Database sideband upgrade 2 million state blocks upgraded (out of 39024411)
[2020-Jan-31 12:57:27.724141]: Database sideband upgrade 3 million state blocks upgraded (out of 39024411)
(...)
[2020-Jan-31 13:38:40.215407]: Database sideband upgrade 39 million state blocks upgraded (out of 39024411)
[2020-Jan-31 13:38:41.990507]: Finished upgrading the sideband
```

nano/secure/blockstore.cpp
```diff
@@ -87,7 +128,7 @@ bool nano::block_sideband::deserialize (nano::stream & stream_a)
 		boost::endian::big_to_native_inplace (timestamp);
 		if (type == nano::block_type::state)
 		{
-			nano::read (stream_a, epoch);
+			details.deserialize (stream_a);
```
wezrule: You should check the return argument here in case of failure. Or in `block_details::deserialize` don't use a try-catch and let the exception propagate (with an appropriate comment)

nano/secure/blockstore.hpp
```diff
@@ -16,6 +16,62 @@
...
+	static constexpr size_t size ()
+	{
+		return 1;
+	}
+	bool operator== (block_details const & other_a) const
+	{
+		return epoch == other_a.epoch && is_send == other_a.is_send && is_receive == other_a.is_receive && is_epoch == other_a.is_epoch;
+	}
+	void serialize (nano::stream &) const;
+	bool deserialize (nano::stream &);
```
wezrule: Some functions are defined inline in the class definition but `serialize`/`deserialize` aren't. Would be good to move them all out, if only for consistency.

nano/secure/blockstore.hpp
```diff
@@ -279,6 +335,21 @@ class db_val
...
+		nano::state_block_w_sideband state_block_w_sideband;
+		state_block_w_sideband.state_block = std::make_shared<nano::state_block> (error, stream);
+		assert (!error);
+
+		state_block_w_sideband.sideband.type = nano::block_type::state;
+		error = state_block_w_sideband.sideband.deserialize (stream);
+		assert (!error);
+
+		return state_block_w_sideband;
+	}
```
wezrule: As this is essentially the same as `state_block_w_sideband_v14` it could be a refactored into a function template and then used in both places.

nano/secure/blockstore.cpp
```diff
@@ -3,14 +3,54 @@
...
+	return result;
+}
+
+void nano::block_details::unpack (uint8_t details_a)
+{
+	is_send = static_cast<bool> (details_a >> 7);
+	is_receive = static_cast<bool> ((static_cast<uint8_t> (details_a << 1)) >> 7);
+	is_epoch = static_cast<bool> (static_cast<uint8_t> (details_a << 2) >> 7);
+	epoch = static_cast<nano::epoch> (static_cast<uint8_t> (details_a << 3) >> 3);
+}
```
wezrule: Can using a `std::bitset<8>` instead of a `uint8_t` make this code a bit simpler? e.g can do:
```
is_send = bits.test (7);
is_receive = bits.test (6);
// etc
```

https://en.cppreference.com/w/cpp/utility/bitset

nano/secure/blockstore.cpp
guilhermelawless: Oops, good catch. I went with the first approach as we test block_details::deserialize independently.

nano/secure/blockstore.cpp
guilhermelawless: Thanks! Means converting to bitset and then to ulong but makes it less prone to errors (which I did initially).

nano/secure/ledger.cpp
```diff
@@ -288,6 +289,7 @@ void ledger_processor::state_block_impl (nano::state_block const & block_a)
 						if (result.code == nano::process_result::progress)
 						{
 							is_send = block_a.hashables.balance < info.balance;
+							is_receive = block_a.hashables.balance > info.balance;
```
SergiySW: Receive block can be 0 balance change (if receiving old type block with 0 send)

nano/node/lmdb/lmdb.cpp
```diff
@@ -781,6 +784,70 @@ void nano::mdb_store::upgrade_v16_to_v17 (nano::write_transaction const & transa
...
+		}
+		if (balance == prev_balance && network_params.ledger.epochs.is_epoch_link (block->link ()))
+		{
+			is_epoch = true;
+		}
+		else if (balance < prev_balance)
+		{
+			is_send = true;
+		}
+		else if (balance > prev_balance)
```
SergiySW: Same here (receive)

nano/secure/ledger.cpp
guilhermelawless: Is this check enough?
`is_receive = block_a.hashables.balance.number () >= info.balance.number () && !block_a.hashables.link.is_zero ();`

guilhermelawless: Added commit that also tests this case in the upgrade test

SergiySW: Seems enough

guilhermelawless: More specifically I did `is_receive = !is_send && !block_a.hashables.link.is_zero ();`

Should we add a `ledger` test for this edge case?

nano/nano_node/entry.cpp
```diff
@@ -1050,6 +1050,44 @@ int main (int argc, char * const * argv)
...
+							}
+							else if (block->balance () == prev_balance && node.node->ledger.is_epoch_link (block->link ()))
+							{
+								// State epoch
+								block_details_error = !sideband.details.is_epoch || sideband.details.is_send || sideband.details.is_receive;
+							}
+							else
+							{
+								// State receive
+								block_details_error = !sideband.details.is_receive || sideband.details.is_send || sideband.details.is_epoch;
```
SergiySW: I think we can add extra check ` !node.node->store.source_exists (transaction, block->link ()) `

guilhermelawless: Updated

guilhermelawless:
Changed recently so that open blocks are also marked as is_receive, for consistency.

@SergiySW performed a full validation of the blocks after bootstrapping from scratch on main net (thank you!).

guilhermelawless:
Opening another PR as Github is being weird no matter what I do, and all commits are signed.
image

PR #2530, opened 2020-01-31
Remove representatives with closed channels

SergiySW:


nano/node/repcrawler.cpp
```diff
@@ -233,16 +235,18 @@ nano::uint128_t nano::rep_crawler::total_weight () const
...
+	if (channel_a->get_tcp_endpoint ().address () != boost::asio::ip::address_v6::any ())
 	{
-		channel_ref_index.modify (itr_pair.first, [](nano::representative & value_a) {
-			value_a.last_request = std::chrono::steady_clock::now ();
-		});
+		probably_rep_t::index<tag_channel_ref>::type & channel_ref_index = probable_reps.get<tag_channel_ref> ();
+
+		// Find and update the timestamp on all reps available on the endpoint (a single host may have multiple reps)
+		auto itr_pair = channel_ref_index.equal_range (*channel_a);
+		for (; itr_pair.first != itr_pair.second; itr_pair.first++)
```
guilhermelawless: Optional cleanup for this section, I think this works:
```cpp
// Find and update the timestamp on all reps available on the endpoint (a single host may have multiple reps)
for (auto channel_i : boost::make_iterator_range (probable_reps.get<tag_channel_ref> ().equal_range (*channel_a)))
{
	channel_i.last_request = std::chrono::steady_clock::now ();
}
```

nano/node/repcrawler.cpp
```diff
@@ -252,9 +256,19 @@ void nano::rep_crawler::cleanup_reps ()
 	{
 		// Check known rep channels
 		nano::lock_guard<std::mutex> lock (probable_reps_mutex);
-		for (auto const & rep : probable_reps.get<tag_last_request> ())
+		auto iterator (probable_reps.get<tag_last_request> ().begin ());
+		while (iterator != probable_reps.get<tag_last_request> ().end ())
 		{
-			channels.push_back (rep.channel);
+			if (iterator->channel->get_tcp_endpoint ().address () != boost::asio::ip::address_v6::any ())
```
guilhermelawless: Would it be easier to keep the loop as it was and check this condition below in the removal loop below?

nano/node/repcrawler.cpp
SergiySW: I think we can't modify boost multiindex in this way because `last_request` is index

nano/node/repcrawler.cpp
SergiySW: It seems we cannot erase channel with empty endpoint by channel key because channel value can be modified outside of rep crawler. I.e. hash_code of channel returned some value (indexed in multiindex), but then became 0: https://github.com/nanocurrency/nano-node/blob/develop/nano/node/transport/tcp.cpp#L34
Probably we can add some random uin64_t to channels for comparision

nano/node/repcrawler.cpp
guilhermelawless: Tested this and you're right!

nano/node/repcrawler.cpp
guilhermelawless: Ok, something to keep in mind

PR #2531, opened 2020-02-01
Bounded memory and redesign in the confirmation height processor

wezrule:
I would recommend not diffing `confirmation_height_processor.cpp` as the algorithm has completely changed. Before the algorithm would start at the block which won the election, check previous until reaching the furthered uncemented block and note:
- All receive blocks
- Call `active.confirm_block` which stopped dependent elections and then stored information relevant for callbacks such as observer type.

It would then start at the bottom most receive block, follow to the sender and repeat the process until it had reached the closest path to genesis (on iterating downwards, no more receives are found). While this worked effectively, it used a lot of memory (especially for the callback data) which does not scale.

The desire was to use bounded memory, but for visualising purposes imagine it is only a buffer of 2 blocks, traversing a chain of millions of blocks, when the first block is confirmed we need to then start again at the top of the chain and repeat the process. This is not very effective, but can be improved by storing checkpoints every so often (size of the circular buffer) and then popping these, however these would also have to be bounded, so once exhausted you are back to using a lot of IO going back to the very beginning.

The algorithm proposed here uses a hybrid of these ideas and works as follows:
- Block is confirmed on network and added to the confirmation height processor.
- Get the cemented frontier of the account and using the sideband successor follow it until the first receive is hit.
- Follow the source block and repeat this process.
- Should there be a large amount of account chains to genesis, add a checkpoint every "size of circular buffer". This is less likely to be exhausted as we are essentially crawling on the floor of the cement.

Active transactions dependency has been removed. The scan for dependent elections now happens when cementing as opposed to iterating. This data can then be accessed when needed and passed back via callbacks.

Fixed a bug with `--confirmation_height_clear`, it was using the `genesis_block` string which isn't the hash of the block. When #2512 is in we can use the cached one rather than parsing it.

Did a live and beta full cement from 0, all observers called + blocks cemented.


nano/core_test/confirmation_height.cpp
```diff
@@ -1130,60 +1250,67 @@ TEST (confirmation_height, dependent_election_after_already_cemented)
...
+		ASSERT_EQ (0, node->active.blocks.size ());
+	}
 
-		system.deadline_set (10s);
-		while (!node->write_database_queue.contains (nano::writer::confirmation_height))
-		{
-			ASSERT_NO_ERROR (system.poll ());
-		}
+	system.deadline_set (10s);
+	auto transaction = node->store.tx_begin_read ();
```
SergiySW: Is it used now?

nano/core_test/confirmation_height.cpp
```diff
@@ -763,23 +735,14 @@ TEST (confirmation_height, pending_observer_callbacks)
...
-		if (node->pending_confirmation_height.size () == 0)
-		{
-			break;
-		}
-	}
-	node->confirmation_height_processor.add (send.hash ());
-	system.deadline_set (5s);
-	while (node->ledger.cache.cemented_count < 3)
+	system.deadline_set (10s);
+	while (1 != node->stats.count (nano::stat::type::http_callback, nano::stat::detail::http_callback, nano::stat::dir::out) || 1 != node->ledger.stats.count (nano::stat::type::observer, nano::stat::detail::all, nano::stat::dir::out))
```
SergiySW: Probably for consistency with other changes `node->stats.count () != 1` ?

nano/core_test/confirmation_height.cpp
wezrule: Removed

nano/core_test/confirmation_height.cpp
wezrule: Yep, thanks!

PR #2535, opened 2020-02-04
Election state refactor

clemahieu:
Converting the election class into a state machine instead of working on rebroadcast iterations.


nano/node/active_transactions.cpp
```diff
@@ -587,7 +529,19 @@ bool nano::active_transactions::add (std::shared_ptr<nano::block> block_a, bool
 			uint64_t difficulty (0);
 			error = nano::work_validate (*block_a, &difficulty);
 			release_assert (!error);
-			roots.get<tag_root> ().emplace (nano::conflict_info{ root, difficulty, difficulty, election });
+			auto inserted = roots.get<tag_difficulty> ().emplace (nano::conflict_info{ root, difficulty, difficulty, election });
+			if (inserted.first != roots.get<tag_difficulty> ().begin ())
+			{
+				--inserted.first;
+				if (!inserted.first->election->idle ())
```
guilhermelawless: This is not immediately obvious at first glance, but as far as I understand it will start the election as passive as long as the election just before it is not idle? Could you add a comment on what's happening, is there any case where an election will be idle?

nano/node/election.cpp
```diff
@@ -3,20 +3,23 @@
...
 
+using namespace std::chrono;
+
 nano::election_vote_result::election_vote_result (bool replay_a, bool processed_a)
 {
 	replay = replay_a;
 	processed = processed_a;
 }
 
 nano::election::election (nano::node & node_a, std::shared_ptr<nano::block> block_a, bool const skip_delay_a, std::function<void(std::shared_ptr<nano::block>)> const & confirmation_action_a) :
```
guilhermelawless: `skip_delay_a` was used to immediately start requesting for elections started from `active_transactions::search_frontiers`. Seems like a similar behavior can be achieved by transitioning to `active` immediately in `active_transactions::add` for these cases, currently it will take 5 seconds to start requesting.

nano/node/election.cpp
```diff
@@ -48,24 +51,168 @@ void nano::election::confirm_once (nano::election_status_type type_a)
...
+		{
+			last_block = std::chrono::steady_clock::now ();
+		}
+	}
+}
+
+bool nano::election::transition_time ()
+{
+	bool result = false;
+	if (std::chrono::minutes (5) < std::chrono::steady_clock::now () - election_start)
```
guilhermelawless: Seems like elections themselves should have an `expired` state_t, and assert that `transition_time` is not called once they are in that state, in the switch below (this code would have to move down below the switch). This ensures the election owner is correctly removing expired elections.

guilhermelawless: `confirmed` would also transition to `expired` it seems.

nano/node/election.cpp
```diff
@@ -48,24 +51,168 @@ void nano::election::confirm_once (nano::election_status_type type_a)
...
+				state_change (nano::election::state_t::passive, nano::election::state_t::active);
+			}
+			break;
+		}
+		case nano::election::state_t::active:
+			broadcast_block ();
+			send_confirm_req ();
+			if (!active_backtrack_once && base_latency () * 10 < std::chrono::steady_clock::now () - state_start)
+			{
+				active_backtrack_once = true;
```
guilhermelawless: Modeling with a flag goes against the state machine design, maybe an `active_backtracked` state?

nano/node/election.cpp
```diff
@@ -48,24 +51,168 @@ void nano::election::confirm_once (nano::election_status_type type_a)
...
+				activate_dependencies ();
```
guilhermelawless: Slightly worried about the 10 seconds here, as we see many elections taking longer than that with mild saturation. Shouldn't this be based at least on their priority? The best approximation seems to be `last_req`

nano/node/election.cpp
```diff
@@ -48,24 +51,168 @@ void nano::election::confirm_once (nano::election_status_type type_a)
...
+			}
+			break;
+		case nano::election::state_t::confirmed:
+			if (base_latency () * 10 < std::chrono::steady_clock::now () - state_start)
+			{
+				result = true;
```
guilhermelawless: Possibly transition to `expired` per the other comment.

nano/node/election.cpp
```diff
@@ -48,24 +51,168 @@ void nano::election::confirm_once (nano::election_status_type type_a)
...
+	}
+	if (escalated_l)
+	{
+		update_dependent ();
+	}
+}
+
+void nano::election::broadcast_block ()
+{
+	if (base_latency () * 5 < std::chrono::steady_clock::now () - last_block)
```
guilhermelawless: Should these multipliers of base_latency be documented constants in election.hpp?

nano/node/election.cpp
```diff
@@ -48,24 +51,168 @@ void nano::election::confirm_once (nano::election_status_type type_a)
 	}
 }
 
-void nano::election::stop ()
+bool nano::election::state_change (nano::election::state_t expected_a, nano::election::state_t desired_a)
```
cryptocode: Would be nice if this asserted on invalid state transitions, to catch (admittedly unlikely) bugs where state goes from, say, confirmed to something else.

nano/node/election.cpp
clemahieu: I changed this to 20x which roughly matches the exiting multipliers. We can tweak these more going forward.

nano/node/election.cpp
```diff
@@ -50,28 +57,258 @@ void nano::election::confirm_once (nano::election_status_type type_a)
...
+}
+
+bool nano::election::state_change (nano::election::state_t expected_a, nano::election::state_t desired_a)
+{
+	bool result = true;
+	if (valid_change (expected_a, desired_a))
+	{
+		if (state_m.compare_exchange_strong (expected_a, desired_a))
+		{
+			state_start = std::chrono::steady_clock::now ();
```
SergiySW: Seeems not protected or atomic?

nano/node/node.cpp
```diff
@@ -1089,8 +1089,11 @@ void nano::node::add_initial_peers ()
 
 void nano::node::block_confirm (std::shared_ptr<nano::block> block_a)
 {
-	active.insert (block_a, false);
-	network.broadcast_confirm_req (block_a);
+	auto election = active.insert (block_a, false);
+	if (election.second)
+	{
+		election.first->transition_active ();
```
SergiySW: Seems not protected & can use/modify some timepoints in election

guilhermelawless: Pushed some new commits. This is handled by making some timepoints atomic (confirm_req timepoint is protected by active mutex).

guilhermelawless: Re-done with a mutex since atomic timepoints are not a guaranteed support.

PR #2539, opened 2020-02-05
Network filter for publish and confirm_ack messages

guilhermelawless:
To-do test cases:
- [x] Block processor full
- [x] Vote processor full
- [x] Unchecked cleanup
- [x] Dropped inactive vote cache
- [x] Dropped active election
- [x] Dropped active election with inactive_vote_cache insertion
- [x] Bootstrap frontier confirmation

Adds duplicate detection of publish and confirm_ack messages before they are deserialized. When they are unique, the digest is saved and passed around to `network` which may drop the messages if the block or vote processor is full.

After processing, there are instances where digests have to be cleared from the filter to ensure the original message can be received again:

- Cleaning up a long unchecked block erases that element from the publish filter
- Reaching max capacity of the inactive votes cache clears the oldest item from the cache and the confirm_ack filter. For this reason, inactive_votes_cache now stores the digests, and votes pulled from this cache for new elections are now erased.
- Interactions between bootstrap frontier confirmation and inactive_votes_cache
- An active election is dropped. The vote digest is now saved here as well. Special case if a vote came from inactive_votes_cache, must ensure the digest is passed along.

The blocks_filter has been removed due to redundancy. 16 extra bytes are used for each hash in an (inactive) vote stored in memory.


nano/secure/stream_filter.cpp
```diff
@@ -0,0 +1,63 @@
...
+		size_t total_read{ 0 };
+		do
+		{
+			bytes_read = stream_a.sgetn (buffer.data (), buffer.size ());
+			total_read += bytes_read;
+			siphash.Update (buffer.data (), bytes_read);
+		} while (bytes_read == buffer.size ());
+
+		// Rewind stream to previous state
+		stream_a.pubseekoff (-total_read, std::ios_base::cur);
```
cryptocode: Since call sites have the raw buffer, I wonder if passing `uint8_t const * buffer_a, size_t size_a` instead of the stream would allow a single `siphash.Update` call for the entire buffer, instead of looping and rewinding at the end. (The loop misses the last few bytes if the stream length is not a multiple of 8, but that might not be important for the filter.)

nano/secure/stream_filter.cpp
```diff
@@ -0,0 +1,63 @@
...
+		siphash.Final (digest.bytes.data ());
```
cryptocode: Since call sites have the raw buffer, I wonder if passing `int8_t const * buffer_a, size_t size_a` instead of the stream would allow a single `siphash.Update` call for the entire buffer, instead of looping and rewinding at the end.

nano/secure/stream_filter.cpp
```diff
@@ -0,0 +1,63 @@
...
+nano::stream_filter::stream_filter (size_t capacity_a) :
+capacity (capacity_a),
+items (capacity_a, item_key_t{ 0 })
+{
+	nano::random_pool::generate_block (key, key.size ());
+}
+
+bool nano::stream_filter::operator() (bool & error_a, nano::stream & stream_a)
+{
+	bool existed;
```
cryptocode: Could be initialized, if only to silence warnings.

nano/secure/stream_filter.cpp
guilhermelawless: Had this initially, there is even a CalculateDigest that does an "all at once" update. `message_parser::deserialize_buffer` first deserializes the header and there doesn't seem to be any `header::size ()` to subtract from the total size, should I make it?

cryptocode: Right, and skip accordingly, e.g `buffer_a + header::size()`

nano/secure/stream_filter.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+	/**
+	 * Reads \p stream_a until end of stream and hashes the resulting contents.
+	 * Then, \p stream_a is rewinded back to its position before the function was called.
+	 * @remark error_a is set to true if an exception occurs when reading the stream, and rewinding is not performed
+	 * @return the siphash digest of the contents in \p stream_a
+	 **/
+	item_key_t hash (bool & error_a, nano::stream & stream_a) const;
+
+	size_t capacity;
```
clemahieu: This can be queried through items.size ().

nano/secure/stream_filter.cpp
guilhermelawless: Done, also now padding the buffer to ensure it has a size multiple of 8 bytes.

guilhermelawless: Reverted in 9d13191

nano/node/active_transactions.cpp
```diff
@@ -938,66 +938,82 @@ size_t nano::active_transactions::inactive_votes_cache_size ()
...
 		{
-			inactive_votes_cache.get<nano::gap_cache::tag_arrival> ().emplace (nano::gap_information{ std::chrono::steady_clock::now (), hash_a, std::vector<nano::account> (1, representative_a) });
+			auto & inactive_votes_by_time (inactive_votes_cache.get<tag_time> ());
+			inactive_votes_by_time.emplace (nano::votes_timepoint{ std::chrono::steady_clock::now (), hash_a, std::vector<std::shared_ptr<nano::vote>> (1, vote_a) });
 			if (inactive_votes_cache.size () > inactive_votes_cache_max)
 			{
-				inactive_votes_cache.get<nano::gap_cache::tag_arrival> ().erase (inactive_votes_cache.get<nano::gap_cache::tag_arrival> ().begin ());
+				auto oldest (inactive_votes_by_time.begin ());
+				auto node_l (node.shared ());
+				node.worker.push_task ([&node_l, votes_l = std::move (oldest->votes)]() {
```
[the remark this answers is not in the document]
guilhermelawless: Always forget this, thanks for spotting.

Srayman:
If a publish message is dropped when the block processor is full, should it be removed from the filter so that it can try to process it the next time it's seen?  How would this impact bootstrapping the block if the publish message is dropped but the filter still contains the block?  

See this line.
https://github.com/guilhermelawless/nano-node/blob/b5f82d75ed5664ecad4016782ebde72780fe542f/nano/node/network.cpp#L446

guilhermelawless:
That's absolutely missing, thank you @Srayman .

guilhermelawless:
Moving with #2643 in V21, filtering votes is left for a future version.

PR #2544, opened 2020-02-07
Use a different confirmation height algorithm when ledger is almost fully cemented

wezrule:
#2531 Introduced a bounded memory algorithm. This algorithm starts from the bottom of account chains. If the starting block when iterating the account chain isn't the first or second most uncemented block, we need to read from the cemented frontier. This results in extra IO. In addition we also do not store the block/sideband data because it's expensive to store this for every block iterated. This also leads to more IO when writing as we traverse the account chains again to extract this data. This is very good for keeping memory utilisation low and minimizing extra IO needed once the circular buffers are exhausted, but I think improvements can be made by using an all in-memory algorithm.

Some statistics from @SergiySW for an arbitrary block confirmed using the algorithm in #2531:
2227 seconds
8898 blocks per second
19816399 blocks cemented

This seems fine for bulk cementing the ledger, but when it is almost cemented I think we can improve IO utilisation by doing a few things:
1 - Iterate top/down
2 - Cache block/sideband for every block when iterating
3 - Don't do validation IO block reads for each pending write. Assume recently read blocks which have been confirmed are still in the ledger.

This means each block only has to be read exactly once from the ledger. During high TPS where disk IO is at a premium from block processing I think any advantage we can get is beneficial.

Separated the algorithms into 2 files, bounded & unbounded. The unbounded is very similar to the original algorithm before #2531 and utilises points 2) and 3) above. This should perform even faster than the original algorithm. This will use more memory of course, which is why we dynamically select the algorithm based on the number of uncemented blocks remaining. This new algorithm is used when the number of uncemented blocks is less than batch_write_size (4096). Which should be sufficient for any major TPS.

When reviewing it is recommended to compare `confirmation_height_unbounded.cpp` with the changes prior to #2531 and `confirmation_height_bounded.cpp` with the existing changes in `develop`. Also recommend turning off whitespace changes for reviewing `confirmation_height.cpp` test

Updated tests to check both algorithms using a new node flag.

Have done a full beta cement and ran all conf height tests on `core_test` with TSAN/ASAN.

Doc update needed: `--confirmation_height_clear` - Make sure this is not called while the node is running otherwise results may be inconsistent.


cryptocode:
Synced with this on beta and unbounded processor kicked in at the end of cementing as expected

wezrule:
> We'll have to do a full sync on live before we do the dev build though I don't see that as blocking this commit.

Just did a fresh live cement, took ~2.5 hours with the `frontiers_confirmation="always"` config option. Most of it is done a lot sooner, but it takes time to get through all the accounts with a low number of blocks

nano/rpc_test/rpc.cpp
```diff
@@ -7027,20 +7027,13 @@ TEST (rpc, block_confirmed)
 
 	// Wait until the confirmation height has been set
 	system.deadline_set (10s);
-	while (true)
+	auto transaction = node->store.tx_begin_read ();
+	while (!node->ledger.block_confirmed (transaction, send->hash ()) || node->confirmation_height_processor.is_processing_block (send->hash ()))
```
SergiySW: Probably can use `nano::node::block_confirmed_or_being_confirmed ()` ?

PR #2546, opened 2020-02-10
Restore max block processor signature verification batch size

SergiySW:
Removed in https://github.com/nanocurrency/nano-node/pull/2279


nano/node/blockprocessor.cpp
```diff
@@ -156,7 +156,18 @@ void nano::block_processor::verify_state_blocks (nano::unique_lock<std::mutex> &
...
 	nano::timer<std::chrono::milliseconds> timer_l (nano::timer_state::started);
 	std::deque<nano::unchecked_info> items;
-	items.swap (state_blocks);
+	if (state_blocks.size () <= max_count)
+	{
+		items.swap (state_blocks);
+	}
+	else
+	{
+		for (auto i (0); i < max_count && !state_blocks.empty (); ++i)
```
clemahieu: !state_blocks.empty() should be false due to the if condition, right? An assert instead inside the loop?

SergiySW: Should be, yes

PR #2549, opened 2020-02-11
Aggressive flooding for local blocks

guilhermelawless:
To-do
- [x] Add contained test

Floods locally produced blocks (going through work watcher or process_local - wallet, RPC process), such that these are sent to all principal representatives in addition to the usual fanout, now targeted at non-PRs.

This allows more protection against Sybil attacks, and more streamlined elections as there are less occasions of votes arriving before blocks.

As a consequence, the average amount of echoes will increase by 1 for PRs. However, with this change in place and used by most of the network, a future enhancement would be to reduce the republishing fanout.

This change is based on a proposal by @Srayman for more information see the original proposal at https://medium.com/nanocurrency/proposal-for-nano-node-network-optimizations-21003e79cdba


Srayman:
Looks good.  One comment on the initial publish, I don't think it's necessary to send to sqrt(non-pr) list.  It's already ample fanout to send to all PR's and then PR's to send to sqrt(peers) until that's changed later with dual phase/DHT to be more efficient.

Particularly for dual-phase in the future I think sending to Non-PR's would be undesirable since we would want to wait until PR's have confirmed there is not a fork before propagating the block to the rest of the network.

guilhermelawless:
@Srayman We are going with the safer approach for now, and re-evaluating when those future changes are done to ensure edge cases such as the one you mentioned are well covered.

guilhermelawless:
Tested on a beta net spam, saw higher bandwidth out as expected but otherwise stats look good.

PR #2555, opened 2020-02-12
Disable UDP by default

wezrule:
Introduce `--enable_udp` CLI flag, keeping but deprecating `--disable_udp` as it is now an unnecessary option but kept for backwards compatibility. Warning if both are used together.

Most changes are updates to tests so the same functionality is tested.

Unrelated: using `nano::writer::testing` in a couple block_processor tests.

### DIFF
--- a/nano/node/cli.cpp
+++ b/nano/node/cli.cpp
@@ -30,6 +30,8 @@ std::string nano::error_cli_messages::message (int ev) const
 			return "Config file read error";
 		case nano::error_cli::disable_all_network:
 			return "Flags --disable_tcp_realtime and --disable_udp cannot be used together";
+		case nano::error_cli::ambiguous_udp_options:
+			return "Flags --disable_udp and --enable_udp cannot be used together";
 	}
 
 	return "Invalid error code";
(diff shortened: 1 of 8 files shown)


nano/node/cli.cpp
```diff
@@ -88,7 +90,8 @@ void nano::add_node_flag_options (boost::program_options::options_description &
 		("disable_wallet_bootstrap", "Disables wallet lazy bootstrap")
 		("disable_bootstrap_listener", "Disables bootstrap processing for TCP listener (not including realtime network TCP connections)")
 		("disable_tcp_realtime", "Disables TCP realtime network")
-		("disable_udp", "Disables UDP realtime network")
+		("disable_udp", "(Deprecated) UDP is disabled by default")
+		("enable_udp", "Enables UDP realtime network")
```
clemahieu: What do you think of making an option like "enable_transport" that's a program_options::multitoken with tcp and udp as options. So it would be "--enable_transport = tcp udp" or "--enable_transport=tcp --enable_transport=udp"

The example they have for setting this up is
("email", value<string>()->multitoken(), "email to send to")
https://www.boost.org/doc/libs/1_71_0/doc/html/program_options/overview.html#id-1.3.32.5.7

wezrule: Do we want to disable tcp by default as well?

clemahieu: I guess we wouldn't want to force the transport to always be chosen.

wezrule:
> LGTM. Slightly unrelated, but could you add a `std::endl` in this line?
> https://github.com/nanocurrency/nano-node/blob/685987c10313e55abe69fb8ed7d879ac0dd54737/nano/nano_node/entry.cpp#L155

As discussed you will add it in another PR

PR #2558, opened 2020-02-13
[RocksDB] Update state block count after batch processing

wezrule:
Edit this is currently waiting on some changes required after https://github.com/nanocurrency/nano-node/pull/2853 is merged.

When profiling the blockprocessor adding 150K state blocks using the following test:
https://gist.github.com/wezrule/2fb14d1877b56c78623229ec939a7306

It showed that about 15% of the time was spent reading the `cached_counts` for state_blocks and updating it for each block added, this is now pretty much eliminated. Blocks are added in batches from the block processor, so I can leverage this knowledge by updating the count in RocksDB after bulk updating the blocks. In an ideal scenario we could just get a count of the blocks at start-up by traversing all the tables and caching the counts in memory like we do in other areas, however it would cause the node to take a while to start up each time, it doesn't scale very well and this approach seems to work well. Results show below in Fig 1:
image
Figure 1 - Left shows develop branch, right shows changes with the PR running the test mentioned above 

Modified a bunch of tests which deal with forks and rollbacks. Have done a successful full beta bootstrap which shows correct block counts, however no rollbacks were present, I have a live bootstrap underway currently.

Not currently doing this for legacy blocks.

Also fixed an issue with "accounts count" not being used correctly.


wezrule:
This is not essential for v21 and would require quite a bit of beta testing so moving to v22 to remove extraneous delays.

wezrule:

PR #2559, opened 2020-02-13
Buffer drop policies

cryptocode:
The socket can drop messages even when the bandwidth limiter is instructed not to, e.g. when the peer isn't keeping up. This PR proposes a buffer drop policy instead of the current droppable flag, so callers can request the socket not to drop as well. 

Policies:
- limiter: Can be dropped by bandwidth limiter (default, as before)
- no_limiter_drop: The limiter should not drop
- no_socket_drop: The limiter should not drop, nor should the socket if possible

The `no_socket_drop` policy is applied to telemetry, otherwise no changes in what is dropped and not. This policy is supposed to be used for low volume messages (the socket impl allows a doubling of the write queue at most, as a safety valve.)

### DIFF
--- a/nano/node/socket.hpp
+++ b/nano/node/socket.hpp
@@ -20,7 +20,7 @@ enum class buffer_drop_policy
 	limiter,
 	/** Should not be dropped by bandwidth limiter */
 	no_limiter_drop,
-	/** Should not be dropped by socket write queue limiter */
+	/** Should not be dropped by bandwidth limiter or socket write queue limiter */
 	no_socket_drop
 };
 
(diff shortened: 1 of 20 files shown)


nano/node/socket.hpp
```diff
@@ -13,6 +13,17 @@
...
 namespace nano
 {
+/** Policy to affects at which stage a buffer can be dropped */
+enum class buffer_drop_policy
+{
+	/** Can be dropped by bandwidth limiter (default) */
+	limiter,
+	/** Should not be dropped by bandwidth limiter */
+	no_limiter_drop,
+	/** Should not be dropped by socket write queue limiter */
```
guilhermelawless: Also not dropped by the bandwidth limiter right?

cryptocode: Correct, will clarify in comment as well, thanks.

PR #2564, opened 2020-02-14
Update --debug_profile_process CLI test

SergiySW:
* fix blocks signatures
* support daemon launch options for profiling
* use block count cache

### DIFF
--- a/nano/nano_node/entry.cpp
+++ b/nano/nano_node/entry.cpp
@@ -708,15 +708,17 @@ int main (int argc, char * const * argv)
 			nano::network_params test_params;
 			nano::block_builder builder;
 			size_t num_accounts (100000);
-			size_t num_interations (5); // 100,000 * 5 * 2 = 1,000,000 blocks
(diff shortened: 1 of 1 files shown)


nano/nano_node/entry.cpp
```diff
@@ -793,25 +795,41 @@ int main (int argc, char * const * argv)
...
+				if ((count % 1500) == 0)
+				{
+					std::cout << boost::str (boost::format ("%1% (%2%) blocks processed (unchecked)") % node->ledger.cache.block_count % node->ledger.cache.unchecked_count) << std::endl;
+				}
+			}
+			// Waiting for final transaction commit
 			uint64_t block_count (0);
+			{
+				auto transaction (node->store.tx_begin_read ());
+				block_count = node->store.block_count (transaction).sum ();
```
guilhermelawless: Is this not redundant given the loop below?

SergiySW: Changed

PR #2565, opened 2020-02-14
Push front blocks from unchecked

SergiySW:
Push blocks from unchecked to front of processing deque to keep more operations with unchecked inside of single write transaction.
It's designed to help with realtime blocks traffic if block processor is not performing large task like bootstrap.
If deque is a quarter full then push back to allow other blocks processing.

Tests improvements with pushing front unchecked blocks (without qiarter full limit):
* around 80% if processing send-receive pairs in incorrect order (receive, then immediately send). Short-living unchecked entries, edge case
* ~0% without unchecked usage or with large bootstrap
* around 120% if process send after receive with delay 50 blocks between


nano/node/blockprocessor.cpp
```diff
@@ -67,7 +67,7 @@ void nano::block_processor::add (std::shared_ptr<nano::block> block_a, uint64_t
 	add (info);
 }
 
-void nano::block_processor::add (nano::unchecked_info const & info_a)
+void nano::block_processor::add (nano::unchecked_info const & info_a, bool push_front_preference)
```
wezrule: Minor but probably want `push_front_preference` to have a trailing `_a` to be consistent with the other parameter

SergiySW: Makes sense, thanks!

PR #2569, opened 2020-02-17
Work version concept

guilhermelawless:
Adds nano::work_version which is passed to all work_generate and work_validate methods (except for tests, ensured with an assert).

The functional changes are quite minimal, as every caller defaults to the only work version available (work_1):

- RPC work_generate, work_validate and block_create can receive an optional "work_version"
- The work websocket outputs the work version

These changes constitute the pipeline necessary for conditional work validation.

------

Have been running the node under Debug and tested some work generation, none of the added asserts hit.


nano/node/json_handler.cpp
```diff
@@ -375,6 +375,24 @@ double nano::json_handler::multiplier_optional_impl (uint64_t & difficulty)
 	return multiplier;
 }
 
+nano::work_version nano::json_handler::work_version_optional_impl (nano::work_version const default_a)
+{
+	nano::work_version result = default_a;
+	boost::optional<std::string> version_text (request.get_optional<std::string> ("version"));
+	if (!ec && version_text.is_initialized ())
+	{
+		if (*version_text == "work_1")
```
wezrule: Could remove duplication & have type safety with: `nano::to_string (nano::work::work_1);`. However It will be more expensive as `std::string` doesn't have a trivial constructor so can't be elided (at least until C++20 has `std::string` constexpr constructors). If this is an issue (it's part of RPC code so don't think so), could make `to_string` a constexpr function and return `const char*`, or perhaps create a new `to_c_string ()` if a `std::string` return arg is still desired in some places.

guilhermelawless: Done, thanks. As you said it's the RPC so just went with the first suggestion 👍

PR #2570, opened 2020-02-18
Parallelize state block signature verification with block processor

wezrule:
While signature verification is done over a lot of threads, the block processor waits until all state blocks have been verified before processing any of the blocks (sometimes waiting up to 2 seconds). This PR parallelizes that process too and feeds `process_batch` smaller chunks (256 * sig threads +1) as opposed to 2048 to speed up initial block processing.

Using the following test (changing blocks from 1M to 150K though):
https://gist.github.com/wezrule/ab90069d569d43f6869af1d21c67ca48

The following CPU profile was made which recorded the stats during high CPU usuage (left is Develop, right is this PR):
image
It can be shown that block processing is done while signature verification is done with the PR, as opposed to blocking it.

Using the test (in the above gist) in its entirety for 1 million blocks it can be seen that rather in spurts of sig verification it is done continuously:
image
Using actual timings these are separate runs made:
Linux, 6 cores (12 threads)
| | PR (ms) | Develop (ms) |
| ---------- | ------------- | ------------- |
| | 14222  | 16739 |
| | 14216  | 17050 |
| | 14513 | 17491 |
| | 14516  | 17558 |
| | 14050 | 16651 |
| | 14156 | 17292 |
| | 14103  | 17449 |
| | 14154  | 17257 |
| | 14143  | 17032 |
| | 14012 | 17164 |
| **Average** | **14208.5** | **17168.3** |

Windows, 8 cores (16 threads). These results were made while working on other stuff, so they are not completely accurate but show the PR on average has an improvement.
| | PR (seconds) | Develop (seconds) |
| ---------- | ------------- | ------------- |
| | 78 | 97 |
| | 88 | 94 |
| | 90 | 88 |
| | 85 | 101 |
| | 85 | 90 |
| **Average** | **85.2** | **94** |

I wonder if we shouldn't use all available threads in the signature checker, otherwise it could block/delay other threads doing any processing. It seems to be quicker than the block processing thread so all blocks could be verified over a smaller number of threads (but will take longer ofc, but it doesn't seem like an issue here). Single verification may have different results though.

Cleaned up the signature checker a bit. 

Moved `collect_container_info` function taking a block_processor into the corresponding correct header/source file instead of `node`.

Did a full beta bootstrap, and ran TSAN/ASAN on the `slow_test` with a lower number of blocks to speed up the test with sanitizers.


guilhermelawless:
Besides the I/O threads also the request loop thread which blocks the active mutex, required to finish processing live blocks.

wezrule:
Changed to use half the amount of threads for signature checking. The highest CPU% reached was 71% under the same conditions as in the PR description, which reached 100% before.

image

guilhermelawless:
This line should be updated upon deciding the default number of threads for the sig checker https://github.com/nanocurrency/nano-node/blob/6b102d6f4e75cf04b25d87e455f8b6a0be054380/nano/node/nodeconfig.cpp#L79

PR #2575, opened 2020-02-20
Add flag for vote_processor capacity and tests

guilhermelawless:
Slightly limited in testing due to being unable to lock the vote processor loop (on the other hand, the block processor can be paused by starting a write transaction).

- Add node flag vote_processor_capacity, similar to the block processor flags
- Remove check to always process for the test network, the capacity is more than enough
- The representative levels stay the same but are now calculated on the fly from the given capacity
- ~Only unlock+notify in vote_processor::vote() if the processor was previously empty. Re-lock removed~ reverted, to be done in a future PR
- ::vote() now returns a boolean representing if the vote was processed, both for testing purposes and will be useful for the upcoming network filter
- ~Adjust vote_processor.flush test as it had a data race~  #2587

Ran all tests under TSAN/ASAN.


nano/node/vote_processor.cpp
```diff
@@ -88,56 +89,48 @@ void nano::vote_processor::process_loop ()
...
-			// Level 3 (> 5%)
-			else if (votes.size () < 144 * 1024)
-			{
-				process = (representatives_3.find (vote_a->account) != representatives_3.end ());
-			}
+			process = true;
 		}
-		else
+		// Level 1 (0.1-1%)
+		else if (votes.size () < 7.0 / 9.0 * max_votes)
```
SergiySW: Not sure if these values can be `static const` to prevent calculation for each vote

nano/node/vote_processor.cpp
```diff
@@ -88,56 +89,48 @@ void nano::vote_processor::process_loop ()
...
+			process = (representatives_3.find (vote_a->account) != representatives_3.end ());
 		}
 		if (process)
 		{
 			votes.emplace_back (vote_a, channel_a);
-
-			lock.unlock ();
-			condition.notify_all ();
-			lock.lock ();
+			if (votes.size () == 1)
```
SergiySW: Need some testing... All other options should be in `!votes.empty ()` `nano::vote_processor::process_loop ()` in theory

nano/node/vote_processor.cpp
guilhermelawless: Static const is pretty bad, could save the values in memory in the class but is memory access better than this? @wezrule what do you think?

wezrule: The easiest gains in this function would be made by passing parameters `vote_a` & `channel_a` by `const ref &` tbh.

Multiplying is pretty cheap, I would personally just do that. I wouldn't use local `static` variables due to the hidden thread synchronization costs which would outweigh it.

However if I was looking to micro-optimize this (as it's a hot path), and profiling showed this to be a bottleneck (I doubt it) then I might consider caching the values as class members, but this would require a lot of thought about cache lines, alignment and data layout; Because by itself it could make the performance worse. Imagine the `vote_processor` is aligned on a 64 byte boundary (typical cache line size) and all data fits neatly into a few cache lines. These extra variables would spill into another one which means if it isn't in the CPU cache already then when this function is called it would require an extra read from main memory (100-200 cycles). During high TPS it will likely be cached already though, as the data would have been recently accessed, but these are the types of things to think about. It would make sense to put all commonly used data members together and organised to reduce padding and make sure they are within the same cache line (align the vote processor on a 64 byte boundary) while you were at it, but any changes to this class in the future would then potentially screw this up (maintenance nightmare).

There's also out of order execution which could make this redundant too. We don't really do this micro-optimisation anywhere else currently, sounds a lot more effort than it's worth at this stage than just multiplying it imo.

guilhermelawless: Thanks for the in-depth details, I agree, having it clear in the code is a larger benefit than the micro optimizations that can be done.

nano/node/vote_processor.cpp
guilhermelawless: Reverted since these changes fall outside the scope of the PR, can re-evaluate in the future.

PR #2588, opened 2020-02-23
 LMDB sync options and new config settings

cryptocode:
Support no-sync options for LMDB for improved commit performance. Introduces an lmdb config section, similar to rocksdb, along with a deprecation for lmdb_max_dbs.

The config options are:

**sync** (applies to ledger, not wallet):
  * always: Default, matches current behaviour
  * nosync_safe: Do not flush meta data eagerly. This may cause loss of transactions, but maintains integrity.
  * nosync_unsafe: Let the OS decide when to flush to disk. On filesystems with write ordering, this has the same guarantees as nosync_safe, otherwise corruption may occur on system crash.
  * nosync_unsafe_large_memory: Use a writeable memory map. Let the OS decide when to flush to disk, and make the request asynchronous. This may give better performance on systems where the database fits entirely in memory, otherwise it may be slower.

**max_databases**: Takes over and deprecates `lmdb_max_dbs`
**map_size**: allows the map size to be changed. This only affects the ledger database

I've been profiling on a system with sufficient memory for nosync_unsafe_large_memory. Differences in profiling performance seem to match what was reported in Discord.

nosync_unsafe_large_memory = MDB_NOSYNC | MDB_WRITEMAP | MDB_MAPASYNC doesn't seem like a scalable option for full ledgers, since it's an optimization for when the database fits in memory - in fact it may give a slowdown as the ledger grows beyond available memory. I've added it as an option for systems with sufficient memory, and it might be a good option in the future for pruned ledgers. 

Note that this option will expand the file size logically to mapsize (currently 128GB), but `du -H data.ldb` will show its actually size. It may expand the file physically on some file systems.

For the nosync options, we could experiment in the future with calling `mdb_env_sync(..., true)` to periodically force a flush - perhaps on every N block process call. Currently it's called when the environment is closed (a no-op for the default always-sync setting)


SergiySW:
`--debug_profile_bootstrap` results
Built with NANO_SIMD_OPTIMIZATIONS
48786092 blocks in ledger
Intel Optane 900p 480GB, 64GB RAM, AMD Ryzen 3900X 12 cores

* always: 3131 seconds, 15578 BPS
* nosync safe: 3076 seconds, 15859 BPS
* nosync unsafe: 2855 seconds, 17082 BPS
* nosync unsafe + large memory: 1589 seconds, 30690 BPS

PR #2596, opened 2020-02-25
Attach sideband to block

guilhermelawless:
Sideband is now an optional directly in `block`, set on ledger processing (only if the result is `progress`) and when deserializing from the database. It can only be retrieved as a const ref, which asserts (debug) if not set before.

Removed `block_sideband::type` as it can always be inferred from the block.

Side effects of this change:
- Epoch code moved to nano_lib
- A `mutable_block_visitor` which can modify blocks, used by the ledger processor
- `block_store::block_get_no_sideband` is used in 2 places to avoid infinite recursion
- Some small versioning changes

Still have to run a main net bootstrap on it but all tests are passing. Some tests had to be changed so that two nodes don't simultaneously process the exact same block object, avoiding races on the sideband.


nano/lib/blocks.hpp
```diff
@@ -63,6 +106,9 @@ class block
 
 protected:
 	mutable nano::block_hash cached_hash{ 0 };
+	// Contextual details about a block, some fields may or may not be set depending on block type
+	// This field is set via sideband_set in ledger processing or deserializing blocks from the database
+	boost::optional<nano::block_sideband> sideband_m{ boost::none };
```
cryptocode: As optional is a value-wrapper, it looks like this will add 100 bytes to every block (block_sideband+block_details), also when sideband is not present (actually +8 bytes extra overhead for optional on my toolchain) 
I wonder if it's better to use a unique_ptr here (which has optional semantics, sort of), possibly using polymorphic sideband types to potentially save more space (may not be worth the inconvenience on call sites, esp since presumably most blocks will be state)

guilhermelawless: `unique_ptr` sounds good, I had a polymorphic sideband as a future task 👍 if you agree with doing it later, as this PR is quite large already.

guilhermelawless: Now using `nano::optional_ptr`, thanks for adding it!

PR #2598, opened 2020-02-25
Simplify telemetry data processing

wezrule:
Currently there are 2 modes of operation, single requests and bulk request rounds. Data from both ends up being used in the main processing code, but this has made it fairly complex. Now only single requests are done, any "bulk" requests just use what ever cached results are available. The ongoing telemetry request loop is now composed of many single requests. It is called more frequently, whenever a peer's telemetry data is out of date. This does mean that results are staggered, but the RPC will contain more up to date info, as it could miss peers before if they are connected to just after the request round, and also message traffic will be sparse rather than done all at the same time. Also the telemetry RPC (not single request) will always return cached data immediately and never wait the response timeout like it did before which improves. Most importantly the code is a lot less complex and I was able to remove a few (now unnecessary) tests too.


nano/node/telemetry.hpp
```diff
@@ -149,51 +85,52 @@ class telemetry
...
-	friend class node_telemetry_ongoing_requests_Test;
-	friend std::unique_ptr<container_info_component> collect_container_info (telemetry &, const std::string &);
+	// clang-format off
+	// This holds the last telemetry data received from peers or can be a placeholder awaiting the first response (check with awaiting_first_response ())
+	boost::multi_index_container<nano::telemetry_info,
+	boost::multi_index::indexed_by<
+		boost::multi_index::hashed_unique<boost::multi_index::tag<tag_endpoint>,
+			boost::multi_index::member<nano::telemetry_info, nano::endpoint, &nano::telemetry_info::endpoint>>,
+		boost::multi_index::ordered_non_unique<boost::multi_index::tag<tag_last_updated>,
+			boost::multi_index::member<nano::telemetry_info, std::chrono::steady_clock::time_point, &nano::telemetry_info::last_request>>>> recent_or_initial_request_telemetry_data;
```
cryptocode: Minor: would be nice with an `namespace mi = boost::multi_index;` alias use here, like we do a few places already, for better readability.

nano/slow_test/node.cpp
```diff
@@ -992,62 +932,46 @@ namespace transport
...
+				}
 
-					shared_data_a.shared_future.wait ();
-					ASSERT_EQ (shared_data_a.count, 0);
-				};
-
-				func (node_data_single, shared_data_single, true);
-				func (node_data_all, shared_data_all, false);
+				shared_data.shared_future.wait ();
+				ASSERT_EQ (shared_data.count, 0);
```
cryptocode: Not necessary now, but it would be nice if we could make the necessary adjustments to make `nano::util::counted_completion` work here; I think it's close.

wezrule: Done

nano/node/telemetry.hpp
wezrule: Done

PR #2603, opened 2020-02-26
Remove peers with different genesis block or invalid telemetry signature

wezrule:
After a tcp node handshake a telemetry data request is now made to more easily detect peers which we may not want to associate with. If the genesis hash is different from our own then the channel is removed. This is checked every message received (Serg suggestion)

Add new node flag to disable this for some node telemetry tests so that it can test the functionality in isolation without polluting expected results

Tested with TSAN with the new test. Also manually modified my genesis hash on a beta node, and it removed any connected peers.


nano/node/telemetry.cpp
```diff
@@ -59,45 +59,62 @@ void nano::telemetry::set (nano::telemetry_ack const & message_a, nano::transpor
...
+		// Disconnect from peer with incorrect telemetry data
+		if (channel_a.get_type () == nano::transport::transport_type::tcp)
+		{
+			network.tcp_channels.erase (channel_a.get_tcp_endpoint ());
+		}
+		else
+		{
+			network.udp_channels.erase (channel_a.get_endpoint ());
+			network.udp_channels.clean_node_id (channel_a.get_node_id ());
+		}
```
guilhermelawless: This could be a `network::erase (channel)` method

wezrule: Done

PR #2618, opened 2020-03-02
Sign telemetry messages

wezrule:
@cryptocode made a point about https://github.com/nanocurrency/nano-node/pull/2603 that a MITM attack could be used to close connections to peers as we are relying on this data having come from the peer; this can be mitigated by the node signing the message.

`size_a` in `telemetry_data::validate_message` is not yet used, but provides scaffolding for future telemetry versions.

Added the IGNORE_GTEST_INCL macro because I want to share testing code (compare default telemetry data function) between `core_test` & `rpc_test` which used google test types however other projects also include this (`node`) which doesn't link against googletest, so this is a way for them to exclude this dependency.


wezrule:
> Can use batcn verification lat I guess? Or it should be low enough count of messages?
> LGTM

Theoretical  maximum would be == total number of peers every (15/60 secs). But requests can go out at different times (same with responses) so it should be ok I think.

PR #2629, opened 2020-03-04
Make bootstrap_processor.process_state more robust under tsan

cryptocode:
This reduces the likelihood of test failure, but it still fails intermittently with the `while (!node1->active.empty ())` pattern.

Marking as wip until it's determined if it's a weakness in the test or if active transactions is not always draining (on testnet).


SergiySW (approved):
Passing TSAN with GCC 9 in multiple runs.
Possible bulk push TSAN socket closing issues worth separate investigation

zhyatt:
@cryptocode Thanks for the changes here. We are looking to commit them but need this commit signed to do so: https://github.com/nanocurrency/nano-node/pull/2629/commits/54c8a66b3fbd9978d3fe0d32c82dcf3c7728f426 - are you able to amend that with your signature?

SergiySW (approved):

wezrule (approved):

PR #2635, opened 2020-03-05
Add direct test to make sure elections are opened for unconfirmed tra…

globotree:
…nsactions if an election is not getting confirmed after enough time.


zhyatt:
This test appears to already be included in the develop branch, so closing the PR. Let us know if you see an issue with this @globotree

PR #2637, opened 2020-03-05
Read config file for CLI commands

wezrule:
#2594 added support for `--config` with inactive node. This PR adds support for reading the `config-node.toml` file too as well as making all CLI commands read it. This is quite prevalent when expecting commands to be used on the RocksDB ledger for instance.

I would suggest testing a couple random CLI commands to make sure the node config and --config CLI options are working as expected. 

Unrelated: `node.bandwidth_limiter` test had a hardcoded port
Unrelated: #2568 modified `assert_internal` to handle a case with Windows on Actions not outputting the stacktrace before exiting. This was resolved in https://github.com/nanocurrency/nano-node/pull/2591 so can be reverted.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -3578,7 +3578,7 @@ TEST (node, bandwidth_limiter)
 	nano::publish message (genesis.open);
 	auto message_size = message.to_bytes ()->size ();
 	auto message_limit = 4; // must be multiple of the number of channels
-	nano::node_config node_config (24000, system.logging);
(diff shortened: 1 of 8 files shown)


cryptocode:
With `--config` support everywhere, should we consider moving node flags into config at some point? 

Since `--config` allows passing in values on the CLI, there doesn't seem to be a good reason to keep node flags (IMO a crowded config file is better than a crowded CLI interface, since the config only needs to contain what's changed)

PR #2638, opened 2020-03-05
Definitions for work thresholds with epoch_2

guilhermelawless:
This article outlines the decision process for these thresholds: [Development Update: V21 PoW Difficulty Increases](https://medium.com/nanocurrency/development-update-v21-pow-difficulty-increases-362b5d052c8e).

Only adds definitions, `nano::work_threshold (version, details)` is not yet used anywhere besides tests.

We can change [these lines](https://github.com/nanocurrency/nano-node/compare/develop...guilhermelawless:work-threshold-definitions?expand=1#diff-5efcd17408c6db0b3d2576f25f90d4c4R87-R90) to early-return only for the live network if we would like to test live thresholds on the beta network.

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -76,7 +76,7 @@ void nano::block_processor::add (std::shared_ptr<nano::block> block_a, uint64_t 
 void nano::block_processor::add (nano::unchecked_info const & info_a)
 {
 	debug_assert (!nano::work_validate (*info_a.block));
-  bool should_notify{ false };
+	bool should_notify{ false };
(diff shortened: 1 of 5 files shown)


nano/node/blockprocessor.cpp
```diff
@@ -71,6 +71,7 @@ void nano::block_processor::add (std::shared_ptr<nano::block> block_a, uint64_t
 void nano::block_processor::add (nano::unchecked_info const & info_a)
 {
 	debug_assert (!nano::work_validate (*info_a.block));
+	if (info_a.block->difficulty () >= nano::work_threshold (info_a.block->work_version ()))
```
guilhermelawless: This line is temporary until the next PR where the check will be in ledger processing, separately for each block variant.

PR #2639, opened 2020-03-05
Make node_telemetry.dos_udp more robust under tsan

cryptocode:
Need to spin on some stat checks, as it otherwise fails intermittently under tsan. I used the opportunity to switch the test to ASSERT_TIMELY etc to reduce line count.

It's quite sensitive to timing, so I also had to do the  `- 1s` thing, otherwise it occasionally hangs forever; maybe there's a better way.


wezrule:
As discussed it doesn't seem to be an issue anymore. Tested a TSAN run with 50 iterations with https://github.com/nanocurrency/nano-node/pull/2669 to make sure it wouldn't break anything too in the future.

PR #2641, opened 2020-03-06
Increase active elections capacity with periodic full checks

guilhermelawless:
This was previously explored in #2440 with good results, but is not necessarily part of those changes.

The goal is to be able to hold more elections even if those are already confirmed. However, looping through all elections every 500ms (our default loop period) is not necessary, so this is only done every ~5 seconds. List of changes:

- `active_elections_size` increased from 10k to 50k. As a side effect, this fixes the slow frontier confirmation issue caused by lingering confirmed elections. `search_frontiers` was adjusted to prevent adding too many.
- Normally check top 5k elections every 500ms, but all 50k every 5 seconds.
- To avoid a situation where most of the top elections are confirmed, the loop counter is only increased for unconfirmed elections.
- Fixed election `confirmed_duration_factor` at 5 (corresponds to 5 seconds), even if saturated (possible due to above changes).
- Added timing logging for large number of elections.
- Adjusted timing logging for the block processor as it too verbose.
- Fixed a bug in RPC confirmation_info, this was introduced by mistake in #2619 

I wasn't sure if we should rename and deprecate the config value. It's a meaningful change and many operators may have overridden the value (e.g., by copying all default values at some point). It's not done in this PR, but something to consider.

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -19,6 +19,7 @@ confirmation_height_processor (confirmation_height_processor_a),
 node (node_a),
 multipliers_cb (20, 1.),
 trended_active_difficulty (node_a.network_params.network.publish_threshold),
+check_all_elections_period (node_a.network_params.network.is_test_network () ? 10ms : 5s),
 election_time_to_live (node_a.network_params.network.is_test_network () ? 0s : 2s),
 thread ([this]() {
 	nano::thread_role::set (nano::thread_role::name::request_loop);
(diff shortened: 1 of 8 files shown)


nano/node/active_transactions.cpp
```diff
@@ -230,11 +231,12 @@ void nano::active_transactions::request_confirm (nano::unique_lock<std::mutex> &
...
-	auto election_ttl_cutoff_l (std::chrono::steady_clock::now () - election_time_to_live);
-	auto roots_size_l (roots.size ());
-	bool saturated_l (roots_size_l > node.config.active_elections_size / 2);
-	auto & sorted_roots_l = roots.get<tag_difficulty> ();
-	size_t count_l{ 0 };
+	auto & sorted_roots_l (roots.get<tag_difficulty> ());
+	auto const election_ttl_cutoff_l (std::chrono::steady_clock::now () - election_time_to_live);
+	bool const check_all_elections_l (std::chrono::steady_clock::now () - last_check_all_elections > check_all_elections_period);
+	size_t const this_loop_target_l (check_all_elections_l ? sorted_roots_l.size () : node.config.active_elections_size / 10);
+	size_t count_l (0);
```
cryptocode: Very minor, renaming this to `unconfirmed_count_l` helps a bit on readability of the loop below

PR #2645, opened 2020-03-07
Rate limiting using token buckets

cryptocode:
There's been requests to add rate limiting to IPC/RPC v2, which I've been experimenting with. Token buckets seems like a good fit, which can also potentially be used for other purposes.

This PR experiments with applying it to the bandwidth limiter (motivation is to allow bursts). Default burst ratio in config is 3, so the long term limit is 5mb/s (configurable), but short outbound burst at 3x that rate is possible (but not sustainable)

My nodes seem to operate fine with various bandwidth limits, would be interesting to see it under load on beta.

(We can add similar rating algorithms to the nano::rate namespace as needed, like leaky buckets, etc, or combinations thereof)

For the IPC/RPC case, the idea is to allow rate limits to be added to access-config.toml (i.e. give roles or users access to certain APIs with an overall rate limit - we could also assign token cost per API if needed) - this would be implemented in a separate PR.


nano/node/transport/transport.cpp
```diff
@@ -218,61 +218,12 @@ bool nano::transport::reserved_address (nano::endpoint const & endpoint_a, bool
 
 using namespace std::chrono_literals;
 
-nano::bandwidth_limiter::bandwidth_limiter (const size_t limit_a) :
-next_trend (std::chrono::steady_clock::now () + 50ms),
-limit (limit_a)
+nano::bandwidth_limiter::bandwidth_limiter (const size_t limit_burst_ratio_a, const size_t limit_a) :
```
guilhermelawless: `limit_burst_ratio_a` could be floating point, with changes required in the config.

PR #2654, opened 2020-03-11
Fix node.aggressive_flooding failing intermittently

cryptocode:
This test fails intermittently in a few different places due to timeouts, especially under tsan, so increasing timeouts.


cryptocode:
@wezrule noticed it still fails under valgrind, every with very large timeouts. I'd say this points to an issue somewhere, so we'll keep looking.

cryptocode:
This will be replaced by a root cause fix

PR #2668, opened 2020-03-17
Add different epoch_2 thresholds for the beta and test networks

guilhermelawless:
Multipliers are 2x / 0.5x for these networks, but these changes allow testing the same functionality.

- Unifies all thresholds in lib/config.hpp similarly to what was done for `publish_threshold`
- Adjusts `nano::work_threshold` to return the appropriate threshold for all networks
- Improves unit tests


guilhermelawless:
To be replaced by a larger set of changes.

PR #2671, opened 2020-03-18
Handle epoch_2 work thresholds in the wallet and most RPCs

guilhermelawless:
This PR is quite extensive with work generation and validation being a central piece of the node.

In lib/config, work thresholds have been added for the beta and test network. They are now placed under a `nano::work_thresholds`. For the beta network, the thresholds are 2x and 1/2x from the base threshold. Tests have slightly higher difficulty threshold but I found this necessary to allow sufficient amplitude between thresholds and not having difficulty be too low for receives.

`nano::work_validate` was renamed to `work_validate_entry`, further validation must be done by using `work_threshold_entry` and `work_threshold_full`.

Work generation was changed to always require `difficulty`, except for tests. The reason is that any caller should not assume a base difficulty anymore.

The internal wallet now validates at the correct difficulties and generates new work when required. Pre-caching is left at the maximum difficulty for now.

These RPCs now fully support the new thresholds and are ready for epoch_2:
- `process` validates at the minimum, entry difficulty, and handles the new `insufficient_work` result from ledger processing (used in https://github.com/nanocurrency/nano-node/pull/2667).
- `send` and `account_representative_set` get the account epoch version and targets the correct difficulty
- `receive` additionally checks if the `source` epoch is higher, allowing new epoch propagation
- `epoch_upgrader`

These RPCs have changes to be compatible but are not yet optimal:
- `block_create` now accepts an optional `difficulty`, and will generate at the highest difficulty otherwise (8x on mainnet).
- `work_generate` also generates at the highest difficulty if `difficulty` is not specified, and multipliers are now off the highest difficulty.
- `work_validate` validates at the highest difficulty unless `difficulty` is specified, meaning this is a breaking change and it can return "not valid" for valid blocks. Multipliers are now off the highest difficulty.

The same multiplier changes affect the websocket (active_difficulty and work).

Support is also limited for active difficulty. It is currently calculated off the `epoch_1` difficulty to avoid any changes.

All these cases will be handled in separate PRs, as this one is large enough as-is.

Unrelated: test `active_transactions.adjusted_difficulty_priority` was generating at large difficulties, made a helper function that allows using 10x less difficulty which makes it much faster with sanitizers.


zhyatt:
@guilhermelawless Can you validate whether this should have a `websockets` label on it?

guilhermelawless:
@zhyatt only indirect changes due to adjusting the base difficculty. Removing the label, thanks for noticing.

PR #2673, opened 2020-03-21
Epoch open blocks should have corresponding pending entries

SergiySW:


nano/secure/ledger.cpp
```diff
@@ -407,6 +407,18 @@ void ledger_processor::epoch_block_impl (nano::state_block & block_a)
...
+					// Non-exisitng account should have pending entries
+					if (result.code == nano::process_result::progress)
+					{
+						bool pending_exists (false);
+						nano::account end (block_a.hashables.account.number () + 1);
+						for (auto i (ledger.store.pending_begin (transaction, nano::pending_key (block_a.hashables.account, 0))), n (ledger.store.pending_begin (transaction, nano::pending_key (end, 0))); i != n && i != ledger.store.pending_end (); ++i)
+						{
+							pending_exists = true;
+							break;
+						}
```
guilhermelawless: Could it be this instead?

```suggestion
						nano::account end (block_a.hashables.account.number () + 1);
						bool pending_exists = ledger.store.pending_begin(transaction, nano::pending_key(block_a.hashables.account, 0)) != ledger.store.pending_begin(transaction, nano::pending_key(end, 0));
```

SergiySW: Thanks, good idea!

PR #2680, opened 2020-03-25
Delay wallet work caching to allow using lower difficulty on demand

guilhermelawless:
This is done by using the node alarm to delay the work cache action by 10 seconds (1 second in tests). Since those can't be canceled, the wallet keeps a map state of each account, and when the alarm triggers, this is checked to only queue the action if the root is the same.

The map is wrapped in `nano::locked`.


guilhermelawless:
This is probably a good candidate for a config value, WDYT?

PR #2691, opened 2020-03-28
Active difficulty normalization

SergiySW:
* use multiplier instead of difficulty in conflict_info as base (adjusted difficulty is still difficulty yet)
* normalize multiplier for different epochs & blocks types
`normalized = (multiplier + (ratio - 1)) / ratio;`
* use adjusted & active multipliers instead of difficulties


nano/node/active_transactions.cpp
```diff
@@ -644,22 +645,102 @@ void nano::active_transactions::update_difficulty (std::shared_ptr<nano::block>
...
+	else if (winner_a != nullptr && winner_a->has_sideband () && winner_a->hash () == block_a->hash ())
+	{
+		threshold = nano::work_threshold (block_a->work_version (), winner_a->sideband ().details);
+	}
+	else
+	{
+		threshold = nano::work_threshold_base (block_a->work_version ());
+	}
+	double multiplier (nano::difficulty::to_multiplier (difficulty, threshold));
+	debug_assert (multiplier >= 1);
```
guilhermelawless: What if `block_a` is an epoch 2 receive (low difficulty) coming from `ledger::old` (no sideband) and different from winner? Or if it's a fork. This assertion seems too strict, we will normally have multipliers below 1 from now on.

nano/node/active_transactions.cpp
```diff
@@ -644,22 +645,102 @@ void nano::active_transactions::update_difficulty (std::shared_ptr<nano::block>
...
+	if (multiplier >= 1)
+	{
+		multiplier_normalization (multiplier, threshold);
+	}
+	return multiplier;
+}
+
+void nano::active_transactions::multiplier_normalization (double & multiplier_a, uint64_t const threshold_a)
```
guilhermelawless: I think this would work better as an overload of `normalized_multiplier` with different parameter types and returning the new multiplier instead of modifying the parameter.

nano/node/active_transactions.cpp
SergiySW: Ok well

nano/node/active_transactions.cpp
SergiySW: Thanks

nano/lib/work.cpp
```diff
@@ -145,6 +145,50 @@ uint64_t nano::work_v1::value (nano::root const & root_a, uint64_t work_a)
...
 
+double nano::normalized_multiplier (double const multiplier_a, uint64_t const threshold_a)
+{
+	static nano::network_constants network_constants;
+	debug_assert (multiplier_a >= 1);
+	auto multiplier (multiplier_a);
+	/* Normalization rules
+	// ratio = multiplier of max threshold with base threshold
+	// normalized = (multiplier + (ratio - 1)) / ratio;
+	Epoch 1
```
guilhermelawless, line 156: Since the numbers below assume a ratio, this comment could include what the ratio is.

nano/node/active_transactions.cpp
```diff
@@ -717,85 +749,88 @@ void nano::active_transactions::update_adjusted_difficulty ()
...
 }
 
 uint64_t nano::active_transactions::active_difficulty ()
 {
-	nano::lock_guard<std::mutex> lock (mutex);
-	return trended_active_difficulty;
+	return nano::difficulty::from_multiplier (active_multiplier (), node.default_difficulty (nano::work_version::work_1));
+}
+
+uint64_t nano::active_transactions::limited_active_difficulty (std::shared_ptr<nano::block> block_a)
```
guilhermelawless, line 815: This could be a `const & nano::block` instead. Similar for `normalized_multiplier`.

SergiySW: Updated

nano/lib/work.cpp
SergiySW: Updated

guilhermelawless:
Removing documentation label as in the end, every change is internal.

PR #2700, opened 2020-04-03
Add Security Policy file

zhyatt:
GPG keys need to be uploaded to `/blob/develop/etc/gpg/` with file name `<github username>.asc`

- [x] @argakiig 
- [x] @guilhermelawless 
- [x] @zhyatt 
- [x] @SergiySW 
- [x] @clemahieu 
- [x] @wezrule


zhyatt:
@argakiig @guilhermelawless Can you verify the updates I made look good?

zhyatt:
Sorry for the confusion here, turns out the Github approach isn't ideal so GPG public key files will need to be uploaded with only the key for the listed email. @SergiySW @clemahieu @wezrule

zhyatt:
Had to sort out my signatures on commits to get verified. @argakiig @guilhermelawless A double-check and stamp please?

PR #2701, opened 2020-04-03
Dont peer with v20 and earlier after epoch 2 block is seen

wezrule:
Once the first epoch block is seen, it invokes a callback which removes any peers lower than v20 nodes (protocol version v18). Any connections which are in flight are not affected, the added complexity in node handshakes didn't seem worth worrying about for this. The minimum protocol version is conditionally set to v17 before any epoch2 blocks are seen and v18 afterwards.

It was tested on the beta network using some code modifications just to make sure that peers actually reconnected once dropped, which they did but on my system (Windows) took about 5 minutes or so.


brunoerg:
Couldn't I run an old version of the node and set my "version constant" to 30, for example? i am asking it because I am thinking of writing my own node implementation and I don't see this mechanism so efficient...

guilhermelawless:
It's an optimization that prevents older nodes from bootstrapping. Anyone can fake it, sure.

brunoerg:
I understand it, but I think it can break light wallets and other implementations of the node. I think Nanollet broke with it, that's the reason for this commit:

https://github.com/brokenbydefault/Nanollet/commit/478f2677574e667cb5868d9a79780635b6e4e38c

PR #2702, opened 2020-04-04
Vote generator session for batch insertions

SergiySW:
Restoring removed in PR https://github.com/nanocurrency/nano-node/pull/2688


guilhermelawless:
For performance, should we make it single thread only and remove the mutex?

PR #2703, opened 2020-04-04
Difficulty calculation for RPC block_create

SergiySW:


nano/node/json_handler.cpp
```diff
@@ -1574,6 +1574,41 @@ void nano::json_handler::block_create ()
 			{
 				if (work == 0)
 				{
+					// Difficulty calculation
+					if (!request.get_optional<std::string> ("difficulty").is_initialized ())
```
guilhermelawless, line 1578: Could use `count` as it would be more clear and shorter.

SergiySW: Thanks!

PR #2704, opened 2020-04-04
Asynchronous epoch upgrade RPC

SergiySW:


nano/node/node.cpp
```diff
@@ -1340,6 +1340,252 @@ bool nano::node::init_error () const
...
+						++workers;
+						while (workers > threads)
+						{
+							nano::unique_lock<std::mutex> lock (mutex);
+							condition.wait (lock);
+						}
+						worker.push_task ([node_l = shared_from_this (), &upgrader_process, &condition, &upgraded_accounts, &workers, &to_next_account, epoch, difficulty, signer, root, account]() {
+							upgrader_process (*node_l, upgraded_accounts, to_next_account, epoch, difficulty, signer, root, account);
+							--workers;
+							condition.notify_all ();
```
guilhermelawless, line 1457: I forgot to add this. Every modification of `workers` must be done under the `mutex` to prevent a deadlock. But in this case it might just be easier to do change the wait conditions to `wait_for`.

guilhermelawless: If we go the protection by mutex route, then `workers` could be non-atomic

SergiySW: Thanks!

PR #2705, opened 2020-04-06
Ensure max_work_generate_difficulty is updated when changing the default difficulty

guilhermelawless:
Moves `max_work_generate_difficulty` out of config and into the node class, now calculated on the fly since it's rarely used.

`limited_active_difficulty` must be updated once #2691 is in.


guilhermelawless:
@SergiySW still seems ok no? https://docs.nano.org/running-a-node/configuration/#nodemax_work_generate_multiplier

Maybe we should change the default value though.

PR #2713, opened 2020-04-10
Optional randomized mode for local work generation

guilhermelawless:
Currently work generation is done FIFO, but in cases where someone is doing requests in burst and using multiple work peers, having the peers pick the next item randomly can increase efficiency, especially due to latency.

While this can be done with a simple middleware, this PR provides the same functionality directly in the node doing local work generation.

This needs an additional synchronization process between the threads while one of them chooses the next item, but the impact is not measurable as work generation uses considerable resources. Retrieving the chosen element is also an O(n) operation vs O(1) in sequenced mode. Doesn't seem worth using a second container or boost multi index container.

Changed to use 2 threads (unless only one is available) in tests as to cover multi-thread behavior


guilhermelawless:
We're leaving this out of V21 but the feature will be available for nano-work-server, see https://github.com/nanocurrency/nano-work-server/pull/12 .

PR #2723, opened 2020-04-16
Some confirmed block observer callbacks being missed

wezrule:
While doing some full cement testing, I noticed that the observer stats didn't match the number of blocks confirmed. This was due to the `election_winner_details` not being up to date when an election was set to the confirmed state. 

Moving the `node.process_confirmed` call out of the background process to a) simple setting of `election_winner_details` & b) it's very simple now (used to be complicated a while ago), there's probably more overhead in scheduling it on the `io_thread` than just calling it on the current thread.

Adding asserts to make sure that the callbacks are sequenced, this means that we can no longer manually add to the confirmation height processor in tests as there will be no election information, instead using `block_confirm`. This made the `pending_observer_callbacks` obsolete too so it has been removed.

The `confirmation_height.many_accounts_many_confirmations` test was also failing on the observer stats which now passes. Have done 50% of a clean cement on beta network with a debug build, but spam test are planned soon with it


nano/slow_test/node.cpp
```diff
@@ -817,21 +824,24 @@ TEST (confirmation_height, dynamic_algorithm_no_transition_while_pending)
...
-		ASSERT_EQ (node->ledger.cache.cemented_count, num_blocks + 1);
 		ASSERT_EQ (node->ledger.stats.count (nano::stat::type::confirmation_height, nano::stat::detail::blocks_confirmed, nano::stat::dir::in), num_blocks);
 		ASSERT_EQ (node->ledger.stats.count (nano::stat::type::confirmation_height, nano::stat::detail::blocks_confirmed_bounded, nano::stat::dir::in), 0);
 		ASSERT_EQ (node->ledger.stats.count (nano::stat::type::confirmation_height, nano::stat::detail::blocks_confirmed_unbounded, nano::stat::dir::in), num_blocks);
-		ASSERT_EQ (node->active.election_winner_details_size (), 0);
+		system.deadline_set (10s);
+		while (node->active.election_winner_details_size () > 0)
+		{
+			ASSERT_NO_ERROR (system.poll ());
+		}
```
guilhermelawless, line 844: Only cosmetic, but these could use ASSERT_TIMELY

nano/node/confirmation_height_bounded.cpp
```diff
@@ -491,7 +491,11 @@ bool nano::confirmation_height_bounded::cement_blocks (nano::write_guard & scope
 		scoped_write_guard_a.release ();
 		notify_observers_callback (cemented_blocks);
 	}
-	debug_assert (ledger.stats.count (nano::stat::type::confirmation_height, nano::stat::detail::blocks_confirmed) == ledger.stats.count (nano::stat::type::observer, nano::stat::detail::all, nano::stat::dir::out));
+	// Tests should check this already at the end, but not all blocks may have elections (e.g from manual calls to confirmation_height_processor::add), this should catch any inconsistencies on live/beta though
+	if (!network_params.network.is_test_network ())
```
guilhermelawless, line 495: These could be a single line if desired with `debug_assert (network_params.network.is_test_network () || condition)`, same for unbounded.

PR #2749, opened 2020-04-29
Multithreaded --validate_blocks

SergiySW:
Block count: 33136176 (November live network snapshot)

develop branch: 8523 seconds validation
2 threads: 4587 seconds validation time
4 threads / 4 cores: 2673 seconds validation time
8 threads (Hyperthreading) / 4 cores: 1814 seconds validation time

Additionally:
- silent command execution to show only validation status & errors count
- alias `--validate_blocks`/`--debug_validate_blocks`


nano/nano_node/entry.cpp
```diff
@@ -1355,29 +1356,81 @@ int main (int argc, char * const * argv)
...
+			if (threads_it != vm.end ())
 			{
+				try
+				{
+					threads_count = boost::lexical_cast<unsigned> (threads_it->second.as<std::string> ());
+				}
+				catch (boost::bad_lexical_cast &)
+				{
+					std::cerr << "Invalid threads count\n";
+					result = -1;
```
guilhermelawless, line 1377: Should exit here, currently it's continuing with 1 thread.

wezrule: Might also consider the try convert method instead of the exception one (`boost::lexical_cast` just calls this under the hood and throws exception if it fails):
```
if (!boost::conversion::try_lexical_convert (threads_it->second.as<std::string> (), threads_count)) {
  std::cerr
}
```

guilhermelawless: Should we also use default hardware_concurrency number of threads for this? It doesn't use the whole CPU on my system, still limited by disk even with NVME

SergiySW: I don't think so. If use all threads, then you probably can't run daemon in background. With fast i/o it can use all CPU. I.e. with Optane 900P I have full usage of 18 threads (according to `top`). I guess same can apply to some slower cloud vCPU + NVME SSD.

guilhermelawless: Fair enough, a callout on documentation instead 👍 
But if I provide `--threads abc` it's not stopping early at the moment.

SergiySW: Updated

PR #2751, opened 2020-04-30
Serialize telemetry as big endian

wezrule:
While our protocol messages don't currently have consistent ordering of bytes, we try to make sure that they are in the network byte order (big endian). That's what this change is mainly about.

There are 2 other unrelated changes:
1 - Move `node_telemetry.many_nodes` to `slow_test`  as it can take > 10 seconds to complete
2 - Explicitly add the telemetry size mask to the header extensions.

Protocol update in https://github.com/nanocurrency/protocol/pull/12


zhyatt:
Related https://github.com/nanocurrency/protocol/pull/12

PR #2754, opened 2020-05-01
Optional "block" given to RPC "work_generate" to infer difficulty

SergiySW:


nano/node/json_handler.cpp
```diff
@@ -4785,7 +4748,32 @@ void nano::json_handler::work_generate ()
 		{
 			ec = nano::error_rpc::difficulty_limit;
 		}
-		if (!ec)
+		// Retrieving optional block
+		std::shared_ptr<nano::block> block;
+		boost::optional<std::string> block_text (request.get_optional<std::string> ("block"));
+		if (!ec && block_text.is_initialized ())
```
guilhermelawless, line 4754: Could do `if (!ec && request.count ("block"))` to avoid retrieving the contents here.

nano/node/json_handler.cpp
```diff
@@ -4785,7 +4748,32 @@ void nano::json_handler::work_generate ()
...
+			block = block_impl (true);
+			if (block != nullptr)
+			{
+				if (hash != block->root ())
+				{
+					ec = nano::error_rpc::block_root_mismatch;
+				}
+				if (!ec && block->difficulty () >= difficulty)
+				{
+					// If optional block difficulty if higher than requested difficulty, send response immediately
```
guilhermelawless, line 4765: Small typo, should read "**is** higher than"

nano/node/json_handler.cpp
SergiySW: Thanks

nano/node/json_handler.cpp

PR #2758, opened 2020-05-06
Store whole block in confirmation height processor queue

wezrule:
Currently only a hash is passed to the confirmation height processor. We always need the block eventually so this would save some read IO. In the unbounded conf height case it is always the first `block_get` call done and may be the only one during real-time transactions. This does have a memory cost, before it was 32 bytes for the hash (+ boost multi-index internals) and now it will be more than 8x that. This processor never really gets more than 10k blocks being processed at any one time, so think this is acceptable for the IO gains especially during high TPS.

I struggled getting boost multi_index to accept a `const_mem_fun` with a reference return argument. I noticed the only other place we do this is with `nano::transport::channel` which uses a `std::reference_wrapper` so I copied that approach and it worked.

Ran with TSAN. No benchmarking has been done.


zhyatt:
@wezrule Are there any blockers or dependencies preventing us from getting this merged (outside of resolving the conflicts)?

wezrule:
@zhyatt no blocker, but should be re-tested with a full cement after merging conflicts.

clemahieu (approved):

PR #2763, opened 2020-05-07
Vote spacing

clemahieu:
Currently there is no rate limiter on vote generation for a particular root and this can lead to increased, unnecessary vote traffic.

The nano::vote_reserver class was created to track the last time a vote was generated for a particular root.

In order to accurately and globally throttle votes, all vote generation was moved to vote_generator::vote and should not be done elsewhere.

nano::vote_generator processes two separate vote pipelines: live traffic and requests.


nano/core_test/vote_generation.cpp
```diff
@@ -0,0 +1,310 @@
...
+	                                                .balance (node.balance (nano::test_genesis_key.pub) - 20)
+	                                                .link (nano::test_genesis_key.pub)
+	                                                .sign (nano::test_genesis_key.prv, nano::test_genesis_key.pub)
+	                                                .work (*system.work.generate (send2->hash ()))
+	                                                .build (ec);
+	ASSERT_FALSE (ec);
+
+	ASSERT_NE (send3->hash (), send3_fork->hash ());
+
+	// Start an election for send3, then publish send3_fork as welll
```
SergiySW: Typo "well"

nano/node/blockprocessor.cpp
```diff
@@ -242,7 +242,6 @@ void nano::block_processor::process_batch (nano::unique_lock<std::mutex> & lock_
 				// Deleting from votes cache & wallet work watcher, stop active transaction
 				for (auto & i : rollback_list)
 				{
-					node.votes_cache.remove (i->hash ());
```
SergiySW: Need to modify comment above I guess

wezrule (dismissed):

SergiySW (approved):

nano/node/voting.hpp
```diff
@@ -122,6 +151,8 @@ class vote_generator final
 	bool started{ false };
 	std::thread thread;
 
+	friend class vote_generator_vote_spacing_Test;
+	friend class vote_generator_vote_rapid_Test;
```
wezrule: There are 2 new `vote_spacing` tests added but it doesn't match these. So it doesn't look like these friend declarations are necessary anymore (and also the nano namespace around the tests)

wezrule (changes requested):

wezrule (dismissed):

argakiig (approved):

PR #2771, opened 2020-05-12
Make entry.sh more portable by removing all bashisms

bbedward:
The docker `entry.sh` script cannot be run in an environment where bash is not present.

This removes all the bashisms to make the script more compatible with more embeddable environments.

An example of why this is might be useful is if you want to use an alpine base image to run a nano node, you can get slightly faster build times and save ~4Mb by not installing bash


MacroChip:
Should likes 16 and 18 then be "sh pass through"?

docker/node/entry.sh
```diff
@@ -1,72 +1,72 @@
...
+	printf "  $0 nano_node [daemon] [cli_options] [-l] [-v size]\n"
+	printf "    daemon\n"
+	printf "      start as daemon\n\n"
+	printf "    cli_options\n"
+	printf "      nano_node cli options <see nano_node --help>\n\n"
+	printf "    -l\n"
+	printf "      log to console <use docker logs {container}>\n\n"
+	printf "    -v<size>\n"
+	printf "      vacuum database if over size GB on startup\n\n"
+	printf "  $0 bash [other]\n"
```
argakiig: remove 16-18 or reference sh

docker/node/entry.sh
```diff
@@ -1,72 +1,72 @@
...
+				echo "log_to_cerr = true"
+				command="${command} --config"
+				command="${command} node.logging.log_to_cerr=true"
+				;;
+			*)
+			 	command="${command} $i"
+				;;
+		esac
 	done
-elif [[ "${TEMP_OPTS[0]}" = 'bash' ]]; then
```
argakiig: was this intentionally removed without replacing with a sh alternative? see previous comment

bbedward:
Re-added passthrough option

image

PR #2774, opened 2020-05-17
Enable WebSocket server by default in Docker image

MacroChip:
The rpc server is enabled by default and port mappings are provided for both the rpc server and the websocket server. Keeping the configuration and documentation consistent between the two will be clearer.

### DIFF
--- a/docker/node/config/config-node.toml
+++ b/docker/node/config/config-node.toml
@@ -4,6 +4,7 @@
 # WebSocket server bind address
 # type:string,ip
 address = "::ffff:0.0.0.0"
+enable = true
 
 [rpc]
 
--- a/docker/node/config/config-node.toml
+++ b/docker/node/config/config-node.toml
@@ -1,13 +1,15 @@
 
 [node.websocket]
 
-# WebSocket server bind address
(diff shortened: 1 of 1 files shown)


docker/node/config/config-node.toml
```diff
@@ -4,6 +4,7 @@
 # WebSocket server bind address
 # type:string,ip
 address = "::ffff:0.0.0.0"
+enable = true
```
wezrule, line 7: Thanks!
`enable` should probably have the corresponding comments above it:
https://github.com/MacroChip/nano-node/blob/patch-1/nano/node/websocketconfig.cpp#L14

guilhermelawless:
```toml
# WebSocket server bind address.
# type:string,ip
address = "::ffff:0.0.0.0"

# Enable or disable WebSocket server.
# type:bool
enable = true
```

MacroChip: Makes sense. Thanks

guilhermelawless:
Thanks @MacroChip , this was a good idea.

PR #2775, opened 2020-05-18
Refactor shared testing code from core_test

wezrule:
In `core_test/testutil.hpp` we have some code which is shared between other test projects as well core libraries. This causes some issues as the core libraries don't include `gtest` so a special MACRO is used so the preprocessor can remove the gtest specific segments. Also because `core_test` is an executable project we cannot link against it, so any changes in `testutil.hpp` may not propagate to these other projects. This file has been stripped out to a new project called `test_common`, this has a gtest dependency, and anything else which is needed by both tests & core libraries has been moved to `testing.hpp/.cpp` of `node` project. Ideally this would be extracted out too, but not done here to keep changes low and it is less of a priority as all test projects currently link against `node`.

`entry.cpp` - Removed `--debug_mass_activity` which was deprecated in v21 (the equivalent `slow_test` can be used instead) which in addition to some other changes completely removes the `system` dependency from this file, which would allow the `system` class to be moved out to a new project entirely in the future (currently `qt_system` is the only non-gtest project which uses `system`. There seems to be an existing issue with `--debug_profile_votes` so this couldn't be tested to completion (active size never goes to 0), the others have been tested.


nano/test_common/testutil.hpp
```diff
@@ -214,61 +212,7 @@ namespace util
...
-	telemetry_data_l.sign (node_id_a);
-	// Signature should be different because uptime/timestamp will have changed.
-	ASSERT_NE (telemetry_data_a.signature, telemetry_data_l.signature);
-	compare_default_telemetry_response_data_excluding_signature (telemetry_data_a, network_params_a, bandwidth_limit_a, active_difficulty_a);
-	ASSERT_EQ (telemetry_data_a.node_id, node_id_a.pub);
-}
-#endif
+void wait_peer_connections (nano::system &);
+void compare_default_telemetry_response_data_excluding_signature (nano::telemetry_data const & telemetry_data_a, nano::network_params const & network_params_a, uint64_t bandwidth_limit_a, uint64_t active_difficulty_a);
+void compare_default_telemetry_response_data (nano::telemetry_data const & telemetry_data_a, nano::network_params const & network_params_a, uint64_t bandwidth_limit_a, uint64_t active_difficulty_a, nano::keypair const & node_id_a);
```
guilhermelawless: As this is now in a new project, should we consider moving these to another header?

wezrule: I didn't want to change structure too much due to the large volume of changes already, but it does make sense to separate this from the util file.

PR #2777, opened 2020-05-18
Add uncemented frontier output CLI command

wezrule:
This adds a CLI command `--debug_unconfirmed_frontiers` to output the frontier and confirmed (cemented) frontier for any accounts which are not yet fully cemented (warning: could output a lot if used on a ledger with large amount of uncemented blocks).


nano/secure/ledger.cpp
```diff
@@ -1203,6 +1203,26 @@ bool nano::ledger::block_not_confirmed_or_not_exists (nano::block const & block_
...
+	auto transaction (store.tx_begin_read ());
+	auto conf_height_i = store.confirmation_height_begin (transaction);
+
+	for (auto i (store.latest_begin (transaction)), n (store.latest_end ()); i != n; ++i, ++conf_height_i)
+	{
+		// Make sure the accounts match
+		debug_assert (conf_height_i->first == i->first);
+		if (i->second.block_count != conf_height_i->second.height)
+		{
+			auto frontier = i->second.head.to_string ();
```
SergiySW: Why to_string? Same for line below

wezrule: Probably artefact from earlier iteration of PR. Doesn't make sense to convert from hash to string back to hash 👎 Good spot!

SergiySW (approved):

PR #2785, opened 2020-05-24
Sequential voting

clemahieu:
There is an increasing amount of information that needs to be tracked for multi-level elections. This is a series of changes in order to decrease the amount of information that needs to be tracked in some of these cases.

* Block post-processing is executed outside of the block's write transaction. This decreases the amount of time the write_transaction needs to be held.
* Extracted dependent_blocks function which iterates over a block's dependencies. The function could_fit is rewritten in terms of this visitor
* Added function can_vote which checks for confirmation on a block's dependencies.
* Modified places where votes can be generated to ensure sequential voting.
* Cleanup on various tests to handle new sequential voting semantics.


nano/node/blockprocessor.hpp
```diff
@@ -27,6 +27,13 @@ enum class block_origin
 	remote
 };
 
+class block_post_events
```
wezrule, line 30: Should make this class `final` as it doesn't have a virtual destructor

nano/node/node.cpp
```diff
@@ -1108,6 +1109,12 @@ void nano::node::block_confirm (std::shared_ptr<nano::block> block_a)
 	}
 }
 
+bool nano::node::block_confirmed (nano::block_hash const & hash_a)
+{
+	auto transaction (store.tx_begin_read ());
+	return store.block_exists (transaction, hash_a) && ledger.block_confirmed (transaction, hash_a);
```
wezrule, line 1115: I've wondered if `block_account_height` (called inside `block_confirmed`) should just return 0 if the block doesn't exist and not actually error. Would save other places having to check if a block exists (like here)

nano/slow_test/node.cpp
```diff
@@ -570,7 +577,11 @@ TEST (confirmation_height, many_accounts_many_confirmations)
 	// Confirm all of the accounts
 	for (auto & open_block : open_blocks)
 	{
-		node->block_confirm (open_block);
+		auto election_insertion_result (node->active.insert (open_block));
+		ASSERT_TRUE (election_insertion_result.inserted);
+		ASSERT_NE (nullptr, election_insertion_result.election);
+		nano::lock_guard<std::mutex> guard (node->active.mutex);
+		election_insertion_result.election->confirm_once ();
```
wezrule, line 584: May need to add this to the new `confirmation_height, many_accounts_send_receive_self` test

nano/core_test/node.cpp
```diff
@@ -2925,7 +2942,7 @@ TEST (node, vote_by_hash_bundle)
 	for (int i = 1; i <= 200; i++)
 	{
 		nano::block_hash hash (i);
```
SergiySW, line 2944: Then this is not required

nano/node/node.cpp
guilhermelawless: Only found this location and it's currently only used for tests, makes sense but could be handled in another PR, what do you think?

wezrule: sure

nano/slow_test/node.cpp
guilhermelawless: Good catch, merged and updated

nano/secure/ledger.cpp
```diff
@@ -1039,57 +1040,90 @@ void nano::ledger::dump_account_chain (nano::account const & account_a, std::ost
...
-		result = ledger.store.block_exists (transaction, block_a.previous ());
+		result[0] = block_a.previous ();
 	}
 	void state_block (nano::state_block const & block_a) override
 	{
-		result = block_a.previous ().is_zero () || ledger.store.block_exists (transaction, block_a.previous ());
-		if (result && !ledger.is_send (transaction, block_a))
+		result[0] = block_a.hashables.previous;
+		result[1] = block_a.hashables.link;
+		if (ledger.is_epoch_link (block_a.hashables.link) || (ledger.store.block_exists (transaction, block_a.hashables.previous) && ledger.is_send (transaction, block_a)))
```
SergiySW, line 1110: I think we need some fail safe version of nano::ledger::is_send () function to save 1 read here while most blocks should have sideband attached

nano/node/active_transactions.cpp
```diff
@@ -183,6 +183,9 @@ void nano::active_transactions::block_cemented_callback (std::shared_ptr<nano::b
 				}
 			}
 		}
+		auto const & account (!block_a->account ().is_zero () ? block_a->account () : block_a->sideband ().account);
+		activate (account);
+		activate (block_a->link ());
```
SergiySW, line 188: Is this function supporting legacy blocks as well? If yes, need destination as well probably. If no, then don't need `block_a->sideband ().account` check

guilhermelawless: Added via `ledger.destination` in https://github.com/nanocurrency/nano-node/pull/2785/commits/c581d1ad6810159210bce80730ca32ad1e0b31ef , which also makes it so that it's not done for link in state receives (we don't want that, will be confirmed by the cementing process). Good catch.

nano/secure/ledger.cpp
guilhermelawless: Nice!

Added a debug_assert to `is_send` in https://github.com/nanocurrency/nano-node/pull/2785/commits/69f4818e84b193443355cd8b09f34709210a1610 similar to `block_source`

The easiest way to save the I/O read imo is an early check that the block has sideband, this is done in https://github.com/nanocurrency/nano-node/pull/2785/commits/a5f8eed2129d75526c2e4c971d96c86c899de947

PR #2793, opened 2020-06-01
Add logging section to rpc config to optionally disable logging

wezrule:
@Json's ramdisk node is crashing when pumping out 2000BPS `process` RPC calls. There are some boost standard error outputs suggesting problems with rotating the files at this high logging output speed.
image

However I cannot reproduce this even, even with spamming 50MB of log messages per second over a few threads. Seems to be something special about his set-up, removing the logging output manually fixed the issue for him, so am adding a logging section to the rpc config which allows these messages to be disabled (they are enabled by default). This can be extended in the future to change the rotation/max size size and maybe some other options, but right now this should be sufficient.


zhyatt:
@guilhermelawless That seems reasonable. Added to items in https://github.com/nanocurrency/nano-docs/pull/271

PR #2795, opened 2020-06-02
Write database queue functions should be NoOps with RocksDB

wezrule:
The `write_database_queue` is currently used by the block processor and confirmation height processor. It's purpose was to provide a way to know if something was holding the LMDB write transaction, so instead of being blocked it could continue doing work. This is not necessary with RocksDB as it doesn't have a global write lock, so these operations can be no-ops essentially.

There were of course some tests relying on this behaviour to test somethings in the confirmation height processor. In those tests (any explicitly using `write_database_queue::contains ()` or `write_database_queue::wait ()`) a new node flag `force_use_write_database_queue` can be set to true.


zhyatt:
Task linked: [CU-dz9fc Currency as a Choice: Part III](https://app.clickup.com/t/dz9fc)

PR #2810, opened 2020-06-12
Checks before overwriting unchecked

guilhermelawless:
**EDIT:** See comment below

As blocks can now enter the node although their work only passes the minimal validation, and not the full ledger work validation, this check prevents replacing unchecked with the same block with insufficient work.

Other checks (thanks @SergiySW) are made related to signatures.


guilhermelawless:
This was a misunderstanding as bootstrap blocks are not duplicate-filtered, so a poisoned chain can always be replaced via a lazy bootstrap on that chain. Closing.

PR #2811, opened 2020-06-12
Prevent lazy bootstrap start with incomplete initial bootstrap

SergiySW:
To prevent possible significant resources usage

### DIFF
--- a/nano/node/bootstrap/bootstrap.cpp
+++ b/nano/node/bootstrap/bootstrap.cpp
@@ -83,30 +83,36 @@ void nano::bootstrap_initiator::bootstrap (nano::endpoint const & endpoint_a, bo
 	condition.notify_all ();
 }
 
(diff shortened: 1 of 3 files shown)


nano/node/bootstrap/bootstrap.cpp
```diff
@@ -83,30 +83,36 @@ void nano::bootstrap_initiator::bootstrap (nano::endpoint const & endpoint_a, bo
...
 			lazy_attempt->lazy_start (hash_or_account_a, confirmed);
 		}
+		condition.notify_all ();
 	}
-	else
-	{
-		lazy_attempt->lazy_start (hash_or_account_a, confirmed);
-	}
-	condition.notify_all ();
+	return start_allowed;
```
guilhermelawless, line 115: Could this return that is used in the RPC ressponse be a bit misleading? Start being allowed doesn't mean it was effectively started, e.g. already ongoing

PR #2826, opened 2020-06-22
Increase expired lazy bootstrap delays

SergiySW:
Speicifcally before finished initial bootstrap
* Additionally RPC "bootstrap_lazy" can report different statues now: if new bootstrap was started & if new lazy key was added as start point


zhyatt:
@SergiySW To pull only RPC change out into new PR

SergiySW:
Replaced with https://github.com/nanocurrency/nano-node/pull/3237

PR #2831, opened 2020-06-30
Fix TCP message manager not notifying producers

guilhermelawless:
Also changing to `notify_one`

### DIFF
--- a/nano/node/network.cpp
+++ b/nano/node/network.cpp
@@ -893,11 +893,11 @@ void nano::tcp_message_manager::put_message (nano::tcp_message_item const & item
 		nano::unique_lock<std::mutex> lock (mutex);
 		while (entries.size () >= max_entries && !stopped)
 		{
-			condition.wait (lock);
+			producer_condition.wait (lock);
 		}
 		entries.push_back (item_a);
 	}
-	condition.notify_one ();
+	consumer_condition.notify_one ();
 }
 
 nano::tcp_message_item nano::tcp_message_manager::get_message ()
@@ -906,7 +906,7 @@ nano::tcp_message_item nano::tcp_message_manager::get_message ()
 	nano::unique_lock<std::mutex> lock (mutex);
 	while (entries.empty () && !stopped)
 	{
-		condition.wait (lock);
+		consumer_condition.wait (lock);
 	}
 	if (!entries.empty ())
 	{
@@ -918,7 +918,7 @@ nano::tcp_message_item nano::tcp_message_manager::get_message ()
 		result = nano::tcp_message_item{ std::make_shared<nano::keepalive> (), nano::tcp_endpoint (boost::asio::ip::address_v6::any (), 0), 0, nullptr, nano::bootstrap_server_type::undefined };
 	}
 	lock.unlock ();
-	condition.notify_one ();
+	producer_condition.notify_one ();
 	return result;
 }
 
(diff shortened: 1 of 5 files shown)


wezrule (dismissed):

nano/node/network.cpp
```diff
@@ -891,32 +891,35 @@ void nano::tcp_message_manager::put_message (nano::tcp_message_item const & item
...
 		nano::unique_lock<std::mutex> lock (mutex);
-		while (entries.size () > max_entries && !stopped)
+		while (entries.size () >= max_entries && !stopped)
 		{
 			condition.wait (lock);
 		}
 		entries.push_back (item_a);
 	}
-	condition.notify_all ();
+	condition.notify_one ();
```
clemahieu: I don't think we want to notify_one here, since multiple threads could be waiting on the queue to be less than max_entries, it might wake up one of those threads instead of the consumer.

nano/node/network.cpp
```diff
@@ -891,32 +891,35 @@ void nano::tcp_message_manager::put_message (nano::tcp_message_item const & item
 {
 	{
 		nano::unique_lock<std::mutex> lock (mutex);
-		while (entries.size () > max_entries && !stopped)
+		while (entries.size () >= max_entries && !stopped)
```
clemahieu, line 894: max_entries seems like it should be a valid size, did we need to change this comparison?

nano/node/network.cpp
```diff
@@ -891,32 +891,35 @@ void nano::tcp_message_manager::put_message (nano::tcp_message_item const & item
...
 		entries.pop_front ();
-		return result;
 	}
 	else
 	{
-		return nano::tcp_message_item{ std::make_shared<nano::keepalive> (), nano::tcp_endpoint (boost::asio::ip::address_v6::any (), 0), 0, nullptr, nano::bootstrap_server_type::undefined };
+		result = nano::tcp_message_item{ std::make_shared<nano::keepalive> (), nano::tcp_endpoint (boost::asio::ip::address_v6::any (), 0), 0, nullptr, nano::bootstrap_server_type::undefined };
 	}
+	lock.unlock ();
+	condition.notify_one ();
```
clemahieu: The same issue as above, this could wake another thread waiting on get_message and not a producer.

nano/node/network.cpp
wezrule: oh good spot, I missed that. I wonder if the producer/consumer should have different condition variables so the correct thread can be woken up instead of all.

nano/node/network.cpp
wezrule: Without the `=` then it's possible to go 1 above `max_entries`. Arguably it could then just be equality `while (entries.size () == max_entries)` because this is the only function which appends entries, so it would never go above the maximum (but possible fail-safe for future changes I guess?).

nano/node/network.cpp
wezrule: Requested from @guilhermelawless to make necessary changes

nano/node/network.cpp
guilhermelawless: Initially the added unit test had to target `max_entries + 1` rather than `max_entries`. Not a significant change but it's technically correct ;)

SergiySW (approved):

clemahieu (approved):

PR #2835, opened 2020-07-01
Allow unescaped quoted strings with --config CLI

wezrule:
This task was originally because of confusion around this error found by @guilhermelawless:
`./nano_node --network live --daemon --config node.online_weight_minimum="40000000000000000000000000000000000000"`
> Error deserializing config: Malformed number (out of range: stoll) at line 1

The [docs](https://docs.nano.org/running-a-node/configuration/#passing-config-values-on-the-command-line) state that strings should be escaped. Would be nice if non-escaped quotes could be done which is what this PR has added.

To solve this we are now separating the key and value, then appending them together with the value being surrounded in escaped quotations if they are not added by the command. Toml seems to be able to handle this fine based on various tests. Reviewers should test a variety of options with different types (bool/int/string for instance).

### DIFF
--- a/nano/core_test/cli.cpp
+++ b/nano/core_test/cli.cpp
@@ -46,26 +46,21 @@ TEST (cli, key_create)
 
 TEST (cli, config_override_parsing)
 {
-	auto error (false);
 	std::vector<nano::config_key_value_pair> key_value_pairs;
-	auto config_overrides = nano::config_overrides (key_value_pairs, error);
+	auto config_overrides = nano::config_overrides (key_value_pairs);
 	ASSERT_TRUE (config_overrides.empty ());
-	ASSERT_FALSE (error);
 	key_value_pairs.push_back ({ "key", "value" });
-	config_overrides = nano::config_overrides (key_value_pairs, error);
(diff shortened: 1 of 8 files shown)


argakiig (dismissed):

guilhermelawless (commented):
Tested:
- node.active_elections_size
- node.httpcallback.address
- node.max_work_generate_multiplier
- node.work_peers

No regressions and all of them now work with unescaped quotes or even no quotes.

However noticed on arrays like `node.work_peers` it's not possible to give `--config node.work_peers=["127.0.0.1:7000"]` , the quotes must be escaped instead: `--config node.work_peers=[\"127.0.0.1:7000\"]`.

Using the approach of this PR, can this case be handled? If not, LGTM but the documentation should note this particular case.

wezrule:
@guilhermelawless I have added support for arrays, thanks for finding that during testing.

guilhermelawless (approved):
Working great!
Could change some `ss << "(char)"` to single quotes but not too important.

A note for documentation is that arrays must not have spaces in between entries.

PR #2837, opened 2020-07-07
UPnP existing lease and other misc enhancements

guilhermelawless:
- Add timed leases (30 minutes) with automatic renewal. The periodic map check will also renew it if something went wrong
- Response from a mapping check was being misinterpreted as remaining lease time, but it returns the total lease time, not remaining (pretty useless)
- Bug: `external_port` was not being set when a lease already exists (e.g., node crashed and restarted). This means self keepalives don't include external address and port information.
- Make it more evident (with debug_assert) that `check_mapping` and `refresh_mapping` can't be used in tests
- Always log on successful mapping rather than being behind the optional `logging.upnp_details` config; otherwise on new leases, it shows the first mapping check as an error and nothing afterwards, causes confusion

Keepalive change:
- 25% of messages are now keepalive_self to ensure peers have updated mapping details.

----

Recommend hiding whitespace changes for `check_mapping()`.

### DIFF
--- a/nano/node/network.cpp
+++ b/nano/node/network.cpp
@@ -145,28 +145,7 @@ void nano::network::send_keepalive (std::shared_ptr<nano::transport::channel> ch
 void nano::network::send_keepalive_self (std::shared_ptr<nano::transport::channel> channel_a)
 {
 	nano::keepalive message;
-	random_fill (message.peers);
-	// Replace part of message with node external address or listening port
-	message.peers[1] = nano::endpoint (boost::asio::ip::address_v6{}, 0); // For node v19 (response channels)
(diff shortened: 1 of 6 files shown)


SergiySW (approved):

argakiig (approved):

NIKOLAY73731 (commented):
nano_3umdphh3t9zszqo45ygghdidddrzxuhtg6opaotbtjxnw6tyarcp9dau6nyf

NIKOLAY73731 (commented):
nano_3umdphh3t9zszqo45ygghdidddrzxuhtg6opaotbtjxnw6tyarcp9dau6nyf

PR #2840, opened 2020-07-10
Different log when the OpenCL library cannot be loaded

guilhermelawless:
Currently the only message is "0 platforms" which is misleading.

Not sure about the method used to do this due to the separation we have between the API wrapper by platform, open to suggestions.

Best reviewed without whitespace diffs.


wezrule (approved):
Works when the library isn't found, but I think it would be nice to another another message if the library is found but there are no compatible platforms/devices.

SergiySW (approved):

PR #2847, opened 2020-07-14
Changelog generator grab pull from commit message

guilhermelawless:
This makes it invariant to the commit hash, all it needs is the PR number in the first line of the commit message (automatic)

### DIFF
--- a/util/changelog.py
+++ b/util/changelog.py
@@ -104,7 +104,7 @@ class cliArgs():
             required=True,
         )
         options = parse.parse_args()
-        self.repo = options.repo
+        self.repo = options.repo.rstrip("/")
         self.start = options.start
         self.end = options.end
         self.pat = options.pat
@@ -121,6 +121,7 @@ class cliArgs():
 class generateTree:
     def __init__(self, args):
         github = Github(args.pat)
+        self.name = args.repo
         self.repo = github.get_repo(args.repo)
         self.start = args.start
         self.end = args.end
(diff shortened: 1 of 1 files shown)


util/changelog.py
```diff
@@ -138,15 +138,21 @@ def __init__(self, args):
...
                     for label in pull.labels:
                         labels.append(label.name)
                     self.commits[pull.number] = {
                         "Title": pull.title,
                         "Url": pull.html_url,
                         "labels": labels
-                    }
+                        }
+                except ValueError:
+                    print("Commit has no associated PR {}: \"{}\"".format(commit.sha, message))
```
argakiig, line 156: Do we want to have this act as a catch for non-PRs, would still include the title and url, but the label would be other. An example of this would be when we increment the CMake versions strings, historically those have not been a specific PR but rather a push to the appropriate releases or develop branch?
Otherwise LGTM

argakiig (dismissed):

util/changelog.py
guilhermelawless: Great idea, pushing a commit that results in the following layout, what do you think?

### Other Updates
  
|Commit|Title  
|:-:|:--  
|[8263eb68](https://github.com/nanocurrency/nano-node/commit/8263eb681c70117b15e826a2e96a50cb261fa27e)|Changelog generator grab pull from commit message

guilhermelawless: Handling with "other" label is quite difficult in the current script unfortunately.

argakiig (approved):

PR #2853, opened 2020-07-24
[RocksDB] Use read-modify-write merge operator for cached counts

wezrule:
Uses the RocksDB merge operator which more effectively increments/decrements the `cache_counts` over the current `get`, update and `put` method which requires more IO operations.


wezrule:

PR #2854, opened 2020-07-24
[RocksDB] Use estimated unchecked count

wezrule:
There are some performance implications when checking whether an unchecked key exists in the block processor. We don't use the unchecked count for anything "exact" so an approximation seems fine.


nano/nano_node/entry.cpp
```diff
@@ -1009,16 +1011,15 @@ int main (int argc, char * const * argv)
...
 			}
-			// Waiting for final transaction commit
+
 			uint64_t block_count (0);
 			while (block_count < max_blocks + 1)
 			{
 				std::this_thread::sleep_for (std::chrono::milliseconds (10));
-				auto transaction (node->store.tx_begin_read ());
-				block_count = node->store.block_count (transaction);
+				block_count = node->ledger.cache.block_count;
```
SergiySW: This is waiting for transaction commit, not cache. Cache is checked before

nano/nano_node/entry.cpp
```diff
@@ -1806,16 +1813,14 @@ int main (int argc, char * const * argv)
...
 				}
 			}
-			// Waiting for final transaction commit
 			uint64_t block_count_2 (0);
 			while (block_count_2 != block_count)
 			{
 				std::this_thread::sleep_for (std::chrono::milliseconds (50));
-				auto transaction_2 (node.node->store.tx_begin_read ());
-				block_count_2 = node.node->store.block_count (transaction_2);
+				block_count_2 = node.node->ledger.cache.block_count;
```
SergiySW: Same here

nano/nano_node/entry.cpp
SergiySW: Entire while loop can be replaces with block_processor.flush () I guess

SergiySW (approved):

PR #2856, opened 2020-07-24
[RocksDB] Use different column family options and prefix extraction for unchecked table

wezrule:
Optimizations for all the different tables we have are made. For instance with the unchecked table, there are a lot of deletions expected, this creates "tombstones" which degrade read performance. So we are using a lower number of level 0 sst files, and using prefix extraction for the keys to improve iterating performance.


clemahieu (dismissed):

nano/node/rocksdb/rocksdb.cpp
```diff
@@ -111,6 +107,100 @@ void nano::rocksdb_store::open (bool & error_a, boost::filesystem::path const &
...
+		std::shared_ptr<rocksdb::TableFactory> table_factory (rocksdb::NewBlockBasedTableFactory (get_active_table_options (block_cache_size_bytes * 2)));
+		cf_options = get_active_cf_options (table_factory, memtable_size_bytes);
+	}
+	else if (cf_name_a == "vote")
+	{
+		// No deletes it seems, only overwrites.
+		std::shared_ptr<rocksdb::TableFactory> table_factory (rocksdb::NewBlockBasedTableFactory (get_active_table_options (block_cache_size_bytes * 2)));
+		cf_options = get_active_cf_options (table_factory, memtable_size_bytes);
+	}
+	else if (cf_name_a == "default")
```
clemahieu: Should this be rocksdb::kDefaultColumnFamilyName.c_str ()

clemahieu (dismissed):

nano/node/rocksdb/rocksdb.cpp
wezrule: Yeh makes more sense than hardcoding it.

clemahieu (approved):

PR #2857, opened 2020-07-24
[RocksDB] Add store hint for cached counts

wezrule:
Inside `put()` we always check if the key exists before updating the `cached_counts` database value. The caller has more information about whether the key might already exist, so we can skip this read IO. Added a "store hint" for this purpose.

### DIFF
--- a/nano/node/rocksdb/rocksdb.cpp
+++ b/nano/node/rocksdb/rocksdb.cpp
@@ -230,7 +230,7 @@ void nano::rocksdb_store::version_put (nano::write_transaction const & transacti
 	debug_assert (transaction_a.contains (tables::meta));
 	nano::uint256_union version_key (1);
 	nano::uint256_union version_value (version_a);
-	auto status (put (transaction_a, tables::meta, version_key, nano::rocksdb_val (version_value)));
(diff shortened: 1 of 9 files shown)


nano/node/rocksdb/rocksdb.cpp
```diff
@@ -290,21 +290,25 @@ int nano::rocksdb_store::decrement (nano::write_transaction const & transaction_
...
 	{
-		if (!exists (transaction_a, table_a, key_a))
+		if (store_hint_a == store_hint::key_not_exists || (store_hint_a == store_hint::none && !exists (transaction_a, table_a, key_a)))
 		{
 			// Adding a new entry so counts need adjusting
 			increment (transaction_a, tables::cached_counts, rocksdb_val (rocksdb::Slice (table_to_column_family (table_a)->GetName ())), 1);
 		}
+		else
+		{
+			debug_assert (store_hint_a == store_hint::key_exists && exists (transaction_a, table_a, key_a));
```
guilhermelawless, line 310: If `store_hint::none` (no hint given) and key does exist it seems like this would assert

wezrule:

PR #2862, opened 2020-07-28
Don't crash using read-only CLI command with no ledger

wezrule:
It now correctly outputs "Genesis block not found. Make sure the node network ID is correct." which indicates that the ledger either doesn't exist or the `data_path` is set incorrectly, when using read-only CLI commands like `--debug_block_count`

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -66,6 +66,15 @@ TEST (node, block_store_path_failure)
 	node->stop ();
 }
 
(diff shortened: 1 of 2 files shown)


guilhermelawless (dismissed):
Prevents the crash, should we consider also noting on the message that the data_path may be incorrect or the database doesn't exist yet (must run daemon first)?

guilhermelawless (approved):

PR #2865, opened 2020-07-29
Move all vote generation to the voting thread

guilhermelawless:
Following https://github.com/nanocurrency/nano-node/pull/2827 , this PR makes additional enhancements and consistency checks to the request aggregator, vote generator and vote history cache.

Changes to which threads vote:
- Request aggregator hashes that have no cached votes are now queued to the vote generator instead. The channel is passed along so that reply can follow. These hashes are queued in bulk, and the number of "bulks" is limited to 2048, erasing the oldest values if going beyond that. Candidates queued for broadcasting always have priority.
- All vote generation is now done from the voting thread. This is enforced with an assert (unless in the test network) inside the vote generation method. By reducing the number of interfaces we reduce the probability of a bug related to vote generation.

Checks to grab votes from cache in more situations:
- Added an early cache check when queueing vote candidates from elections.
- Added late cache checks in the vote generation loop, to prevent re-generating votes when they were queued both for broadcasting (a new election) and/or multiple requests. Given the resource usage of vote generation, this might not be that uncommon. Added new stats to track in beta testing.

Local votes cache consistency enhancements:
- Votes cached are now enforced to be unique by account. Since we *now* have a single thread adding to the cache, along with the late cache checks, this should not happen, but the cost to enforce is negligible.
- Expanded the vote cache consistency checks performed in debug builds.

Others:
- Added an early return to `stats:add` if the value is 0, so the caller doesn't have to make that check.

Note the minimal amount of changes to existing tests, which should give some confidence to this approach.


SergiySW (dismissed):

nano/node/voting.cpp
```diff
@@ -32,8 +46,9 @@ void nano::local_vote_history::add (nano::root const & root_a, nano::block_hash
 		}
 	}
 	auto result (history_by_root.emplace (root_a, hash_a, vote_a));
+	(void)result;
 	debug_assert (result.second);
-	debug_assert (std::all_of (history_by_root.equal_range (root_a).first, history_by_root.equal_range (root_a).second, [&hash_a](local_vote const & item_a) -> bool { return item_a.vote != nullptr && item_a.hash == hash_a; }));
+	debug_assert (consistency_check (root_a));
```
clemahieu, line 49: Is there a reason for consistency_check to be ifdef'ed out in non-debug mode? Since debug_assert will already be eliminated in that case. The less preprocessor magic the better.

nano/node/voting.hpp
```diff
@@ -68,31 +73,50 @@ class local_vote_history final
...
+private:
+	using candidate_t = std::pair<nano::root, nano::block_hash>;
+	using request_t = std::pair<std::vector<candidate_t>, std::shared_ptr<nano::transport::channel>>;
+
 public:
-	vote_generator (nano::node_config const & config_a, nano::ledger & ledger_a, nano::wallets & wallets_a, nano::vote_processor & vote_processor_a, nano::local_vote_history & history_a, nano::network & network_a);
+	vote_generator (nano::node_config const & config_a, nano::ledger & ledger_a, nano::wallets & wallets_a, nano::vote_processor & vote_processor_a, nano::local_vote_history & history_a, nano::network & network_a, nano::stat & stats_a);
+	/** Queue items for vote generation, or broadcast votes already in cache */
 	void add (nano::root const &, nano::block_hash const &);
+	/** Queue blocks for vote generation, returning the number of successfull candidates.*/
```
clemahieu: Spelling, 'successful'

nano/node/voting.cpp
guilhermelawless: Good point, removed the preprocessor conditionals

wezrule (approved):

SergiySW (approved):

PR #2871, opened 2020-08-05
Combine alarm and worker threads into a pool

wezrule:
The number of alarm and worker operations can spike at some points which can lead to a backlog as they only run in 1 thread each. This PR combines them and allows delayed tasks and normal ones to be spread across a pool. Just using 2 threads in the thread pool for now so that the number of total threads in the system remains the same, but in future we can combine more of the other threads into this to reduce overall thread count.

Currently it uses the `boost::asio::steady_timer` with the `boost::asio::thread_pool` executor which only has a constructor in boost 1.7, so this will currently fail until we move to 1.7. Some tests are also currently failing so marking this as a draft.

Fixes a TSAN error in `wallets.vote_minimum`


wezrule:
Currently waiting on https://github.com/nanocurrency/nano-node/pull/3085 which should resolve the failing tests

clemahieu (approved):
Nice to see the alarm class go. Deleted code, my favorite thing.

PR #2876, opened 2020-08-10
Improve ledger cache initialization with multiple threads

guilhermelawless:
Got a **10x** speed-up on my system (Ryzen 3600, NVME SSD) on cold initialization, and a small gain on warm initialization.

Checked consistency against RPC *account_count*, *block_count* and some entries from *representatives*.

A system with 4 physical threads scaled well to 40 threads here, so range of values here is [10, 10*hardware_concurrency(), 40] for simplicity.


guilhermelawless:
Last commit moves these parallel methods to the block store, so we can use them in various CLI commands for example. Only read_transaction is accepted and they are suffixed by `_par`

nano/secure/blockstore_partial.hpp
```diff
@@ -834,3 +860,31 @@ class block_predecessor_set : public nano::block_visitor
 	nano::block_store_partial<Val, Derived_Store> & store;
 };
 }
+
+namespace
+{
+template <typename T, typename U, typename V>
+void parallel_traversal (std::function<void(V const &, V const &, bool const)> action)
```
clemahieu: action can be taken by reference instead of copied.

nano/secure/blockstore_partial.hpp
```diff
@@ -834,3 +860,31 @@ class block_predecessor_set : public nano::block_visitor
 	nano::block_store_partial<Val, Derived_Store> & store;
 };
 }
+
+namespace
+{
+template <typename T, typename U, typename V>
```
clemahieu: Are types T and U used?

SergiySW (approved):
Interesting why it's faster while it should be sequential read of accounts list

clemahieu (approved):

PR #2879, opened 2020-08-11
Packed timestamp

clemahieu:
This adds a class that generates a 64-bit timestamp.

The timestamp is a combination of:
* A 44-bit clock portion which is the number of milliseconds since the UNIX epoch when measured in UTC
* A 20-bit monotonic counter for each sample within the same millisecond window.


nano/lib/timestamp.hpp
```diff
@@ -0,0 +1,83 @@
...
+	}
+	static uint64_t now_base ()
+	{
+		uint64_t ms_since_epoch (std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now ().time_since_epoch ()).count ());
+		uint64_t result (timestamp_from_ms (ms_since_epoch));
+		return result;
+	}
+	uint64_t now ()
+	{
+		static_assert (time_bits + count_bits == 64);
```
guilhermelawless: CI has been updated in `develop` with cpp17 which is needed to support this no-message version of `static_assert`

guilhermelawless:
File `lib/timestamp.cpp` is marked for adding but has nothing inside

SergiySW:
Sounds like last date can be 06/23/2527

nano/core_test/ledger.cpp
```diff
@@ -852,7 +852,7 @@ TEST (votes, add_existing)
 	ASSERT_EQ (nano::vote_code::vote, node1.active.vote (vote1));
 	// Block is already processed from vote
 	ASSERT_TRUE (node1.active.publish (send1));
-	ASSERT_EQ (1, election1.election->votes ()[nano::dev_genesis_key.pub].sequence);
+	ASSERT_EQ (1, election1.election->last_votes[nano::dev_genesis_key.pub].timestamp);
```
SergiySW, line 856: I think better use `votes ()` that has election mutex lock. Same below

nano/node/voting.cpp
```diff
@@ -280,12 +281,9 @@ void nano::vote_generator::vote (std::vector<nano::block_hash> const & hashes_a,
 {
 	debug_assert (hashes_a.size () == roots_a.size ());
 	std::vector<std::shared_ptr<nano::vote>> votes_l;
-	{
-		auto transaction (ledger.store.tx_begin_read ());
-		wallets.foreach_representative ([this, &hashes_a, &transaction, &votes_l](nano::public_key const & pub_a, nano::raw_key const & prv_a) {
-			votes_l.emplace_back (this->ledger.store.vote_generate (transaction, pub_a, prv_a, hashes_a));
-		});
-	}
+	wallets.foreach_representative ([this, &hashes_a, &votes_l](nano::public_key const & pub_a, nano::raw_key const & prv_a) {
```
SergiySW, line 284: Should it capture `timestamps`?

nano/node/request_aggregator.hpp
```diff
@@ -59,7 +60,7 @@ class request_aggregator final
 	// clang-format on
 
 public:
-	request_aggregator (nano::network_constants const &, nano::node_config const & config, nano::stat & stats_a, nano::vote_generator &, nano::local_vote_history &, nano::ledger &, nano::wallets &, nano::active_transactions &);
+	request_aggregator (nano::timestamp_generator & timestamps_a, nano::network_constants const &, nano::node_config const & config, nano::stat & stats_a, nano::vote_generator &, nano::local_vote_history &, nano::ledger &, nano::wallets &, nano::active_transactions &);
```
SergiySW: Are timespamps not used here?

nano/node/voting.cpp
clemahieu: timestamps is accessed via this capture.

nano/node/request_aggregator.hpp
clemahieu: Fixed.

SergiySW (dismissed):

SergiySW (dismissed):

SergiySW (approved):

Issue #1094, opened 2018-08-23
Support Ledger Pruning


rkeene:
As a part of node stratification (#1092) not all nodes will need the full ledger, and a mechanism will need to be created to determine which parts of the ledger can be safely discarded on each node.  This mechanism will be called pruning.

More details will be added as time goes on.

Omarino99:
Are you also thinking of a mechanism to make pending deposits compatible with pruning?

rkeene:
All mechanisms that involve pruning must not break the ledger.  This includes not pruning the sending block for blocks that have no corresponding receiving block but which could in the future (i.e., we are free to locally prune sends to address 0, since there can be no receive block generated from that address).

Omarino99:
Has the core team ever discussed the possibility of making such that the receive block is automatically generated after its send block, by making changes in the protocol?

Omarino99:
@rkeene on the roadmap I find this "Ledger Pruning allows for a much smaller database size. This is done through pruning off blocks other than the frontier (head block) +1 or 2 previous for each account. This feature is dependent on Universal Blocks as they create a snapshot of an account on a per block basis requiring fewer blocks to verify an account.". Isn't this just plain wrong? It would work only on certain accounts. Everyone who doesn't know the protocol believes that it would be possible to prune all accounts, while it's not possible.

rkeene:
It can be either considered incorrect or incomplete.

Considering it incomplete is the most charitable interpretation, so we shall proceed under that interpretation.  The most likely reason it is incomplete is because the author elided the detail that pending information would also be included in the ledger for simplicity or simply failed to consider this additional detail or simply thought it was implied that referable blocks could not be elided.

Ultimately the number of pending blocks are they can go down over time, so they are a very minor factor regarding ledger pruning.

I think the best thing we can do regarding this information is update the roadmap documentation to be more specific.

funkspock:
@kaiquenunes23 account creation and transaction creation is controlled/limited by PoW.

@rkeene Ledger pruning will prune off blocks other than the frontier (head block) +1 or 2 previous for each account. Which to my understanding **[Limits history of transactions]**
1. In addition to this, will pruning take place (periodically?) for accounts below a certain threshold (dust accounts)? or like other coins require that "dust accounts" be topped first before they can be used in transactions (for example XRP requires a base reserve of 20). **[Limits number of accounts]**

[...]

PR #2881, opened 2020-08-13
Experimental ledger pruning

SergiySW:
This is the initial, experimental implementation of pruning, related to https://github.com/nanocurrency/nano-node/issues/1094. Node ledger pruning is designed to safely discard parts of the ledger. Now ledger size can be reduced from 23.8 GB to 7.5GB after vacuum. And from 18.8 GB to 5.2 GB after database rebuild.

**Configuration and considerations**
- This version of ledger pruning is for experimental use only:
- Enable optional ledger pruning with node CLI `--debug_prune` for manual pruning & node launch flag `--enable_pruning` for automatic node background actions.
- Pruning rules can be configured in [node.experimental] config with “max_pruning_age” (Time in seconds to limit for blocks age after pruning. Also period for automatic node pruning task, default 1 day, 5 minutes for beta network) & “max_pruning_depth” (Limit for full blocks in chain after pruning, default 0 - unlimited). By default all confirmed blocks older than 1 day can be pruned, except genesis block & last confirmed block in each account chain.
- Last confirmed block in account chain & all unconfirmed blocks cannot be pruned from ledger. Pruning is relying on confirmation height, so confirmation height reset or too low online_weight_minimum config can cause issues.
- Pruned blocks hashes are located in database table “pruned” & can be used by node internally (https://github.com/nanocurrency/nano-node/pull/2946).

**RPC, CLI and other changes**
- Several RPC calls are modified to support pruned blocks related to main request block (https://github.com/nanocurrency/nano-node/pull/2977).
- RPC “history”: if previous block is pruned, “amount” field is not returned for all blocks, “account” field is not returned for state blocks, also state block type/subtype is “unknown” . If source is pruned, then “account” field is not returned for receive/open legacy & for receive/open state blocks (https://github.com/nanocurrency/nano-node/pull/2977).
- Same as above for QT wallet history (https://github.com/nanocurrency/nano-node/pull/2977).
- RPC “block_info”/“blocks_info”: if previous block is pruned, “amount” field is not returned. RPC “block_count”: for nodes with enabled pruning returns 2 new fields: “full” blocks & “pruned” blocks. “full” + “pruned” = total “blocks” (https://github.com/nanocurrency/nano-node/pull/2977).
- Block count for CLI/RPC/telemetry is sum of full blocks & pruned blocks (https://github.com/nanocurrency/nano-node/pull/2977).
- RPC "pruned_exists" to chekc pruned block existence in local database.

**Other notes**
- Node is aggressively pruning ledger during initial bootstrap (until hardcoded block count) to limit node disk space usage, pruning config settings are ignored. If you want to avoid this behavior, finish full bootstrap before pruning start.
-  `--enable_pruning` flag & config option “enable_voting” are mutually exclusive to prevent representatives start with pruned ledger. Pruning is not available for any representatives (https://github.com/nanocurrency/nano-node/pull/2947).
- Pruned node can serve as bootstrap server for remaining blocks in ledger (https://github.com/nanocurrency/nano-node/pull/2976).
- Multiple internal functions are modified to use pruned blocks in block processor (https://github.com/nanocurrency/nano-node/pull/2974), internal node wallet actions (https://github.com/nanocurrency/nano-node/pull/2992), confirmation height processor (https://github.com/nanocurrency/nano-node/pull/2978) and rep crawler (https://github.com/nanocurrency/nano-node/pull/2975).
- Development: consider using ledger safe functions for code with possible pruned blocks (https://github.com/nanocurrency/nano-node/pull/2968).

**Additional updates being reviewed**
- Include pruned count in telemetry.
- Improve RocksDB pruned block count calculation.


nano/node/confirmation_height_bounded.cpp
```diff
@@ -78,11 +78,18 @@ void nano::confirmation_height_bounded::process ()
 		auto block = ledger.store.block_get (transaction, current);
 		if (!block)
 		{
-			auto error_str = (boost::format ("Ledger mismatch trying to set confirmation height for block %1% (bounded processor)") % current.to_string ()).str ();
-			logger.always_log (error_str);
-			std::cerr << error_str << std::endl;
+			if (ledger.enable_pruning && ledger.store.pruned_exists (transaction, current))
+			{
+				continue;
```
wezrule: As discussed this can cause infinite loop if the source hash is pruned.

nano/node/node.cpp
```diff
@@ -977,6 +986,125 @@ void nano::node::ongoing_unchecked_cleanup ()
...
+		while (pruning_targets.size () < batch_size_a && !target_finished && !stopped)
+		{
+			target_finished = collect_ledger_pruning_targets (pruning_targets, last_account, batch_size_a * 2, max_depth, cutoff_time);
+		}
+		// Pruning write operation
+		transaction_write_count = 0;
+		if (!pruning_targets.empty () && !stopped)
+		{
+			auto scoped_write_guard = write_database_queue.wait (nano::writer::pruning);
+			auto write_transaction (store.tx_begin_write ({ tables::blocks, tables::pruned }));
```
wezrule, line 1058: This is the only place that writes pruned block? Could be put into the second argument which doesn't require locking (with RocksDB), i.e `store.tx_begin_write ({ tables::blocks},{ tables::pruned }))`

nano/node/rocksdb/rocksdb.cpp
```diff
@@ -110,6 +110,18 @@ void nano::rocksdb_store::open (bool & error_a, boost::filesystem::path const &
 			logger.always_log (boost::str (boost::format ("The version of the ledger (%1%) is too high for this node") % version_l));
 		}
 	}
+
+	// Exact pruned block count
+	if (!error_a && !enable_pruning_a)
```
wezrule: I'm not sure this is working correctly :(, will investigate further

nano/secure/blockstore_partial.hpp
```diff
@@ -552,9 +552,46 @@ class block_store_partial : public block_store
 		release_assert (success (status));
 	}
 
+	void pruned_put (nano::write_transaction const & transaction_a, nano::block_hash const & hash_a) override
+	{
+		// Disabled path if TSAN/ASAN/Valgrind complain
+		//nano::db_val<Val> zero (static_cast<uint8_t> (0));
+		//auto status = put (transaction_a, tables::pruned, hash_a, zero);
```
wezrule: Commented out?

nano/secure/blockstore_partial.hpp
```diff
@@ -610,6 +647,26 @@ class block_store_partial : public block_store
...
+			existing = make_iterator<nano::block_hash, nano::db_val<Val>> (transaction_a, tables::pruned);
+		}
+		if (existing != end)
+		{
+			return existing->first;
+		}
+		else
+		{
+			return 0;
+		}
```
wezrule: Could remove the `else` and just have `return 0;` at end, or even use a ternary operator `? :` with single return statement

nano/secure/ledger.cpp
```diff
@@ -905,6 +938,35 @@ nano::block_hash nano::ledger::block_source (nano::transaction const & transacti
...
+{
+	if (!enable_pruning)
+	{
+		auto block (store.block_random (transaction_a));
+		return std::make_pair (block->hash (), block->root ());
+	}
+	else
+	{
+		nano::block_hash hash (0);
+		nano::block_hash root (0);
```
wezrule: What do you think about moving `hash`/`root` out, setting them in `if` block too and having single `return` at end of function?

nano/secure/ledger.cpp
```diff
@@ -1202,6 +1293,39 @@ bool nano::ledger::block_confirmed (nano::transaction const & transaction_a, nan
 	return confirmed;
 }
 
+uint64_t nano::ledger::prune (nano::write_transaction & transaction_a, nano::block_hash const & hash_a, uint64_t const bath_size_a)
```
wezrule: `bath_size_a` typo should be `batch_size_a`

nano/node/node.cpp
```diff
@@ -977,6 +986,125 @@ void nano::node::ongoing_unchecked_cleanup ()
...
+		}
+		else
+		{
+			++i;
+		}
+	}
+	return !finish_transaction || last_account_a.is_zero ();
+}
+
+void nano::node::ledger_pruning (uint64_t const batch_size_a, bool bootstrap_weight_reached, bool log_to_cout)
```
wezrule: should have consistent `_a` for all arguments

nano/secure/ledger.cpp
SergiySW: Thanks!

nano/node/confirmation_height_bounded.cpp
SergiySW: Fixed with your commit :)

nano/node/node.cpp
SergiySW: Updated

nano/secure/blockstore_partial.hpp
SergiySW: It sounds like null insert value is OK for sanitizers as well, removing

nano/secure/blockstore_partial.hpp
SergiySW: Updated

nano/secure/ledger.cpp
SergiySW: Updated

guilhermelawless:
QT wallet could add "(Pruned)" in status text, also explain in tooltip that amounts and source accounts can be wrong.

nano/secure/ledger.hpp
```diff
@@ -67,6 +75,7 @@ class ledger final
 	std::atomic<size_t> bootstrap_weights_size{ 0 };
 	uint64_t bootstrap_weight_max_blocks{ 1 };
 	std::atomic<bool> check_bootstrap_weights;
+	std::atomic<bool> enable_pruning{ false };
```
guilhermelawless: Does it need too be atomic? Is it not always modified in single thread (node init) or tests?

guilhermelawless: Maybe short-handing to just `pruning` as well?

nano/node/node.cpp
```diff
@@ -977,6 +1004,125 @@ void nano::node::ongoing_unchecked_cleanup ()
 	});
 }
 
+bool nano::node::collect_ledger_pruning_targets (std::deque<nano::block_hash> & pruning_targets_a, nano::account & last_account_a, uint64_t const batch_read_size_a, uint64_t const max_depth_a, uint64_t const cutoff_time_a)
```
guilhermelawless, line 984: This can now be written with the parallelized table loop, but maybe not worth it if the time to read the table is much less than the time to delete.

nano/secure/ledger.hpp
SergiySW: Updated

nano/node/node.cpp
SergiySW: I think with frequent pruning deleting can take much less time while search would be long

wezrule (dismissed):

wezrule (dismissed):

wezrule (approved):

PR #2883, opened 2020-08-15
Remove conf height put during ledger processing on a newly opened account

wezrule:
When an account is opened the confirmation height database adds the account with height/frontier of 0. This is unnecessary as the lack of this existing means these values are implicitly 0. This PR makes that change. 

This allows us to set `allow_concurrent_memtable_writes` to false in RocksDB and enables use of the `HashSkipListFactory` for memtables enabling faster point reads. This does mean that we can no longer write multiple independent things in RocksDB, but we don't currently do that with anything else.

In the future this could enable a separate LMDB database store to be used to get better block processing/conf height performance with LMDB because they currently block each other on writes and could also remove the write database queue (although it's useful for teats still). This is not done here.

Added a test for https://github.com/nanocurrency/nano-node/pull/2777 `unconfirmed frontiers`
Clearing confirmation height now just removes the whole table rather than reset every account individually.


nano/secure/ledger.cpp
```diff
@@ -1257,7 +1253,6 @@ void nano::ledger::update_account (nano::write_transaction const & transaction_a
 	}
 	else
 	{
-		store.confirmation_height_del (transaction_a, account_a);
```
SergiySW, line 1282: Why not remove it?

wezrule: There will only be a conf height entry for this account if the open block is cemented, so there should never be a scenario where this account is being rolled back so no need to try delete from conf height table as it shouldn't be there. I will add a `debug_assert` though

SergiySW: Confirmation height can be manually cleared I guess

wezrule: Yeh it should delete the conf height entry, good spot!

SergiySW (approved):

PR #2884, opened 2020-08-18
Confirm delta

clemahieu:
online_weight_minimum is already considered in the calculation of node::delta. Using online_weight_minimum directly can cause quorum requirements to be unnecessarily high as actual online vote weight approaches online_weight_minimum.

Replacing references to online_weight_minimum with node:delta and adding tests to ensure rollbacks and confirmation is done precisely on the correct conditions.


guilhermelawless:
Spots that might also need a change (or at least can be optimized) as part of this:
- `peers_stake_required` in the `confirmation_quorum` RPC response
- `gap_cache::bootstrap_check` could possibly use this threshold and not online weight minimum
- Similarly for `active_transactions::inactive_votes_bootstrap_check` , for legacy bootstrap when lazy is disabled

Srayman:
This handles the case where online weight is reduced and approaches the minimum, but what happens in the opposite scenario where a large rep comes back online or a large pending balance is received?  The peers_stake would increase immediately but online_weight is trended over 2 weeks so there would be a lag before delta is updated to account for the new weight.

Should delta be the greater of 50% of online weight or 50% of peers stake to account for the rapid increase in peers weight?  As the weight is reduced delta would slowly go down, but it would respond more dramatically to a rapid increase in weight since peers weight is real-time.

Alternatively modify the online weight process to move up quickly if the weight is increased but down slowly over 2 weeks if it decreases.

clemahieu:
> This handles the case where online weight is reduced and approaches the minimum, but what happens in the opposite scenario where a large rep comes back online or a large pending balance is received? The peers_stake would increase immediately but online_weight is trended over 2 weeks so there would be a lag before delta is updated to account for the new weight.
> 
> Should delta be the greater of 50% of online weight or 50% of peers stake to account for the rapid increase in peers weight? As the weight is reduced delta would slowly go down, but it would respond more dramatically to a rapid increase in weight since peers weight is real-time.
> 
> Alternatively modify the online weight process to move up quickly if the weight is increased but down slowly over 2 weeks if it decreases.

@Srayman  Added some new changes to address this issue.

nano/core_test/election.cpp
```diff
@@ -13,3 +13,155 @@ TEST (election, construction)
...
+	nano::block_builder builder;
+	std::shared_ptr<nano::block> send1 = builder.state ()
+	                                     .account (nano::dev_genesis_key.pub)
+	                                     .previous (nano::genesis_hash)
+	                                     .representative (nano::dev_genesis_key.pub)
+	                                     .balance (node1.online_reps.delta ())
+	                                     .link (key1.pub)
+	                                     .work (0)
+	                                     .sign (nano::dev_genesis_key.prv, nano::dev_genesis_key.pub)
+	                                     .build ();
```
guilhermelawless: Can use `.build_shared()` on these which allows replacing `std::shared_ptr<nano::block>` with `auto`

nano/node/vote_processor.hpp
```diff
@@ -33,7 +33,7 @@ namespace transport
 class vote_processor final
 {
 public:
-	explicit vote_processor (nano::signature_checker & checker_a, nano::active_transactions & active_a, nano::node_observers & observers_a, nano::stat & stats_a, nano::node_config & config_a, nano::node_flags & flags_a, nano::logger_mt & logger_a, nano::online_reps & online_reps_a, nano::ledger & ledger_a, nano::network_params & network_params_a);
+	explicit vote_processor (nano::node & node_a);
```
guilhermelawless: Many class forward declarations can be removed in this file, and `node` should be declared. Does this change to reference `node` go against our efforts to not have `node` be referenced by other classes to reduce compilation time?

clemahieu: I think this was some cruft left over from building the PR. At any rate it's an unrelated change so removed.

nano/node/online_reps.cpp
```diff
@@ -1,101 +1,131 @@
...
 }
 
-nano::uint128_t nano::online_reps::trend (nano::transaction & transaction_a)
+nano::uint128_t nano::online_reps::calculate_trend (nano::transaction & transaction_a) const
 {
 	std::vector<nano::uint128_t> items;
-	items.reserve (network_params.node.max_weight_samples + 1);
-	items.push_back (minimum);
+	items.push_back (config.online_weight_minimum.number ());
+	items.reserve (config.network_params.node.max_weight_samples + 1);
```
wezrule: Most implementations tend to give `std::vector` a default size of 0, so could reduce an allocation swapping these lines back (i.e `reserve` first)

nano/node/vote_processor.cpp
```diff
@@ -15,16 +15,24 @@
...
 #include <boost/format.hpp>
 
-nano::vote_processor::vote_processor (nano::node & node_a) :
-node{ node_a },
-max_votes (node_a.flags.vote_processor_capacity),
-started{ false },
-stopped{ false },
-is_active{ false },
-thread{ [this]() {
+nano::vote_processor::vote_processor (nano::signature_checker & checker_a, nano::active_transactions & active_a, nano::node_observers & observers_a, nano::stat & stats_a, nano::node_config & config_a, nano::node_flags & flags_a, nano::logger_mt & logger_a, nano::online_reps & online_reps_a, nano::ledger & ledger_a, nano::network_params & network_params_a) :
```
wezrule: Nice job removing the `node` dependency! Should be able to remove the `node.hpp` include now too

nano/node/online_reps.cpp
```diff
@@ -1,101 +1,131 @@
...
 
 void nano::online_reps::observe (nano::account const & rep_a)
 {
 	if (ledger.weight (rep_a) > 0)
 	{
+		auto now = std::chrono::steady_clock::now ();
 		nano::lock_guard<std::mutex> lock (mutex);
-		reps.insert (rep_a);
+		auto new_insert = reps.get<tag_account> ().erase (rep_a) == 0;
+		reps.insert ({ now, rep_a });
```
wezrule: Should `now` be captured inside the critical section in case it takes a while to grab the mutex?

SergiySW: Is emplace () better here?

nano/node/online_reps.cpp
```diff
@@ -1,101 +1,131 @@
...
 	}
 }
 
 void nano::online_reps::sample ()
 {
-	auto transaction (ledger.store.tx_begin_write ({ tables::online_weight }));
-	// Discard oldest entries
-	while (ledger.store.online_weight_count (transaction) >= network_params.node.max_weight_samples)
+	nano::unique_lock<std::mutex> lock (mutex);
+	nano::uint128_t online_l;
```
SergiySW: Can be `nano::uint128_t online_l (online_m);` or `nano::uint128_t online_l = online_m;` probably

nano/node/online_reps.cpp
clemahieu: Emplace can't be used with POD type initialization.

wezrule (approved):

SergiySW (approved):

PR #2885, opened 2020-08-18
Add cemented frontier successor confirmation in request loop

wezrule:
This adds the following functionality:
- New flag in elections `is_optimistic` which is set to true for any elections inserted from the optimistic frontiers confirmation. New confirm_req rounds are sent every 10 seconds (as opposed to 5 for normal elections), and expire after 1 minute
- The frontiers confirmation procedure now goes: wallet frontiers -> frontiers -> expired cemented frontier successors. The last one is new and is to confirm blocks at cemented height + 1 as the network has likely not confirmed this frontier (due to sequential voting) so it is not possible to confirm blocks for this account without backtracking. If it is a receive block the source must be confirmed otherwise an election is not attempted, there is no complex dependency traversal done or any prioritisation of accounts, it simply iterates from the beginning of the accounts database in sorted order and starts elections for ones that it can. The pessimistic elections are not set to active immediately.

(Unrelated) - Moved a comment I noticed was misplaced in `confirmation_height_bounded.cpp`


nano/node/active_transactions.cpp
```diff
@@ -297,16 +336,121 @@ void nano::active_transactions::frontiers_confirmation (nano::unique_lock<std::m
...
+				{
+					auto source{ block->source () };
+					if (source.is_zero ())
+					{
+						source = block->link ();
+					}
+
+					auto should_start_election{ false };
+
+					// Check if the source is confirmed (if open/receive) before trying to confirm this block
```
clemahieu: Can we reuse can_vote here? Possibly renaming that function to be more descriptive? That should be checking existence/confirmation height of dependencies.

Srayman:
It can sometimes take 2+ rounds to complete bootstrapping frontier confirmations as it's limited to just the top 30 reps each round.  If I'm reading this correctly it now extends frontiers to 1 minute in between confirm_req that could extend the frontier confirmation process quite a bit for initial bootstrap.  Also as the representatives become more decentralized the number of reps required for quorum will continue to increase.  Right now on the live network the top 30 reps make up 66.59M weight which is quite close to the required minimum 60M.  Even though the minimum will be adjusted https://github.com/nanocurrency/nano-node/pull/2884 to be lower and more dynamic there's still the risk of the required reps needing to be more than 30.

Potential ideas
1. Allow 2 rounds before delaying to 1 minute intervals
2. Increase max reps above 30
3. Dynamically adjust max reps to account for the number that are required to reach 2/3 majority for example, to account for potential network or communication issues from some nodes.

nano/node/active_transactions.cpp
wezrule: So I originally didn't use it because we already know that the previous dependent is already confirmed (comes from cemented frontier), so it seems a waste to check it again. However it does better encapsulate the logic, rather than try to change it to fit the needs here I'm just gonna use it. Changed name to `ledger::dependents_confirmed` and also replaced contents with `block_confirmed` (could possibly use `block_confirmed_or_confirming` too but this at least matches behaviour which is currently there). Thanks.

wezrule:
@Srayman made some good points. After some discussion with him we decided that optimistic elections would do 5 second requests for 1 minute then once a minute until the 5 minute election expiration time is reached. Also will wait 5 seconds before starting pessimistic elections to allow nodes enough time to generate and publish votes if they're all moving forward with the same blocks.

nano/node/active_transactions.hpp
```diff
@@ -245,14 +260,33 @@ class active_transactions final
 		mi::ordered_non_unique<mi::tag<tag_uncemented>,
 			mi::member<nano::cementable_account, uint64_t, &nano::cementable_account::blocks_uncemented>,
 			std::greater<uint64_t>>>>;
+
+	boost::multi_index_container<nano::expired_optimistic_election_info,
+	boost::multi_index::indexed_by<
+		boost::multi_index::ordered_non_unique<boost::multi_index::tag<tag_expired_time>,
+			boost::multi_index::member<expired_optimistic_election_info, std::chrono::steady_clock::time_point, &expired_optimistic_election_info::expired_time>>,
+		boost::multi_index::hashed_unique<boost::multi_index::tag<tag_account>,
+			boost::multi_index::member<expired_optimistic_election_info, nano::account, &expired_optimistic_election_info::account>>>>
```
guilhermelawless: Can use the alias `mi` here for `boost::multi_index`

nano/node/election.hpp
```diff
@@ -69,9 +69,10 @@ class election final : public std::enable_shared_from_this<nano::election>
 	void generate_votes (nano::block_hash const &);
 	void remove_votes (nano::block_hash const &);
 	std::atomic<bool> prioritized_m = { false };
+	bool optimistic_m = { false };
```
guilhermelawless: Thoughts on making this an enum `election_behavior {default, optimistic}` ? Feels better passing that around and not a bool to election insertion methods in active_transactions.

nano/node/active_transactions.cpp
```diff
@@ -256,6 +280,34 @@ void nano::active_transactions::request_confirm (nano::unique_lock<std::mutex> &
 		bool const overflow_l (unconfirmed_count_l > node.config.active_elections_size && election_l->election_start < election_ttl_cutoff_l && !node.wallets.watcher->is_watched (i->root));
 		if (overflow_l || election_l->transition_time (solicitor))
 		{
+			if (election_l->failed () && election_l->optimistic ())
```
guilhermelawless: We've had this function growing in size in the past and it would be nice to keep it tighter, this looks like a good candidate to extract (although the number of functions in active_transactions is astronomic 😅 )

nano/node/active_transactions.cpp
```diff
@@ -297,16 +349,89 @@ void nano::active_transactions::frontiers_confirmation (nano::unique_lock<std::m
...
+ * confirming blocks at cemented height + 1 (cemented frontier successor) for an account only if all dependent blocks already
+ * confirmed.
+ */
+void nano::active_transactions::confirm_expired_frontiers_pessimistically (nano::transaction const & transaction_a, uint64_t max_elections_a, uint64_t & elections_count_a)
+{
+	auto i{ node.store.latest_begin (transaction_a, next_frontier_account) };
+	auto n{ node.store.latest_end () };
+	nano::timer<std::chrono::milliseconds> timer (nano::timer_state::started);
+	nano::confirmation_height_info confirmation_height_info;
+	uint64_t elections_count{ 0 };
```
SergiySW: Not used here?

wezrule: Removed

nano/node/active_transactions.cpp
wezrule: Extracted to new function

nano/node/election.hpp
wezrule: Good suggestion

nano/node/active_transactions.hpp
wezrule: Done thanks :)

SergiySW (dismissed):

SergiySW (dismissed):

SergiySW (dismissed):

SergiySW (dismissed):

guilhermelawless (dismissed):

SergiySW (dismissed):

guilhermelawless (dismissed):

guilhermelawless (approved):

clemahieu (approved):

SergiySW (approved):

PR #2886, opened 2020-08-19
Inactive votes election hint

guilhermelawless:
Allows nodes to lazily insert an election that is likely to confirm.

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -657,40 +657,40 @@ TEST (active_transactions, inactive_votes_cache_election_start)
 	nano::send_block_builder send_block_builder;
 	nano::state_block_builder state_block_builder;
 	auto send1 = send_block_builder.make_block ()
-	            .previous (latest)
-	            .destination (key1.pub)
-	            .balance (nano::genesis_amount - 2000 * nano::Gxrb_ratio)
-	            .sign (nano::dev_genesis_key.prv, nano::dev_genesis_key.pub)
-	            .work (*system.work.generate (latest))
-	            .build_shared ();
+	             .previous (latest)
+	             .destination (key1.pub)
+	             .balance (nano::genesis_amount - 2000 * nano::Gxrb_ratio)
+	             .sign (nano::dev_genesis_key.prv, nano::dev_genesis_key.pub)
+	             .work (*system.work.generate (latest))
+	             .build_shared ();
 	auto send2 = send_block_builder.make_block ()
-	            .previous (send1->hash ())
-	            .destination (key2.pub)
-	            .balance (nano::genesis_amount - 4000 * nano::Gxrb_ratio)
-	            .sign (nano::dev_genesis_key.prv, nano::dev_genesis_key.pub)
-	            .work (*system.work.generate (send1->hash ()))
-	            .build_shared ();
+	             .previous (send1->hash ())
+	             .destination (key2.pub)
+	             .balance (nano::genesis_amount - 4000 * nano::Gxrb_ratio)
+	             .sign (nano::dev_genesis_key.prv, nano::dev_genesis_key.pub)
+	             .work (*system.work.generate (send1->hash ()))
+	             .build_shared ();
 	auto send3 = send_block_builder.make_block ()
-	            .previous (send2->hash ())
-	            .destination (key3.pub)
-	            .balance (nano::genesis_amount - 6000 * nano::Gxrb_ratio)
-	            .sign (nano::dev_genesis_key.prv, nano::dev_genesis_key.pub)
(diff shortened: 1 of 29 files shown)


nano/core_test/active_transactions.cpp
```diff
@@ -646,6 +646,164 @@ TEST (active_transactions, inactive_votes_cache_multiple_votes)
 	ASSERT_EQ (2, node.stats.count (nano::stat::type::election, nano::stat::detail::vote_cached));
 }
 
+TEST (active_transactions, inactive_votes_cache_election_start)
+{
+	nano::system system (1);
```
wezrule: Doesn't seem like you need the node created by the system constructor

guilhermelawless: Thanks, removed it

SergiySW (dismissed):

nano/core_test/active_transactions.cpp
```diff
@@ -646,6 +646,164 @@ TEST (active_transactions, inactive_votes_cache_multiple_votes)
...
+	// A late block arrival also checks the inactive votes cache
+	ASSERT_TRUE (node.active.empty ());
+	auto send7_cache (node.active.find_inactive_votes_cache (send7->hash ()));
+	ASSERT_EQ (6, send7_cache.voters.size ());
+	ASSERT_TRUE (send7_cache.status.bootstrap_started);
+	ASSERT_TRUE (send7_cache.status.confirmed);
+	ASSERT_TRUE (send7_cache.status.election_started); // already marked even though the block does not exist
+	node.process_active (send6);
+	node.block_processor.flush ();
+	// An election is started for send6 but does not confirm
```
clemahieu, line 799: We can add a check here for cemented count to prove send6 did not confirm.

nano/core_test/toml.cpp
```diff
@@ -404,6 +405,7 @@ TEST (toml, daemon_config_deserialize_no_defaults)
 	network_threads = 999
 	online_weight_minimum = "999"
 	online_weight_quorum = 99
+	election_hint_weight_percent = 19
```
clemahieu, line 408: This is 19 here, and 10 below. Is this supposed to be 10?

nano/node/nodeconfig.hpp
```diff
@@ -56,6 +56,7 @@ class node_config
 	unsigned vote_generator_threshold{ 3 };
 	nano::amount online_weight_minimum{ 60000 * nano::Gxrb_ratio };
 	unsigned online_weight_quorum{ 50 };
+	unsigned election_hint_weight_percent{ 10 };
```
clemahieu, line 59: And 10 here. Are these supposed to be using a common constant in case it needs to be changed and stay in sync in the future?

nano/node/gap_cache.cpp
```diff
@@ -88,29 +88,34 @@ bool nano::gap_cache::bootstrap_check (std::vector<nano::account> const & voters
...
-				{
-					node_l->bootstrap_initiator.bootstrap_lazy (hash_a);
-				}
-				else if (!node_l->flags.disable_legacy_bootstrap)
-				{
-					node_l->bootstrap_initiator.bootstrap ();
-				}
-			}
-		});
+		bootstrap_start (hash_a);
```
clemahieu: Are there any changes here or just extracting a function? Can this be done in a separate PR to simplify this?

nano/node/blockprocessor.cpp
```diff
@@ -307,6 +307,10 @@ void nano::block_processor::process_live (nano::block_hash const & hash_a, std::
 			election.election->try_generate_votes (block_a->hash ());
 		}
 	}
+	else
+	{
+		node.active.check_inactive_votes_cache_election (block_a);
```
clemahieu: I think the work "trigger" makes more sense than "check" because this is taking an action based on status.

nano/node/gap_cache.cpp
SergiySW: Just extraction. Can be separate

nano/core_test/toml.cpp
guilhermelawless: Oh this is a config test to ensure the values are all correctly deserialized, see the values above which are also not default.

nano/node/nodeconfig.hpp
guilhermelawless: 10 is the default and can be configured, no constants necessary I think.

nano/node/gap_cache.cpp
guilhermelawless: Reverted 👍

clemahieu (approved):

SergiySW (approved):

PR #2891, opened 2020-08-21
Remove election fallback behaviors and adjusted multiplier calculations

guilhermelawless:
Removes backtracking and dependency activation. The move to sequential elections also allows removing the adjusted multiplier from chains.

Frontier successor insertion (https://github.com/nanocurrency/nano-node/pull/2885) is the new fallback behavior which does not escalate.


SergiySW (approved):
If we think that dependent elections are not possible anymore

clemahieu (approved):

PR #2893, opened 2020-08-24
Node initialization refactor and cleanup

clemahieu:
This is a major refactor in order to clean up the initialization of nano::node objects and eliminates a significant amount of boilerplate code needed for testing.

The nano::environment class
- Holds shared io objects, work threads, and alarm objects which are shared between nodes in a testing context
- Holds the data_path, named 'path', for all running objects.
- While testing, each node will execute in a unique subdirectory
- Holds code related to processing node_config overrides from the config file and command line arguments.

Test initialization:
- Port and data path determination while in a testing context is made inside the node class which eliminates boilerplate code needed to select unique ports in tests.
- nano::node_config::port and nano::node_config::path are used as overrides to default path and port determination. If they are set to 0/empty, the default node port and nano::working_path() will be used

The nano::node_flags class is merged in to the nano::config class. Both of these classes are related to per-node initialization information and keeping them together simplifies referencing node configuration objects.

The nano::network_params class is renamed to nano::environment_constants, and are shared constants between possibly multiple nodes in a testing context.


clemahieu:
I'll break this up in to smaller commits but I'm interested on feedback about the general design, it seems to simplify everything.

clemahieu:
This PR has been superseded by https://github.com/nanocurrency/nano-node/pull/3420/files

PR #2898, opened 2020-08-28
Docker base update

argakiig:


guilhermelawless (approved):
LGTM, implicit gcc upgrade to gcc7

PR #2899, opened 2020-08-28
Implement state block version 2

wezrule:
This adds 5 new fields to the state block:

1. Height - The height of the block
2. Signer - Who signed the block? Either account owner or epoch currently.
3. Link interpretation - How should the link field be interpreted? either send, receive or noop (change/epoch)
4. Upgrade - Whether this block is upgrading an account, always set for open blocks
5. Version - The version of the block

- Changes to existing blocks updated with `refresh()` now update the block signature and changed function to `rebuild()`
- Encapsulate block::account/block::height with using the sideband too, rather than checking block followed by checking sideband if it doesn't exist.
- Epoch upgrades not required to be sequential anymore
- A new function `simple_block_validation ()` is used in various places which can do some rudimentary checks on state v2 blocks, anything which doesn't require ledger processing (of previous/source blocks for instance).
- Ledger - epoch_link has a bit of a different meaning now. Self-signed epoch blocks are allowed, and epoch links were, but now the epoch link can be an account (for self-signed epochs), so passing the whole block in to a new function `has_epoch_link` which can determine if the link field belongs to an epoch link or a self-signed epoch.
- Updated flatbuffer definitions and block builder to support state v2 blocks.
- RPC - block_create, if "type" == "state2" the only required field is "link_interpretation" (being either "send", "receive" or "noop"). The rest of the fields can be implicitly determined assuming the previous block is in the ledger, if using a pruned node this will only work for frontier blocks. Also `"version"` has been changed to `"work_version"`.

(Bug) - The public key used when verifying a send to an epoch link is the epoch account, so it will always fail signature validation currently

TSAN run was clean.

I have made a diagram of the core test `ledger.state_blocks_v2` here:
State v2 (1)


nano/secure/ledger.cpp
```diff
@@ -1044,18 +1208,53 @@ bool nano::ledger::can_vote (nano::transaction const & transaction_a, nano::bloc
...
+	{
+		auto state_block = dynamic_cast<nano::state_block const *> (&block_a);
+		if (state_block)
+		{
+			return (state_block->hashables.flags.link_interpretation () == nano::link_flag::noop && block_a.link () == block_a.account ());
+		}
+	}
+
+	return false;
 }
```
cryptocode: This PR nicely fixes some dynamic_cast's in ipc::flatbuffers_builder, and it would be good to avoid it here as well.

What do you think about making `has_epoch_link` and `is_self_signed_epoch` polymorphic `block` functions? The default implementation would return the non-state2 case, with state_block overriding it.

The reason I'm asking is that `dynamic_cast` has shown up in profiling on some compilers in the past (like #831). I imagine these functions will be called frequently.

It can be argued these doesn't make sense for non-state blocks, but the ledger functions are called on all block types anyway, so I think that's fine (and possibly the lesser code smell vs dynamic_cast). Maybe rename `is_self_signed_epoch` to `has_self_signed_epoch` if the move is done.

In addition to improved performance, this should lead to a tad less code, and there's one less branch for checking the dynamic_cast. For `has_epoch_link`, network_params can be passed in as an argument, as having it as a static local adds synchronization overhead.

api/flatbuffers/nanoapi.fbs
```diff
@@ -93,6 +93,7 @@ enum BlockSubType: byte {
 /** Block union */
 union Block {
 	BlockState,
+	BlockState2,
```
cryptocode: Strictly speaking, this should be appended at the end of the union to ensure backwards compatibility, but I don't think fb is used yet, so maybe it's fine for now?

wezrule: Good point, could cause some confusion in the future if we add another state block version and just append it in the middle so will move to end.

nano/secure/ledger.cpp
wezrule: I guess they could have been `static_cast` too, but your idea of making them members of blocks helps with avoiding downcasts, updated.

rotilho:
What's the plan for this change? Will it be included in v22 or v23?

wezrule:
> What's the plan for this change? Will it be included in v22 or v23?

It won't be in v22, on the right it states the milestone as v23 but it's possible to change.

rotilho:
Crossing my fingers to be included in v23, this is one of the most positive additions IMO. I can see sooo many optimization after this change is deployed in the live net

Great job guys and @wezrule

PR #2903, opened 2020-09-01
Provide network_receive_minimum in RPC/WS active_difficulty

guilhermelawless:
On mainnet I get the following response:

```json
{
    "multiplier": "1.5",
    "network_current": "fffffffaaaaaaaab",
    "network_minimum": "fffffff800000000",
    "network_receive_current": "fffffff07c1f07c2",
    "network_receive_minimum": "fffffe0000000000"
}
```

The multiplier between `network_receive_minimum` and `network_receive_current` is 33, plugging that into our normalization expression `(33 + (64 - 1)) / 64` yields 1.5 as expected.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -4529,6 +4529,21 @@ TEST (node, deferred_dependent_elections)
 }
 }
 
+TEST (node, default_difficulty)
+{
+	nano::system system (1);
+	auto & node (*system.nodes[0]);
+	auto const & thresholds = nano::network_params{}.network.publish_thresholds;
+	ASSERT_EQ (thresholds.epoch_1, node.default_difficulty (nano::work_version::work_1));
+	ASSERT_EQ (thresholds.epoch_1, node.default_receive_difficulty (nano::work_version::work_1));
+	nano::upgrade_epoch (system.work, node.ledger, nano::epoch::epoch_1);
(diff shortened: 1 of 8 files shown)


argakiig (dismissed):

SergiySW (dismissed):

PlasmaPower:
How about a `network_receive_current` too? That seems a bit more useful to me, unless current is inaccurate to the degree where it's better to use minimum anyways, which would be a separate issue.

guilhermelawless:
@PlasmaPower thanks for the suggestion and for the original suggestion of adding this to the RPC (was already in the works, just waiting for epoch 2 to finish 👍).

The node applies a function to normalize the prioritization with respect to the block type and even the epoch, so that a receive block doesn't have 10x the priority of a send block with less work generated. Essentially this means `network_current` applies to all blocks with some small margin of error. The normalization function is here: https://github.com/nanocurrency/nano-node/blob/V21.1/nano/lib/work.cpp#L148

guilhermelawless:
@PlasmaPower thanks for clarifying via DM, indeed it is possible to take advantage of the normalization to still provide less work for receive blocks and obtain the same priority. Added `network_receive_current`

argakiig (dismissed):

SergiySW (approved):

PR #2906, opened 2020-09-02
Add support for stackful coroutines in the nano::socket class

clemahieu:
This change adds support for stackful coroutines into the node::socket class.

This removes the write queue in favor of a write counter. The queue is unnecessary because requests are already queued in the socket's strand.

This also removes the socket concurrency component as all the writes are serialized anyway.


guilhermelawless:
To be rebased after #2788

clemahieu:
The windows failures seem due to memory leak. The lifetime management around the current bootstrapping code is really complicated so I'll have to split up the conversion into smaller patches.

clemahieu:
This effort will be pushed off until either boost ASIO / stackful coroutines are stable together or the c++ networking library standardization.

PR #2909, opened 2020-09-02
Add support for defer_lock in nano::unique_lock

guilhermelawless:
Needs NANO_TIMED_LOCKS CMake variable > 0 to test, otherwise it defaults to using std::unique_lock

This is necessary to build since #2901 but not caught by CI due to not using this developer-oriented feature


guilhermelawless:
Modified branch for a CI test here passed: https://github.com/guilhermelawless/nano-node/tree/ci/locks/defer-lock-test

argakiig (approved):

PR #2916, opened 2020-09-07
Fix error with `work_cancel` RPC request

koczadly:
`work_cancel` currently returns no response data when successful, which results in an "Empty response" error being returned. Appears to have been broken from PR #1032.

**This PR adds:**
- Return `{"success": ""}` as the response for `work_cancel`
- Add `nano::error_rpc::empty_response` error code, which is returned for empty responses (previously only an error string was passed, but without an error code).
- Update existing tests which expected an empty response error

Unfortunately I've been unable to test this as I couldn't get boost to link up with VS correctly. Not sure what I'm doing wrong, so you'll have to rely on the CI results.

### DIFF
--- a/nano/core_test/fakes/work_peer.hpp
+++ b/nano/core_test/fakes/work_peer.hpp
@@ -133,6 +133,7 @@ private:
 		std::stringstream ostream;
 		ptree::write_json (ostream, message_l);
 		beast::ostream (response.body ()) << ostream.str ();
+		write_response ();
 	}
 
 	void handle_generate (nano::block_hash const & hash_a)
--- a/nano/core_test/fakes/work_peer.hpp
+++ b/nano/core_test/fakes/work_peer.hpp
@@ -33,7 +33,6 @@ enum class work_peer_type
 class work_peer_connection : public std::enable_shared_from_this<work_peer_connection>
 {
 	const std::string generic_error = "Unable to parse JSON";
-	const std::string empty_response = "Empty response";
 
 public:
 	work_peer_connection (asio::io_context & ioc_a, work_peer_type const type_a, nano::work_version const version_a, nano::work_pool & pool_a, std::function<void(bool const)> on_generation_a, std::function<void()> on_cancel_a) :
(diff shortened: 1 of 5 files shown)


clemahieu (dismissed):

guilhermelawless:
> Return {"success": ""} as the response for work_generate

@koczadly did you mean work_cancel ?

Thanks for the fix, we require however that commits are GPG-signed. If you don't mind we can alternatively commit the patch ourselves with you as co-author.

koczadly:
@guilhermelawless 

Yeah sorry, I meant work_cancel. I've squashed and re-committed it with a signature, apologies.

nano/core_test/fakes/work_peer.hpp
```diff
@@ -176,9 +185,7 @@ class work_peer_connection : public std::enable_shared_from_this<work_peer_conne
 		}
 		else if (action_text == "work_cancel")
 		{
-			error (empty_response);
-			on_cancel ();
-			write_response ();
+			handle_cancel ();
```
guilhermelawless, line 189: This isn't being tested here which is fine, but to keep the current behavior, `write_response ()` should not be removed.

nano/rpc_test/rpc.cpp
```diff
@@ -2999,6 +2999,8 @@ TEST (rpc, work_cancel)
 		ASSERT_TIMELY (10s, response1.status != 0);
 		ASSERT_EQ (200, response1.status);
 		ASSERT_NO_ERROR (ec);
+		std::string success (response1.json.get<std::string> ("success"));
```
guilhermelawless, line 3002: Before doing `json.get...` you should check that "success` is there with e.g. `json.count ("success")`, otherwise it will throw an exception if not found (instead of properly failing the test). An alternative is using `get_optional` and checking that the return is initialized.

guilhermelawless (commented):
Just some minor suggestions

nano/core_test/fakes/work_peer.hpp
koczadly: It's all been delegated to the handle_cancel method instead, write_response is still there.

nano/rpc_test/rpc.cpp
koczadly: Hmm.. do exceptions not fail the test properly? Sorry, I'm a java dev and don't work with C languages or frameworks very often.

None of the other tests appear to check the presence of an attribute, unless I'm reading them wrong. I had just copy-pasted the check from bootstrap_any.

nano/core_test/fakes/work_peer.hpp
koczadly: Oops, seems you're right. I forgot to append it to the end.

nano/rpc_test/rpc.cpp
guilhermelawless: We haven't enforced it in the past so no worries.

clemahieu (approved):

guilhermelawless (approved):

PR #2918, opened 2020-09-08
Move channel drop policy off of nano::socket and remove channel concurrency value as redundant

clemahieu:
This commit moves the responsibility for limiting writes to a socket from nano::socket on to nano::channel_tcp. The intended behavior of nano::socket is to be a thin wrapper around a raw tcp socket with the addition of queueing on a strand plus statistics gathering.
The drop policy is part of the channel abstraction which varies depending on the channel used. channel_udp uses the operating system to drop writes it can't service and channel_tcp needs to track the number of outstanding writes to determine what should be dropped.
The drop policy was being passed all the way through to nano::socket, punching a hole through the nano::socket abstraction.

This commit also removes the concurrency enum as redundant. All socket operations were being serialized through the strand implicitly and the only other function of having a non-concurrent channel was to bypass the socket rate limiter. Non-concurrent users of channel_tcp should manage their own rate limiting if they don't want their buffers dropped. The only users of non-concurrent channels appear to be bootstrapping and this limitation already appears to be respected.

This also removes the need for queueing the buffer and callback in a separate container as they're already captured in objects to be executed on the socket's strand. This removes the need to flush the buffers on socket close.

### DIFF
--- a/nano/core_test/socket.cpp
+++ b/nano/core_test/socket.cpp
@@ -18,19 +18,11 @@ TEST (socket, drop_policy)
 
 	std::vector<std::shared_ptr<nano::socket>> connections;
 
-	// We're going to write twice the queue size + 1, and the server isn't reading
-	// The total number of drops should thus be 1 (the socket allows doubling the queue size for no_socket_drop)
-	size_t max_write_queue_size = 0;
-	{
-		auto client_dummy (std::make_shared<nano::socket> (*node, boost::none, nano::socket::concurrency::multi_writer));
-		max_write_queue_size = client_dummy->get_max_write_queue_size ();
-	}
-
 	auto func = [&](size_t total_message_count, nano::buffer_drop_policy drop_policy) {
 		auto server_port (nano::get_available_port ());
-		boost::asio::ip::tcp::endpoint endpoint (boost::asio::ip::address_v4::any (), server_port);
+		boost::asio::ip::tcp::endpoint endpoint (boost::asio::ip::address_v6::any (), server_port);
 
-		auto server_socket (std::make_shared<nano::server_socket> (*node, endpoint, 1, nano::socket::concurrency::multi_writer));
+		auto server_socket = std::make_shared<nano::server_socket> (*node, endpoint, 1);
 		boost::system::error_code ec;
 		server_socket->start (ec);
 		ASSERT_FALSE (ec);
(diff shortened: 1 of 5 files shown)


nano/node/socket.hpp
```diff
@@ -84,25 +76,23 @@ class socket : public std::enable_shared_from_this<nano::socket>
...
 	std::atomic<bool> closed{ false };
 	void close_internal ();
-	void write_queued_messages ();
 	void start_timer ();
 	void stop_timer ();
 	void checkup ();
-	void flush_send_queue_callbacks ();
+
+public:
+	size_t const queue_size_max = 128;
```
guilhermelawless: This could be `static` and `constexpr`

SergiySW (dismissed):

cryptocode:
> This commit also removes the concurrency enum as redundant.

The purpose was to avoid enqueuing bootstrap writes. In that case it was `async_write->handler->async_write->...` (such as `bulk_pull_server::send_next`), while now it'll be `post (strand queuing)->async_write->handler->post->async_write->...`

Not sure if it'll affect performance/memory, just wanted to point it out since it's a significant amount of calls/data being processed.

clemahieu:
> es. In that case it was `async_write->handler->async_write->...` (such as `bulk_pull_server::send_next`), while now it'll be `post (stran`

Good point. I'm working on a refactor of the bootstrapping code, hopefully using coroutines. I'll keep that in mind to profile.

SergiySW (dismissed):

nano/core_test/socket.cpp
```diff
@@ -41,22 +37,25 @@ TEST (socket, drop_policy)
...
 				},
 				drop_policy);
 			}
 		});
-		write_completion.await_count_for (std::chrono::seconds (5));
+		ASSERT_FALSE (write_completion.await_count_for (std::chrono::seconds (5)));
+		ASSERT_EQ (1, client.use_count ());
 	};
 
 	func (max_write_queue_size * 2 + 1, nano::buffer_drop_policy::no_socket_drop);
```
SergiySW: Can it use `nano::socket::queue_size_max` directly? It seems `max_write_queue_size `is not used in other places

guilhermelawless (approved):
TSAN and ASAN passing, tested a short bootstrap on Debug also. Great change!

PR #2921, opened 2020-09-10
Deadline timer cleanup

clemahieu:
This is a follow on to other tcp socket changes. This removes the use of boost::optional for the io_timeout, which is unnecessary as a max value can take the place of boost::none. It also makes the io_timeout atomic which removes the need for the change to be enqueued on the socket's strand.


nano/node/socket.cpp
```diff
@@ -72,69 +67,39 @@ void nano::socket::async_read (std::shared_ptr<std::vector<uint8_t>> buffer_a, s
...
-					if (queue_size < this_l->queue_size_max || (drop_policy_a == nano::buffer_drop_policy::no_socket_drop && queue_size < (this_l->queue_size_max * 2)))
-					{
-						this_l->send_queue.emplace_back (nano::socket::queue_item{ buffer_a, callback_a });
-					}
-					else if (auto node_l = this_l->node.lock ())
+		++queue_size;
+		boost::asio::post (strand, boost::asio::bind_executor (strand, [buffer_a, callback_a, this_l]() {
+			if (!this_l->closed)
+			{
+				this_l->start_timer ();
```
SergiySW: If deadline is also atomic, is it required to be in strand?

clemahieu: It seems like there are other things to clean up before we can remove the strand entirely.

SergiySW (dismissed):

SergiySW (dismissed):

thsfs:
@clemahieu, my last changes on this PR was for the only purpose of resolving its conflicts with the current develop branch. This PR seems (on my perspective) to be only a re-factory. Do you see any change for which we need to improve our testing coverage?

The changes made here may also conflict with the PR: https://github.com/nanocurrency/nano-node/pull/2965

theohax (approved):

PR #2925, opened 2020-09-10
Election mutexes to remove heavy active mutex contention

guilhermelawless:
The heavy active mutex contention from processing blocks, votes, confirmation requests and also for requesting confirmations has been manifesting as one of our largest bottlenecks, both in runtime performance and design flexibility.

This adds a mutex for each election and attempts to unlock the active_transactions mutex for the aforemetioned operations. From the election point of view, the active mutex is only required when confirming. A few locations remain where both the active mutex and an election mutex are held. These are pointed out in the code with a comment. The active mutex is always held first, which prevents a lock inversion issue.

Heavy use of the election atomics was made to reduce mutex locks. The reasoning for this is the confirmation request loop still needs to poll through each election in the loop. Solving that, possibly with a scheduling algorithm, is left as a future enhancement.

Note the mutex is not necessarily held in `election::transition_time`. This means the state can be changed (to confirmed) while in that method. For this reason, `election::state_change` no longer asserts on invalid state changes. Whatever the case, a confirmed election will always remain as confirmed.

This PR is marked as experimental, indicating it must undergo heavy testing. Available unit tests are comprehensive, especially with TSAN.


wezrule:
Ran TSAN/ASAN with RelWithDebInfo on gcc 9.2 Ubuntu 19.10, was clean

SergiySW (dismissed):

wezrule (dismissed):

SergiySW (dismissed):

SergiySW (dismissed):

SergiySW (dismissed):

SergiySW (dismissed):

nano/node/active_transactions.cpp
```diff
@@ -1369,6 +1452,9 @@ nano::inactive_cache_status nano::active_transactions::inactive_votes_bootstrap_
 	nano::inactive_cache_status status (previously_a);
 	constexpr unsigned election_start_voters_min{ 5 };
 	nano::uint128_t tally;
+
+	lock_a.unlock ();
```
wezrule, line 1457: Should probably add a check for `debug_assert (lock_a.owns_lock ());` before calling this, it is a pre-condition for this function. We do similar elsewhere with mutexes

SergiySW: Updated

SergiySW (approved):

wezrule (approved):

PR #2933, opened 2020-09-14
Fix core_test build with Boost 1.74

SergiySW:
Sample: https://www.boost.org/doc/libs/1_74_0/libs/beast/example/doc/http_examples.hpp
Since at least: https://www.boost.org/doc/libs/1_66_0/libs/beast/doc/html/beast/ref/boost__beast__http__message/content_length.html


guilhermelawless (approved):
We have boost 1.69 min, LGTM

argakiig (approved):

PR #2937, opened 2020-09-14
Use node::receive_confirmed synchronously

guilhermelawless:
The confirmation height processor already calls `node::receive_confirmed()` synchronously, since it uses `wallet::receive_async()` internally, so the actual block creation happens elsewhere. This change makes it so the same thing happens from `wallet::search_pending()`.

Also noticed there was no test going through that path from `wallet::search_pending()`, only activating elections for unconfirmed blocks. Added a test.


clemahieu (dismissed):

nano/node/wallet.cpp
```diff
@@ -1194,11 +1194,7 @@ bool nano::wallet::search_pending ()
 						if (wallets.node.ledger.block_confirmed (block_transaction, hash))
 						{
 							// Receive confirmed block
-							auto node_l (wallets.node.shared ());
-							wallets.node.background ([node_l, block, hash]() {
-								auto transaction (node_l->store.tx_begin_read ());
-								node_l->receive_confirmed (transaction, block, hash);
-							});
+							wallets.node.receive_confirmed (block_transaction, block, hash);
```
SergiySW: I think it should pass wallet transaction as well, otherwise there will be 2 wallet read transactions in 1 threads (next in scan_receivable)

guilhermelawless: Good catch, thanks! This makes me think if it might be possible to have a transaction tracker that would issue a stacktrace when more than one transaction of the same store is open in a thread. @wezrule you usually like trackers 😄 is it feasible?

SergiySW (approved):

nano/node/wallet.cpp
wezrule: Sure it's possible, you could have a `thread_local` count which is incremented and decremented in read transaction constructor/destructor, when constructing one if the count is non-zero it means a transaction is already created.

PR #2939, opened 2020-09-15
Convert nano::network periodic functions to use a steady timer instead of callbacks

clemahieu:
This change pulls in the boost coroutine library into the node.

Additionally, it rewrites the ongoing_cleanup function in nano::network in terms of steady timers, which is easier to reason around instead of callback loops through nano::alarm.

### DIFF
--- a/nano/node/network.cpp
+++ b/nano/node/network.cpp
@@ -754,7 +754,7 @@ void nano::network::ongoing_syn_cookie_cleanup ()
 		boost::system::error_code ec;
 		while (!stopped && !ec)
 		{
-			syn_cookies.purge (std::chrono::steady_clock::now () - nano::transport::syn_cookie_cutoff);
+			this->syn_cookies.purge (std::chrono::steady_clock::now () - nano::transport::syn_cookie_cutoff);
 			cookie_timer.expires_from_now (nano::transport::syn_cookie_cutoff * 2);
 			cookie_timer.async_wait (yield[ec]);
 		}
--- a/nano/node/network.cpp
+++ b/nano/node/network.cpp
@@ -5,6 +5,7 @@
 #include <nano/node/telemetry.hpp>
 #include <nano/secure/buffer.hpp>
 
+#include <boost/asio/steady_timer.hpp>
 #include <boost/format.hpp>
 #include <boost/variant/get.hpp>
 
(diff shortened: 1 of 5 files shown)


nano/node/network.cpp
```diff
@@ -727,14 +731,18 @@ void nano::network::cleanup (std::chrono::steady_clock::time_point const & cutof
...
 {
-	cleanup (std::chrono::steady_clock::now () - node.network_params.node.cutoff);
-	std::weak_ptr<nano::node> node_w (node.shared ());
-	node.alarm.add (std::chrono::steady_clock::now () + node.network_params.node.period, [node_w]() {
-		if (auto node_l = node_w.lock ())
+	boost::asio::spawn (
+	node.io_ctx,
+	[this](boost::asio::yield_context yield) {
+		boost::system::error_code ec;
+		while (!stopped && !ec)
```
guilhermelawless, line 740: Should we debug_assert at the end that `!ec || ec == boost::system::errc::operation_canceled` ?

nano/node/network.cpp
```diff
@@ -727,14 +731,18 @@ void nano::network::cleanup (std::chrono::steady_clock::time_point const & cutof
...
 		{
-			node_l->network.ongoing_cleanup ();
+			cleanup (std::chrono::steady_clock::now () - node.network_params.node.cutoff);
+			cleanup_timer.expires_from_now (node.network_params.node.period);
+			cleanup_timer.async_wait (yield[ec]);
 		}
-	});
+	},
+	boost::coroutines::attributes (128 * 1024));
```
guilhermelawless: Is this a leftover?

nano/node/network.cpp
```diff
@@ -132,6 +135,7 @@ void nano::network::stop ()
 		{
 			thread.join ();
 		}
+		cleanup_timer.cancel ();
```
guilhermelawless, line 139: Does this operation to be queued in a strand?
Also think we should use the overload with error_code to swallow exceptions.

clemahieu: The documentation says it kills any asynchronous operation and the error_code variant is deprecated.
https://www.boost.org/doc/libs/1_71_0/doc/html/boost_asio/reference/basic_waitable_timer/cancel.html

guilhermelawless (approved):
This introduces a thread sanitizer segfault on daemon shutdown.

On inspection it does not look like an issue with the implementation. It happens when multiple threads run the io context and stackful coros were spawned. It appears the segfault happens after all I/O is handled which makes it a non-issue.

wezrule (approved):
LGTM

PR #2943, opened 2020-09-16
Add RocksDB as a submodule

wezrule:
Adds RocksDB as a submodule. It uses a custom branch with some fixes and improvements: https://github.com/nanocurrency/rocksdb/tree/6.11.4
Increases MacOS version from 10.12 to 10.14 to fix a build error.
`NANO_ROCKSDB_TOOLS` CMake option added to build the tools to help with RocksDB sst file inspection.
Removes `NANO_ROCKSDB` CMake option and from the code as it will always be built now.


clemahieu (dismissed):
Tested on OSX, Ubuntu + ASAN/TSAN.

Srayman:
Compiled on Windows.  nano_node would not run, was getting errors generating default configs and process would not start with --daemon.  I was able to resolve this by building with -DPORTABLE=1 as it appears my CPU does not support AVX2 which is default for RocksDB.

https://github.com/facebook/rocksdb/wiki/Building-on-Windows

Apparently that was already mentioned here, but maybe needs to be included in documentation for building the node locally with this change.  https://github.com/nanocurrency/nano-node/pull/2288

argakiig (commented):
Updated PR with the rest of the cached rocksdb artifacts being removed,
also added PORTABLE=1 to all tests and build artifacts

guilhermelawless (approved):

PR #2944, opened 2020-09-16
Use empty uint8_t peer data

SergiySW:
to save disk space

ASAN/TSAN/Valgrind tests with different compilers are welsome


wezrule (dismissed):

cryptocode:
Would a key-only `put` overload make sense so the caller doesn't need to know the no-data representation? Could be used for the future `pruned_put` as well I think.

SergiySW:
@cryptocode sure, make sense

wezrule (approved):

PR #2946, opened 2020-09-16
[Pruning] Basic pruning blocks storage & database upgrade

SergiySW:
Splitting https://github.com/nanocurrency/nano-node/pull/2881

PR #2947, opened 2020-09-16
[Pruning] Pruning configuration settings

SergiySW:
Splitting https://github.com/nanocurrency/nano-node/pull/2881

* Enable optional ledger pruning with node launch flag `--enable_pruning` for automatic node background actions.
* Pruning rules can be configured in [node.experimental] config with “max_pruning_age” (Time in seconds to limit for blocks age after pruning. Also period for automatic node pruning task, default 1 day, 5 minutes for beta network) & “max_pruning_depth” (Limit for full blocks in chain after pruning, default 0 - unlimited). By default all confirmed blocks older than 1 day can be pruned, except genesis block & last confirmed block in each account chain.
* `--enable_pruning` flag & config option “enable_voting” are mutually exclusive to prevent representatives start with pruned ledger. Pruning is not available for any representatives.


guilhermelawless:
Best if #2946 merges first as you noted in the code

wezrule (dismissed):

guilhermelawless (approved):

PR #2949, opened 2020-09-17
block destination () function

SergiySW:
for send blocks


clemahieu:
I guess there's not a way to determine a send block's destination without loading the previous block. This would have to be a ledger function in that case.

Are you going to use this code in places?

SergiySW:
@clemahieu is yourt comment still valid?

SergiySW:
Contains in https://github.com/nanocurrency/nano-node/pull/2992

PR #2965, opened 2020-09-22
This converts the socket IO timer to a timer started/stopped with raii.

clemahieu:
Each socket has a deadline_next value that specifies when the next i/o deadline is to occur.
Deadlines are either: a timeout for an I/O operation or, a timeout for inactivity. When an I/O operation completes, a deadline for inactivity is started.
Each creation of a timer stores a timestamp and sets the deadline in the socket to be the same value. Upon timer destruction, iff the socket value equals the deadline timer, we'll set a new timeout deadline, otherwise we have overlapping I/O operations and another timer will do the deadline setting.


SergiySW (dismissed):

dsiganos:
This change brings code complexity and it is not clear that it gives us much in return.
So we decided to cleanup and document the existing code instead (#3704).

PR #2966, opened 2020-09-22
Check if awaiting collection is empty before waiting on condition in conf height processor

wezrule:
@guilhermelawless found that the `confirmation_height.unbounded_block_cache_iteration` test failed intermittently.

It was because a check is made to see if the `awaiting_processing` collection is empty before waiting on the condition variable. However the `lock` is unlocked before this action is made (while other actions are performed) and it's possible `confirmation_height_processor::add` gets called (which notifies the condition), however we `condition_variable::wait` after this, so it ends up not being processed until another block is confirmed.


guilhermelawless (dismissed):

nano/node/confirmation_height_processor.cpp
```diff
@@ -118,7 +118,11 @@ void nano::confirmation_height_processor::run (confirmation_height_mode mode_a)
 				else
 				{
 					lock_and_cleanup ();
-					condition.wait (lk);
```
clemahieu, line 121: Does this condition need to fire on either original_hash having values, as below, or awaiting_processing having values? It seems like at the top of the loop we could have one wait_until, which checks if either is non-empty, and the body determines which to process.

wezrule: It doesn't need to fire on `original_hash` changing that is just the last popped off value of `awaiting_processing`. The condition below is only hit if the confirmation height processor is paused which is only utilised in a couple tests. I have updated it to make that more clear.

clemahieu (approved):

PR #2982, opened 2020-09-25
Add timing_logging check before conf height timing output

wezrule:
This is to prevent unwanted output in logs, originally I thought 50ms was enough but even on my windows machine cementing 1 block can take over 50ms, so now guarding it with the `timing_logging` config option.


guilhermelawless (approved):
LGTM if CI passes, I have observed the same on my windows node, though RocksDB is fast enough.

PR #2997, opened 2020-10-09
Fix repcrawler loopback channel guard

cryptocode:
Based on a betanet crash report, it seems like loopback channels (from #2970) may release assert in repcrawler. This expands the loopback check to cover sending the confirm_ack as well. 
Review without whitespaces changes.


guilhermelawless:
Good catch, thank you!

clemahieu (approved):

Issue #3017, opened 2020-10-23
debug_assert in work_generate_blocking using qt wallet


wezrule:
Reported by @cryptocode

```
boost::optional<uint64_t> nano::node::work_generate_blocking (nano::block & block_a)
{
    debug_assert (network_params.network.is_dev_network ());
    return work_generate_blocking (block_a, default_difficulty (nano::work_version::work_1));
}
```

This assert hits when creating a block in the qt wallet. This variant of `work_generate_blocking` is only meant for tests. Spoke with @guilhermelawless who said it looks like it was missed in https://github.com/nanocurrency/nano-node/pull/2569 and it should call the same overload (taking difficulty) as the non-qt wallet.

guilhermelawless:
@wezrule @zhyatt I should be able to give this a go. The node wallet behavior is to grab cached work and, [if the difficulty isn't high enough, obtain it again](https://github.com/nanocurrency/nano-node/blob/V21.2/nano/node/wallet.cpp#L1151-L1158).

~~It looks like the QT wallet doesn't grab any cached work, could you confirm this and let me know if that would be the desired behavior instead? If so, it seems reasonable to extract the node wallet behavior and use that from the QT wallet as well.~~

^ nevermind, this is only an issue in block creation

zhyatt:
Thanks @guilhermelawless - assigned to you. Let us know if you need anything from our end.

PR #3018, opened 2020-10-24
Fix qt wallet hitting a debug_assert when creating blocks

guilhermelawless:
This changes the block creation path of the Qt wallet to use the appropriate difficulty for work generation. By doing so, it fixes the assert reported in #3017.

The normal sending process (clicking Send in the wallet) and receive process (automatically) use the node wallet and thus the appropriate difficulty.

A test is added to make some epoch_2 validation where difficulty is different across block types. There is a similar test for the node internal wallet.

### DIFF
--- a/nano/qt/qt.cpp
+++ b/nano/qt/qt.cpp
@@ -2265,7 +2265,11 @@ void nano_qt::block_creation::create_send ()
 						(void)error;
 						debug_assert (!error);
 						nano::state_block send (account_l, info.head, info.representative, balance - amount_l.number (), destination_l, key, account_l, 0);
-						if (wallet.node.work_generate_blocking (send).is_initialized ())
+						nano::block_details details;
+						details.is_send = true;
+						details.epoch = info.epoch ();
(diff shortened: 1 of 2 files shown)


nano/qt/qt.cpp
```diff
@@ -2500,7 +2514,11 @@ void nano_qt::block_creation::create_open ()
 							if (!error)
 							{
 								nano::state_block open (pending_key.account, 0, representative_l, pending.amount, source_l, key, pending_key.account, 0);
-								if (wallet.node.work_generate_blocking (open).is_initialized ())
+								nano::block_details details;
+								details.is_receive = true;
+								details.epoch = pending.epoch;
```
SergiySW, line 2519: I guess need
`etails.epoch = std::max (pending.epoch, info.epoch ());`

SergiySW (approved):

wezrule (approved):

zhyatt:
@wezrule @SergiySW If this is in good shape to merge, feel free to do so!

PR #3022, opened 2020-10-26
Add include_only_confirmed to account_balance & account_info

wezrule:
This adds `include_only_confirmed` to account_balance and account_info RPC calls for `balance` & `pending`. `account_balance` is useful as it can give pending amounts for not-yet opened accounts and is quite a good utility.


nano/secure/ledger.cpp
```diff
@@ -826,26 +826,47 @@ nano::uint128_t nano::ledger::balance_safe (nano::transaction const & transactio
...
 {
 	nano::uint128_t result (0);
 	nano::account end (account_a.number () + 1);
 	for (auto i (store.pending_begin (transaction_a, nano::pending_key (account_a, 0))), n (store.pending_begin (transaction_a, nano::pending_key (end, 0))); i != n; ++i)
 	{
 		nano::pending_info const & info (i->second);
-		result += info.amount.number ();
+		if (only_confirmed_a)
+		{
+			if (block_confirmed (transaction_a, i->first.hash))
```
cryptocode: I implemented something similar to this ledger function for a PR I'm working on, though I added a pruning check - that's needed right? Something like

`if (block_confirmed (transaction_a, i->first.hash) || (pruning && store.pruned_exists (transaction_a, i->first.hash)))`

wezrule: Oh really! Good to see you saw value in it too 👍 Oh yeh probably, wonder if it should be part of block_confirmed. There's also the conf height processor to consider too (`is_being_confirmed`). Will have a look, thanks!

wezrule: So I think not using conf height in-progress checking is fine. Added the pruning check though thanks! We also shouldn't error in account_info if not using include_only_confirmed, as there are still valid non-confirmed entries. The docs should also make clear that "include_only_confirmed" only affects balance and pending count (if requested). Maybe it should be called `include_only_confirmed_balance` but then inconsistent with rest, arghh

SergiySW (approved):
LGTM

cryptocode (approved):

PR #3023, opened 2020-10-26
Proposal: Websocket Payment Tracking

cryptocode:
This PR attempts to provide an easy and reliable way to get payment notifications, through a Websocket subscription. It's an alternative to block confirmation/pending/balance checks, and supports single-use accounts and block hand-off.

Preliminary documentation: https://github.com/cryptocode/notes/wiki/Websocket-Payment-Tracking

- [x] It adds ledger functionality similar to https://github.com/nanocurrency/nano-node/pull/3022 - will move this PR to those ledger functions once they're merged.

### DIFF
--- a/nano/node/node.hpp
+++ b/nano/node/node.hpp
@@ -45,6 +45,7 @@ namespace nano
 namespace websocket
 {
 	class listener;
+	class payment_validator;
 }
 class node;
 class telemetry;
@@ -156,6 +157,7 @@ public:
 	nano::node_config config;
 	nano::stat stats;
 	nano::thread_pool workers;
+	std::shared_ptr<nano::websocket::payment_validator> websocket_payment_validator;
 	std::shared_ptr<nano::websocket::listener> websocket_server;
 	nano::node_flags flags;
 	nano::work_pool & work;
--- a/nano/node/websocket.cpp
+++ b/nano/node/websocket.cpp
@@ -6,12 +6,90 @@
 #include <nano/node/transport/transport.hpp>
 #include <nano/node/wallet.hpp>
 #include <nano/node/websocket.hpp>
+#include <nano/secure/ledger.hpp>
 
 #include <boost/algorithm/string.hpp>
 #include <boost/property_tree/json_parser.hpp>
 
 #include <algorithm>
 #include <chrono>
+
+nano::websocket::payment_tracking_options::payment_tracking_options (boost::property_tree::ptree const & options_a)
+{
+	tracked_account = (options_a.get<std::string> ("account", ""));
+	max_tracking_duration = std::chrono::seconds (options_a.get<uint64_t> ("timeout_seconds", 0));
+
+	auto policy_l (options_a.get<std::string> ("track", ""));
+	if (policy_l == "account")
+	{
+		tracking_policy = nano::websocket::payment_tracker::policy::account;
+
+		// Minimum of 0 raw is invalid, which will cause an error ack
+		std::string amount_string_l (options_a.get<std::string> ("minimum_amount", "0"));
+		if (minimum_amount.decode_dec (amount_string_l, true))
+		{
+			minimum_amount.clear ();
+		}
+	}
+	else if (policy_l == "block")
+	{
+		tracking_policy = nano::websocket::payment_tracker::policy::block;
+		minimum_amount.clear ();
+		auto block_node_l (options_a.get_child_optional ("block"));
+		if (block_node_l)
+		{
+			// Decode block. If invalid or missing, validate() will fail and cause an error ack
+			auto block = std::make_unique<nano::state_block> ();
+			if (!block->deserialize_json (*block_node_l))
+			{
+				tracked_block = std::move (block);
+				tracked_account = tracked_block->link ().to_account ();
+			}
+		}
+
+		watch_work = options_a.get<bool> ("watch_work", false);
+	}
+	else
+	{
+		tracking_policy = nano::websocket::payment_tracker::policy::invalid;
+	}
+}
+
+nano::error nano::websocket::payment_tracking_options::validate () const
+{
+	nano::error res;
+	nano::account account_l;
+	bool valid_account = !account_l.decode_account (tracked_account);
+
+	if (max_tracking_duration.count () == 0)
+	{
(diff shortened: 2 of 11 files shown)


nano/node/websocket_payment_tracking.cpp
```diff
@@ -0,0 +1,192 @@
...
+	std::unordered_map<std::string, payment_tracking_info> currently_tracked_l;
+	{
+		auto tracked_l = tracked_accounts.lock ();
+		currently_tracked_l = *tracked_l;
+	}
+
+	for (auto item : currently_tracked_l)
+	{
+		callback_a (item.first, item.second);
+	}
```
wezrule: Can use C++17 structured bindings now in the for loop range declaration

nano/node/websocket_payment_tracking.hpp
```diff
@@ -0,0 +1,109 @@
...
+	payment_validator (nano::node & node_a);
+
+	/** Check if the payment conditions are met, and if so, send notification to websocket client */
+	void check_payment (nano::account const & destination_account_a, nano::block_hash const & block_hash_a, std::shared_ptr<nano::websocket::session> const & session_a);
+
+	/** Publish a send state block to the network */
+	void publish_block (std::shared_ptr<nano::block> const & block_a, bool const work_watcher_a);
+
+private:
+	nano::node & node;
```
wezrule: It would be nice not to have to depend on the whole node for the payment validator (although we still do this in a lot of places). Most of the uses of the node look easy enough to remove and use the objects themselves (`ledger`, `logger` etc). The use of functions on the node aren't so easy, e.g `process_local`, probably could pass in a `publish_block` callback for that. It also makes it clearer what dependencies this class actually has. But it's quite a bit more effort, and I don't think there will be any initialization issues like we've had in other areas, so don't worry about it unless you feel inclined.

nano/node/websocket_payment_tracking.hpp
```diff
@@ -0,0 +1,109 @@
...
+{
+class session;
+class payment_tracking_options;
+
+/** Per-session tracking of payment destination accounts. This class is thread-safe. */
+class payment_tracker
+{
+public:
+	/** Payment tracking policies */
+	enum policy
```
wezrule: Should use `enum class`

nano/node/websocket_payment_tracking.hpp
```diff
@@ -0,0 +1,109 @@
...
+		account,
+		/** Track confirmation of a send state block (hand-off use-case) */
+		block
+	};
+
+	/** Tracking info based on payment subscription options */
+	class payment_tracking_info
+	{
+	public:
+		payment_tracking_info (std::string id_a, std::shared_ptr<nano::state_block> const & tracked_block_a, nano::amount const & minimum_amount_a, nano::websocket::payment_tracker::policy tracking_policy_a, std::chrono::seconds track_until_a) :
```
wezrule: Could pass the `std::string` by const ref or move it into `id`

nano/node/websocket.hpp
```diff
@@ -136,6 +142,39 @@ namespace websocket
...
+		nano::error validate () const;
+
+		/** We keep the subscribe message id so it can be included in notifications. This can be used by clients for message correlation. */
+		std::string id;
+
+		/** The tracked desination account */
+		std::string tracked_account;
+
+		/** Set only if tracking a block */
+		std::shared_ptr<nano::state_block> tracked_block;
```
wezrule: Is it necessary to have this be `shared_ptr`?

nano/node/websocket.cpp
```diff
@@ -613,6 +752,36 @@ void nano::websocket::listener::on_accept (boost::system::error_code ec)
...
+{
+	std::vector<std::shared_ptr<nano::websocket::session>> sessions_l;
+	nano::lock_guard<std::mutex> lk (sessions_mutex);
+	for (auto & weak_session : sessions)
+	{
+		auto session_ptr (weak_session.lock ());
+		if (session_ptr)
+		{
+			auto subscription_l (session_ptr->subscriptions.find (topic_a));
+			if (subscription_l != session_ptr->subscriptions.end ())
```
wezrule: Could use `if (session_ptr->subscriptions.count (topic_a) > 0)` to save a line as you don't need the iterator and still seems readable.

nano/node/node.hpp
```diff
@@ -163,6 +164,7 @@ class node final : public std::enable_shared_from_this<nano::node>
 	nano::network_params network_params;
 	nano::node_config config;
 	nano::stat stats;
+	std::shared_ptr<nano::websocket::payment_validator> websocket_payment_validator;
```
wezrule, line 160: Is it necessary to store this here? Seems it can just be created and passed to `websocket_server` which then has a handle to it. And probably can be `std::unique_ptr` there too.

wezrule (commented):
This is really nice!! Having recently had the headache of doing something similar recently I can definitely see the usefulness.

wezrule:
I also agree it would be nice if wallets supported the "blob" qr code for sending from an account, for instance if the customer needs a refund can get it from the account (assuming the funds are pocketed first), and also for payouts from things like arcade machine.

nano/node/websocket_payment_tracking.hpp
cryptocode: Yeah exactly, tried to avoid it, but felt like process_local and friends should be factored out into something that can be passed along. I can take a look at a callback.

cryptocode: Now passing in everything needed + callback for process_local.

nano/node/websocket.hpp
cryptocode: It was shared with payment_tracking_info, but moving to tracking only hash, so it can become a unique_ptr here and be moved into   👍

nano/node/node.hpp
cryptocode: Gone through the node ref removal refactoring

cryptocode:
CI suddenly started complaining about rocksdb... clang got updated? https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=249428

wezrule (dismissed):

PR #3026, opened 2020-10-28
Storing blocks according to tally in elections with forks

SergiySW:
Additionally store tally in inactive votes cache

### DIFF
--- a/nano/node/election.cpp
+++ b/nano/node/election.cpp
@@ -568,9 +568,9 @@ bool nano::election::replace_by_weight (nano::unique_lock<std::mutex> & lock_a, 
 	std::vector<std::pair<nano::block_hash, nano::uint128_t>> sorted;
 	sorted.reserve (last_tally.size ());
 	std::copy (last_tally.begin (), last_tally.end (), std::back_inserter (sorted));
+	lock_a.unlock ();
 	// Sort in ascending order
 	std::sort (sorted.begin (), sorted.end (), [](auto const & left, auto const & right) { return left.second < right.second; });
-	lock_a.unlock ();
 	// Replace if lowest tally is below inactive cache new block weight
 	auto inactive_existing (node.active.find_inactive_votes_cache (hash_a));
 	auto inactive_tally (inactive_existing.status.tally);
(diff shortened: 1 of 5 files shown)


wezrule (dismissed):

wezrule (dismissed):

nano/node/election.cpp
```diff
@@ -528,6 +535,83 @@ void nano::election::remove_votes (nano::block_hash const & hash_a)
...
+	debug_assert (lock_a.owns_lock ());
+	nano::block_hash replaced_block (0);
+	auto winner_hash (status.winner->hash ());
+	// Sort existing blocks tally
+	std::vector<std::pair<nano::block_hash, nano::uint128_t>> sorted;
+	sorted.reserve (last_tally.size ());
+	std::copy (last_tally.begin (), last_tally.end (), std::back_inserter (sorted));
+	// Sort in ascending order
+	std::sort (sorted.begin (), sorted.end (), [](auto const & left, auto const & right) { return left.second < right.second; });
+	lock_a.unlock ();
```
guilhermelawless: Seems like it could be unlocked before sorting

nano/node/election.cpp
```diff
@@ -528,6 +535,83 @@ void nano::election::remove_votes (nano::block_hash const & hash_a)
...
+	}
+	else if (inactive_tally > 0 && inactive_tally > sorted.front ().second)
+	{
+		if (sorted.front ().first != winner_hash)
+		{
+			replaced_block = sorted.front ().first;
+		}
+		else if (inactive_tally > sorted[1].second)
+		{
+			// Avoid removing winner
```
guilhermelawless, line 593: Could the winner change while the mutex is unlocked and another vote is received? Meaning it would assert in `remove_block`

SergiySW: Thanks, was fixed in one of the previous commits, but then rolled back again

nano/node/election.cpp
SergiySW: Updated

wezrule (approved):

PR #3028, opened 2020-10-30
Persistent TCP channel endpoint value

SergiySW:


wezrule (approved):
Tested in a while loop for an hour, no crash, compared with a few minutes before

PR #3032, opened 2020-11-02
Secure Websocket support and certificate config improvements

cryptocode:
This PR adds wss:// support to the Websocket implementation, complementing the existing TLS support for RPC.

### Configuration
Having a single config/certificate setup for both RPC and Websockets is desirable, so the existing [secure] portion of config-rpc.tml has been moved to a new (but optional) `config-tls.toml` file. It allows a node operator to flip on wss and https individually. One scenario is using plain http for local rpc, but require a secure Websocket from an external payment processor. 

There's no automated [secure] config migration, but the node won't start if the configuration isn't correct. The [secure] section isn't populated by default, so I don't think this a problem as there are currently few, if any, https users (it requires building the node with the NANO_SECURE_RPC flag)

### Implementation notes:
1. The Boost Websocket implementation doesn't have a common type for ssl and non-ssl streams, so I added a type-erased stream to handle both configurations (a common approach is using variants, but I think a stream type is cleaner)
2. tls_config holds the ssl context, which must be shared between sessions. That's why there's a shared_ptr to a single tls_config instance, which is then connected to the rpc and websocket configuration objects on startup.
3. The cli is updated to be able to generate a default tls config file.

To encourage the use of secure rpc/websockets, I would like to see the node/wallet eventually being built with openssl by default (that is, with NANO_SECURE_RPC on), unless that's problematic in terms of redistribution/CI. The node operator would still need to manually enable wss/https, so it shouldn't be a breaking change.

### DIFF
--- a/nano/core_test/toml.cpp
+++ b/nano/core_test/toml.cpp
@@ -1,5 +1,6 @@
 #include <nano/lib/jsonconfig.hpp>
 #include <nano/lib/rpcconfig.hpp>
+#include <nano/lib/tlsconfig.hpp>
 #include <nano/lib/tomlconfig.hpp>
 #include <nano/node/daemonconfig.hpp>
 #include <nano/secure/utility.hpp>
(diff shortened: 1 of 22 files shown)


zhyatt:
@cryptocode Can you rebase this? We are looking at possible inclusion of this in V23

cryptocode:
> Can you rebase this? We are looking at possible inclusion of this in V23

Will rebase after the merge of #3460 as most conflicts are formatting related.

EDIT: Rebased

clemahieu (approved):

PR #3034, opened 2020-11-03
github container repo

argakiig:
changes made to account for https://docs.docker.com/docker-hub/download-rate-limit/
using ghcr.io
update dockerfiles to include required label for github container repo reference(for ghcr.io)
image built tagged with docker.pkg.github.com and docker hub variants
pull artifacts in ci from ghcr.io instead of docker hub
push to ghcr.io using module for login in and push
continue to deploy to docker hub when password is present

When merged into develop new PRs will pull from ghcr.io for build environment


zhyatt (approved):
I'm not the best with these pieces, but nothing stood out as an issue to me in a brief review of the changes.

PR #3042, opened 2020-11-07
Make fuzz_buffer compile again after message_parser changes

cryptocode:


wezrule:
Should this PR (https://github.com/nanocurrency/nano-node/pull/3038) also solve this? message_parser function prototype is removing the bool

cryptocode:
@wezrule Looks like it, wasn't aware of the change, though the include removal is still needed (the include is not necessary, but points to a header that is moved). I can adjust this PR once #3038 is merged.

cryptocode:
Given #3038, this is now just a header removal to make it compile again

wezrule (approved):

argakiig (approved):

PR #3047, opened 2020-11-11
Send only confirmed frontiers for bootstrap

SergiySW:
In frontier_req client/server & bulk_pull server


wezrule (dismissed):

clemahieu (dismissed):

PlasmaPower:
Could you add a flag in the header extensions to bypass this? I've been using the bootstrapping APIs to diagnose network issues, and this would significantly limit my capabilities.

Srayman:
Is bulk pull server used for lazy bootstraps also?  Wondering if this change really only needs to be made for when frontiers are requested for legacy bootstraps since that’s what causes nodes to ban each other if they send bad frontiers, but I don’t think the same happens for lazy bootstraps.  Would that solve the comment from @PlasmaPower to allow bulk pull to send any blocks but only send confirmed frontiers for legacy attempts?

That would maybe help with fork resolution also when a lazy attempt is made after seeing enough votes on the other fork?

PlasmaPower:
I think either of these changes (or maybe both?) make sense:
- If the bulk pull start is a hash, go ahead regardless of confirmation status
- If the header extensions contains a flag, ignore confirmation status

SergiySW:
@PlasmaPower Now frontier_req_server and bulk_pull_server can ignore confirmation status with header extension flag

PlasmaPower:
Thanks! I appreciate it.

PlasmaPower:
Thanks! I appreciate it.

zhyatt (commented):
@PlasmaPower Disregard this previous comment, didn't have the comments above on my side :)

wezrule (dismissed):

nano/node/bootstrap/bootstrap_attempt.cpp
```diff
@@ -282,8 +282,12 @@ void nano::bootstrap_attempt_legacy::request_push (nano::unique_lock<std::mutex>
 void nano::bootstrap_attempt_legacy::add_frontier (nano::pull_info const & pull_a)
 {
 	nano::pull_info pull (pull_a);
```
clemahieu: There doesn't seem to be a reason to copy this here.

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -305,7 +305,14 @@ void nano::bootstrap_server::receive_bulk_pull_action (boost::system::error_code
...
 			if (node->config.logging.bulk_pull_logging ())
 			{
-				node->logger.try_log (boost::str (boost::format ("Received bulk pull for %1% down to %2%, maximum of %3%") % request->start.to_string () % request->end.to_string () % (request->count ? request->count : std::numeric_limits<double>::infinity ())));
+				if (!request->start.is_zero ())
+				{
+					node->logger.try_log (boost::str (boost::format ("Received bulk pull for %1% down to %2%, maximum of %3%") % request->start.to_string () % request->end.to_string () % (request->count ? request->count : std::numeric_limits<double>::infinity ())));
+				}
+				else
+				{
+					node->logger.try_log (boost::str (boost::format ("Received bulk pull for %1% down to %2%, maximum of %3% from %4%") % request->start.to_string () % request->end.to_string () % (request->count ? request->count : std::numeric_limits<double>::infinity ()) % remote_endpoint));
```
clemahieu: These clauses are almost identical, can we just always log the remote endpoint?

nano/node/bootstrap/bootstrap_bulk_pull.cpp
```diff
@@ -622,6 +652,7 @@ nano::bulk_pull_server::bulk_pull_server (std::shared_ptr<nano::bootstrap_server
 connection (connection_a),
 request (std::move (request_a))
 {
+	send_unconfirmed_blocks = request->header.bootstrap_are_unconfirmed_blocks_present ();
```
clemahieu: Rather than being a member variable, I think this should be a function that returns request->header.bootstrap_are_unconfirmed_blocks_present (), then there's no risk of this value being incorrectly set.

clemahieu (changes requested):

nano/node/bootstrap/bootstrap_attempt.cpp
SergiySW: Updated, but probably there was some reason long ago (?)

nano/node/bootstrap/bootstrap_bulk_pull.cpp
SergiySW: Udpated

nano/node/bootstrap/bootstrap_server.cpp
SergiySW: Udpated

clemahieu (dismissed):

wezrule (dismissed):

wezrule (dismissed):

SergiySW:
For v22 replaced with https://github.com/nanocurrency/nano-node/pull/3152

Issue #3030, opened 2020-11-01
Race leading to LMDB assert failed on wallet creation


PlasmaPower:
**Description of bug:**

Occasional crash when creating a wallet, due to yet another bad side effect of the dual wallet memory and LMDB storage.

**Steps to reproduce the issue:**

Can't be reliably reproduced, but try creating wallets en masse during bootstrapping or high tps.

**Describe the results you received:**

```
Assertion (status == 0) failed
nano::mdb_iterator<T, U>::mdb_iterator(const nano::transaction&, MDB_dbi, const MDB_val&) [with T = nano::public_key; U = nano::wallet_value; MDB_dbi = unsigned int; MDB_val = MDB_val]
/home/lee/programming/cpp/nano-node/nano/node/lmdb/lmdb_iterator.hpp:39
```

**Describe the results you expected:**

The node to not crash

**Additional information you deem important (e.g. issue happens only occasionally):**

I believe I've traced this issue. The generic problem is the following situation:
- Wallet transaction A created
- Wallet transaction B created
- Wallet created with transaction B
- Transaction B committed
- Wallet saved to memory
- Attempt to access wallet with transaction A, probably because of looping over in-memory wallets, but the wallet didn't exist at the time of transaction A's creation

I believe this is how it happened for me, given the situation and the exact assert that failed (copied from my Discord post):

`active_transactions::block_cemented_callback` creates a wallet transaction then calls `node::receive_confirmed` with it, which then iterates over the wallets in memory via `wallets::get_wallets` and calls `wallet_store::exists` with the transaction and the wallet store including the `MDB_dbi`, which then calls `wallet_store::begin`, which then calls `mdb_iterator::mdb_iterator`, which fails because the wallet db didn't exist when the transaction was created, which leads to the failed assert in the crash

and I should note that this race is made more likely by the mutex which protects both `wallets::create` and `wallets::get_wallets`, meaning that the `get_wallets` call waits for the wallet to finish being created, making this race significantly more likely

**Environment**:

Local beta net release build of V22.0DB8

**logs**

Ends with:
```
[2020-Nov-01 11:26:41.953555]: 0x5602b9060210 {"action":"wallet_create"}
[2020-Nov-01 11:26:42.047889]: RPC request 0x5602b9060210 completed in: 94357 microseconds
```

wezrule:
[...]

PR #3051, opened 2020-11-17
Wallet create causing crash when confirming blocks

wezrule:
This bug was found by @PlasmaPower and nicely documented [here](https://github.com/nanocurrency/nano-node/issues/3030).

I think this can be resolved by checking that the `handle` is valid inside `wallet_store::begin`. There doesn't seem to be a standard way to test that so I'm just checking that the `mdb_dbi_flags` doesn't return an error (that function calls `TXN_DBI_EXIST` which checks that the transaction and dbi arguments are correct and returns EINVAL if they aren't), a few others did that too but this one seemed to make the most sense.

The new slow test added seems to more often than not hit the bug on a single iteration, but doing 5 in a loop just to increase the probability.

Couple unrelated changes I made: Added `[[maybe_unused]` in `receive_confirmed` & removed an unnecessary `std::move` in `wallets::queue_wallet_action`

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -364,7 +364,7 @@ TEST (node, search_pending)
 	system.wallet (0)->insert_adhoc (nano::dev_genesis_key.prv);
 	ASSERT_NE (nullptr, system.wallet (0)->send_action (nano::dev_genesis_key.pub, key2.pub, node->config.receive_minimum.number ()));
 	system.wallet (0)->insert_adhoc (key2.prv);
-	ASSERT_FALSE (system.wallet (0)->search_pending ());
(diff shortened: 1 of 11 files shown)


SergiySW (dismissed):

PlasmaPower:
Given the latest commit, I should note that I'm not sure I found the only possible scenario where a wallet transaction is created and then get_wallets is called.

wezrule:
The only other place is `search_pending` which was already creating the transaction after calling `get_wallets ()`. However they both didn't account for deletes so now getting the read snaphot under the wallets mutex as well to make sure they are in a valid state.

SergiySW (approved):

PR #3058, opened 2020-11-26
Various consistency changes

wezrule:
`migrate_working_path` can be removed as we only upgrade from v19 nodes now.
`nano::collect_container_info` should should east const ref for the `name` parameter
Moved `bootstrap_attempt_legacy` class to its own file, as `boostrap_attempt_lazy` has its own
Removed `~frontier_req_client ();` destructor as it's empty (and unnecessary)


argakiig (dismissed):

nano/nano_node/entry.cpp
```diff
@@ -65,6 +65,7 @@ class address_library_pair
 int main (int argc, char * const * argv)
 {
 	nano::set_umask ();
+	//
```
clemahieu: Unused comment.

nano/node/lmdb/lmdb.cpp
```diff
@@ -269,6 +269,7 @@ bool nano::mdb_store::do_upgrades (nano::write_transaction & transaction_a, bool
 		case 17:
 			upgrade_v17_to_v18 (transaction_a);
 			needs_vacuuming = true;
+			// Upgrades to v19 & v20 are both part of the v22 node release
```
clemahieu: Should we start using [[fallthrough]]; ?

clemahieu (dismissed):

nano/nano_node/entry.cpp
wezrule: Removed

nano/node/lmdb/lmdb.cpp
wezrule: Yeh we should, added

argakiig (approved):

PR #3090, opened 2021-01-28
Add recent constref change to rpc secure as well

cryptocode:
Makes develop branch compile with NANO_SECURE_RPC


wezrule (approved):
Woops, I need to remember to build with the secure option. Thanks!

PR #3098, opened 2021-02-10
Use frontiers age option for frontier_req client

SergiySW:
to reduce bandwidth usage and failure rate for millions of frontiers.

- ~66.7%~ 75% of bootstrap attempts after initial bootstrap and after node warm up after startup to request only frontier modified since 1 day ago (1 hour for beta)
- for long inactive node use last stored online weight timestamp minus 1 hour

Additionally disable bulk push for requests containing age option.


wezrule (dismissed):

Srayman:
Curious, what's the thinking of having the frontier requests pull all frontiers 1/3 of the time instead of using the modified time every time?  Lazy bootstrapping and regular block propagation should be sufficient to keep a node in sync once it's fully bootstrapped correct?

SergiySW:
> Curious, what's the thinking of having the frontier requests pull all frontiers 1/3 of the time instead of using the modified time every time? Lazy bootstrapping and regular block propagation should be sufficient to keep a node in sync once it's fully bootstrapped correct?

We can't guarantee it's sufficient (yet) I think

clemahieu (approved):

PR #3100, opened 2021-02-12
Reverse store iterator and selector for last record

SergiySW:
Can be useful in certain conditions, i.e. select last online weight timestamp.


wezrule (dismissed):

wezrule (dismissed):

nano/secure/blockstore_partial.hpp
```diff
@@ -756,6 +756,16 @@ class block_store_partial : public block_store
 		return make_iterator<nano::qualified_root, nano::block_hash> (transaction_a, tables::final_votes);
 	}
 
+	nano::store_iterator<nano::account, nano::account_info> accounts_last (nano::transaction const & transaction_a) const override
```
clemahieu: I think these should follow the STL naming convention rbegin/rend http://www.cplusplus.com/reference/vector/vector/rbegin/

Since we prefix the table this would be accounts_rbegin(), accounts_rend().

nano/node/lmdb/lmdb_iterator.hpp
```diff
@@ -10,35 +10,20 @@ template <typename T, typename U>
...
-			if (current.first.size () != sizeof (T))
-			{
-				clear ();
-			}
+			current.first = val_a;
 		}
 		else
 		{
-			clear ();
+			operation = use_first_a ? MDB_FIRST : MDB_LAST;
```
clemahieu: Instead of use_first it seems these should be named "direction".

clemahieu (changes requested):

nano/node/lmdb/lmdb_iterator.hpp
SergiySW: Replaced with `direction_asc`

nano/secure/blockstore_partial.hpp
SergiySW: Updated

clemahieu (approved):

PR #3107, opened 2021-03-01
Check open file descriptor limit, warn if low

cryptocode:
Sample output (stderr and log)

```
WARNING: The file descriptor limit on this system may be too low (512) 
and should be increased to at least 16384.
```

This can be tested with `docker run --ulimit nofile=512:1024 ...`

### DIFF
--- a/nano/lib/utility.cpp
+++ b/nano/lib/utility.cpp
@@ -4,6 +4,7 @@
 #include <boost/filesystem.hpp>
 
 #include <iostream>
+#include <limits>
 #include <sstream>
 #include <string_view>
 #include <thread>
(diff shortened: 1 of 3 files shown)


SergiySW:
Could `rlim_cur ` be lower than `rlim_max ` ?

cryptocode:
@SergiySW yes, but the soft limit (rlim_cur) is what processes run into, rlim_max is the system hard limit, for processes trying to increase the limit programmatically. Do you mean rlim_max should be checked still?

wezrule (approved):

SergiySW:
> @SergiySW yes, but the soft limit (rlim_cur) is what processes run into, rlim_max is the system hard limit, for processes trying to increase the limit programmatically. Do you mean rlim_max should be checked still?

Yes, I wonder if need to check it as well

argakiig (approved):

SergiySW (approved):

Issue #3099, opened 2021-02-11
C++17ify nano/ "lib, node, rpc & secure subfolders"


LenWilliamson:
Hello,

as suggested in [this post](https://forum.nano.org/t/best-way-to-contribute-as-a-new-joiner/1252/2), I would like to C++17ify everything under the nano/ directory. I will start with lib, node, rpc & secure subfolders. As suggested by @wezrule I will:

1. Add structured bindings
2. Add `[[nodiscard]]` to functions which return a `bool`
3. Inline class static variables which are declared in source files

If there are any other C++17 features I should implement, please let me know.

PR #3119, opened 2021-03-07
issue 3099 - C++17ify

LenWilliamson:
Added [features](https://github.com/nanocurrency/nano-node/issues/3099).


LenWilliamson:
Okay, I forget to remove some merge conflict markers. I was only looking in the files I did changes. I fixed it and run `make` again. Then I will run `core_test`. But my laptop (not the most powerful machine) often crashes while running `make core_test -j`. I don't know why...

wezrule:
Thanks! Looks like there are still some merging issues (std::mutex should be nano::mutex for instance). Also clicking `Details` on our CI checks will show there are some no discard checks being missed. In some cases where the result is actually not cared about it can be cast to `(void)`.

wezrule:
I don't think we should have `[[nodiscard]]` in the source files too, header file is sufficient. I was originally thinking of just having it on functions that return error codes which might accidentally not get checked, but maybe having them on all (like you've done) is more appropriate

LenWilliamson:
Hi, I will check `Details`. Also I will remove `[[nodiscard]]` in the `.cpp` files. Okay, if there is anything else let me know.

LenWilliamson:
Regarding the `[[nodiscard]]` part, if you think having it on all is more appropriate than it is fine. But to be consistent one would have to put  `[[nodiscard]]` in front of every function that returns `bool`. And I don't know if this adds the desired readability? I'm currently still very inexperienced. In my opinion it would be better to put `[[nodiscard]]` on functions that return error codes, as you originally thought. For instance, why shouldn't one want to discard `nano::state_block::operator==`. It doesn't cause me any trouble to revert these changes. Maybe you can give me an example of a function where it truly makes sense to put a `[[nodiscard]]` attribute. I would think in nano/lib/logger_mt.hpp `bool try_log` maybe?

LenWilliamson:
1. Is it not sufficient to execute `/nano-node$ ci/clang-format-all.sh` for the formatting?
2. Why are so many tests failing? When I rerun the test on my computer I got:

grafik

anishsingh935:
Can I fix this problem ??

anishsingh935:
Please

LenWilliamson:
Yes, please. It would be great if you could also share the reason for the problem and how you fixed it. So I don't run into it again.

PR #3133, opened 2021-03-11
Limit bulk push targets count

SergiySW:
With 32k max default


SergiySW:
Already limited by bulk_push_cost (max 100-200 targets)

PR #3134, opened 2021-03-11
Final votes

clemahieu:
This patch is a mitigation when handling specific fork situations and uses a double consensus pass before arriving at confirmation.

Thank you to @PlasmaPower for the initial design and @Srayman for help with patch testing.


SergiySW (dismissed):

Srayman:
Can both the sequence and timestamps be used but have the timestamp include the actual timestamp the vote was generated?  

This would help in analyzing votes for confirmations to know when a final vote was generated compared to others.

The sequence could continue to be used to differentiate newer/final votes.

clemahieu:
> Can both the sequence and timestamps be used but have the timestamp include the actual timestamp the vote was generated?
> 
> This would help in analyzing votes for confirmations to know when a final vote was generated compared to others.
> 
> The sequence could continue to be used to differentiate newer/final votes.

That would require a second field in the vote and a new vote format with an upgrade etc.

SergiySW (approved):

Issue #3139, opened 2021-03-14
Enforcing const correctness and other good practices


marioortizmanero:
**Description of bug:**

[Const correctness](https://isocpp.org/wiki/faq/const-correctness) is quite lacking in the current codebase, and could help maintainability. There are instances like [this line](https://github.com/nanocurrency/nano-node/blob/develop/nano/node/cli.cpp#L212) where a variable could be `const` but it isn't. It doesn't offer performance gains, but it helps avoiding errors -- something important in C++. I propose we use `const` as a default, and then remove it if it's not needed. That's what many style guides recommend.

I would like to bring up this issue again so that we can decide on what can be improved (or at least the most relevant points): https://github.com/nanocurrency/nano-node/issues/297 (that way it can finally be closed, as it was opened back in 2017)

* Empty lines to separate huge blocks of code
* Early returns
* Consistency: returning a bool means that it went OK if `true`, and it failed if `false`
* Indentation in constructor definitions (they're really hard to read otherwise IMO)

I will make a few PRs myself for some of these (including const correctness), but I'd like an opinion beforehand. Not all of them are important at all.

clemahieu:
Const-correctness definitely.

We usually return true to signify an error has occurred though this is not done consistently which should be cleaned up. Using false for success and true for failure is done to be consistent with error codes where 0 means success and non-zero indicates the type of error.

Indentation for initializers yes. The clang-format script should be updated to enforce this.

marioortizmanero:
> We usually return true to signify an error has occurred though this is not done consistently which should be cleaned up. Using false for success and true for failure is done to be consistent with error codes where 0 means success and non-zero indicates the type of error.

As long as it's consistent it's perfectly fine. You didn't mention anything about early returns but in the older issue you said it was ok, so I'll assume there's no problem with that.

Thanks! I'll create a couple PRs soon™. I would say #297 can be closed now.

zhyatt:
Are there docs updates in https://docs.nano.org/core-development/code-standards/ that should be made from this? Open to suggestions there.

Issue #297, opened 2017-12-27
Coding remarks


paulmelis:
I have seen quite a bit of C++ over the years, but RaiBlocks' coding style really makes me stretch my brain to comprehend what is going on in the sources (in `rai/`). Below are some notes I made while looking through the code. 

I don't have a specific point to make (or issue to fix) here, but I was wondering if it is only me that is having trouble with this. For some of these things I can see the benefit, but my god, it makes me cringe thinking about the discipline required to type and read some of this stuff in this way (although a reformatting tool might be involved). Also, as there doesn't seem to be a readme on the code (or something like that) I have to guess a bit in certain areas as to why things are the way they are.

* Virtually no comments, nor any high-level descriptions of how things are organized
* No empty lines in method bodies and class definitions anywhere, to separate things like variable declarations from method calls using them
* `rai::` namespace prefixes, even for method definitions and variable types that are clearly part of that namespace
* Constructor calls instead of assignments, e.g.
    `auto node_l (shared_from_this ());`

    `for (auto i (peers.begin ()), n (peers.end ()); i != n; ++i)`    
* A space directly before the opening parenthesis of a method call and bracket of a template specialization, making complex statement especially hard to parse, as whitespace can't be trusted to separate logical elements, while also mimicing the frequently used C++ lambda functions, e.g.
    `endpoint_l = rai::endpoint (boost::asio::ip::address_v6::v4_mapped (endpoint_l.address ().to_v4 ()), endpoint_l.port ());`

    `auto existing (blocks.get <1> ().find (hash));`

    `send_buffer (buffer_a->data (), buffer_a->size (), endpoint_a, [buffer_a, node_w, endpoint_a] (boost::system::error_code const & ec, size_t size)
    {...}`
    
* Unexplained suffixes on variables, `_a` ("a ..."?) and `_l` (local?)
* In constructor definitions member initializations before the body are not indented, leading to hard to parse method definitions, as the constructor definition doesn't stand out. e.g.
    ```
    rai::network::network (rai::node & node_a, uint16_t port) :
    socket (node_a.service, rai::endpoint (boost::asio::ip::address_v6::any (), port)),
    resolver (node_a.service),
    node (node_a),
    bad_sender_count (0),
    on (true),
    insufficient_work_count (0),
    error_count (0)
    {
    }
   ```
* Methods often return `false` in case of success. The `result` variable that is returned from methods is therefore best to think of as being called "failure". For example, 
    ```
[...]

PR #3143, opened 2021-03-15
Some const correctness, upgrades to c++17, and lint suggestions

marioortizmanero:
I've been manually reviewing the code for some improvements regarding #3099, #3139 and #297. Here are the main changes:

* Added `explicit` to constructors with a single parameter. [Reference](https://clang.llvm.org/extra/clang-tidy/checks/google-explicit-constructor.html). All of them suggested by `clang-tidy`.
* Some early returns: most make the code considerably easier to read. But they're a bit tricky in reviews so not all are included. Applied manually from `rg 'if \(!error\)'` and `rg 'if \(!result\)'`.
* ~~Added `[[nodiscard]]` to getters and similar functions. [Reference](https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nodiscard.html). All of them suggested by `clang-tidy`.~~
* Added `const` to a good chunk of the codebase. I haven't found a way to do this automatically so not all of them are included yet. I've enforced that all of them are in the `T const` order, as it was the most commonly used one. Unfortunately, this can't be checked automatically with `clang-format` [for now](https://bugs.llvm.org/show_bug.cgi?id=34220).
* Removed unused headers. All of them recommended by the CLion IDE.
* Removed `const T` parameters in headers, as they have no effect. [Reference](https://clang.llvm.org/extra/clang-tidy/checks/readability-avoid-const-params-in-decls.html). All of them suggested by `clang-tidy`.

I'm not adding more changes related to early returns because there are so many lines modified already and I wanted to keep the PR smaller. If it's too much I can split it up or something. Sorry about that.

### DIFF
--- a/nano/nano_node/entry.cpp
+++ b/nano/nano_node/entry.cpp
@@ -748,7 +748,7 @@ int main (int argc, char * const * argv)
 			}
 			else
 			{
-				for (const auto & text : results)
+				for (auto const & text : results)
 				{
 					uint64_from_hex address_hex;
 					if (boost::conversion::try_lexical_convert (text, address_hex))
@@ -1351,7 +1351,7 @@ int main (int argc, char * const * argv)
 			++errors;
 		};
 
-		auto start_threads = [node, &threads_count, &threads, &mutex, &condition, &finished](const auto & function_a, auto & deque_a) {
+		auto start_threads = [node, &threads_count, &threads, &mutex, &condition, &finished](auto const & function_a, auto & deque_a) {
 			for (auto i (0U); i < threads_count; ++i)
 			{
 				threads.emplace_back ([&function_a, node, &mutex, &condition, &finished, &deque_a]() {
@@ -1612,7 +1612,7 @@ int main (int argc, char * const * argv)
 				if (accounts.size () > accounts_deque_overflow)
 				{
 					auto wait_ms (250 * accounts.size () / accounts_deque_overflow);
-					const auto wakeup (std::chrono::steady_clock::now () + std::chrono::milliseconds (wait_ms));
+					auto const wakeup (std::chrono::steady_clock::now () + std::chrono::milliseconds (wait_ms));
 					condition.wait_until (lock, wakeup);
 				}
 				accounts.emplace_back (i->first, i->second);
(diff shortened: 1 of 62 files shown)


nano/nano_node/entry.cpp
```diff
@@ -142,1894 +140,1870 @@ int main (int argc, char * const * argv)
...
+				platform = boost::lexical_cast<unsigned short> (platform_it->second.as<std::string> ());
 			}
-
-			nano::work_pool work (std::numeric_limits<unsigned>::max (), pow_rate_limiter);
-			nano::change_block block (0, 0, nano::keypair ().prv, 0, 0);
-			if (!result)
+			catch (boost::bad_lexical_cast &)
 			{
-				std::cerr << boost::str (boost::format ("Starting generation profiling. Difficulty: %1$#x (%2%x from base difficulty %3$#x)\n") % difficulty % nano::to_string (nano::difficulty::to_multiplier (difficulty, network_constants.publish_full.base), 4) % network_constants.publish_full.base);
-				while (!result)
```
marioortizmanero, line 499: This `while(!result)` and the previous `if (!result)` didn't make much sense because the code guaranteed `result` would be untouched and thus equal to `0`. So I just ignored the `if` and added a `while (true)` instead.

marioortizmanero:
I've also removed the `[[nodiscard]]` attributes, as #3119 is already working on that (although I added very few of them in comparison).

nano/node/wallet.cpp
```diff
@@ -1633,17 +1633,17 @@ void nano::wallets::reload ()
 	}
 	// Delete non existing wallets from memory
 	std::vector<nano::wallet_id> deleted_items;
-	for (auto i : items)
```
marioortizmanero, line 1636: I changed the name of `i` here because it was overshadowing the previously declared `i`, but perhaps that's not necessary

marioortizmanero:
OK I've reviewed my own PR and the changes seem to be correct. I won't be making any more modifications other than possible suggestions from reviews.

The only hard to check parts of this PR are the early returns, as most others would not compile if invalid, and because the git diff here on GitHub is a bit messed up. Specially `entry.cpp` was a bit tricky because it was a huge function, so let me know if I missed something there.

clemahieu:
Thank you very much for doing this cleanup, the changes look good on brief inspection.

I'll figure out when is a good time to include this as it churns a significant amount of files

zhyatt:
@theohax We are interested in having you look at the easier of these various changes to remove the conflicts on and clean up for inclusion.

theohax:
Superseded by https://github.com/nanocurrency/nano-node/pull/3510 . Thanks again @marioortizmanero for your contribution!

marioortizmanero:
Glad I could be of help! Thanks for cleaning up the PR and fixing the build as well.

PR #3144, opened 2021-03-15
The build info shows following description: \

skynxt:
Build Info: 5160073 "GNU C++ version " "5.4.0 20160609" "BOOST 107000" BUILT "Sep  3 2020" \
This file has default set to boostVersion=1.69 which is inconsitent with above build info which shows 1.70. Hence, updating default to latest Boost version to 1.75

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -375,7 +375,7 @@ set(Boost_USE_MULTITHREADED ON)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules")
 
-find_package(Boost 1.69.0 REQUIRED COMPONENTS filesystem log log_setup thread
+find_package(Boost 1.70.0 REQUIRED COMPONENTS filesystem log log_setup thread
                                               program_options system)
 
 # RocksDB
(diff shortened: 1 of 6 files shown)


cryptocode:
@skynxt Very nice. Could you use this opportunity to also remove 1.69 as an option in `case "${boostVersion}"` as it's no longer a supported version?

skynxt:
@cryptocode thanks for comment.
Yep, can do that, but had couple of points to discuss.
The following files(for osx, windows) has a reference to 1.70V Boost, is this a binary hosted in cloud? Perhaps version needs to change to the latest 1.75V Boost? nano-node is being successfully being built with 1.75V Boost. If yes, how can this be changed?
https://github.com/nanocurrency/nano-node/blob/develop/util/build_prep/fetch_boost.sh
https://github.com/nanocurrency/nano-node/blob/develop/ci/actions/windows/install_deps.ps1

Also, this minimum requirement need to be changed to 1.75v Boost too in the following link, or just leave as is?
https://docs.nano.org/integration-guides/build-options/#general-build-instructions

Looking forward to your comments.

cryptocode:
@skynxt we probably want 1.70 as the minimum, see https://github.com/nanocurrency/nano-docs/issues/485 about your last point.

argakiig (approved):

PR #3145, opened 2021-03-15
Disable frontiers confirmation check as ineffective

SergiySW:


SergiySW:
Replaced with : https://github.com/nanocurrency/nano-node/pull/3149

PR #3147, opened 2021-03-15
Limit frontier request block count by default

SergiySW:
Default limit is 1M blocks per frontier request, then after bulk pulls next request

Changes to RPC:
- allow optional "account" for RPC "bootstrap_any" for starting attempt account
- return "frontiers_age" & "last_account" for RPC "bootstrap_status" in legacy attempts


SergiySW:
make configurable with lower limit 1024

clemahieu (approved):

PR #3151, opened 2021-03-16
Lower frontier request server resource usage

SergiySW:
Config option to reduce 1 block account frontiers send probability


SergiySW:
Complicates config settings

PR #3159, opened 2021-03-19
Expired elections restart souldn't target same block

SergiySW:


zhyatt:
When moving to prioritization the tracking of expired elections for possible restart will no longer be needed.

zhyatt:
@clemahieu Can you confirm this is no longer needed with https://github.com/nanocurrency/nano-node/pull/3190?

SergiySW:
Closed due to https://github.com/nanocurrency/nano-node/pull/3208

PR #3165, opened 2021-03-23
Block rebroadcasting changes from V21.3

Srayman:
Rebroadcasting blocks only when an election is prioritized or confirmed. This will focus block propagation on blocks with an open election.  Included in V21.3, creating PR to merge into Develop. https://github.com/nanocurrency/nano-node/pull/3140/commits/c21ec84cfe49bd545198f2cbfa64095bd714e4b8


zhyatt:
Covered already by https://github.com/nanocurrency/nano-node/pull/3173

PR #3173, opened 2021-03-29
Removing broadcasting state

clemahieu:
Putting block announcing in the active election state and removing the broadcasting state entirely.

This was depending on a certain count of successful confirmation requests to move to the rebroadcasting state. This could leave elections hanging if confirmation reqs can't be published to another PR for whatever reason. This additionally only publishes blocks when an election has started which is the new desired behavior.

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -360,10 +360,6 @@ void nano::block_processor::process_live (nano::transaction const & transaction_
 	{
 		node.network.flood_block_initial (block_a);
 	}
-	else if (!node.flags.disable_block_processor_republishing)
-	{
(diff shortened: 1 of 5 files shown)


SergiySW (approved):

nano/node/blockprocessor.cpp
```diff
@@ -360,10 +360,6 @@ void nano::block_processor::process_live (nano::transaction const & transaction_
 	{
 		node.network.flood_block_initial (block_a);
 	}
-	else if (!node.flags.disable_block_processor_republishing)
-	{
-		node.network.flood_block (block_a, nano::buffer_drop_policy::no_limiter_drop);
```
SergiySW, line 365: What other places can propagate block properly?

PR #3174, opened 2021-03-30
Election start encapsulation

clemahieu:
This is a step to moving election starting responsibility into only one location in the active_transactions class. Elections are started either through pessimistic or optimistic election starting.


SergiySW (approved):
Need frontier cementing until initial bootstrap finished (and can be separate PR), other than that good to merge

PR #3178, opened 2021-04-08
start an election when a vote request is seen for an unconfirmed block with dependencies already confirmed

Srayman:
When nodes are attempting to confirm many different blocks from the unconfirmed pool/backlog they can be out of sync on which elections are active resulting increased vote requests rather than using the preferred push method.  This change activates an election when a vote request is processed if the node has dependencies confirmed already.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -2745,7 +2745,7 @@ TEST (node, local_votes_cache)
 	{
 		ASSERT_NO_ERROR (system.poll (node.aggregator.max_delay));
 	}
-	ASSERT_TIMELY (3s, node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_votes) == 3);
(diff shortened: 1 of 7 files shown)


nano/core_test/node.cpp
```diff
@@ -2745,7 +2745,7 @@ TEST (node, local_votes_cache)
 	{
 		ASSERT_NO_ERROR (system.poll (node.aggregator.max_delay));
 	}
-	ASSERT_TIMELY (3s, node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_votes) == 3);
+	ASSERT_TIMELY (3s, node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_votes)+node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_cached_late_votes) == 3);
```
SergiySW, line 2748: `+` -> ` + ` I guess. Same below

zhyatt:
This is no longer applicable due to prioritization changes earlier this year.

PR #3206, opened 2021-04-17
Fixing static initialization order on Window

clemahieu:


theohax (approved):
Static initialization order fiasco at its finest :D

PR #3217, opened 2021-04-20
Signal manager, handle signals using boost asio

dsiganos:
Currently the nano_node has 2 problems:
* it uses the function signal which is not very portable
* it calls possibly blocking code from signal handler context

These problems will become even more crucial when we start using SIGHUP to
reload config. In addition, we plan to introduce more processes and it
makes sense to have a common signal framework for all the processes.

This commit introduces the signal manager, which fixes all the problems
listed above.

### DIFF
--- a/nano/core_test/signal_manager.cpp
+++ b/nano/core_test/signal_manager.cpp
@@ -1,0 +1,106 @@
+/**
+ * IMPORTANT NOTE:
+ * These unit tests may or may not work, gtest and boost asio signal handling are not really compatible.
+ * The boost asio signal handling assumes that it is the only one handling signals but gtest
+ * also does some signal handling of its own. In my testing this setup works although in theory
+ * I am playing with unspecified behaviour. If these tests start causing problems then we should
+ * remove them and try some other approach.
+ * The tests are designed as death tests because, as normal tests, the boost library asserts
+ * when I define more than one test case. I have not investigated why, I just turned them into death tests.
+ *
+ * Update: it appears that these tests only work if run in isolation so I am disabling them.
+ */
+
+#include <nano/lib/signal_manager.hpp>
+
+#include <gtest/gtest.h>
+
+#include <boost/format.hpp>
+
+#include <csignal>
+#include <iostream>
+
+static void handler_print_signal (int signum)
+{
+	std::cerr << "boost signal handler " << signum << std::endl
+			  << std::flush;
+}
+
+static int wait_for_sig_received (int millisecs, int & sig_received)
+{
+	for (int i = 0; i < millisecs && sig_received == 0; i++)
+	{
+		std::this_thread::sleep_for (std::chrono::microseconds (1));
+	}
+	return sig_received;
+}
+
+static int trap (int signum)
+{
+	nano::signal_manager sigman;
+	int sig_received = 0;
+
+	std::function<void (int)> f = [&sig_received] (int signum) {
+		std::cerr << "boost signal handler " << signum << std::endl
+				  << std::flush;
+		sig_received = signum;
+	};
+
+	sigman.register_signal_handler (signum, f, false);
+
+	raise (signum);
+
+	exit (wait_for_sig_received (10000, sig_received));
+}
+
+static void repeattest (int signum, bool repeat)
+{
+	nano::signal_manager sigman;
+	int sig_received = 0;
+
+	std::function<void (int)> f = [&sig_received] (int signum) {
+		std::cerr << "boost signal handler" << std::flush;
+		sig_received = signum;
+	};
+
+	sigman.register_signal_handler (signum, f, repeat);
+
+	for (int i = 0; i < 10; i++)
+	{
+		sig_received = 0;
+		raise (signum);
+		if (wait_for_sig_received (10000, sig_received) != signum)
+		{
+			exit (1);
+		}
+	}
+
+	exit (0);
+}
+
+TEST (DISABLED_signal_manager_test, trap)
+{
+	int signum;
+
+	signum = SIGINT;
+	ASSERT_EXIT (trap (signum), ::testing::ExitedWithCode (signum), "");
+
+	signum = SIGTERM;
+	ASSERT_EXIT (trap (signum), ::testing::ExitedWithCode (signum), "");
+}
+
+TEST (DISABLED_signal_manager_test, repeat)
+{
+	int signum;
+
+	signum = SIGINT;
+	ASSERT_EXIT (repeattest (signum, true), ::testing::ExitedWithCode (0), "");
+}
+
+TEST (DISABLED_signal_manager_test, norepeat)
+{
+	int signum;
+
+	signum = SIGINT;
+	ASSERT_DEATH (repeattest (signum, false), "^boost signal handler$");
+}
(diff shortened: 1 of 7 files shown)


nano/lib/signal_manager.hpp
```diff
@@ -0,0 +1,89 @@
...
+	void log (const std::string);
+
+	/**
+	 * This is the actual handler that is registered with boost asio.
+	 * It logs any signals, calls the caller supplied function (if one is given) and sets the handler to repeat (or not).
+	 */
+	static void base_handler (nano::signal_manager::signal_descriptor descriptor, const boost::system::error_code & error, int signum);
+
+	/** reference to logger object that logs signal events */
+	nano::logger_mt * logger;
```
clemahieu: This should be nano::logger_mt & logger.

nano/nano_node/daemon.cpp
```diff
@@ -49,6 +43,7 @@ void nano_daemon::daemon::run (boost::filesystem::path const & data_path, nano::
 	{
 		config.node.logging.init (data_path);
 		nano::logger_mt logger{ config.node.logging.min_time_between_log_output };
+		sigman.set_logger (&logger);
```
clemahieu: This should be sigman.set_logger (logger) once the parameter is changed to a reference.

nano/lib/signal_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+	// ensure the signal set and descriptors live long enough
+	descriptor_list.push_back (descriptor);
+
+	// asynchronously listen for signals from this signal set
+	sigset->async_wait (boost::bind (nano::signal_manager::base_handler, descriptor, _1, _2));
+
+	log (boost::str (boost::format ("Registered signal handler for signal %d") % signum));
+}
+
+void nano::signal_manager::set_logger (nano::logger_mt * logger_a)
```
clemahieu: Parameter here should be a reference, tying in to the comment below.

nano/lib/signal_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+nano::signal_manager::signal_descriptor::signal_descriptor (std::shared_ptr<boost::asio::signal_set> sigset_a, signal_manager & sigman_a, std::function<void (int)> handler_func_a, bool repeat_a) :
+	sigset (sigset_a), sigman (sigman_a), handler_func (handler_func_a), repeat (repeat_a)
+{
+}
+
+void nano::signal_manager::register_signal_handler (int signum, std::function<void (int)> handler, bool repeat)
+{
+	// create a signal set to hold the mapping between signals and signal handlers
+	std::shared_ptr<boost::asio::signal_set> sigset (new boost::asio::signal_set (ioc, signum));
```
clemahieu: This should make use of the make_shared function. This is an exception-safe way to ensure no leaks:

auto signet = std::make_shared<boost::asio::signal_set> (ioc, signum);

nano/lib/signal_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+	// a signal descriptor holds all the data needed by the base handler including the signal set
+	// working with copies of a descriptor is OK
+	signal_descriptor descriptor (sigset, *this, handler, repeat);
+
+	// ensure the signal set and descriptors live long enough
+	descriptor_list.push_back (descriptor);
+
+	// asynchronously listen for signals from this signal set
+	sigset->async_wait (boost::bind (nano::signal_manager::base_handler, descriptor, _1, _2));
```
clemahieu: Using boost::bind is for pre-c++11 compatibility and can be replaced with a lambda:

sigset->async_wait ([this, descriptor] (const boost::system::error_code & error, int signum) { base_handler (descriptor, error, signum); });

nano/lib/signal_manager.hpp
```diff
@@ -0,0 +1,89 @@
...
+	/** stops the signal manager io context and wait for teh thread to finish */
+	~signal_manager ();
+
+	/** Register a handler for a signal to be called from a safe context.
+	 *  The handler will be called from one of the contexts of the io context supplied.
+	 */
+	void register_signal_handler (int signum, std::function<void (int)> handler, bool repeat);
+
+	/** set the logger to use, without a logger, logs go to std:cerr */
+	void set_logger (nano::logger_mt * logger_a);
```
clemahieu: This can be passed by reference.

nano/lib/signal_manager.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+		/** indicates if the signal handler should continue handling a signal after receiving one */
+		bool repeat;
+	};
+
+	/**
+	 * Logging function of signal manager. At the start, until a logger is given, it logs to standard error.
+	 * If it is given an logger then it logs to it. This is done because at startup we dont have a logger.
+	 */
+	void log (const std::string);
```
clemahieu: The string should be passed by reference.

nano/lib/signal_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+std::function<void (int)> nano::signal_manager::get_debug_files_handler (void)
+{
+	return [] (int) {
+		nano::dump_crash_stacktrace ();
+		nano::create_load_memory_address_files ();
+	};
+}
+
+void nano::signal_manager::log (const std::string msg)
```
clemahieu: The string should be passed by reference otherwise its copied when calling this function.

Since const can be used in multiple points in a type declaration, we stick it to the right of the thing we're consting:

void nano::signal_manager::log (std::string const & msg)

nano/lib/signal_manager.cpp
```diff
@@ -0,0 +1,101 @@
...
+		// call the user supplied function, if one is provided
+		if (descriptor.handler_func)
+		{
+			descriptor.handler_func (signum);
+		}
+
+		// continue asynchronously listening for signals from this signal set
+		if (descriptor.repeat)
+		{
+			descriptor.sigset->async_wait (boost::bind (base_handler, descriptor, _1, _2));
```
clemahieu, line 76: boost::bind should be replaced with a lambda as above.

nano/lib/signal_manager.hpp
dsiganos: I made it a pointer so that it can be unassigned. If it is unassigned then we can use signal manager before we have a logger.
If the logger pointer is null then stderr is used for logging.

nano/core_test/signal_manager.cpp
```diff
@@ -0,0 +1,112 @@
...
+{
+	std::cerr << "boost signal handler " << signum << std::endl
+			  << std::flush;
+}
+
+static int wait_for_sig_received (int millisecs, int & sig_received)
+{
+	for (int i = 0; i < millisecs && sig_received == 0; i++)
+	{
+		usleep (1000);
```
clemahieu: This can be replaced with std::this_thread::sleep (std::chrono::microseconds (1));

nano/lib/signal_manager.hpp
dsiganos: I can also use std::optional reference.

nano/core_test/signal_manager.cpp
dsiganos: Fixed.

nano/lib/signal_manager.hpp
dsiganos: Fixed

clemahieu (dismissed):

clemahieu (approved):

PR #3221, opened 2021-04-21
Refactory of block processor thread

thsfs:
Moved block processor's thread from node class to block processor class.


nano/node/blockprocessor.hpp
```diff
@@ -88,6 +89,7 @@ class block_processor final
 	nano::write_database_queue & write_database_queue;
 	nano::mutex mutex{ mutex_identifier (mutexes::block_processor) };
 	nano::state_block_signature_verification state_block_signature_verification;
+	std::thread thread;
```
theohax: Maybe `processing_thread`? Or any other name just to actually say something about the purpose.

theohax (dismissed):
Looks good!

nano/node/blockprocessor.cpp
```diff
@@ -44,11 +44,19 @@ nano::block_processor::block_processor (nano::node & node_a, nano::write_databas
...
+	thread = std::thread ([this] () {
+		nano::thread_role::set (nano::thread_role::name::block_processing);
+		this->process_blocks ();
+	});
 }
 
 nano::block_processor::~block_processor ()
 {
 	stop ();
+	if (thread.joinable ())
```
dsiganos: Should the thread join be done in the stop function maybe? I am simply wondering and sharing my thoughts.
This is my thinking:
* The join used to be done in the stop function of node class and now it has moved in the destructor.
* The class state_block_signature_verification joins its thread at the stop function too.

clemahieu: Putting the thread join inside ::stop makes it block between when the stop is signaled and when the thread actually exits. The idea of putting it in the destructor means we can signal all the different components to stop, and let each thread shut down simultaneously.

nano/node/blockprocessor.hpp
thsfs: Way better than just "thread".

theohax (approved):

Issue #3235, opened 2021-04-26
Failing test case: qt_test wallet.status


dsiganos:
The test is failing because it is waiting for a wallet to attain the synchronising state.
Synchronising state is when a bootstrap attempt is in progress.

A boostrap attempt starts but it is abandoned immediately due to "pulling" being false.
The failure is caused by the commit:
https://github.com/nanocurrency/nano-node/commit/a3f15515431769467b9c7be1dbf7d0163cbdad64
a3f15515 ("Combine alarm and worker threads into a pool (#2871)", 2021-02-01)

In software before that commit, the synchronising state stays in the wallet's active status just about long enough to be noticed by the test function. That's due to more context switches happening due to the use of the alarm class and Qt processEvents not processing both the attempt started and attempt ended in one call.

After that commit, the processing of events is more efficient and the synchronising is added and removed in one call of processEvents and the test function does not see the synchronising state spike.

But, having said all that, I believe the commit above is not the problem but a trigger for the problem that had already existed. It seems to me that the test case is wrong and needs fixing.

dsiganos:
Possible solution. Calling it a possible solution because I do not think it is a good solution.
https://github.com/dsiganos/nano-node/commit/f7f1134d8b9cfce8a62eb5268212f9a48f3a89cd
It seems wrong to bypass the status check and add an observer directly into the node.

PR #3236, opened 2021-04-26
Qt test wallet status with peer

dsiganos:
Create a new test, wallet.status_with_peer, modelled on the test case qt_test wallet.status.

The expected behaviour of the test wallet.status is not defined so it cannot be fixed at the moment and we have bigger problems to work on. See #3235

So, I disable test wallet.status and introduce a new test wallet.status_with_peer, which performs a similar test but with defined behaviour.

### DIFF
--- a/nano/qt_test/qt.cpp
+++ b/nano/qt_test/qt.cpp
@@ -67,15 +67,14 @@ TEST (wallet, status_with_peer)
 {
 	nano_qt::eventloop_processor processor;
 	nano::system system (2);
-	auto wallet_l (system.nodes[0]->wallets.create (nano::random_wallet_id ()));
+	auto wallet_l = system.nodes[0]->wallets.create (nano::random_wallet_id ());
 	nano::keypair key;
 	wallet_l->insert_adhoc (key.prv);
-	auto wallet (std::make_shared<nano_qt::wallet> (*test_application, processor, *system.nodes[0], wallet_l, key.pub));
+	auto wallet = std::make_shared<nano_qt::wallet> (*test_application, processor, *system.nodes[0], wallet_l, key.pub);
 	wallet->start ();
 	auto wallet_has = [wallet] (nano_qt::status_types status_ty) {
 		return wallet->active_status.active.find (status_ty) != wallet->active_status.active.end ();
 	};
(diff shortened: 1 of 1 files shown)


nano/qt_test/qt.cpp
```diff
@@ -60,6 +62,38 @@ TEST (wallet, status)
 	ASSERT_TRUE (wallet_has (nano_qt::status_types::disconnected));
 }
 
+// this test is modelled on wallet.status but it introduces another node on the network
+TEST (wallet, status_with_peer)
+{
+	nano_qt::eventloop_processor processor;
+	nano::system system (2);
+	auto wallet_l (system.nodes[0]->wallets.create (nano::random_wallet_id ()));
```
clemahieu: For a new test, we should use assignment initialization:

auto wallet_l = system.nodes[0]->wallets.create (nano::random_wallet_id ());

nano/qt_test/qt.cpp
```diff
@@ -60,6 +62,38 @@ TEST (wallet, status)
...
+	nano::keypair key;
+	wallet_l->insert_adhoc (key.prv);
+	auto wallet (std::make_shared<nano_qt::wallet> (*test_application, processor, *system.nodes[0], wallet_l, key.pub));
+	wallet->start ();
+	auto wallet_has = [wallet] (nano_qt::status_types status_ty) {
+		return wallet->active_status.active.find (status_ty) != wallet->active_status.active.end ();
+	};
+	system.nodes[0]->network.udp_channels.insert (nano::endpoint (boost::asio::ip::address_v6::loopback (), 10000), 0);
```
clemahieu: Let's use a TCP variant since we want to remove the UDP code soon.

nano/qt_test/qt.cpp
dsiganos: OK

nano/qt_test/qt.cpp
dsiganos: Removed the line alltogether since it turned out it is not necessary.
 	system.nodes[0]->network.udp_channels.insert

clemahieu (approved):

PR #3253, opened 2021-04-30
Extend block_store.unchecked_begin_search test

SergiySW:


SergiySW:
Seems related to: https://github.com/nanocurrency/nano-node/pull/2856/files

zhyatt:
@SergiySW Can you explain more about what this test is targeting and what remains?

SergiySW:
@zhyatt there was some RocksDB failure in iterators. Only in "unchecked_keys" RPC, starting with iterator like this one:
https://github.com/nanocurrency/nano-node/pull/3253/files#diff-bfaabe7ea1654a097673911617a55f8e2759509d10e2d0d6c08bf224c5c347efR594.
So for RocksDB backend this RPC always returns 0 entries regardless of unchecked entries count

zhyatt:
Great, thanks @SergiySW

dsiganos:
I am closing this PR for the following reasons:
* it is too old
* it does not explain enough why it is done and what exactly it is 
* it is about rocksdb which we do not recommend using anymore

PR #3265, opened 2021-05-06
Election flush

clemahieu:
This corrects the implementation of flushing the election scheduler. Flushing will terminate when either:
1) The election scheduler is stopped
2) there are no more elections to schedule
3) the active transaction container has no vacancy.


nano/core_test/election_scheduler.cpp
```diff
@@ -118,3 +121,25 @@ TEST (election_scheduler, no_vacancy)
 	auto election4 = node.active.election (block1->qualified_root ());
 	ASSERT_NE (nullptr, election4);
 }
+
+TEST (election_scheduler, flush_vacancy)
```
thsfs, line 126: What assertions does it check?

thsfs (commented):
I'm not able to evaluate the whole functionality, but the changes seem ok. My only question is regarding the assertion check for the flush_vacancy unit test.

theohax (dismissed):

nano/core_test/election_scheduler.cpp
clemahieu: It's testing that election_scheduler::flush does not hang when it cannot activate an election.

thsfs (approved):

PR #3279, opened 2021-05-12
Improve confirmation_height.gap_live core test

SergiySW:
Prevent possible auto-received blocks conflicts with manually created blocks

### DIFF
--- a/nano/core_test/confirmation_height.cpp
+++ b/nano/core_test/confirmation_height.cpp
@@ -310,11 +310,11 @@ TEST (confirmation_height, gap_live)
 		system.wallet (1)->insert_adhoc (destination.prv);
 
 		nano::genesis genesis;
-		auto send1 (std::make_shared<nano::state_block> (nano::genesis_account, genesis.hash (), nano::genesis_account, nano::genesis_amount - nano::Gxrb_ratio, destination.pub, nano::dev_genesis_key.prv, nano::dev_genesis_key.pub, 0));
+		auto send1 (std::make_shared<nano::state_block> (nano::genesis_account, genesis.hash (), nano::genesis_account, nano::genesis_amount - 1, destination.pub, nano::dev_genesis_key.prv, nano::dev_genesis_key.pub, 0));
 		node->work_generate_blocking (*send1);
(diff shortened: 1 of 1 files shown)


dsiganos:
Is this change just simply sending much smaller amounts?

SergiySW:
Yes, to be less then default auto receive (can be nano::xrb_ratio as well, or can change receive minimum in config, or don't add destination private key to local wallet)

dsiganos:
Is this solving a problem or is it simply making the numbers easier?

SergiySW:
> Is this solving a problem or is it simply making the numbers easier?

It should solve this test failures (unchecked == 1 instead of expected 0). Because auto-receiving can generate slightly different chain & manual blocks open, receive1, receive2 can be considered as forks.
I.e. if there is fork conflict between receive1 & auto-received recieve1B, then receive2 will remain in unchecked table (previous block for it == receive1 wasn't processed).

dsiganos:
For the record, I do not know what auto-receive is. I assume it is the feature of the node to publish receive blocks when it notices that there is a confirmed pending send for an account in its wallet.

But I still do not understand how simply changing the amounts sent, fixes the problem.

SergiySW:
Correct.

And very small values are not received without config change or manual/script intervention:
https://github.com/nanocurrency/nano-node/blob/9982779c1bbf8c108ade266352f62abab5ca9083/nano/node/wallet.cpp#L1187
https://github.com/nanocurrency/nano-node/blob/9982779c1bbf8c108ade266352f62abab5ca9083/nano/node/nodeconfig.hpp#L53

dsiganos:
OK, I think I am starting to understand now. By sending very small amounts, we ensure that auto-receive will not publish receive blocks for the money sent and it will not interfere with the manual receive blocks that the test publishes.

However, it seems to me that to depend on such a little detail is too subtle. In my opinion, the test should either explicitly disable auto-receive or add a comment explaining the auto-receive problem.

Some questions:
* Is the auto-receive problem due to auto-receive receiving the blocks in an order which is different to the order the test expects?
* Could we change the test to depend on auto-receive and not publish its own receive blocks?

dsiganos (approved):

clemahieu (approved):

Issue #3255, opened 2021-04-30
Break up the block_store tables into their respective classes


thsfs:
The current implementation of block_store class is too big. It could be improved by cutting off any duplicated code for table handling, and also by making the classes to follow the SRP.

PR #3287, opened 2021-05-14
Frontier store break up

thsfs:
Moves out the frontier functions to a frontier store class. It is part of issue #3255

### DIFF
--- a/nano/node/lmdb/lmdb.cpp
+++ b/nano/node/lmdb/lmdb.cpp
@@ -361,7 +361,7 @@ void nano::mdb_store::upgrade_v14_to_v15 (nano::write_transaction & transaction_
 
 		nano::mdb_val value{ data.size (), (void *)data.data () };
 		auto s = mdb_put (env.tx (transaction_a), state_blocks_new, nano::mdb_val (hash), value, MDB_APPEND);
-		release_assert_success (s);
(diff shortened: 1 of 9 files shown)


nano/node/lmdb/lmdb.cpp
```diff
@@ -361,7 +361,7 @@ void nano::mdb_store::upgrade_v14_to_v15 (nano::write_transaction & transaction_
 
 		nano::mdb_val value{ data.size (), (void *)data.data () };
 		auto s = mdb_put (env.tx (transaction_a), state_blocks_new, nano::mdb_val (hash), value, MDB_APPEND);
-		release_assert_success (s);
+		release_assert_success<MDB_val, mdb_store> ((*this), s);
```
thsfs: The way it was implemented, needed to pass the template values. Please check whether we want this.

nano/secure/blockstore_partial.hpp
```diff
@@ -27,15 +23,35 @@ namespace nano
 template <typename Val, typename Derived_Store>
 class block_predecessor_set;
 
+template <typename Val, typename Derived_Store>
```
thsfs, line 26: Attention here, moved the macro to a friend function declared outside class block_store_partial.

clemahieu (approved):

PR #3292, opened 2021-05-16
clang-format check against intial commit

argakiig:
Instead of checking clang-format against files changed over HEAD~1 this goes clear back to the initial commit showing any changes.
Also included were two changes required to pass


argakiig:
closed and re-openned to try and fix ci, which appears to have hung

argakiig:
closed in favor of #3395

PR #3297, opened 2021-05-20
Split election drop stats to be either overflow or expired

MajorChump:
At the moment it's not clear if elections are simply expiring after 5 minutes or being dropped due to the container overflowing. This should give us a better insight and allow us to monitor the situation.

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -349,39 +349,40 @@ void nano::active_transactions::request_confirm (nano::unique_lock<nano::mutex> 
 	}
 }
 
-void nano::active_transactions::cleanup_election (nano::unique_lock<nano::mutex> & lock_a, nano::election_cleanup_info const & info_a)
+void nano::active_transactions::cleanup_election (nano::unique_lock<nano::mutex> & lock_a, nano::election const & election)
 {
-	debug_assert (lock_a.owns_lock ());
-
-	if (!info_a.confirmed)
+	if (!election.confirmed ())
 	{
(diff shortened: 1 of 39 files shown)


nano/node/active_transactions.cpp
```diff
@@ -330,6 +330,10 @@ void nano::active_transactions::request_confirm (nano::unique_lock<nano::mutex>
 			}
 
 			// Locks active mutex, cleans up the election and erases it from the main container
+			if (!election_l->confirmed())
```
theohax: PR looks good! Don't forget to run `ci/clang-format-all.sh` so that formatting rules get applied.

theohax (dismissed):
LGTM!

nano/node/active_transactions.cpp
```diff
@@ -330,6 +330,10 @@ void nano::active_transactions::request_confirm (nano::unique_lock<nano::mutex>
 			}
 
 			// Locks active mutex, cleans up the election and erases it from the main container
+			if (!election_l->confirmed ())
```
SergiySW: Can it be `confirmed_l` ?

nano/node/active_transactions.cpp
```diff
@@ -316,6 +316,7 @@ void nano::active_transactions::request_confirm (nano::unique_lock<nano::mutex>
 	for (auto const & election_l : elections_l)
 	{
 		bool const confirmed_l (election_l->confirmed ());
+		unconfirmed_count_l += !confirmed_l;
```
MajorChump, line 319: This is needed for logging on L349

clemahieu (approved):

PR #3301, opened 2021-05-24
Refactor using guard clauses and add additional logging

MajorChump:


MajorChump:
New PR to be raised with improved logging and improved socket connection handling

PR #3302, opened 2021-05-24
Election cleanup cleanup

clemahieu:
Merges duplicate cleanup code, simplifies lock logic, and removes unneeded cleanup_info function.

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -347,39 +347,40 @@ void nano::active_transactions::request_confirm (nano::unique_lock<nano::mutex> 
 	}
 }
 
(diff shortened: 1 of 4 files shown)


nano/node/active_transactions.cpp
```diff
@@ -347,39 +347,40 @@ void nano::active_transactions::request_confirm (nano::unique_lock<nano::mutex>
...
 		}
 
-		if (!info_a.confirmed)
+		if (!election.confirmed ())
 		{
 			// Clear from publish filter
 			node.network.publish_filter.clear (block);
 		}
 	}
-	lock_a.lock ();
```
thsfs, line 382: Is this supposed to be removed?

thsfs (approved):

SergiySW (approved):

PR #3304, opened 2021-05-25
Increase soft limit for number of open file descriptors

theohax:


SergiySW:
I wonder if we can use/modify already existing code. Or at least create new function in the same place for consistency
https://github.com/nanocurrency/nano-node/blob/9a9168f823cb40479efa1f89c940b671cfa88e87/nano/nano_node/daemon.cpp#L74
https://github.com/nanocurrency/nano-node/blob/f7f83e79cbf2f6edf30460fcd77a4283bffa2d5e/nano/lib/utility.cpp#L38

theohax:
> get_filedescriptor_limit

Oh...sure, let me have a look. I hadn't noticed it.

theohax:
@SergiySW addressed, thanks.

clemahieu (approved):

PR #3305, opened 2021-05-25
Network peer fixes

MajorChump:
This resolves two issues which result in the same outcome. The node will continue to accept connections but none are handled. At some point the node stops accepting new connections probably because the previous ones haven’t been handled and connections timeout. 

To recreate: 

1.  Accept error, set file limits to 512 and run node it will occur within 30 minutes
2. tcp_incoming_connections_max, set to default of 1024 it will occur within 30 minutes

Using the network_logging branch and `watch -n 0.1 'cat node.log | grep Network'` both of the above tests will result in 

1. Network: Stopping to accept connections
2. Network: max_inbound_connections reached, unable to open new connection (primary)

Using the network_peer_fixes branch the following errors will appear in the logs

1. Network: Unable to accept connection: file limit reached
2. Network: max_inbound_connections reached, unable to open new connection

But the node should continue to operate and allow new connections.


nano/node/socket.cpp
```diff
@@ -238,55 +238,110 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+				boost::asio::post (this_l->strand, boost::asio::bind_executor (this_l->strand, [this_l, callback_a] () {
+					this_l->on_connection (callback_a);
 				}));
+				return;
 			}
-		}
+
+			if (this_l->is_system_error (ec_a))
+			{
+				boost::this_thread::sleep (boost::posix_time::milliseconds (500));
```
ghost: If we should sleep, need to make sure what is the proper way do to it from within a strand. V21.3 did it using a async_wait, when needed. Good if someone with more experience in this code can recommend the best way. 

This also applies to the case in the max incoming connection limit check (where I think we should sleep a bit as well).

nano/node/socket.cpp
```diff
@@ -238,55 +238,110 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+			this_l->node.logger.always_log ("Network: Stopping to accept connections. Node may need restarting.");
+		}));
 	}));
 }
 
+bool nano::server_socket::is_temporary_error (boost::system::error_code const ec_a)
+{
+	switch (ec_a.value ())
+	{
+#if EAGAIN != EWOULDBLOCK
```
ghost, line 326: Think no need to check here. Common practice is to check for both values, no matter the system.

nano/node/socket.hpp
```diff
@@ -124,5 +124,7 @@ class server_socket final : public socket
 	boost::asio::ip::tcp::endpoint local;
 	size_t max_inbound_connections;
 	void evict_dead_connections ();
```
ghost, line 126: Possible improvement: These don’t have to be part of the class. Can be free functions in the .cpp file.

nano/node/socket.cpp
```diff
@@ -238,55 +238,110 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
-							this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_success, nano::stat::dir::in);
-							this_l->connections.push_back (new_connection);
-						}
-						else
-						{
-							this_l->node.logger.try_log ("Unable to accept connection: ", ec_a.message ());
-						}
+				this_l->node.logger.always_log ("Network: max_inbound_connections reached, unable to open new connection");
+				this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_failure, nano::stat::dir::in);
+				boost::asio::post (this_l->strand, boost::asio::bind_executor (this_l->strand, [this_l, callback_a] () {
```
ghost: I think we should go back to V21.3 behavior and sleep for 2 seconds here. Keep the accept loop going, but reduce how often it’s called until the connections are not at peak capacity anymore.

nano/node/socket.cpp
```diff
@@ -238,55 +238,110 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+			//EINTR
+			//EINVAL
+			//ENOTSOCK
+			//EOPNOTSUPP
+			//EPROTO
+			//EPERM
+			// Check how the listener wants to handle these error
+			if (callback_a (new_connection, ec_a))
+			{
+				this_l->on_connection (callback_a);
```
ghost: Sometimes we call on_connection() directly, and sometimes via boost:asio::post. Can we align on one correct way?

nano/node/socket.cpp
MajorChump: Results in compilation error on systems which match as a switch can't have the same value twice

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+
+			if (this_l->is_system_error (ec_a))
+			{
+				return this_l->on_connection_requeue_delayed (callback_a);
+			}
+
+			// Unhandled errors above
+			//EBADF
+			//EFAULT
+			//EINTR
```
ghost: At least EINTR should lead to direct retry

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+				this_l->connections.push_back (new_connection);
+				callback_a (new_connection, ec_a);
+				return this_l->on_connection_requeue_instant (callback_a);
+			}
+
+			this_l->node.logger.always_log ("Network: Unable to accept connection: ", ec_a.message ());
+			this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_failure, nano::stat::dir::in);
+			if (this_l->is_temporary_error (ec_a))
+			{
+				return this_l->on_connection_requeue_instant (callback_a);
```
ghost: Confusing to have return statements with result of void functions. Think it's more clear to call the function, then call return standalone on the next line. Seen in a few places.

nano/node/socket.cpp
ghost: Ok. Using if statement would work. But would make the overall code more unclear I think, so agree this is better.

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
 
+void nano::server_socket::on_connection_requeue_instant (std::function<bool (std::shared_ptr<nano::socket> const &, boost::system::error_code const &)> callback_a)
+{
+	return on_connection (callback_a);
+}
+
+void nano::server_socket::on_connection_requeue_delayed (std::function<bool (std::shared_ptr<nano::socket> const &, boost::system::error_code const &)> callback_a)
+{
+	auto this_l (std::static_pointer_cast<nano::server_socket> (shared_from_this ()));
+	return node.workers.add_timed_task (std::chrono::steady_clock::now () + std::chrono::milliseconds (500), [this_l, callback_a] () {
```
ghost: No need with return statement since function is void return type.

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+			this_l->node.logger.always_log ("Network: Acceptor is not open");
+			return;
+		}
+
+		// Prepare new connection
+		auto new_connection = std::make_shared<nano::socket> (this_l->node, boost::none);
+		this_l->acceptor.async_accept (new_connection->tcp_socket, new_connection->remote,
+		boost::asio::bind_executor (this_l->strand,
+		[this_l, new_connection, callback_a] (boost::system::error_code const & ec_a) {
+			this_l->evict_dead_connections ();
```
dsiganos, line 252: Add an empty line after evict_dead_connections for readability

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+
+			if (!ec_a)
+			{
+				// Make sure the new connection doesn't idle. Note that in most cases, the callback is going to start
+				// an IO operation immediately, which will start a timer.
+				new_connection->checkup ();
+				new_connection->start_timer (this_l->node.network_params.network.is_dev_network () ? std::chrono::seconds (2) : this_l->node.network_params.node.idle_timeout);
+				this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_success, nano::stat::dir::in);
+				this_l->connections.push_back (new_connection);
+				callback_a (new_connection, ec_a);
```
dsiganos, line 269: Why are we ignoring the result of callback_a, we should use the return code to decide if we should carry on accepting connections.

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+				return this_l->on_connection_requeue_instant (callback_a);
+			}
+
+			this_l->node.logger.always_log ("Network: Unable to accept connection: ", ec_a.message ());
```
dsiganos, line 274: We are leaking new_connection after this point.

nano/node/socket.cpp
MajorChump: Agreed, will update now.

nano/node/socket.cpp
MajorChump: Agreed, will update now.

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+			this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_failure, nano::stat::dir::in);
+			if (this_l->is_temporary_error (ec_a))
+			{
+				return this_l->on_connection_requeue_instant (callback_a);
+			}
+
+			if (this_l->is_system_error (ec_a))
```
dsiganos, line 282: I do not agree with the concept of system error.
I'd like the system error checking section removed.
I do not see how it adds any value.
It adds unnecessary complexity.

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -22,22 +22,17 @@ void nano::bootstrap_listener::start ()
...
 		throw std::runtime_error (ec.message ());
 	}
 	debug_assert (node.network.endpoint ().port () == listening_socket->listening_port ());
 	listening_socket->on_connection ([this] (std::shared_ptr<nano::socket> const & new_connection, boost::system::error_code const & ec_a) {
-		bool keep_accepting = true;
 		if (ec_a)
 		{
-			keep_accepting = false;
-			this->node.logger.try_log (boost::str (boost::format ("Error while accepting incoming TCP/bootstrap connections: %1%") % ec_a.message ()));
+			return false;
```
dsiganos, line 32: The callback should always return true (continue to accept).
There is no point stopping to accept accepts.
We have the half a second throttling to avoid fast loops.

nano/node/socket.cpp
```diff
@@ -238,55 +238,117 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
+				this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_success, nano::stat::dir::in);
+				this_l->connections.push_back (new_connection);
+				callback_a (new_connection, ec_a);
+				this_l->on_connection_requeue_instant (callback_a);
+				return;
+			}
+
+			this_l->node.logger.always_log ("Network: Unable to accept connection: ", ec_a.message ());
+			this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_failure, nano::stat::dir::in);
+			if (this_l->is_temporary_error (ec_a))
```
dsiganos, line 276: Since I am proposing that the callback always returns "continue accepting"  then there is no need for the temporary errors checking either. Let's remove the temporary error checking.

nano/node/socket.cpp
ghost, line 274: Same comment as above regarding leaking connection

nano/node/socket.cpp
```diff
@@ -238,55 +238,112 @@ void nano::server_socket::on_connection (std::function<bool (std::shared_ptr<nan
...
-					}
-					else
-					{
-						this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_accept_failure, nano::stat::dir::in);
-						boost::asio::post (this_l->strand, boost::asio::bind_executor (this_l->strand, [this_l, callback_a] () {
-							this_l->on_connection (callback_a);
-						}));
-					}
-				}));
+				this_l->node.logger.always_log ("Network: max_inbound_connections reached, unable to open new connection");
```
ghost, line 255: As I read it, there is no leak as new_connection is a shared_ptr. When all references have gone out of scope it will be freed automatically and the destructor will close the socket. You see somewhere a reference is kept on to?

dsiganos: Correct, my mistake.

nano/node/socket.cpp
dsiganos: Correct.

nano/node/bootstrap/bootstrap_server.cpp
dsiganos: On reflection, it does make sense to keep the temporary error check.
We do not want to sleep for a half a second just because of EAGAIN.
I would ideally like all errors to be reported to the callback and the callback to decide if we should carry on or not but it is not abnormal to treat EAGAIN a special maskable error. So I suggest we keep the temporary errors check but only check for EAGAIN/EWOULDBLOCK.

ghost: My thougts around this. I think there is a point. If it's not considered "temporary" (retry directly) or "system" (retry with some delay) (maybe some update of the terminology could make it clearer?) then it's a "permanent" error related to the socket/ip address changed etc. (e.g. EBADF, ENOTSOCK etc), which would for proper handling require tearing down the socket and recreating it. There is no such functionality from before, and I think we decided it is not in scope for this fix. So in those "permanent" error cases the option is to stop accepting (as was done before for ANY error), or keep accepting with no chance of it ever working again. Both will result in loss of ability to accept any more connections until the process is restarted. 

I feel both ways could work for this "hotfix", and we should implement the one with least amount of changed code/flows needed.

MajorChump:
Duplicate of https://github.com/nanocurrency/nano-node/pull/3315

PR #3306, opened 2021-05-25
Account store break up

thsfs:
Moves out the account functions to an account store class. It is part of issue #3255

### DIFF
--- a/nano/node/common.cpp
+++ b/nano/node/common.cpp
@@ -1246,7 +1246,7 @@ nano::error nano::telemetry_data::serialize_json (nano::jsonconfig & json, bool 
 	json.put ("block_count", block_count);
 	json.put ("cemented_count", cemented_count);
 	json.put ("unchecked_count", unchecked_count);
-	json.put ("count", account_count);
+	json.put ("account_count", account_count);
 	json.put ("bandwidth_cap", bandwidth_cap);
 	json.put ("peer_count", peer_count);
 	json.put ("protocol_version", protocol_version);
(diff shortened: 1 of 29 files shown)


nano/node/common.cpp
```diff
@@ -1246,7 +1246,7 @@ nano::error nano::telemetry_data::serialize_json (nano::jsonconfig & json, bool
 	json.put ("block_count", block_count);
 	json.put ("cemented_count", cemented_count);
 	json.put ("unchecked_count", unchecked_count);
-	json.put ("account_count", account_count);
```
thsfs: ops, this wasn't supposed to be changed. I'll fix.

nano/secure/blockstore.hpp
```diff
@@ -632,13 +632,28 @@ class frontier_store
 	virtual void for_each_par (std::function<void (nano::read_transaction const &, nano::store_iterator<nano::block_hash, nano::account>, nano::store_iterator<nano::block_hash, nano::account>)> const & action_a) const = 0;
 };
 
+class account_store
```
dsiganos, line 641: Can you add some inline comments explaining a little what each of the methods of account_stores does?

thsfs: Added a simple comment there while I don't have a better one. After these break up PRs for each class I'm intending to do another one placing each class in a single file and starting a second round of code organization.

clemahieu (dismissed):

clemahieu (approved):

PR #3310, opened 2021-05-27
Warn if the limit for open FDs is lower than recommended

theohax:
@SergiySW as per discussed :D


SergiySW (approved):

zhyatt:
This has been noted in the docs here: https://docs.nano.org/integration-guides/advanced/#known-issues

PR #3312, opened 2021-05-27
Remove full-check for triggering successive active blocks as this is handled by the scheduler.

jamescoxon:
Current code has a low active election check where if active elections > 500 it doesn’t allow sequential elections along a chain. This patch removes low active election check as this is limiting progression through the backlog (its currently relevant post 2021 spam but might well need to be reimplemented at a later date).


clemahieu:
Since this is a behavior change it seems better suited for v23 though the commit is based on v22. The PR was based on develop, that's why it had a funky change list.

Removing this code also didn't break any tests which means it didn't have any test coverage. We need to validate there is a test covering the sequential election case even when the active container is full, this should be fairly easy to implement by setting the active_transaction container size to 1 and disabling the ongoing_backlog_population. This will cause a sequence of transactions to be confirmed iff it's activating sequences correctly.

clemahieu (approved):

PR #3314, opened 2021-05-27
Remove election::blocks precondition.

clemahieu:
election::blocks is now used by cleanup_election, not just in unit tests so remove debug_assert precondition.


dsiganos:
@SergiySW did it already. I just merged it into develop. It's what I was saying this morning, by the time we think it, Sergei has already done it! :-D
https://github.com/nanocurrency/nano-node/pull/3311

PR #3324, opened 2021-06-07
Depth-first search ledger walking

theohax:
This is the first iteration of the algorithm. Planning to add disk-based hashtables to it and bloom filters at some point but first to get it stable, well tested and profiled.


nano/node/ledger_walker.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include <utility>
+
+nano::ledger_walker::ledger_walker (nano::ledger const & ledger_a) :
+	ledger{ ledger_a },
+	walked_blocks{},
+	blocks_to_walk{}
+{
+	if (ledger.store.init_error ())
+	{
+		throw nano::error{ "Ledger store initialization error" };
```
clemahieu: This should be a debug_assert since it's a correctness issue the caller can check. We don't use exceptions so if this had needed to signal an error it should either return it or use a by-reference error code.

nano/node/ledger_walker.cpp
```diff
@@ -0,0 +1,70 @@
...
+			enqueue_block (ledger.store.block_get (transaction, block->previous ()));
+		}
+
+		if (block->sideband ().details.is_receive)
+		{
+			enqueue_block (ledger.store.block_get (transaction, block->link ().as_block_hash ()));
+		}
+	}
+
+	walked_blocks.clear ();
```
clemahieu: It doesn't seem right to clear walked_blocks and fully reset block_to_walk. I assume the swap is to free the memory. It seems like both should be swapped or neither.

It also needs a unit test to make sure multiple walks using a single ledger_walker instance works correctly.

nano/core_test/ledger_walker.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include <nano/node/testing.hpp>
+#include <nano/test_common/testutil.hpp>
+
+#include <gtest/gtest.h>
+
+#include <numeric>
+
+using namespace std::chrono_literals;
+
+TEST (leger_walker, first_test)
```
dsiganos: TYPO: leger -> ledger

nano/core_test/ledger_walker.cpp
```diff
@@ -0,0 +1,79 @@
...
+	ASSERT_FALSE (last_destination_read_error);
+
+	/*
+    This is how we expect chains to look like (for 3 accounts and 10 amounts to be sent).
+    k1: Gx     SEND     02     SEND     05     SEND     08     SEND
+    k2: Gx     00       SEND   03     SEND     06     SEND     09
+    k3: Gx     01       SEND   04     SEND     07     SEND
+    */
+
+	std::vector<nano::uint128_t> amounts_expected{ nano::Mxrb_ratio + 9, nano::Mxrb_ratio + 8, nano::Mxrb_ratio + 7, nano::Mxrb_ratio + 4, nano::Mxrb_ratio + 3,
```
dsiganos: For improved readability, you can use
```
node_config.receive_minimum = 1;
```
to avoid having to multiply all amounts by nano::Mxrb_ratio.

nano/node/ledger_walker.cpp
theohax: Additionally to the clear, a shrink_to_fit would do nicely, but I don't think that's part of the unordered_set API. Yeah, perhaps swapping both would be a good idea.

CMakeLists.txt
```diff
@@ -378,6 +378,23 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules")
 find_package(Boost 1.70.0 REQUIRED COMPONENTS filesystem log log_setup thread
                                               program_options system)
 
+# Diskhash
```
theohax: About to change these few lines to actually use cmake for building the library -- current form was just easier for testing.

nano/node/ledger_walker.cpp
```diff
@@ -0,0 +1,122 @@
...
+{
+	if (add_to_walked_blocks (block_hash_a))
+	{
+		blocks_to_walk.emplace (std::move (block_hash_a));
+	}
+}
+
+void nano::ledger_walker::enqueue_block (std::shared_ptr<nano::block> const & block_a)
+{
+	if (block_a)
```
clemahieu: This seems like it should be a debug_assert. The caller can filter nullptr as a precondition.

theohax:
Adding @thsfs at reviewers -- let me know if integrating diskhash from their public repository and building it like this makes sense to you too. And, of course, as per discussed, having our own fork of it does not sound like a bad idea :D.

thsfs:
> Adding @thsfs at reviewers -- let me know if integrating diskhash from their public repository and building it like this makes sense to you too. And, of course, as per discussed, having our own fork of it does not sound like a bad idea :D.

Having our own repository for this is not required for now. It'd be in case we need to send changes to it (when we do), or for safety, as this code seems to be maintained by a few people.

thsfs (dismissed):

nano/core_test/ledger_walker.cpp
```diff
@@ -0,0 +1,220 @@
...
+	nano::keypair key{};
+	system.wallet (0)->insert_adhoc (key.prv);
+	ASSERT_TRUE (system.wallet (0)->send_action (nano::dev_genesis_key.pub, key.pub, 1));
+	ASSERT_TIMELY (3s, 5 == node->ledger.cache.cemented_count);
+
+	const auto transaction = node->ledger.store.tx_begin_read ();
+	nano::account_info account_info{};
+	ASSERT_FALSE (node->ledger.store.account.get (transaction, key.pub, account_info));
+
+	//    TODO: check issue with account head
```
thsfs, line 110: I'd prefer not adding it as it won't run, rather we could add an issue on GitHub to track this.

theohax: These are a couple of extra type of checks for unit tests that for some reason do not work as I expected them to. The ledger walker is most likely going to be changed again soon anyways (in different PR) so I don't think this would sit around too much commented out, but yeah maybe I should just take it out? @clemahieu what do you think?

thsfs (dismissed):

nano/node/ledger_walker.cpp
theohax: Addressed, thanks

theohax:
@thsfs @clemahieu one more review/approval after last pushes please, thx.

thsfs (approved):

clemahieu (approved):

PR #3326, opened 2021-06-09
Removing redundant check in telemetry limiter.

clemahieu:
Time points are default constructed at the lowest time point which will already pass the duration check making it redundant.


dsiganos:
Similar logic exists in udp.cpp.

dsiganos (approved):

SergiySW (approved):

PR #3329, opened 2021-06-11
Block store break up

thsfs:
As part of the issue #3255:
- Moves out version store functions to a version class.
- Splits block_store class into two classes, store / block_store class. This last now just have block related functions.


nano/secure/store_partial.hpp
```diff
@@ -0,0 +1,217 @@
...
+	bool exists (nano::transaction const & transaction_a, tables table_a, nano::db_val<Val> const & key_a) const
+	{
+		return static_cast<const Derived_Store &> (*this).exists (transaction_a, table_a, key_a);
+	}
+
+	int const minimum_version{ 14 };
+
+protected:
+	nano::network_params network_params;
+	int const version_number{ 21 };
```
thsfs, line 136: Because it was conflicting with version store, changed the variable name to version_number.

clemahieu (dismissed):

clemahieu (approved):

PR #3331, opened 2021-06-13
Moving the check "using_rocksdb_in_tests" on to rocksdb_config

clemahieu:
Moving the check "using_rocksdb_in_tests" onto rocksdb_config, which is logically where it belongs. Also, this moves the check from the node library into the nano_lib library. This addresses a circular reference where test code is referencing the node library.


nano/node/node.cpp
```diff
@@ -1781,7 +1781,7 @@ nano::node_wrapper::node_wrapper (boost::filesystem::path const & path_a, boost:
 	}
 
 	auto & node_config = daemon_config.node;
-	node_config.peering_port = nano::get_available_port ();
+	//node_config.peering_port = nano::get_available_port ();
```
thsfs: Can we just remove this?

clemahieu: Uncommented the line.

thsfs (approved):

Issue #3330, opened 2021-06-12
Limit and add optional "offset" & "count" parameters for "delegators" RPC


shryder:
### Summary

Limit the amount of delegators returned by the "delegators" RPC and add the optional parameters "offset" and "count".

### What problem would be solved by this feature?

If I send multiple "delegators" RPC commands to my node for the rep NF#1 for example (it has ~16 million delegators)
1. If the server is not strong enough, it would crash after few requests. 100% disk and cpu load. (Is this a small attack vector maybe? create millions of dust accounts and delegate to nf#1 and call delegators rpc on public nodes)
2. If the server is strong enough, it would take a really long time to fetch all delegators.

The second problem is for services having to cache this data instead of getting up-to-date information through the RPC when desired.

### Are there any previous requests for this feature?

I don't know

### Do you have a suggested solution?

So my suggestion is 
1. Set a limit to the amount of delegators returned by this RPC command 
2. Add the optional fields "offset" and "count" to facilitate viewing this data. 
3. Exclude dust amounts from the response (make the dust amount configurable maybe)

The first suggestion is so the weak nodes don't crash 
The second and third is so the user/service doesn't have to wait a long time to get the current delegators list. 

Explorers currently make the request every once in a while and cache the delegators instead of showing live data.

### If this feature is approved, would you be willing to submit a pull request with the solution?

I would

shryder:
The `account_history` RPC action has the `count` parameter required. Should it be the same for this RPC?

PR #3333, opened 2021-06-13
Improve delegators rpc by adding "count", "start", and "threshold" parameters (#3330)

shryder:


nano/node/json_handler.cpp
```diff
@@ -2069,19 +2069,37 @@ void nano::json_handler::database_txn_tracker ()
...
+		}
+	}
 	if (!ec)
 	{
-		boost::property_tree::ptree delegators;
 		auto transaction (node.store.tx_begin_read ());
-		for (auto i (node.store.account.begin (transaction)), n (node.store.account.end ()); i != n; ++i)
+		boost::property_tree::ptree delegators;
+		uint64_t delegators_count = 0;
+		for (auto i (node.store.account.begin (transaction, head_account.number () + 1)), n (node.store.account.end ()); i != n && delegators.size () < count; ++i)
```
dstnbrkr: Where does the variable 'count' get assigned a value?

shryder: On line 2072:
`auto count (count_optional_impl ());`

nano/node/json_handler.cpp
```diff
@@ -2069,19 +2069,37 @@ void nano::json_handler::database_txn_tracker ()
...
+			ec = nano::error_common::bad_account_number;
+		}
+	}
 	if (!ec)
 	{
-		boost::property_tree::ptree delegators;
 		auto transaction (node.store.tx_begin_read ());
-		for (auto i (node.store.account.begin (transaction)), n (node.store.account.end ()); i != n; ++i)
+		boost::property_tree::ptree delegators;
+		uint64_t delegators_count = 0;
```
clemahieu: This variable seems assigned but otherwise unused.

shryder: True, let me remove it.

nano/node/json_handler.cpp
```diff
@@ -2069,19 +2069,35 @@ void nano::json_handler::database_txn_tracker ()
 void nano::json_handler::delegators ()
 {
 	auto account (account_impl ());
+	auto count (count_optional_impl ());
```
SergiySW: Probably can have default value like 1024 or something (i.e. `auto count (count_optional_impl (1024));`). By default it's 2^64-1

nano/node/json_handler.cpp
```diff
@@ -2069,19 +2069,35 @@ void nano::json_handler::database_txn_tracker ()
...
+	auto threshold (threshold_optional_impl ());
+	auto head_str (request.get_optional<std::string> ("head"));
+
+	nano::account head_account (0);
+	if (head_str)
```
SergiySW: Probably shorter variant
```
	if (!ec && head_str.is_initialized ())
	{
		head_account = account_impl (head_str.get ());
	}
```

SergiySW: It seems like we have at least 4 places using this logic, probably would make sense to use some separate function then in next PRs...

nano/node/json_handler.cpp
```diff
@@ -2069,19 +2069,35 @@ void nano::json_handler::database_txn_tracker ()
...
 			{
-				std::string balance;
-				nano::uint128_union (info.balance).encode_dec (balance);
-				nano::account const & account (i->first);
-				delegators.put (account.to_account (), balance);
+				if (info.balance.number () >= threshold.number ())
+				{
+					std::string balance;
+					nano::uint128_union (info.balance).encode_dec (balance);
+					nano::account const & account (i->first);
```
SergiySW: I think we have names conflict in this RPC (not related to PR, but would be good to resolve). Probably `auto account (account_impl ());` can be renamed to `auto representative` or this one (`nano::account const & account (i->first)`) to something else

nano/node/json_handler.cpp
```diff
@@ -2069,19 +2069,35 @@ void nano::json_handler::database_txn_tracker ()
 void nano::json_handler::delegators ()
 {
 	auto account (account_impl ());
+	auto count (count_optional_impl ());
+	auto threshold (threshold_optional_impl ());
+	auto head_str (request.get_optional<std::string> ("head"));
```
SergiySW: As for name I would probably prefer "start" or may be even "start_account" to differentiate it from head block & "head" in history. Similar name can be used later for "frontiers" & "ledger" I guess, because it's not really "account" we use in wallet actions or account_info

nano/node/json_handler.cpp
shryder: I don't think we can do this because the "account" parameter is already reserved for the representative's address:

```
auto representative (account_impl ());
```

SergiySW:
```
	if (!ec && start_account_text.is_initialized ())
	{
		start_account = account_impl (start_account_text.get ());
	}
```
In this way it uses `start_account_text` value & not "account" field.
https://github.com/nanocurrency/nano-node/blob/4d915c1fa3f7741299cae041f992dae283319e74/nano/node/json_handler.cpp#L232-L236
Probbaly need different function for such things later...

SergiySW: For current variant it still probably need `if (!ec && start_account_text)` to not replace error code if count or threshold are invalid

shryder: Ah I see now, makes sense, thanks.

>  Probably need different function for such things later...

Agree

SergiySW (dismissed):

SergiySW:
@Shryder I included small test rpc.delegators_parameters if you don't mind

SergiySW (approved):

PR #3335, opened 2021-06-15
This fixes an issue where nano_node couldn't be built with sanitizers

clemahieu:
This fixes an issue where nano_node couldn't be built with sanitizers enabled with clang because a post-build step would execute a nano_node command, get a sanitizer error, and fail. The failure originated inside the boost/any.hpp file. This seems to have been an issue in versions earlier than 1.69 though it appears to be an issue at least in 1.74. http://boost.2283326.n4.nabble.com/undefined-behavior-in-Boost-Program-Options-td4708585.html

This hooks the asan_blacklist file up as an exclusion list in the CMakeLists.txt file and excludes boost/any.hpp from sanitizer instrumentation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,8 +240,12 @@ else()
   if(${USING_ASAN} OR ${USING_ASAN_INT})
     if(${USING_ASAN_INT})
       add_compile_options(-fsanitize=address,undefined,integer)
+      add_compile_options(
+        -fsanitize-blacklist=${PROJECT_SOURCE_DIR}/asan_blacklist)
     else()
       add_compile_options(-fsanitize=address,undefined)
+      add_compile_options(
+        -fsanitize-blacklist=${PROJECT_SOURCE_DIR}/asan_blacklist)
     endif()
     add_definitions(-DED25519_NO_INLINE_ASM)
   elseif(${USING_TSAN})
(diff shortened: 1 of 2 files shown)


dsiganos:
Do we know what caused the difference in behaviour?
AHow can I recreate the build failure?

clemahieu:
> Do we know what caused the difference in behaviour?
> AHow can I recreate the build failure?

I'm able to produce this on Ubuntu 20.04 building nano_node with NANO_ASAN=1. After the linking step it will execute the post-build event and should give the UBSAN errors.

dsiganos:
I am not able to reproduce this problem and I am on ubuntu 20.04.
I do not see anything wrong with the change.

CMakeLists.txt
```diff
@@ -240,8 +240,12 @@ else()
   if(${USING_ASAN} OR ${USING_ASAN_INT})
     if(${USING_ASAN_INT})
       add_compile_options(-fsanitize=address,undefined,integer)
+      add_compile_options(
```
dsiganos, line 243: Can we move this compile option, one if statement level up, and avoid having to repeat it?

dsiganos (approved):

clemahieu:
Unable to reproduce on a different ubuntu system. Likely a build error.

PR #3336, opened 2021-06-15
Fix for network.peer_max_tcp_attempt

clemahieu:
This test starts several node instances which can be slow to start, especially with TEST_USE_ROCKSDB=1. This would cause connections to be cleaned up before the substantial ASSERTs were checked. This splits the node construction and connection loops so connections are done immediately before the network checks.


dsiganos (approved):
Looks good to me.

argakiig (approved):

PR #3343, opened 2021-06-16
Fix RPC "active_difficulty" returning empty string as "difficulty_trend" (#3323)

### LINKED ISSUE #3323: Bug: Mismatching type of difficulty_trend in deprecated active_difficulty call
### Summary

`difficulty_trend` returns a `string` rather than an `array`, which caused issues in some wallets relying on it, e.g gonano wallet:

> json: cannot unmarshal string into Go struct field .diffulty_trend of type []string

### Node version

Nano V22.0

### Build details

several community-ran nodes

### OS and version

several community-ran nodes

### Steps to reproduce the behavior

make an RPC call
```json
{
  "action": "active_difficulty",
  "include_trend": "true"
}
```

### Expected behavior

expected to produce a list as per [docs page](https://docs.nano.org/commands/rpc-protocol/#active_difficulty)

image

### Actual behavior

produces an empty string instead
```json
"difficulty_trend": ""
```

### Possible solution

return a single-item list
```json
"difficulty_trend": [
  "1.000000000000000"
]
```

### Supporting files

_No response_

shryder:


SergiySW:
I guess it will not pass without change here:
https://github.com/nanocurrency/nano-node/blob/develop/nano/rpc_test/rpc.cpp#L7119

shryder:
Yeah I just realized after I made the pull request, looking into it...

clemahieu (approved):

PR #3364, opened 2021-07-01
Message sink interface

clemahieu:
This change is an application of dependency injection to remove coupling between the tcp_channels and udp_channels classes and the network class. An interface called message_sink has been created and an implementation that forwards network message matching existing behavior has been put it its place.


nano/node/network.cpp
```diff
@@ -20,7 +25,7 @@ nano::network::network (nano::node & node_a, uint16_t port_a) :
 	node (node_a),
 	publish_filter (256 * 1024),
 	udp_channels (node_a, port_a),
```
thsfs: Why not passing it also to udp_channels, so you don't need to rely on node's dependency?

nano/node/network.hpp
```diff
@@ -118,6 +118,19 @@ class syn_cookies final
 class network final
 {
 public:
+	class inbound_t : public nano::transport::message_sink
```
thsfs: What about just passing a function to tcp/udp instead of a nested class?

thsfs (approved):

PR #3383, opened 2021-07-20
Cleaning up genesis_dev constants

clemahieu:
A number of constants used in unit testing are spread between ledger_constants and the nano:: namespace. Some are duplicated in testutil.hpp.

These changes unify the unit test constants under nano::dev.


cryptocode (approved):
LGTM, and successfully compiled/ran tests locally

PR #3387, opened 2021-07-22
wallet_pending and accounts_pending handle empty results differently

clemahieu:
Accounts_pending inserted an entry with an empty array and wallet_pending didn't insert a result at all. This caused wallet_pending to fail check_block_response_count.

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -142,7 +142,15 @@ boost::property_tree::ptree wait_response (nano::system & system, std::shared_pt
 void check_block_response_count (nano::system & system, std::shared_ptr<nano::rpc> const & rpc, boost::property_tree::ptree & request, uint64_t size_count)
 {
 	auto response (wait_response (system, rpc, request));
-	ASSERT_EQ (size_count, response.get_child ("blocks").front ().second.size ());
+	auto & blocks = response.get_child ("blocks");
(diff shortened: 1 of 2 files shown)


nano/rpc_test/rpc.cpp
```diff
@@ -142,7 +142,15 @@ boost::property_tree::ptree wait_response (nano::system & system, std::shared_pt
 void check_block_response_count (nano::system & system, std::shared_ptr<nano::rpc> const & rpc, boost::property_tree::ptree & request, uint64_t size_count)
 {
 	auto response (wait_response (system, rpc, request));
-	ASSERT_EQ (size_count, response.get_child ("blocks").front ().second.size ());
```
theohax, line 145: Looks like this used to be UB before with an empty blocks and then calling `.second` on the `.front()`. Did it ever crash though? 🤔

theohax (approved):

nano/rpc_test/rpc.cpp
clemahieu: Yes, it was periodically crashing or giving uninitialized results.

zhyatt:
@clemahieu This is all internal, no external RPC impacts, correct?

zhyatt:
> @clemahieu This is all internal, no external RPC impacts, correct?

Colin confirmed no external RPC impacts to note in documentation

PR #3388, opened 2021-07-23
Refer to empty/null accounts as nano::account{nullptr}

theohax:


nano/secure/common.hpp
```diff
@@ -401,13 +401,17 @@ namespace dev
 	extern nano::uint128_t genesis_amount;
 }
 
-/** Constants which depend on random values (this class should never be used globally due to CryptoPP globals potentially not being initialized) */
-class random_constants
+/** Constants which depend on random values (always used as singleton) */
+class hardened_constants
```
theohax, line 378: The new `hardened_constants`, adding a comment here for spotting this easily.

theohax:
Just rebased this so looking forward to getting it merged in a few days so that it doesn't get stale again.

@ reviewers: main addition here is `class hardened_constants` and a few usages of it, all the other changes are just sort of boilerplate

theohax:
Addressed code review as per discussed on Slack, please have another look when you can guys, thanks!

@clemahieu @dsiganos

nano/lib/blocks.cpp
```diff
@@ -151,8 +152,10 @@ bool nano::block::has_sideband () const
 
 nano::account const & nano::block::representative () const
 {
-	static nano::account rep{ 0 };
-	return rep;
+	/* return nano::hardened_constants::get ().not_an_account; */
```
thsfs: Vote for removing this comment.

thsfs (dismissed):

nano/lib/blocks.cpp
theohax: Oh, good catch. If I remember right, I put that comment there because it seemed to me like a good place to use the sentinel, although for some reason I could not find the right places to use it too "on the other side of the == operation". In like, if I changed it there, tests were failing and I couldn't tell what was missing.

Maybe the sentinel isn't fit for this specific site... @clemahieu do you have any idea?

clemahieu: I'm not sure what would be causing it without digging in but we don't need to fix up all call sites in one go. I'd say remove the comments also if it's not the implementation we're using.

clemahieu (dismissed):

nano/lib/blocks.cpp
dsiganos: We can remove the comments and replace them with a improvement github issue.

nano/lib/numbers.cpp
```diff
@@ -61,6 +62,16 @@ std::string nano::public_key::to_account () const
 	return result;
 }
 
+nano::public_key::public_key () :
+	uint256_union{ 0 }
+{
+}
+
+nano::public_key::public_key (std::nullptr_t) :
```
dsiganos: I thought we were supposed to remove the nullptr_t constructor to avoid the 0 constructor problem?

nano/lib/blocks.cpp
theohax: Ok, I'll do that

nano/lib/numbers.cpp
theohax: No, I understood we leave this constructor but make the default one behave as if it was constructed with a `0` argument.

dsiganos (commented):
I thought we agreed to do this:

Remove the nullptr constructor to avoid the zero constructor ambiguity
Construct a null account via a static method call
A null account should be equal to the nullptr when compared
We can leave the default constructor as is, if we want (I have no strong opinion on that although in general i like things to be initialised)
We can remove the static casts to uint64

theohax:
> I thought we agreed to do this:
> 
> Remove the nullptr constructor to avoid the zero constructor ambiguity Construct a null account via a static method call A null account should be equal to the nullptr when compared We can leave the default constructor as is, if we want (I have no strong opinion on that although in general i like things to be initialised) We can remove the static casts to uint64

I don't see a big difference nor a big benefit in obtaining null accounts via a static method call as opposed to the nullptr_t constructor we already have, but ok, I will address that as well.

dsiganos:
The big difference is that nano::account{0} will not be confusing.

theohax:
> The big difference is that nano::account{0} will not be confusing.

As of this PR, it is no longer neccessary to ever say `nano::account{0}` and even after I address your requested changes `nano::account{0}` should not ever be needed (nor even "acceptable", since we have the default constructor). I'll post another comment when changes are addressed.

dsiganos:
At the moment, nano::account{0} does not create an account with zero destination, which is very confusing.

theohax:
Addressed requested changes, rebased the branch, fixed the conflicts. Ready for review again.

nano/node/active_transactions.hpp
```diff
@@ -279,7 +279,7 @@ class active_transactions final
 	void confirm_expired_frontiers_pessimistically (nano::transaction const &, uint64_t, uint64_t &);
 	void frontiers_confirmation (nano::unique_lock<nano::mutex> &);
 	bool insert_election_from_frontiers_confirmation (std::shared_ptr<nano::block> const &, nano::account const &, nano::uint128_t, nano::election_behavior);
-	nano::account next_frontier_account{ 0 };
```
dsiganos, line 282: We do not need to change zero constructors to default constructors because a zero constructor is not ambiguous after moving the construction of a null function to static class function. This change could be dramatically simplified if we just left the zero constructors as they are. The vast majority of the changes, are rewriting the zero constructors, which is no longer necessary.

nano/lib/numbers.cpp
```diff
@@ -61,6 +62,16 @@ std::string nano::public_key::to_account () const
 	return result;
 }
 
+nano::public_key::public_key () :
+	uint256_union{ 0 }
+{
+}
+
+const nano::public_key & nano::public_key::null () const
```
dsiganos: The function that creates a null public key is not used anywhere other than the equality function.
Is that expected?

nano/lib/numbers.cpp
```diff
@@ -61,6 +62,16 @@ std::string nano::public_key::to_account () const
...
+{
+	return nano::hardened_constants::get ().not_an_account;
```
dsiganos, line 72: I expected this to return a reference to a static public_key rather construct a temporary public key.

nano/lib/numbers.cpp
```diff
@@ -939,6 +955,16 @@ nano::public_key::operator nano::hash_or_account const & () const
 	return reinterpret_cast<nano::hash_or_account const &> (*this);
 }
 
+bool nano::public_key::operator== (std::nullptr_t) const
+{
+	return bytes == null ().bytes;
```
dsiganos, line 960: If we have a static object for the null account then this comparison can become a simple pointer comparison and there will be no temporary object involved. It is more efficient and it removes the theoretical possibility of a value clash.

nano/node/active_transactions.hpp
theohax: We discussed to make the default constructor equivalent to calling the uint64 constructor with a 0 value, therefore code that was previously explicitly saying `account{0}` now just says `account{}`. The changes had been there from the first version of the PR and I see no reason to revert them, they are making the code less verbose and taking out boilerplate.

nano/lib/numbers.cpp
theohax: Yes, the PR provides the support, we will change occurrences (to nullptr accounts) as we see fit in other PRs.

nano/lib/numbers.cpp
theohax: It is not constructing a temporary key. It is returning a lazily initialized reference.

nano/lib/numbers.cpp
theohax: I don't believe there's a huge penalty being paid for this comparison, but since you shouldn't even be able to create the null account in a different way than using `nano::account::null ()` (or directly `nano::hardened_constants::get ().not_an_account` for that matter), then I guess it's very safe to use an address comparison -- therefore I addressed this as well.

nano/lib/numbers.cpp
dsiganos: But, will it return the same exact object every time?
It would be great if there was one and only one null object.

dsiganos:
I made a mistake and flagged up the wrong comparison function.
It was the other comparison that could be done with a pointer check.
The one that compared to nullptr.
I think the latest changes do not make sense at all. 
So I suggest you revert it and go back to what you had.

nano/lib/numbers.cpp
dsiganos: The difference is that this PR would affect only a 3-4 files whereas now it affects 36 files. But, I do not feel very strongly about it.

nano/lib/numbers.cpp
theohax: Yeah, that's exactly the way it behaves.

nano/secure/common.cpp
```diff
@@ -115,7 +115,7 @@ nano::ledger_constants::ledger_constants (nano::work_thresholds & work, nano::ne
 	: network_a == nano::networks::nano_test_network                                                                           ? nano_test_genesis
 																															   : nano_live_genesis),
 	genesis_amount{ std::numeric_limits<nano::uint128_t>::max () },
-	burn_account (0),
+	burn_account{ static_cast<uint64_t> (0) },
```
dsiganos: Do we need this uint64 cast?

nano/secure/common.hpp
```diff
@@ -99,7 +99,7 @@ class account_info final
 	size_t db_size () const;
 	nano::epoch epoch () const;
 	nano::block_hash head{ 0 };
-	nano::account representative{ 0 };
+	nano::account representative{ static_cast<uint64_t> (0) };
```
dsiganos: Do we need this cast?

dsiganos:
Why do we declare not_an_account inside hardened constants and then pass it through public_key class?
Shouldn't not_an_account be encapsulated implementation detail of class public_key?
For example, not_an_account is directly accessed in a few places:
* election::election()
* election::log_votes()
* election::votes_with_weights()

It seems that we created the public_key null API but we are bypassing it and going straight to hardened_constants::not_an_account.

By comparing directly to hardened_constants::not_an_account we are introducing 2 different ways to check for not an account, once with pointer comparison and another with account number comparison. That introduces unnecessary complexity. We should either do pointer only or value only comparisons, not both. I am happy to go back to comparing public key values only, I can see that comparing pointers is more risky.

theohax:
> Why do we declare not_an_account inside hardened constants and then pass it through public_key class? Shouldn't not_an_account be encapsulated implementation detail of class public_key? For example, not_an_account is directly accessed in a few places:
> 
> * election::election()
> * election::log_votes()
> * election::votes_with_weights()
> 
> It seems that we created the public_key null API but we are bypassing it and going straight to hardened_constants::not_an_account.
> 
> By comparing directly to hardened_constants::not_an_account we are introducing 2 different ways to check for not an account, once with pointer comparison and another with account number comparison. That introduces unnecessary complexity. We should either do pointer only or value only comparisons, not both. I am happy to go back to comparing public key values only, I can see that comparing pointers is more risky.

Addressed this as well. Let us merge this PR.

clemahieu (approved):

PR #3394, opened 2021-07-26
Moving networking related parameters off of the node config

clemahieu:
Moving networking related parameters off of the node config class an on to the networking config class where they belong.


theohax (approved):
Refactoring looks good (taking what is not node-config related off it), but if anything, I find the following access rather cumbersome from a semantic perspective:

`node.network_params.network`

Not a deal breaker, just thinking out loud.

zhyatt:
@clemahieu Was this all internal changes or are there some external toml file impacts to note?

zhyatt:
> @clemahieu Was this all internal changes or are there some external toml file impacts to note?

Colin confirmed no external configuration file impacts

PR #3400, opened 2021-07-26
Use ledger_walker from within the confirmation_height_processor

theohax:
The ugly ifdefs can be taken out (together with the ones from `ledger_walker.cpp`) pretty soon once everything is stable with diskhash on Windows.

Besides that, this PR adds the `run_walker` method to the `confirmation_height_processor`, but does not attempt to remove already used code -- @clemahieu that's the bit of refactoring we were discussing that you could help with. Please let me know if something is not clear or not complete before that final refactoring and the removal of the bounded/unbounded processors can be done. Additionally, I would like to make `awaiting_processing` a simple std::deque or something like that instead of the complex `boost::multi_index_container` that we have right now, because we will likely no longer need that (currently it's used just to be able to prevent double callback for cemented blocks).

### DIFF
--- a/nano/core_test/confirmation_height.cpp
+++ b/nano/core_test/confirmation_height.cpp
@@ -1374,7 +1374,7 @@ TEST (confirmation_height, unbounded_block_cache_iteration)
 		ASSERT_EQ (nano::process_result::progress, ledger.process (transaction, *send1).code);
 	}
 
-	nano::confirmation_height_processor confirmation_height_processor (ledger, write_database_queue, 10ms, logging, logger, initialized_latch, nano::confirmation_height_mode::unbounded);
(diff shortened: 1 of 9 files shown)


theohax:
@ reviewers: I rebased this one as well and solved its conflicts and freshened it up a little bit.

There's definitely work to be done to fully replace the logic in `confirmation_height_processor` with `run_walker` implementation from this PR. But to be honest I'd rather see it done in two steps, so if you could review this one and if we could merge it, at a later point we can look at opening another PR that gradually removes the existing logic, eventually based on some kind of config parameter if we're not sure about the implications of `diskhash`.

@clemahieu Especially regarding the heavy refactoring that we were discussing a while back, I think we're better off moving that into a separate PR. So let's try to get some focus on this one and see if we can merge it so that it doesn't become stale.

Thanks for the upcoming reviews guys!

PR #3402, opened 2021-07-27
Passing nano::ledger_constants in to nano::ledger

clemahieu:
Passing nano::ledger_constants in to nano::ledger rather than it being statically initialized. Further work in the effort to reduce static initialization.


theohax (approved):
Looks good!

PR #3407, opened 2021-07-29
remove errexit ghcr.io has some edgecase rules, and isnt required for…

### LINKED ISSUE #3399: Tests failing because my Github username is capitalized
Example: https://github.com/shryder/nano-node/runs/3163839612

> invalid argument "ghcr.io/Shryder/nano-node/nano-env:base" for "-t, --tag" flag: invalid reference format: repository name must be lowercase
> ...
> invalid argument "ghcr.io/Shryder/nano-node/nano-env:gcc" for "-t, --tag" flag: invalid reference format: repository name must be lowercase

I've changed my username now its all lower-case.

argakiig:
… CI, just convenient at times
closes #3399

### DIFF
--- a/ci/build-docker-image.sh
+++ b/ci/build-docker-image.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -eu
+set -u
 
 if [ "$#" -lt 2 ]; then
 	echo 'Usage: build-docker-image.sh <dockerFile> <dockerImageTag> [<dockerBuildArgs>...]' >&2


dsiganos (commented):
What are the problems caused by ```set -e```?
In general, I think it is a good thing to have in a build script, because it is too difficult to check all exit conditions in a script, so I wonder why we are removing it.

argakiig:
The script this modification was made in should never fail just because a username is Capitalized.
ghcr.io is used as a caching mechanism with a fallback to dockerhub, in the event neither of these pulls work it should then attempt to build the docker images from source, set -e would short out in the referenced issue and so the PR would not finish building

dsiganos:
I'm not fully understanding the issue but your explanation makes sense and I understand why removing -e helps.
However, is the real problem here that we are using github names as docker tags, when docker tags have strict rules?

argakiig:
this is specific to Githubs container registry, see https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images for details about why the convention we follow is done.
The only reason ghcr.io was added was due to the changes with dockerhub pulls and rate limiting

dsiganos (approved):

shryder (approved):

PR #3413, opened 2021-08-04
Removing static initialization of nano::network_params

clemahieu:
This is a series of commits to remove static instantiation of nano::network_params. Static initialization is a bad practice in general and prevents more complex configurations, hindered config reloading, and is generally counterintuitive.

### DIFF
--- a/nano/node/common.cpp
+++ b/nano/node/common.cpp
@@ -56,7 +56,6 @@ nano::message_header::message_header (bool & error_a, nano::stream & stream_a)
 
 void nano::message_header::serialize (nano::stream & stream_a) const
 {
-	static nano::network_params network_params;
 	nano::write (stream_a, boost::endian::native_to_big (static_cast<uint16_t> (network)));
 	nano::write (stream_a, version_max);
 	nano::write (stream_a, version_using);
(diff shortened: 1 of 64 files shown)


thsfs (dismissed):

nano/node/common.cpp
```diff
@@ -51,9 +42,6 @@ uint64_t nano::ip_address_hash_raw (boost::asio::ip::address const & ip_a, uint1
 
 nano::message_header::message_header (nano::message_type type_a) :
 	network (nano::network_constants::active_network),
-	version_max (get_protocol_constants ().protocol_version),
```
dsiganos, line 54: Are version_max, version_using, version_min left uninitialised?
That doesn't sound like a good idea.
I agree that it is a pain to provide these 3 numbers every time but leaving them uninitialised also sounds bad.

clemahieu: The goal with leaving them uninitialized is that a compiler tool could detect uninitialized usages. Alternatively we could initialized them with a value like 0xff and then debug_assert that they are not so when serializing.

dsiganos: It's just that it is another exception that we have to remember - that these fields are not always set.
And it is a random exception. For example, why set version only bytes in the network send function? Why not set the magic and the networks ID byte there too? In fact, I wonder if it would be possible to set the whole message header there.

dsiganos (approved):

PR #3420, opened 2021-08-10
Remove static instantiation of network_constants

clemahieu:
This change removes static references to network_constants that were littered around the code. Additionally, free-functions related to work difficulty and value calculations were moved on to the work_thrsholds class.

This PR is best reviewed on a per-commit basis, each commit is self-contained with clean test runs.

### DIFF
--- a/nano/secure/ledger.cpp
+++ b/nano/secure/ledger.cpp
@@ -346,8 +346,7 @@ void ledger_processor::state_block_impl (nano::state_block & block_a)
 				if (result.code == nano::process_result::progress)
 				{
 					nano::block_details block_details (epoch, is_send, is_receive, false);
-					nano::network_constants constants;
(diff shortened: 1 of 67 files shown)


nano/secure/ledger.cpp
```diff
@@ -869,7 +869,8 @@ nano::uint128_t nano::ledger::account_pending (nano::transaction const & transac
 
 nano::process_return nano::ledger::process (nano::write_transaction const & transaction_a, nano::block & block_a, nano::signature_verification verification)
 {
-	debug_assert (!nano::work_validate_entry (block_a) || constants.genesis == nano::dev::genesis);
+	static nano::network_constants network_constants;
```
dsiganos: is this need?

dsiganos (approved):

thsfs (approved):

nano/secure/ledger.cpp
clemahieu: This was added back in a later commit.

PR #3422, opened 2021-08-12
Fix daemon_config active_network.

clemahieu:
This is a fix where daemon_config was not switching its active networ…k off of nano_dev_network.


argakiig (approved):
ran through paces locally on all OS's seems to be working as expected

Issue #3389, opened 2021-07-23
Reduce compile times by using forward-declarations when possible.


clemahieu:
When profiling with -ftime-trace, it was found that parsing node.hpp represents a significant portion of compile time. The class nano::node has many of its other components as values and as such, requires the implementation to be included in the node.hpp file.

This can be avoided by forward declaring the internal components and nano::node can hold unique_ptr or references to them.

An example of how this can be done is with the nano::node::store member:
```
	std::unique_ptr<nano::store> store_impl;
	nano::store & store{ *store_impl };
```

This should be done with any members that take a significant time to parse.

thsfs:
I've been working on applying IWYU (https://include-what-you-use.org/). It is a tool that suggests better includes and forward-references that can replace the includes you already have. Discovered that when applying it automatically it will surely mess with the code. Better way is to do it manually, so we have more control to what is going to be changed. Applying to headers that are included a lot of times might obligate you to apply the changes in all the depending source code, as a cascade effect, so you may end up changing your entire code base at once.

A very helpful tool to be used together is ClangBuildAnalyzer (https://github.com/aras-p/ClangBuildAnalyzer), so you can check before which headers have more impact on the build time. There is also a tweak that can be done on its source code so it can return the least included headers. It is useful to start changing the code which have least impact on the build, so you avoid the cascade effect when applying the changes.

Tweak suggestion to be applied on "Analysis.cpp" upon the current ClangBuildAnalyzer source-code:
```
std::vector<std::pair<std::string_view, int64_t>> leastUsedHeaders;
void Analysis::FindLeastUsedHeaders()
{
  leastUsedHeaders.reserve(headerMap.size());
  for (const auto& kvp : headerMap)
  {
    if (config.onlyRootHeaders && !kvp.second.root)
      continue;
    if (kvp.first.find("nano-node/nano/") == std::string::npos)
      continue;
    leastUsedHeaders.push_back(std::make_pair(kvp.first, kvp.second.count));
  }
  std::sort(leastUsedHeaders.begin(), leastUsedHeaders.end(), [&](const auto& a, const auto& b)
  {
    if (a.second != b.second)
      return a.second > b.second;
    return a.first < b.first;
  });
  if (leastUsedHeaders.size() > config.headerCount)
    leastUsedHeaders.resize(config.headerCount);
}
```

```

  FindLeastUsedHeaders();

  if (!leastUsedHeaders.empty())
  {
[...]

PR #3426, opened 2021-08-18
Applies IWYU suggestions

thsfs:
- Applies IWYU (include-what-you-use) that is a tool that suggests better includes and possible forward-references.
- Intended to speed-up the node's build time.


thsfs:
Cancelled, as mentioned in https://github.com/nanocurrency/nano-node/issues/3389.

PR #3427, opened 2021-08-20
Initial draft for bottom up (upward) bootstrapping

theohax:
This is a very early version of the bottom up bootstrapping. Mainly put it out with the intention of gathering early feedback.

It contains skeleton code as well as some core functionality.

Lifetime management isn't being taken care of, perhaps many of the involved objects will need to extend their lifetimes via the the `shared_from_this()-kept-in-a-lambda` idiom.

I'm interested in having a bunch of discussions around the existing skeleton, answer the questions/uncertainties described in the comments, improve things here or there, etc.

### DIFF
--- a/nano/node/bootstrap/upward/bootstrapper.cpp
+++ b/nano/node/bootstrap/upward/bootstrapper.cpp
@@ -1,0 +1,156 @@
+#include <nano/lib/logger_mt.hpp>
+#include <nano/lib/threading.hpp>
+#include <nano/node/bootstrap/upward/bootstrapper.hpp>
+#include <nano/node/bootstrap/upward/peer_manager.hpp>
+#include <nano/node/bootstrap/upward/pull_client.hpp>
+#include <nano/node/bootstrap/upward/pull_info.hpp>
+#include <nano/secure/store.hpp>
+
+#include <boost/format.hpp>
+
+#include <utility>
+
+nano::bootstrap::upward::bootstrapper::bootstrapper (nano::store const & store_a,
+                                                     nano::bootstrap::upward::peer_manager & peer_manager_a,
+                                                     nano::network_constants const & network_constants_a,
+                                                     nano::logger_mt & logger_a,
+                                                     nano::thread_pool & thread_pool_a) :
+store{ store_a },
+peer_manager { peer_manager_a },
+network_constants{ network_constants_a },
+logger{ logger_a },
+thread_pool { thread_pool_a },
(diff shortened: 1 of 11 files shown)


qwahzi:
For clarity, bottom up here refers to bootstrapping from the highest account block heights to the lowest, right? The reverse of the current genesis-to-frontiers bootstrapping model?

Are there any thoughts on how this would interact with pruning? I'm guessing that due to block dependencies + ready to receive blocks it might still require bootstrapping the entire ledger and then pruning, but it would be amazing to bootstrap straight to a pruned state if possible (faster, less bandwidth usage, less disk io, etc)

dsiganos:
@qwahzi The bottom is the genesis block. And top are the frontiers.

qwahzi:
> @qwahzi The bottom is the genesis block. And top are the frontiers.

Isn't that how bootstrapping already works (genesis->frontiers)? In that case, how does this new bootstrapping process differ from the old process?

EDIT - more details from @dsiganos on Discord:

> The legacy bootstrapper works from the frontiers down to genesis.
> There is also lazy and wallet bootstrappers. I;m not sure what they do. But by default it is the legacy one working.
> There is nothing wrong with the legacy bootstrapper in principle, we would just like to have one that also grows from the bottom upwards.
> To avoid the scenario where you have millions of blocks downloaded but none cemented.
> The idea is that they will both work in parallel.

nano/node/bootstrap/upward/bootstrapper.cpp
```diff
@@ -0,0 +1,156 @@
...
+    const auto transaction = store.tx_begin_read ();
+
+    nano::lock_guard<nano::mutex> lock{ mutex };
+    while (!accounts_to_follow.empty ())
+    {
+        const auto & account = accounts_to_follow.back ();
+        recently_followed_accounts.emplace (account);
+
+        nano::account_info account_info{};
+        if (!store.account.get (transaction, account, account_info))
```
clemahieu, line 107: The account being followed won't always be findable, e.g. following the genesis account to accounts that are destinations from genesis transactions.

theohax: Makes sense, I updated the code so that `account_info`s are now `std::optional` and assume that if we can't find the account in the storage then we'll be expecting an open block as the first block received for that account.

dsiganos:
This has been superseded by the ascending bootstrap. Closing.

PR #3429, opened 2021-08-26
Add "confirmed" field to "account_history" RPC response (#3424)

### LINKED ISSUE #3424: Add block confirmation status to blocks returned by `account_history`.
### Summary

Currently, to get the confirmation status of blocks returned by `account_history`, `account_info` must be requested at the same time in order to infer block confirmation status based on confirmation height. It would be nice if `account_history` included confirmation status for each block. 

I'm writing a wallet which stores blocks locally. This feature would halve the number of calls required to sync accounts with a node.

Not as far as I can tell. 

Add an optional `confirmed` field to blocks in the response given by `account_history`. Looking at the rpc implementation, the confirmation height could be obtained in the same way it is for `account_info` then a `confirmed` field could be added. 

I would not

### Possible solution

_No response_

### Supporting files

_No response_

shryder:
Closes #3424

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -2547,6 +2547,7 @@ void nano::json_handler::account_history ()
 					entry.put ("local_timestamp", std::to_string (block->sideband ().timestamp));
 					entry.put ("height", std::to_string (block->sideband ().height));
 					entry.put ("hash", hash.to_string ());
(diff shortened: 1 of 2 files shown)


thsfs (approved):
Hi @zhyatt , the code is ok. Just remind to update the documentation as well.

theohax (approved):

PR #3431, opened 2021-08-31
fix failure to build with GCC11

slact:
GCC11 has a change that breaks the build:

>Header dependency changes
>
>Some C++ Standard Library headers have been changed to no longer include other headers that they do need to depend on. As such, C++ programs that used standard library components without including the right headers will no longer compile.
>
>The following headers are used less widely in libstdc++ and may need to be included explicitly when compiled with GCC 11:
>  ```
>    <limits> (for std::numeric_limits)
>    <memory> (for std::unique_ptr, std::shared_ptr etc.)
>    <utility> (for std::pair, std::tuple_size, std::index_sequence etc.)
>    <thread> (for members of namespace std::this_thread.)
```
(see https://gcc.gnu.org/gcc-11/porting_to.html )

### DIFF
--- a/nano/lib/threading.hpp
+++ b/nano/lib/threading.hpp
@@ -5,6 +5,7 @@
 #include <nano/lib/utility.hpp>
 
 #include <boost/thread/thread.hpp>
+#include <thread>
 
 namespace nano
 {


argakiig:
@slact just a heads up `master` has not been used in a bit, `develop` is the current development branch with releases being branched off `releases/vxx`

clemahieu:
I tried to change the base branch with GitHub but it created chaos with the merge.

For your two PRs can you cherry pick them on top of develop?

slact:
yep i'll close and reopen against the `develop` branch

PR #3436, opened 2021-09-01
Remove temporary concept from channel_tcp

clemahieu:


nano/node/transport/tcp.cpp
```diff
@@ -316,7 +309,6 @@ void nano::transport::tcp_channels::process_message (nano::message const & messa
 					debug_assert (endpoint_a == temporary_channel->get_tcp_endpoint ());
```
dsiganos, line 304: There are references to temporary channels here, should they still be here?

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -123,7 +123,6 @@ nano::bootstrap_server::~bootstrap_server ()
 		auto exisiting_response_channel (node->network.tcp_channels.find_channel (remote_endpoint));
```
dsiganos, line 159: The comment above, refers to temporary channel.

nano/node/repcrawler.cpp
```diff
@@ -117,7 +117,7 @@ std::vector<std::shared_ptr<nano::transport::channel>> nano::rep_crawler::get_cr
 	required_peer_count += required_peer_count / 2;
 
 	// The rest of the endpoints are picked randomly
-	auto random_peers (node.network.random_set (required_peer_count, 0, true)); // Include channels with ephemeral remote ports
+	auto random_peers (node.network.random_set (required_peer_count, 0)); // Include channels with ephemeral remote ports
```
dsiganos, line 151: Does the comment about ephemeral ports still apply? I think not.

dsiganos (commented):
I do not understand exactly what temporary channels were intended to achieve.
Do you understand Colin?

Issue #3411, opened 2021-08-02
Replacing `pending` with `receivable`/`ready to receive`


zhyatt:
Update node references, RPC calls, etc. to use `receivable` to indicate that an incoming send block on an account is ready to have the corresponding receive block published for it (after which the newly added balance can be sent by the receiving account). Previously the less clear term `pending` was used.

Any updates to the QT wallet UI referencing the same should use the term `ready to receive` for end users.

Audit of locations requiring change TBD.

See https://www.reddit.com/r/nanocurrency/comments/ov3fro/term_pending_to_be_replaced_with_ready_to_receive/ for broader ecosystem request to make this change.

shryder:
These are the references I found in the QT wallet:
https://github.com/nanocurrency/nano-node/blob/develop/nano/qt/qt.cpp#L118
https://github.com/nanocurrency/nano-node/blob/develop/nano/qt/qt.cpp#L275
https://github.com/nanocurrency/nano-node/blob/develop/nano/qt/qt.cpp#L780
https://github.com/nanocurrency/nano-node/blob/develop/nano/qt/qt.cpp#L1911
https://github.com/nanocurrency/nano-node/blob/develop/nano/qt/qt.cpp#L2356
https://github.com/nanocurrency/nano-node/blob/develop/nano/qt/qt.cpp#L2521

These are the references I found in the RPC handler (will be changed in 2.0 only as these would all be breaking changes):
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L529
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L613
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L695
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L898
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L1209
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L1270
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L2642
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L2661
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L2713
https://github.com/nanocurrency/nano-node/blob/develop/nano/node/json_handler.cpp#L4317
[...]

PR #3437, opened 2021-09-04
Replace "pending" with "receivable"/"ready to receive" in QT wallet (#3411)

shryder:
(#3411)

### DIFF
--- a/nano/qt/qt.cpp
+++ b/nano/qt/qt.cpp
@@ -1919,7 +1919,7 @@ nano_qt::advanced_actions::advanced_actions (nano_qt::wallet & wallet_a) :
 	});
 
 	bootstrap->setToolTip ("Multi-connection bootstrap to random peers");
-	search_for_receivables->setToolTip ("Search for ready to receive blocks");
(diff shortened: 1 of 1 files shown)


nano/qt/qt.cpp
```diff
@@ -1919,7 +1919,7 @@ nano_qt::advanced_actions::advanced_actions (nano_qt::wallet & wallet_a) :
 	});
 
 	bootstrap->setToolTip ("Multi-connection bootstrap to random peers");
-	search_for_receivables->setToolTip ("Search for pending blocks");
+	search_for_receivables->setToolTip ("Search for ready to receive blocks");
```
keerifox: i'd suggest `"Search for ready-to-receive blocks"` here so you can't read it as "Search for ready, to receive blocks"

zhyatt: Alternatively this could be `Search for ready to be received blocks` as that phrase is used in other changes in this PR.

argakiig (approved):

PR #3439, opened 2021-09-05
Add `receivable` along with `pending` to RPC responses (#3411)

shryder:
#3411


nano/node/json_handler.cpp
```diff
@@ -527,6 +527,7 @@ void nano::json_handler::account_balance ()
 		auto balance (node.balance_pending (account, include_only_confirmed));
 		response_l.put ("balance", balance.first.convert_to<std::string> ());
 		response_l.put ("pending", balance.second.convert_to<std::string> ());
```
thsfs, line 529: Is it kept for compatibility?

thsfs (commented):
Why we are keeping the "pending" field?

nano/node/json_handler.cpp
shryder: @thsfs Correct

thsfs (approved):

theohax (approved):

PR #3441, opened 2021-09-08
Removing null checks for nano::bootstrap::socket

clemahieu:
Removing null checks for nano::bootstrap::socket since it's required to be non-null on construction and is never nulled out during execution.


theohax (approved):
`bootstrap_server.cpp` needs every tiny bit of "less code, less complex" it can get! Looks good.

PR #3444, opened 2021-09-12
Improve how the flatbuffers api file is located

cryptocode:
A small improvement to deal with how various build configs/IDEs puts the node binary in different paths (so this is just about developer convenience) I noticed this issue after upgrading xcode / latest develop.

@zhyatt, this is the PR I mentioned in the notes after going through the RPC2 stuff.

### DIFF
--- a/nano/node/ipc/flatbuffers_handler.cpp
+++ b/nano/node/ipc/flatbuffers_handler.cpp
@@ -39,21 +39,23 @@ std::string make_error_response (std::string const & error_message)
  */
 boost::optional<boost::filesystem::path> get_api_path ()
 {
+	boost::filesystem::path const fb_path = "api/flatbuffers";
 	auto parent_path = boost::dll::program_location ().parent_path ();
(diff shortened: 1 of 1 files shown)


clemahieu:
This looks like it keeps searching directories upward until it finds one named api. Is it more appropriate to instruct CMake to place the file in a fixed location relative to the executable rather than making the searching logic more complex?

cryptocode:
It searches for "api/flatbuffers", not just api, which is where nanoapi.fbs is located (like before, it just searches further). 

IIRC the CMake config is set up to distribute these files under install-dir/api/flatbuffers, so this is just a dev convenience (no searching happens in correct production setups)

But yeah, CMake could copy these files to whatever the output directory is (like build/Debug) as an alternative to this. Let me know if you prefer that and I'll change the PR.

cryptocode:
This PR adjusts existing behavior, but it should really be CMake copying these files (so closing per discussions with colin, will create an alternative pr)

PR #3445, opened 2021-09-13
Fix fill_keepalive_self so the low index entries aren't clobbered.

clemahieu:
Fill_keepalive_self sets special values in index 0 and 1 to refer to the peering port of the node. Values are also filled from the lowest index so the lowest entries would get clobbered. Preserve these values in index 2 and 3.

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -1262,3 +1262,11 @@ TEST (network, filter)
 	node1.network.inbound (keepalive, std::make_shared<nano::transport::channel_loopback> (node1));
 	ASSERT_EQ (1, node1.stats.count (nano::stat::type::message, nano::stat::detail::invalid_network));
 }
+
+TEST (network, fill_keepalive_self)
+{
+	nano::system system{ 2 };
+	std::array<nano::endpoint, 8> target;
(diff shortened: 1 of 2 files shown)


nano/core_test/network.cpp
```diff
@@ -1262,3 +1262,11 @@ TEST (network, filter)
 	node1.network.inbound (keepalive, std::make_shared<nano::transport::channel_loopback> (node1));
 	ASSERT_EQ (1, node1.stats.count (nano::stat::type::message, nano::stat::detail::invalid_network));
 }
+
+TEST (network, fill_keepalive_self)
+{
+	nano::system system{ 2 };
+	std::array<nano::endpoint, 8> target;
+	system.nodes[0]->network.fill_keepalive_self (target);
+	ASSERT_TRUE (target[2].port () == system.nodes[1]->network.port);
```
dsiganos, line 1271: It would be better to check that system.nodes[1]->network.port is somewhere in the target without specifying its exact position. But it is a very minor point and I have no string opinion on it just thought I'd mention it because our tests in general have too implementation detail.

dsiganos (approved):

theohax (approved):

PR #3460, opened 2021-09-17
Make code formatting compliant with clang-format 12

theohax:
Just ran the new `./ci/clang-format-all.sh` script with version 12 of `clang-format` and this is what it produced.

### DIFF
--- a/nano/core_test/work_pool.cpp
+++ b/nano/core_test/work_pool.cpp
@@ -90,8 +90,8 @@ TEST (work, opencl)
 		{
 			// 0 threads, should add 1 for managing OpenCL
 			nano::work_pool pool{ nano::dev::network_params.network, 0, std::chrono::nanoseconds (0), [&opencl] (nano::work_version const version_a, nano::root const & root_a, uint64_t difficulty_a, std::atomic<int> & ticket_a) {
-				return opencl->generate_work (version_a, root_a, difficulty_a);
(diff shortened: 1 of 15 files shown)


thsfs (commented):
image

It is better because it fits better the screen, but I'd suggest two changes for ternary operators upon these current changes:

1. Avoid using spaces when it is possible, as we prefer using tabs;
2. If it is something easily feasible, fitting a condition per line, so it'd like below:

image

theohax:
> image
> 
> It is better because it fits better the screen, but I'd suggest two changes for ternary operators upon these current changes:
> 
> 1. Avoid using spaces when it is possible, as we prefer using tabs;
> 2. If it is something easily feasible, fitting a condition per line, so it'd like below:
> 
> image

I agree, I don't really like how ternary operator usages look like either.

But for the time being I'd rather merge this as-is in order to prevent other conflicts and let's bring up the subject in the dev call so that we can eventually open a subsequent PR in which we add a clang-format rule for this and also modify all occurrences across all the codebase.

theohax:
@argakiig @clemahieu @dsiganos a green button would be welcome when you can guys, there's not a lot to actually review, it's just changes done by clang-format itself. I'd like to get this in in order to prevent conflicts with other PRs. Thanks!

clemahieu (approved):

Issue #3294, opened 2021-05-17
Bootstrapping blocks slowly crawl to a halt


DarrenAlex:
### Summary

When bootstrapping blocks, slowly the number of blocks fetched/downloaded per second goes down.
As an example, when I first started fetching blocks from other peers, I downloaded 2,000 blocks/second.
After hitting 21.3 million blocks (60 million for someone else), it slowed down to 20-30 blocks/second.

Local device resources seem to be fine

### Node version

V22.0

### Build details

Docker container, nothing out of the ordinary.

### OS and version

Linux 20.04

### Steps to reproduce the behavior

1. Follow installation instruction in the docs
2. Start bootstrapping
3. Monitor block/sec progress

### Expected behavior

The software continues to sync blocks at a similar/ascending rate, not a descending rate

### Actual behavior

The software syncs blocks at a descending rate

### Possible solution

Honestly, I don't have any idea

qwahzi:
Same issue for me with multiple nodes/bootstrapping attempts (Centos 8). The slowdown happens around ~59-60M blocks for me though. In the last 4 hours, I've only gained ~100 counted blocks.

```
$ cat startTime.txt
           Universal time: Thu 2021-05-13 20:49:04 UTC

$ curl -d '{"action": "block_count"}' 127.0.0.1:7076
{
    "count": "60680998",
    "unchecked": "60623590",
    "cemented": "37743403"
}

$ timedatectl
           Universal time: Mon 2021-05-17 16:15:28 UTC
```

Not sure if related, but I see a bunch of these messages in the logs:
```[2021-May-17 16:12:27.089344]: Error initiating bootstrap connection to [::ffff:194.195.243.169]:57930: No route to host
[2021-May-17 16:12:33.072637]: Error initiating bootstrap connection to [::ffff:143.198.231.110]:54182: Connection reset by peer
[2021-May-17 16:12:33.839897]: 27416 accounts in pull queue
[2021-May-17 16:12:44.109434]: Error initiating bootstrap connection to [::ffff:194.195.243.169]:57930: No route to host
```

DarrenAlex:
I also receive those messages in the log file. I've uploaded the log files (only lines which contain the word "bootstrap") to https://node.erawan.me/log.txt

The only difference is that the counted and cemented blocks are stuck at 1 for me. 
If you sum them all up, it's close to the network's total blocks. Is the cemented block number going up?

atomlab:
The same problem!

[...]

PR #3461, opened 2021-09-21
Fix the bootstrap request queue processing

theohax:
This is the fix we already discussed/tested, we're still looking at a failing QT unit test but other than that it seems good.


clemahieu (approved):

qwahzi:
Hey @theohax, do you know if this PR is related to https://github.com/nanocurrency/nano-node/issues/3294? Or is this something completely different?

dsiganos:
@qwahzi, the bug probably doesn't help with the bootstrapping issues but we do not think this is the primary cause of the bootstrapping problems.

PR #3462, opened 2021-09-21
Improve clang-format detection & enforcement

theohax:
This gives better error messages, gets rid of some Python script that we weren't really using (at least not the way it was supposed to be used, for our purpose we can get away much quicker & lighter) and also fixes the actual enforcement which wasn't being done properly for the past few PRs, now it should be all good.


theohax:
Had couple hiccups but everything seems fine now -- this is reviewable/mergeable. Thanks.

argakiig (approved):

dsiganos (approved):

PR #3463, opened 2021-09-21
Compile code that uses diskhash on Windows

theohax:
Such code was being taken out of the build before because diskhash wasn't building on Windows a few months back, but now it is, so this PR should build/run fine on Win.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -389,7 +389,8 @@ find_package(Boost 1.70.0 REQUIRED COMPONENTS filesystem log log_setup thread
 
 # diskhash
 if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
-  add_library(diskhash STATIC ${CMAKE_SOURCE_DIR}/diskhash/src/diskhash.c)
+  add_library(diskhash STATIC ${CMAKE_SOURCE_DIR}/diskhash/src/os_wrappers.c
+                              ${CMAKE_SOURCE_DIR}/diskhash/src/diskhash.c)
   include_directories(diskhash/src)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -388,11 +388,9 @@ find_package(Boost 1.70.0 REQUIRED COMPONENTS filesystem log log_setup thread
                                               program_options system)
 
 # diskhash
-if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
-  add_library(diskhash STATIC ${CMAKE_SOURCE_DIR}/diskhash/src/os_wrappers.c
-                              ${CMAKE_SOURCE_DIR}/diskhash/src/diskhash.c)
-  include_directories(diskhash/src)
-endif()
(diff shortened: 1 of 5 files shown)


theohax:
@thsfs looks like Windows build still broken, but not because of compilation issues, it doesn't even get there, it can't include `diskhash.hpp` for some reason. You said you were able to use it in your Windows tests?

thsfs:
> @thsfs looks like Windows build still broken, but not because of compilation issues, it doesn't even get there, it can't include `diskhash.hpp` for some reason. You said you were able to use it in your Windows tests?

For Windows I just ran the DiskHash unittests, that is not sufficient, but it was supposed to at least build. I'm going to take a look on this error.

theohax:
> > @thsfs looks like Windows build still broken, but not because of compilation issues, it doesn't even get there, it can't include `diskhash.hpp` for some reason. You said you were able to use it in your Windows tests?
> 
> For Windows I just ran the DiskHash unittests, that is not sufficient, but it was supposed to at least build. I'm going to take a look on this error.

Thanks! Hope it'll run fine after your changes.

CMakeLists.txt
```diff
@@ -388,11 +388,9 @@ find_package(Boost 1.70.0 REQUIRED COMPONENTS filesystem log log_setup thread
                                               program_options system)
 
 # diskhash
-if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
```
theohax, line 391: That's right, these were added by me and forgot to take them out now... Thanks!

thsfs: No problem, there is still an error for MacOSX, but seems to be a trivial thing to be fixed on DiskHash library. Let's see if there will be more errors on Windows, so I can send a new pack of changes.

theohax: Sounds good, thanks for your swift intervention and yeah let's see how the Windows tests behave first.

Later edit: CI just failed at the build step: `D:\a\nano-node\nano-node\diskhash\src\diskhash.c(156,72): error C2036: 'void *': unknown size [D:\a\nano-node\nano-node\build\diskhash.vcxproj]`. Can we make a PR to our local diskhash repo to fix that MSVS compilation error?

thsfs: Yes, sure. BTW, do you think the current failure on develop branch will also impact this PR? Does this branch have the same clang change?

theohax: Hmm no, not really, that one got solved anyway. Let's look into the latest diskhash-related build fails on Win and other than that we should be fine.

thsfs: DiskHash is indeed broken on Windows. It will require a more deep investigation. I suggest making this PR a draft, so we can resume it once we solve the DH problems.

theohax: Agree, let's put it on hold therefore.

thsfs:
Closing this PR.  It doesn't change much the diskhash code. The reason for it to fail is that the code for Windows is not working properly. This should be reviewed and tested first before coming back to integration.

PR #3470, opened 2021-09-22
Fix cmake-format-all.sh script + friendlier xargs usage

theohax:
@cryptocode @clemahieu let me know guys if `./ci/clang-format-all.sh` and especially `./ci/cmake-format-all.sh` run fine on your Macbooks. Just tested on mine and looks good.

### DIFF
--- a/ci/check-cmake-format.sh
+++ b/ci/check-cmake-format.sh
@@ -1,16 +1,22 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-if ! command -v cmake-format &>/dev/null; then
-    echo "pip install cmake-format and try again"
+set -e
+
+if [[ ! -z $(git status --untracked-files=no --porcelain) ]]; then
+    echo "Unable to run script: working directory not clean (see git status)"
     exit 1
 fi
-REPO_ROOT=$(git rev-parse --show-toplevel)
-cd "${REPO_ROOT}"
-find "${REPO_ROOT}" -iwholename "${REPO_ROOT}/nano/*/CMakeLists.txt" -o -iwholename "${REPO_ROOT}/CMakeLists.txt" -o -iwholename "${REPO_ROOT}/coverage/CMakeLists.txt" | xargs cmake-format --check &>.cmake_format_check
-if [[ $(wc -l .cmake_format_check | cut -f1 -d ' ') != 0 ]]; then
-    echo
-    echo "Code formatting differs from expected - please run \n\t'bash ci/cmake-format-all.sh'"
-    RET=1
+
+source "$(dirname "$BASH_SOURCE")/common.sh"
+
+"$REPO_ROOT/ci/cmake-format-all.sh"
+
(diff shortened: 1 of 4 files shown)


cryptocode:
This makes the cmake related scripts execute, but running `ci/check-cmake-format.sh` after `ci/cmake-format-all.sh` still results in `Code formatting differs from expected - please run \n\t'bash ci/cmake-format-all.sh'`

theohax:
> This makes the cmake related scripts execute, but running `ci/check-cmake-format.sh` after `ci/cmake-format-all.sh` still results in `Code formatting differs from expected - please run \n\t'bash ci/cmake-format-all.sh'`

Thanks for reporting, I haven't looked into that one but will and then update this accordingly!

theohax:
@cryptocode Both checks seem happy when done by the CI following the last commit. Could you run the check scripts on your OSX locally as well?

```
Run ci/check-commit-format.sh
Detected 'clang-format' with version 'Ubuntu clang-format version 11.0.0-2~ubuntu20.04.1'  (different than '12'), skipping it.
Using 'clang-format-12' version '12'
clang-format passed
```

```
Run bash ci/check-cmake-format.sh
Using 'cmake-format' version '0.6.13'
cmake-format passed
```

cryptocode:
I now get `Unable to run script: working directory not clean (see git status)` on a fresh checkout due to 

```
git status
	modified:   miniupnp (modified content, untracked content)
	modified:   rocksdb (modified content)
```
I haven't actually modified anything manually there, but I think running CMake does affect makefiles in submodules.

But shouldn't these check scripts work on dirty working copies anyway? For instance, I check clang formatting before committing (I have a pre-commit hook so I don't forget)

theohax:
> I now get `Unable to run script: working directory not clean (see git status)` on a fresh checkout due to
> 
> ```
> git status
> 	modified:   miniupnp (modified content, untracked content)
> 	modified:   rocksdb (modified content)
> ```
> 
> I haven't actually modified anything manually there, but I think running CMake does affect makefiles in submodules.
> 
> But shouldn't these check scripts work on dirty working copies anyway? For instance, I check clang formatting before committing (I have a pre-commit hook so I don't forget)

Hi @cryptocode, after redesigning the scripts I changed that behaviour so that you'd be able to "update/do/perform" formatting via the script at anytime, but you'd only be able to "check" if the formatting is right when in a clean `git status` state. That's because the "check" step is actually doing the same "update" step and then a `git diff` to tell if you were alright or not.

But after seeing your pre-commit hook argument and thinking more about it, I realised the "check" step is definitely not robust. So I've already re-redesigned the "doing" and the "checking" steps for both clang and cmake, but I'd rather put that up in a different PR in order to keep things clean and tidy. Would you mind approving this as-is, getting it in, so that I can open the subsequent one and you can have a look at how the new mechanism works?

Thanks for your valuable reports btw!

cryptocode (approved):

theohax:
Thanks for approving, new PR is https://github.com/nanocurrency/nano-node/pull/3486

PR #3486, opened 2021-09-29
Redesign CI format check/update scripts according to review

theohax:
As per discussed in #3470, this is the new redesign of the format update/check CI scripts.

The main difference is that now the checking scripts can be run even on a dirty working copy (whereas before `git status` had to be clean in order to be able to run the check script). They're also a bit more organized and tidied up.

@cryptocode let me know how all 4 of them behave for you (macOS), I've only tested Ubuntu so far. They are:
- `ci/clang-format-do.sh`
- `ci/clang-format-check.sh`
- `ci/cmake-format-do.sh`
- `ci/cmake-format-check.sh`


cryptocode:
The cmake scripts seems to work fine on my end 👍 

`clang-format-check.sh` seems to trigger checks on generated files now (don't think that was the case before):

```
generated/flatbuffers/nanoapi_generated.h:2408:39: error: code should be clang-formatted [-Wclang-format-violations]
  flatbuffers::String *mutable_account() {
```

theohax:
> The cmake scripts seems to work fine on my end +1
> 
> `clang-format-check.sh` seems to trigger checks on generated files now (don't think that was the case before):
> 
> ```
> generated/flatbuffers/nanoapi_generated.h:2408:39: error: code should be clang-formatted [-Wclang-format-violations]
>   flatbuffers::String *mutable_account() {
> ```

So exactly which folders we want to include in the update/check clang-format and which we don't? Currently, the following are searched for: `nano-node/nano/*/(*.h, *.hpp, *.cpp)`

cryptocode:
> So exactly which folders we want to include in the update/check clang-format and which we don't? Currently, the following are searched for: `nano-node/nano/*/(*.h, *.hpp, *.cpp)`

The folder seems correct, probably just have to filter out nanoapi_generated.h. Not sure why this hasn't been a problem before, maybe things have moved around since I last worked on the codebase.

The clang-format-do.sh script is emitting quite a few `Is a directory` messages by the way.

cryptocode:
Actually, the reason this worked before is that .h files was previously not included in the check script: https://github.com/nanocurrency/nano-node/blob/V21.0/ci/check-commit-format.sh

theohax:
Ok so should we only search for *.hpp and *.cpp again or all of them and just ignore the generated header (if it's just one)? 

Weird about the warnings, I'm not getting them, but I'll tweak the `find` commands (for both clang/cmake) to only look for files.

theohax:
Updated scripts as per review/discussions, let me know how it looks now :D decided to go only with *.hpp/*.cpp files instead of trying to ignore generated files which would also have to be maintained.

cryptocode (approved):
All scripts work for me 👍

PR #3500, opened 2021-10-08
Endianess fix for vote timestamp field (does not affect x86)

dsiganos:
This change will have no effect on little endianess machines (x86).
But it would fix big endian machines.


clemahieu (commented):
I think this needs the associated endian swap for deserialization on 687

dsiganos:
Yeap, and in it is missing from the second serialisation function too...

theohax (approved):

PR #3505, opened 2021-10-15
Add [[nodiscard]] attribute to node::process method

theohax:
Compiling locally I did not see any build failures, but let's see the CI result first.

### DIFF
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -268,8 +268,8 @@ TEST (bootstrap_processor, process_state)
 
 	node0->work_generate_blocking (*block1);
 	node0->work_generate_blocking (*block2);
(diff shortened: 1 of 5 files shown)


clemahieu (changes requested):
Can we get the associated warning silenced by handling the return value. In tests we can do ASSERT_EQ to check for the correct process result.

theohax:
> Can we get the associated warning silenced by handling the return value. In tests we can do ASSERT_EQ to check for the correct process result.

Ohh that's why I didn't see it, I was expecting an error, but we're not compiling with -Werror, yeah...

I'll search manually for the usages and fix them.

theohax:
@clemahieu seems good now, can you have another look? thanks!

clemahieu (approved):

Issue #3419, opened 2021-08-10
Immediately broadcast blocks on arrival


zhyatt:
As part of the V21.3 service release the [rebroadcasting of blocks upon immediate arrival was removed](https://github.com/nanocurrency/nano-node/pull/3140) (specifically commit c21ec84cfe49bd545198f2cbfa64095bd714e4b8) but this change reduces block propagation and may be contributing to delays in non-PR nodes receiving blocks initially. This rebroadcasting of blocks immediately on arrival should be added back in to improve block propagation.

Exxenoz:
If I publish blocks on other nodes than my own (for example MyNanoNinja using the process RPC) it takes around 1-5 minutes until my node marks them as confirmed. I even setup a second Nano node on a fresh server - same behaviour. Seems like all non-PR nodes are currently affected by this. Confirmations are almost instant if I publish the blocks directly on my node, ~~except for open blocks, which confirm somehow as slow as if I had published them using an external node (1-5mins).~~

dsiganos:
@Exxenoz , if you wish to help debug this, the next step would be to enable detailed logs and repeat your test case.
To enable detailed logs , you can add these entries to your config-node.toml file.

```
[node.logging]
active_update = true
bulk_pull = true
election_expiration = true
election_fork = true
ledger = true
ledger_rollback = true
ledger_duplicate = true
log_ipc = false
log_to_cerr = false
max_size = 4294967296
min_time_between_output = 0
network = true
network_keepalive = true
network_message = true
network_packet = true
network_publish = true
network_rejected = true
network_telemetry = true
network_timeout = true
node_lifetime_tracing = false
rotation_size = 268435456
single_line_record = true
stable_log_filename = true
timing = true
upnp_details = false
vote = true
flush = true
```

zhyatt:
@Exxenoz We would expect you to have less impacts now, as we are now temporarily rebroadcasting blocks to help improve these situations (until a fix can be deployed broadly). This just started consistently about an hour or so ago. Let us know if you keep experiencing these same delays.

Exxenoz:
@zhyatt

> except for open blocks, which confirm somehow as slow as if I had published them using an external node (1-5mins).

^ This was my fault. I accidentally published all open blocks to MyNanoNinja yesterday.

Now:
Test1: Send & Open block published to node1. -> Send & Open block got instantly confirmed on node1.
Test2: Send & Open block published to node1. -> Send & Open block got instantly confirmed on node1.
[...]

PR #3507, opened 2021-10-18
Brings back broadcasting blocks on process_live

thsfs:
This shall increase block propagation and reduce the delays in non-PR nodes.
Partially reverts https://github.com/nanocurrency/nano-node/commit/c21ec84cfe49bd545198f2cbfa64095bd714e4b8.
Fixes issue https://github.com/nanocurrency/nano-node/issues/3419.

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -328,7 +328,7 @@ void nano::block_processor::process_live (nano::transaction const & transaction_
 	}
 	else if (!node.flags.disable_block_processor_republishing)
 	{
-		node.network.flood_block (block_a, nano::buffer_drop_policy::no_limiter_drop);
+		node.network.flood_block (block_a, nano::buffer_drop_policy::limiter);
 	}
 
 	if (node.websocket_server && node.websocket_server->any_subscriber (nano::websocket::topic::new_unconfirmed_block))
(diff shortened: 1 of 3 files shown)


nano/node/blockprocessor.cpp
```diff
@@ -326,6 +326,10 @@ void nano::block_processor::process_live (nano::transaction const & transaction_
 	{
 		node.network.flood_block_initial (block_a);
 	}
+	else if (!node.flags.disable_block_processor_republishing)
+	{
+		node.network.flood_block (block_a, nano::buffer_drop_policy::no_limiter_drop);
```
dsiganos: Should this be "no limiter drop", since this isn't an absolutely essential activity for the stability of the network?

zhyatt (commented):
I believe block flooding is done on election activation from the scheduler (line 317), so could this cause double-publishing of the block?

nano/node/blockprocessor.cpp
thsfs: Changed to allow drops.

theohax (dismissed):
Following the in-call discussion, looks good.

zhyatt (dismissed):

clemahieu (approved):

PR #3509, opened 2021-10-19
Fix Windows build (signal handling)

theohax:
This should fix the [build error seen here](https://github.com/nanocurrency/nano-node/runs/3942518654?check_suite_focus=true#step:5:943)

### DIFF
--- a/nano/nano_node/daemon.cpp
+++ b/nano/nano_node/daemon.cpp
@@ -27,6 +27,20 @@ void nano_abort_signal_handler (int signum)
 	raise (signum);
 }
 
+void install_abort_signal_handler ()
+{
+#ifndef _WIN32
+	// We catch signal SIGSEGV and SIGABRT not via the signal manager because we want these signal handlers
+	// to be executed in the stack of the code that caused the signal, so we can dump the stacktrace.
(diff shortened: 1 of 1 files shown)


dsiganos:
Theo, do you know why this doesn't this work on Windows?

theohax:
> Theo, do you know why this doesn't this work on Windows?

Because it's using UNIX signals. `sigaction` seems to be a problem as well, it might work with the plain old `signal()` function but I haven't tested that. First trying to get the build working, then we'll see if it's still possible to keep this signal handling in an Windows-equivalent form.

dsiganos:
I believe that windows is able to handle to SIGV and SIGABRT. We just need to find out how to do on windows.
Previously we used to use to plain old signal() function. That must have worked on windows too.

theohax:
> I believe that windows is able to handle to SIGV and SIGABRT. We just need to find out how to do on windows. Previously we used to use to plain old signal() function. That must have worked on windows too.

As I said, I believe too that `signal` might work. Let us get the build green, we'll then worry about the handling.

dsiganos:
OK

dsiganos (approved):

PR #3510, opened 2021-10-20
Improvements in const correctness and 'const' positioning

theohax:
Co-authored-by: marioortizmanero <marioortizmanero@gmail.com>

Supersedes https://github.com/nanocurrency/nano-node/pull/3143

Interesting fact: starting with [`clang-format-14`, we'll be able to automatically check & enforce](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) the type modifiers positioning:
Screenshot from 2021-10-20 16-16-07

### DIFF
--- a/nano/lib/asio.cpp
+++ b/nano/lib/asio.cpp
@@ -1,6 +1,6 @@
 #include <nano/lib/asio.hpp>
 
-nano::shared_const_buffer::shared_const_buffer (const std::vector<uint8_t> & data) :
+nano::shared_const_buffer::shared_const_buffer (std::vector<uint8_t> const & data) :
 	m_data (std::make_shared<std::vector<uint8_t>> (data)),
 	m_buffer (boost::asio::buffer (*m_data))
 {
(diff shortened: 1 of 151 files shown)


nano/lib/asio.cpp
```diff
@@ -29,17 +29,17 @@ nano::shared_const_buffer::shared_const_buffer (std::shared_ptr<std::vector<uint
...
 }
 
-const boost::asio::const_buffer * nano::shared_const_buffer::end () const
+boost::asio::const_buffer const * nano::shared_const_buffer::end () const
 {
 	return &m_buffer + 1;
 }
 
-size_t nano::shared_const_buffer::size () const
+std::size_t nano::shared_const_buffer::size () const
```
clemahieu, line 42: It looks like it's adding std:: In the front of size_t. Is clang-format doing this?

theohax: Oh no, that was me, I've seen mixed `size_t` and `std::size_t` and I thought to be consistent. Admittedly, there were way more `size_t`s than `std::` ones but since we're all about using full namespaces I thought it makes more sense. Do you believe it makes code harder to read or anything?

nano/lib/errors.cpp
```diff
@@ -486,6 +486,7 @@ nano::error & nano::error::clear ()
 	return *this;
 }
 
+// TODO: theoretically, nothing besides template (partial) specializations should ever be added inside std...
```
thsfs, line 489: Vote for removing this comment and opening an issue if there is something else to be done here.

nano/lib/asio.cpp
clemahieu: Others have moved to using std::size_t  so this is fine https://github.com/dartsim/dart/pull/681

clemahieu (approved):

thsfs (approved):

nano/lib/errors.cpp
theohax: I'll put a "create an issue" on my list for this, thanks.

PR #3511, opened 2021-10-20
Handle SIGABRT and SIGSEGV on Windows

theohax:
Follow-up of https://github.com/nanocurrency/nano-node/pull/3509

@dsiganos I think this is how you were thinking to do it as well, right? Curious to see if the CI is green on Windows. It used to be like that a while ago so it should be fine I hope.

### DIFF
--- a/nano/nano_node/daemon.cpp
+++ b/nano/nano_node/daemon.cpp
@@ -29,7 +29,10 @@ void nano_abort_signal_handler (int signum)
 
 void install_abort_signal_handler ()
 {
-#ifndef _WIN32
+#ifdef _WIN32
+	signal (SIGSEGV, nano_abort_signal_handler);
+	signal (SIGABRT, nano_abort_signal_handler);
+#else
 	// We catch signal SIGSEGV and SIGABRT not via the signal manager because we want these signal handlers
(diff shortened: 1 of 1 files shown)


dsiganos:
Yes, this is probably good enough.
Alternatively, we could use signal() everywhere although it is supposedly deprecated for sigaction.

dsiganos:
We just need to make sure we do not enter into a signal raising loop.
When using signal() function what will do the equivalent of SA_RESETHAND?

dsiganos:
This is what the old signal handler used to do:
```
void my_abort_signal_handler (int signum)
{
	std::signal (signum, SIG_DFL);
	nano::dump_crash_stacktrace ();
	nano::create_load_memory_address_files ();
}
```

So, we maybe need to call `std::signal (signum, SIG_DFL)` inside the signal handler when in windows.

dsiganos:
I think what was happening with the old signal handler was this:
* The signal handler was executed and it removed it self from the signal handling
* then it saved the stack trace and returned
* the program re-tries  the last operation and it fails again and raises the signal again but this time the default signals is executed and the program ends

theohax:
@dsiganos alright, sounds good, incorporated that feedback into the last commit, please have another look. Thanks.

dsiganos (approved):

PR #3512, opened 2021-10-22
Disable (intermittently) failing unit tests

theohax:
T̶h̶i̶s̶ ̶P̶R̶ ̶m̶a̶k̶e̶s̶ ̶t̶h̶e̶ ̶C̶I̶ ̶t̶i̶m̶e̶o̶u̶t̶ ̶f̶o̶r̶ ̶u̶n̶i̶t̶ ̶t̶e̶s̶t̶s̶ ̶a̶ ̶b̶i̶t̶ ̶l̶a̶r̶g̶e̶r̶.̶

Initially, the PR was about enlarging the timeout, but in fact, the problem for which some jobs were timing out was because there was an intermittent unit test failure and the Bash script kept rerunning `core_test` until it succeeded, which led to the timeout.

What the PR does now is: disable all unit tests that fail intermittently across Linux/OSX/Windows. Purpose for this is to keep a base of tests that always (100% of the runs) pass and make a green CI light a mandatory step for merging PRs. 

I will then open a couple issues to look at disabled tests and see how we can get them enabled again.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1682,7 +1682,7 @@ TEST (node, DISABLED_fork_stale)
 	}
 }
 
-TEST (node, broadcast_elected)
+TEST (node, DISABLED_broadcast_elected)
 {
 	std::vector<nano::transport::transport_type> types{ nano::transport::transport_type::tcp, nano::transport::transport_type::udp };
 	for (auto & type : types)
@@ -1904,7 +1904,7 @@ TEST (node, DISABLED_bootstrap_no_publish)
 }
 
 // Check that an outgoing bootstrap request can push blocks
-TEST (node, bootstrap_bulk_push)
+TEST (node, DISABLED_bootstrap_bulk_push)
 {
 	nano::system system0;
 	nano::system system1;
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1682,6 +1682,9 @@ TEST (node, DISABLED_fork_stale)
 	}
 }
 
+// Test disabled because it's failing intermittently.
+// PR in which it got disabled: https://github.com/nanocurrency/nano-node/pull/3512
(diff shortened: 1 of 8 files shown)


dsiganos (dismissed):

nano/core_test/node.cpp
```diff
@@ -1682,7 +1682,7 @@ TEST (node, DISABLED_fork_stale)
 	}
 }
 
-TEST (node, broadcast_elected)
+TEST (node, DISABLED_broadcast_elected)
```
dsiganos, line 1688: Can you add a comment explaining why you are disabling it?

theohax: I changed the scope of the PR, please have a look at the newly added description.

Long story short: disable all failing tests, aim for an always green CI, make it a mandatory step for merging future PRs.
Debt: open issues for the disabled tests, investigate and see how/if they can be re-enabled.

dsiganos:
Let's add comments and raise issues for every disabled test case and then this can be committed.

theohax:
Leaving some links & references here for posterity.

In this PR, the following unit tests have been disabled:
- rpc.wallet_history -- issue #3514 
- node.bootstrap_bulk_push -- issue #3515 
- node.broadcast_elected -- issue #3516 
- bootstrap_processor.push_diamond_pruning -- issue #3517
- active_transactions.confirm_active -- issue #3522

dsiganos (dismissed):

theohax:
Caught `active_transactions.confirm_active` [failing](https://github.com/nanocurrency/nano-node/runs/3995990187?check_suite_focus=true#step:6:43) as well, therefore [disabled](https://github.com/nanocurrency/nano-node/pull/3512/commits/3a4866a650b20a44b77fbec1d457efb17f52354b) it too.

One more approval please @dsiganos, thanks.

dsiganos (dismissed):

dsiganos (approved):

PR #3526, opened 2021-10-26
Disable intermittently failing unit test

theohax:
Disabled two more tests:
- `node.fork_invalid_block_signature` (failed [here](https://github.com/nanocurrency/nano-node/runs/4003377949?check_suite_focus=true#step:6:1850), issue [here](https://github.com/nanocurrency/nano-node/issues/3527))
- `election.quorum_minimum_update_weight_before_quorum_checks` (failed [here](https://github.com/nanocurrency/nano-node/runs/4008874110?check_suite_focus=true#step:6:562), issue [here](https://github.com/nanocurrency/nano-node/issues/3528))

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -3159,7 +3159,10 @@ TEST (node, epoch_conflict_confirm)
 	}
 }
 
-TEST (node, fork_invalid_block_signature)
+// Test disabled because it's failing intermittently.
+// PR in which it got disabled: https://github.com/nanocurrency/nano-node/pull/3526
+// Issue for investigating it: https://github.com/nanocurrency/nano-node/issues/3527
+TEST (node, DISABLED_fork_invalid_block_signature)
 {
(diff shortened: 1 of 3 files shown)


nano/core_test/node.cpp
```diff
@@ -3159,6 +3159,9 @@ TEST (node, epoch_conflict_confirm)
 	}
 }
 
+// Test disabled because it's failing intermittently.
+// PR in which it got disabled: https://github.com/nanocurrency/nano-node/pull/3526
+// Issue for investigating it: https://github.com/nanocurrency/nano-node/issues/3527
 TEST (node, fork_invalid_block_signature)
```
dsiganos: you forgot to actually disabled it ;-)

theohax: Good part about that is that another one failed and I added it to the PR. Thanks.

dsiganos (approved):

PR #3531, opened 2021-10-27
Rep crawler weight minimum

### LINKED ISSUE #2176: confirmation_quorum not including all representatives
It seems that the `confirmation_quorum` RPC call with `peer_details: true` only includes principal representatives.
At https://mynano.ninja/ I use this call to get the reps IP to guess their provider and location.
To get a better overview of the location and providers of _all_ representatives it would be good to have that information available.

I know that it includes only reps who are "considered in the summation of peers_stake_total" (https://docs.nano.org/commands/rpc-protocol/#confirmation_quorum) and non-principal reps weight is not counted in an election.

It would be good to either add that to the existing `confirmation_quorum` RPC or create a new RPC action which combines the `peers` and `confirmation_quorum` actions and includes all available information of the connected representatives. The output could be the following:
```
{
  "peers": [
    {
      "ip": "[::ffff:5.189.128.113]:34632",
      "account": "nano_1ninja7rh37ehfp9utkor5ixmxyg8kme8fnzc4zty145ibch8kf5jwpnzr3r",
      "weight": "258668230880283508504634562507801736",
      "protocol_version": "17",
      "type": "tcp"
    },
    {
      "ip": "[::ffff:5.189.144.220]:7075",
      "account": "nano_3u7d5iohy14swyhxhgfm9iq4xa9yibhcgnyj697uwhicp14dhx4woik5e9ek",
      "weight": "165487346227148931481124621329038766",
      "protocol_version": "17",
      "type": "tcp"
    }
  ]
}
```

dsiganos:
Introduce a feature to allow the rep crawler to track reps with weight lower than principal rep.
This is for debugging purposes and for websites that would like to track all online reps irrespective of weight.

I also introduce the framework for rep crawler log messages.
And I updated the Representative Found" to print the channel info and also to distinguish between an addition
and an update. And on update it prints the channel before and after to help us diagnose  problems with reps not having
the correct IP address.

Fixes #2176

### DIFF
--- a/nano/node/repcrawler.cpp
+++ b/nano/node/repcrawler.cpp
@@ -33,7 +33,9 @@ void nano::rep_crawler::validate ()
 		responses_l.swap (responses);
 	}
 
-	auto minimum = node.minimum_principal_weight ();
+	// normally the rep_crawler only tracks principal reps but it can be made to track
+	// reps with less weight by setting rep_crawler_weight_minimum to a low value
+	auto minimum = std::min (node.minimum_principal_weight (), node.config.rep_crawler_weight_minimum.number ());
 
 	for (auto const & i : responses_l)
 	{
--- a/nano/node/repcrawler.cpp
+++ b/nano/node/repcrawler.cpp
@@ -60,20 +60,26 @@ void nano::rep_crawler::validate ()
 			continue;
 		}
 
-		auto updated_or_inserted = false;
+		// temporary data used for logging after dropping the lock
+		auto inserted = false;
+		auto updated = false;
+		std::shared_ptr<nano::transport::channel> prev_channel;
+
 		nano::unique_lock<nano::mutex> lock (probable_reps_mutex);
+
 		auto existing (probable_reps.find (vote->account));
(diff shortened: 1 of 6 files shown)


nano/node/repcrawler.cpp
```diff
@@ -35,6 +35,13 @@ void nano::rep_crawler::validate ()
 
 	auto minimum = node.minimum_principal_weight ();
 
+	// normally the rep_crawler only tracks principal reps but it can be made to track
+	// reps with less weight by setting rep_crawler_weight_minimum to a low value
+	if (node.config.rep_crawler_weight_minimum < minimum)
```
clemahieu: Putting this in to a single expression might be more understandable.

auto minimum = std::min (node.minimum_principal_weight (), node.config.rep_crawler_weight_minimum);

clemahieu (dismissed):

nano/node/repcrawler.cpp
dsiganos: Done

clemahieu (approved):

theohax (approved):

Issue #3523, opened 2021-10-24
Websocket message is missing the `source_account` on receive blocks


tombertrand:
### Summary

When receiving messages from the websocket, a `send` transaction between 2 accounts can easily be identified using the `account` and `link_as_account` fields. The issue is when getting `receive` transaction the receiver can be easily identified from the `account` field but the `link_as_account` field doesn't correspond to the sender. When grabbing the block from RPC I notice the extra `source_account` field that corresponds to the sender that is not present on the socket message.

### Node version

V22.1

### Build details

V22.1

### OS and version

Any OS

### Steps to reproduce the behavior

1. Listen the websocket for a receive tx

### Expected behavior

The `receive` subtype message should contain the same format as the block and include `source_account` which is the sender.

### Actual behavior

`source_account` is not present in the message

### Supporting files

[Discord discussion](https://discord.com/channels/370266023905198083/405506987045158914/901611588740603985)

PR #3538, opened 2021-10-31
Websocket confirmation topic : Add source_account on receive blocks / open blocks

gr0vity-dev:
In response to issue 3523 ( https://github.com/nanocurrency/nano-node/issues/3523 )
I added the source_account into the websocket confirmation topic.

The source_account is added for each block type. It follows the same logic as "block_info" rpc call.
For "open" and "receive" blocks, we show the nano_address corresponding to the sender account.
Else (for "send", "change" and "epoch") we show "0"

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -128,7 +128,7 @@ nano::node::node (boost::asio::io_context & io_ctx_a, boost::filesystem::path co
 		if (config.websocket_config.enabled)
 		{
 			auto endpoint_l (nano::tcp_endpoint (boost::asio::ip::make_address_v6 (config.websocket_config.address), config.websocket_config.port));
-			websocket_server = std::make_shared<nano::websocket::listener> (logger, wallets, io_ctx, endpoint_l);
+			websocket_server = std::make_shared<nano::websocket::listener> (logger, *this, wallets, io_ctx, endpoint_l);
 			this->websocket_server->run ();
 		}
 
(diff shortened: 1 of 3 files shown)


clemahieu:
The concept for this PR makes sense and I agree it'd be useful to implementations.

The websocket.confirmation_options test fails which needs to be addressed. This seems like a possible race condition in the test that can be fixed.

I think this information is more generally applicable, specifically it's useful in both RPCs and web socket callbacks.  Additionally, it seems useful to have a single data item that identifies the counterparty, rather than using two depending if it's a send or a receive.

I've made two changes I think will make this most useful. These changes would deprecate "link_as_account" which is only a translation of the link field to an account number and only works for send blocks, and adds "linked_account" which is the counterparty for any balance-changing transaction, omitted if it is not a balance changing transaction.

https://github.com/nanocurrency/nano-node/tree/websocket_confirmation_source_account

zhyatt:
After discussion, we need to further define the structure for this and will do so soon, but the PR will be merged for a version after V23.0

Exxenoz:
Is there already a decision on this matter? At the moment, you still have to send a request to the node for each receive block in order to get the sender account. The "linked_account" solution looks good. Would you merge this PR if I rebase it to the current node version?

gr0vity-dev:
If we do this, we might also include `receive_hash` (displays the hash of the send block's corresponding receive if any).

But it adds a read transaction which might have a performance impact. 
When websocket is enabled, I believe outgoing messages are on the hot path and can slow down the node significantly. 
Not sure if that has been migrated to its own background thread.

Exxenoz:
> If we do this, we might also include `receive_hash` (displays the hash of the send block's corresponding receive if any).
> 
> But it adds a read transaction which might have a performance impact. When websocket is enabled, I believe outgoing messages are on the hot path and can slow down the node significantly. Not sure if that has been migrated to its own background thread.

I would not include `receive_hash` for two reasons:

- Corresponding receive blocks are most likely not confirmed or stored yet when the send block was processed and send out via the web socket module (except during bootstrapping)
- Searching for the corresponding receive block by going through large blockchains could slow down the web socket thread

gr0vity-dev:
> Corresponding receive blocks are most likely not confirmed or stored yet when the send block was processed

You are right! It doesn't make much sense to add it.

Issue #2985, opened 2020-09-26
clang/boost 1.73 compile error on case-insensitive file systems


cryptocode:
Here's a cute problem when I compile the node with the latest clang version/boost 1.73 on macOS.

So `boost/config/select_stdlib_config.hpp` contains this:

```
#if defined(__cplusplus) && defined(__has_include)
#  if __has_include(<version>)
#    include <version>
#  else
#    include <cstddef>
#  endif
#elif defined(__cplusplus)
#  include <cstddef>
#else
#  include <stddef.h>
#endif
```

where `include <version>` picks up... `miniupnp/miniupnpc/VERSION` 🙄 🤣  This happens because APFS is case-insensitive.

Renaming that file to VERSION_ or whatever "fixes" the issue, but surely there must be a better way.

clemahieu:
I worked around this by deleting that VERSION file, it doesn't appear to be needed.

Since someone else is running in to this we'll get a fix.

cryptocode:
This problem should go away soon: https://github.com/miniupnp/miniupnp/issues/540 😀 
Might require a minor update to the cmake script when upgrading.

clemahieu:
Fixed with https://github.com/nanocurrency/nano-node/pull/3541

PR #3541, opened 2021-11-01
Updating miniupnp to v2.2.3

clemahieu:
This solves https://github.com/nanocurrency/nano-node/issues/2985 which is includes https://github.com/miniupnp/miniupnp/issues/540 which is included in miniupnp v2.2.3

### DIFF
--- a/nano/node/portmapping.cpp
+++ b/nano/node/portmapping.cpp
@@ -1,8 +1,8 @@
 #include <nano/node/node.hpp>
 #include <nano/node/portmapping.hpp>
 
-#include <miniupnp/miniupnpc/upnpcommands.h>
-#include <miniupnp/miniupnpc/upnperrors.h>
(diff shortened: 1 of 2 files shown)


zhyatt:
Are the test failures here intermittently failing ones we want to disable?

clemahieu:
> Are the test failures here intermittently failing ones we want to disable?

One seems to be ledger_walker.ladder_geometry.

dsiganos (approved):

theohax (approved):
Always nice to get up to date with upstream. Looks good.

theohax:
> > Are the test failures here intermittently failing ones we want to disable?
> 
> One seems to be ledger_walker.ladder_geometry.

Oh, finally seeing this fail again. Seems like it indeed is an intermittent failure there, so we can add it to our list.

PR #3544, opened 2021-11-05
Vote timestamp encapsulation

clemahieu:
This is the first commit related to https://github.com/nanocurrency/nano-node/issues/3453

It encapsulates the nano::vote::timestamp field and rewrites tests to use the encapsulated field.

### DIFF
--- a/nano/secure/common.cpp
+++ b/nano/secure/common.cpp
@@ -473,7 +473,7 @@ bool nano::vote::operator== (nano::vote const & other_a) const
 			}
 		}
 	}
-	return timestamp == other_a.timestamp && blocks_equal && account == other_a.account && signature == other_a.signature;
+	return timestamp_m == other_a.timestamp_m && blocks_equal && account == other_a.account && signature == other_a.signature;
 }
 
 bool nano::vote::operator!= (nano::vote const & other_a) const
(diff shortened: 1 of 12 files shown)


thsfs (approved):

nano/secure/common.hpp
```diff
@@ -264,8 +264,15 @@ class vote final
 	boost::transform_iterator<nano::iterate_vote_blocks_as_hash, nano::vote_blocks_vec_iter> begin () const;
 	boost::transform_iterator<nano::iterate_vote_blocks_as_hash, nano::vote_blocks_vec_iter> end () const;
 	std::string to_json () const;
+	uint64_t timestamp () const;
+	static uint64_t constexpr timestamp_max = { 0xffff'ffff'ffff'ffffULL };
```
theohax, line 268: Would it be maybe any better to use the numeric limits? 

Also, on a different note, is `timestamp_m` (the suffix `_m`) for member? I don't remember seeing it a lot.

theohax (approved):

nano/secure/common.cpp
```diff
@@ -485,8 +485,8 @@ void nano::vote::serialize_json (boost::property_tree::ptree & tree) const
 {
 	tree.put ("account", account.to_account ());
 	tree.put ("signature", signature.number ());
-	tree.put ("sequence", std::to_string (timestamp));
-	tree.put ("timestamp", std::to_string (timestamp));
+	tree.put ("sequence", std::to_string (timestamp ()));
```
dsiganos, line 488: Why are we writing both a sequence and a timestamp?
For backwards compatibility?

dsiganos (approved):

nano/secure/common.cpp
clemahieu: Yes though we could probably remove sequence at this point.

nano/secure/common.hpp
clemahieu: This number will be different with the actual vote duration changes.

PR #3545, opened 2021-11-05
Fix conflicts.add_two unit test

### LINKED ISSUE #3535: Unit test conflicts.add_two failed in CI
[ RUN      ] conflicts.add_two
/workspace/nano/core_test/conflicts.cpp:68: Failure
  2
  node1.active.size ()
    Which is: 1
[  FAILED  ] conflicts.add_two (226 ms)

https://github.com/nanocurrency/nano-node/runs/4038550570?check_suite_focus=true

To reproduce this, let the CPU load to be high and run ./core_test with the options:
`./core_test --gtest_filter=conflicts.add_two --gtest_repeat=100 --gtest_throw_on_failure`

theohax:
Addresses https://github.com/nanocurrency/nano-node/issues/3535

I added a new one instead of removing the old one just for easier review (as it's all new code, instead of seeing interleaved additions and removals). After the PR is reviewed, I can delete the old disabled one. 

@dsiganos @thsfs if you guys wanna have a look on it together, let me know

### DIFF
--- a/nano/core_test/conflicts.cpp
+++ b/nano/core_test/conflicts.cpp
@@ -76,6 +76,10 @@ TEST (conflicts, add_two)
 	nano::system system{};
 	auto const & node = system.add_node ();
 
+	// define a functor that sends from given account to given destination,
+	// optionally force-confirming the send blocks *and* receiving on the destination account;
+	// the functor returns a pair of the send and receive blocks created or nullptrs if something failed
+	//
 	auto const do_send = [&node] (auto const & previous, auto const & from, auto const & to, bool forceConfirm = true)
 	-> std::pair<std::shared_ptr<nano::block>, std::shared_ptr<nano::block>> {
 		auto send = std::make_shared<nano::send_block> (previous->hash (), to.pub, 0, from.prv, from.pub, 0);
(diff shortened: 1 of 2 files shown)


nano/core_test/conflicts.cpp
```diff
@@ -71,6 +71,79 @@ TEST (conflicts, DISABLED_add_two)
...
 }
 
+TEST (conflicts, add_two)
+{
+	nano::system system{};
+	auto const & node = system.add_node ();
+
+	const auto do_send = [&node] (const auto & previous, const auto & from, const auto & to, bool forceConfirm = true)
+	-> std::pair<std::shared_ptr<nano::block>, std::shared_ptr<nano::block>> {
+		auto send = std::make_shared<nano::send_block> (previous->hash (), to.pub, 0, from.prv, from.pub, 0);
```
dsiganos, line 85: If we are re-writing it, should we use a state block instead of of legacy block?

nano/core_test/conflicts.cpp
```diff
@@ -71,6 +71,79 @@ TEST (conflicts, DISABLED_add_two)
 	ASSERT_EQ (2, node1.active.size ());
 }
 
+TEST (conflicts, add_two)
```
dsiganos, line 74: Can you add a comment in english that explains what this test case does?
Also a comment about the subtle but important fact that the node cannot vore because it does not have a signing key.
Like the annotations here:
https://github.com/nanocurrency/nano-node/issues/3535#issuecomment-959827386

nano/core_test/conflicts.cpp
```diff
@@ -71,6 +71,79 @@ TEST (conflicts, DISABLED_add_two)
...
+		ASSERT_TIMELY (3s, 5 == node->ledger.cache.cemented_count);
+	}
+
+	node->scheduler.activate (account3.pub, node->store.tx_begin_read ());
+	node->scheduler.flush ();
+
+	// wait 3s before asserting just to make sure there would be enough time
+	// for the AEC to evict both elections in case they would wrongfully get confirmed
+	//
+	std::this_thread::sleep_for (3s);
```
dsiganos, line 163: I do not know if this check is worth the extra 3 seconds of delay in executing core_test.

nano/core_test/conflicts.cpp
theohax: Unfortunately that was not as trivial as it seems -- I tried for a few good hours hitting various hurdles and when everything seemed fine, I could not get passed the "work requirement". We can review it in the dev call if you want, but otherwise I'd leave it like this working rather than not working.

nano/core_test/conflicts.cpp
theohax: So was the case before and it used to pass about 50% of the time. I actually think a 3s delay in test execution is a small price to pay for ensuring there are no race conditions.

nano/core_test/conflicts.cpp
theohax: Sounds good, I'll add a couple comments.

thsfs: I think this duplicates https://github.com/nanocurrency/nano-node/issues/3535.

nano/core_test/conflicts.cpp
```diff
@@ -71,6 +71,79 @@ TEST (conflicts, DISABLED_add_two)
...
+		const auto [send4, _] = do_send (receive2, account2, account3, false);
+		ASSERT_TRUE (send4);
+		ASSERT_TIMELY (3s, 5 == node->ledger.cache.cemented_count);
+	}
+
+	node->scheduler.activate (account3.pub, node->store.tx_begin_read ());
+	node->scheduler.flush ();
+
+	// wait 3s before asserting just to make sure there would be enough time
+	// for the AEC to evict both elections in case they would wrongfully get confirmed
```
thsfs: Suggest writing Active Elections Container instead.

thsfs:
I ran this test in a very very loaded system `stress -m 40` in a mid-life laptop, 50 times (sufficient to see this type of failure for major of the tests I have checked before). I haven't seem any failure. It looks good!

thsfs (dismissed):

nano/core_test/conflicts.cpp
clemahieu: If we want to permit a delay I think this should be rolled in to an ASSERT_TIMELY so it will exit quickly on success.

nano/core_test/conflicts.cpp
theohax: Addressed

theohax:
@dsiganos @thsfs -- addressed CR guys, looking forward to merging this and eventually maybe the three of us can have a short live session to try the state blocks and delete the old disabled test once everything is settled.

thsfs (approved):

clemahieu (approved):

PR #3546, opened 2021-11-05
Fix crash in nano_wallet when ran with --config option

### LINKED ISSUE #3494: ./nano_wallet --config rpc.enable=true crashes on linux
`./nano_wallet --config rpc.enable=true` crashes on linux (ubuntu 20.04)

On develop branch commit 1eefcaf70da5dd9b1f8e05f31ffd17478c827b2c

theohax:
Addresses https://github.com/nanocurrency/nano-node/issues/3494

There were 2 issues that are fixed in this PR:
- config parsing was not done correctly, hence an exception being thrown that was correctly caught and tried to be shown in a modal message box, no crash until here
- the modal message box would be instantiated without an alive `QApplication` object, thus the crash

Both issues should be fixed now.

### DIFF
--- a/nano/nano_wallet/entry.cpp
+++ b/nano/nano_wallet/entry.cpp
@@ -74,11 +74,11 @@ int run_wallet (QApplication & application, int argc, char * const * argv, boost
 	boost::filesystem::create_directories (data_path);
 	nano::set_secure_perm_directory (data_path, error_chmod);
 	QPixmap pixmap (":/logo.png");
-	QSplashScreen * splash = new QSplashScreen (pixmap);
+	auto * splash = new QSplashScreen (pixmap);
 	splash->show ();
-	application.processEvents ();
+	QApplication::processEvents ();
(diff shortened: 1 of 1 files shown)


nano/nano_wallet/entry.cpp
```diff
@@ -256,9 +256,11 @@ int main (int argc, char * const * argv)
 {
 	nano::set_umask ();
 	nano::node_singleton_memory_pool_purge_guard memory_pool_cleanup_guard;
+
+	QApplication application (argc, const_cast<char **> (argv));
```
clemahieu, line 260: I tried making this change only and it seemed to address the issue of the dialog not popping up correctly.

Specifically it didn't require making the substitution: application./QApplication:: anywhere

theohax: Ah, the `QApplication`/`application.` is unrelated, that was a clang-tidy warning telling us that we're invoking a static function via an object uselessly, the actual reason for which the crash happened was instantiating `QMessageBox` without a `QApplication` instance.

clemahieu (approved):

PR #3548, opened 2021-11-07
Adding duration to vote timestamps.

### LINKED ISSUE #3453: Add vote duration to low order bits of timestamp

clemahieu:
This patch creates backward/future-compatibility with associating a duration to a vote/timestamp. It uses the low-order 4 bits of the timestamp field in vote messages to pack a duration. Nodes that don't understand the duration field will see these bits as a small deviation in the timestamp, <16ms, which won't affect vote processing.

This value is packed in the 4 low order bits of the timestamp field. This forces the timestamp to have a minimum granularity of 16ms. The duration bits encode a duration = 2^(x+4) which represents 16-524,288ms in power of 2 increments.

The duration emitted after this change is 0x9 which encodes as 8192ms.

This implements https://github.com/nanocurrency/nano-node/issues/3453

### DIFF
--- a/nano/node/voting.cpp
+++ b/nano/node/voting.cpp
@@ -362,7 +362,8 @@ void nano::vote_generator::vote (std::vector<nano::block_hash> const & hashes_a,
 	std::vector<std::shared_ptr<nano::vote>> votes_l;
 	wallets.foreach_representative ([this, &hashes_a, &votes_l] (nano::public_key const & pub_a, nano::raw_key const & prv_a) {
 		auto timestamp = this->is_final ? nano::vote::timestamp_max : nano::milliseconds_since_epoch ();
-		votes_l.emplace_back (std::make_shared<nano::vote> (pub_a, prv_a, timestamp, hashes_a));
+		uint8_t duration = this->is_final ? nano::vote::duration_max : /*8192ms*/ 0x9;
+		votes_l.emplace_back (std::make_shared<nano::vote> (pub_a, prv_a, timestamp, duration, hashes_a));
 	});
 	for (auto const & vote_l : votes_l)
 	{
--- a/nano/secure/common.cpp
+++ b/nano/secure/common.cpp
@@ -756,8 +756,7 @@ bool nano::vote::validate () const
 
 uint64_t nano::vote::packed_timestamp (uint64_t timestamp, uint8_t duration) const
 {
-	debug_assert ((timestamp & 0xf) == 0 && "Invalid timestamp");
-	debug_assert ((duration & 0xf0) == 0 && "Invalid duration");
+	debug_assert (duration <= duration_max && "Invalid duration");
 	debug_assert ((!(timestamp == timestamp_max) || (duration == duration_max)) && "Invalid final vote");
 	return (timestamp & timestamp_mask) | duration;
 }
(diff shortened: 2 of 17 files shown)


thsfs (dismissed):

nano/secure/common.cpp
```diff
@@ -738,6 +754,14 @@ bool nano::vote::validate () const
 	return nano::validate_message (account, hash (), signature);
 }
 
+uint64_t nano::vote::packed_timestamp (uint64_t timestamp, uint8_t duration) const
+{
+	debug_assert ((timestamp & 0xf) == 0 && "Invalid timestamp");
```
dsiganos: This check looks wrong. A timestamp could be any number including all zeros at the lowest bits.

nano/secure/common.cpp
```diff
@@ -738,6 +754,14 @@ bool nano::vote::validate () const
...
+	debug_assert ((duration & 0xf0) == 0 && "Invalid duration");
```
dsiganos: Since you have `duration_max`, it would be slightly better to compare against it here.

dsiganos:
Conceptually how does this relate to final votes?
What value does a final vote have after this change?

Answer:
Final votes are represented in the same way as before, with all bits turned on.
Therefore, a final vote has a duration of approximately 8 minutes.

nano/secure/common.cpp
clemahieu: This check is to force the caller to discard the low 4 bits rather than packed_timestamp discarding them silently. This seemed less confusing but it could be done either way if there's a strong reason to truncate internally.

dsiganos: I see, I was expecting it to be the other way and I misread the assert.

I see now that you set the timestamp using this:
`nano::milliseconds_since_epoch () & nano::vote::timestamp_mask `

Personally, I would have gone with the option of silently dropping the timestamp inside the timestamp class just to keep the code simpler. So it would be treating the dropping of the 4 LSBs as a implementation detail and the creator of votes wouldn't need to do the masking of of LSBs. But there is no right or wrong in these things.

theohax (dismissed):

nano/node/voting.cpp
```diff
@@ -362,7 +362,9 @@ void nano::vote_generator::vote (std::vector<nano::block_hash> const & hashes_a,
 	std::vector<std::shared_ptr<nano::vote>> votes_l;
 	wallets.foreach_representative ([this, &hashes_a, &votes_l] (nano::public_key const & pub_a, nano::raw_key const & prv_a) {
 		auto timestamp = this->is_final ? nano::vote::timestamp_max : nano::milliseconds_since_epoch ();
-		votes_l.emplace_back (std::make_shared<nano::vote> (pub_a, prv_a, timestamp, hashes_a));
+		timestamp &= nano::vote::timestamp_mask;
+		uint8_t duration = this->is_final ? nano::vote::duration_max : /*8192ms*/ 0x9;
```
dsiganos, line 365: Should the default duration value be configurable?

Ultimatetinycellstuan (dismissed):

nano/secure/common.cpp
clemahieu: Changed this so the caller doesn't need to truncate the bits.

nano/node/voting.cpp
clemahieu: No the duration needs to be software controlled.

dsiganos (approved):

PR #3553, opened 2021-11-19
Adds an Unchecked Map - wrapper for the unchecked table outside the store class

thsfs:
This PR creates a wrapper for the unchecked table, so it starts getting decoupled from the store logic, since this table is not part of the ledger. This will facilitate further refactoring like supporting more than type of DB instantiated.

It also simplifies the LMDB and RocksDB related implementation by removing source code from our side.

### DIFF
--- a/nano/core_test/CMakeLists.txt
+++ b/nano/core_test/CMakeLists.txt
@@ -42,6 +42,7 @@ add_executable(
   toml.cpp
   timer.cpp
   uint256_union.cpp
+  unchecked_map.cpp
   utility.cpp
   vote_processor.cpp
   voting.cpp
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -22,6 +22,8 @@
 
 #include <stdlib.h>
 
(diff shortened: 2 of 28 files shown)


thsfs:
for the unchecked_map change, there is still an error on block_or_pruned_exists when running the tests. These won't fail with ASSERT_TIMELY. As the DB writing happens in a concurrent thread, should this now be the proper way to check the blocks are there?

```
2021-12-13T14:46:55.1189467Z [ RUN      ] bootstrap_processor.lazy_unclear_state_link
2021-12-13T14:46:55.2258106Z /workspace/nano/core_test/bootstrap.cpp:928: Failure
2021-12-13T14:46:55.2259741Z Value of: node2->ledger.block_or_pruned_exists (open->hash ())
2021-12-13T14:46:55.2260336Z   Actual: false
2021-12-13T14:46:55.2260727Z Expected: true
2021-12-13T14:46:55.4969842Z [  FAILED  ] bootstrap_processor.lazy_unclear_state_link (378 ms)
```

clemahieu (approved):

PR #3567, opened 2021-11-22
Turns off sub-network connection limiting for IPv4

thsfs:
The current prefix limit for a IPv6 sub-network is 64, pretty similar to the most of the ISP provided IP sub-networks. Since the IPv4 address space is much smaller than this it is not necessary to limit it by sub-network.

So, what it does is to turn-off the limiting for when the endpoint IP of the incoming connection is IPv4 or v4 mapped IPv6.
The limit will only work for IPv6 only IPs.

### DIFF
--- a/nano/node/socket.cpp
+++ b/nano/node/socket.cpp
@@ -304,14 +304,15 @@ size_t network_prefix)
 bool nano::server_socket::limit_reached_for_incoming_subnetwork_connections (std::shared_ptr<nano::socket> const & new_connection)
 {
 	debug_assert (strand.running_in_this_thread ());
-	if (node.flags.disable_max_peers_per_subnetwork)
+	if (node.flags.disable_max_peers_per_subnetwork || nano::transport::is_ipv4_or_v4_mapped_address (new_connection->remote.address ()))
 	{
 		// If the limit is disabled, then it is unreachable.
+		// If the address is IPv4 we don't check for a network limit, since its address space isn't big as IPv6 /64.
 		return false;
 	}
 	auto const counted_connections = socket_functions::count_subnetwork_connections (
 	connections_per_address,
-	nano::transport::mapped_from_v4_or_v6 (new_connection->remote.address ()),
+	new_connection->remote.address ().to_v6 (),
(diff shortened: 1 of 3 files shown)


nano/node/socket.cpp
```diff
@@ -304,14 +304,15 @@ size_t network_prefix)
 bool nano::server_socket::limit_reached_for_incoming_subnetwork_connections (std::shared_ptr<nano::socket> const & new_connection)
 {
 	debug_assert (strand.running_in_this_thread ());
-	if (node.flags.disable_max_peers_per_subnetwork)
+	if (node.flags.disable_max_peers_per_subnetwork || nano::transport::is_ipv4_or_v4_mapped_address (new_connection->remote.address ()))
```
dsiganos, line 307: I do not think we need to do a source code change to handle this.
We could set the subnetwork to default to /32 (/128 for ipv6 ipv4-mapped)

dsiganos: We do not have a separate subnetwork limit for ipv4 addresses, so my comment does not make any sense.

nano/node/socket.cpp
```diff
@@ -304,14 +304,15 @@ size_t network_prefix)
...
 	{
 		// If the limit is disabled, then it is unreachable.
+		// If the address is IPv4 we don't check for a network limit, since its address space isn't big as IPv6 /64.
 		return false;
 	}
 	auto const counted_connections = socket_functions::count_subnetwork_connections (
 	connections_per_address,
-	nano::transport::mapped_from_v4_or_v6 (new_connection->remote.address ()),
+	new_connection->remote.address ().to_v6 (),
```
dsiganos, line 315: I think we should convert to v6 higher up the stack.

dsiganos: Hmmm, I can;t see an obvious place to do it. So never mind.

thsfs: I see. The problem is that the socket class holds a boost endpoint that has the address type (convertible to either v4 or v6).

dsiganos (approved):

clemahieu (approved):

PR #3569, opened 2021-11-23
Refine election scheduler lock.

clemahieu:
This is a change to release the election scheduler lock while doing active_transactions operations. This ensures the election scheduler is always available to producers while the queue is being consumed.


thsfs (approved):

thsfs:
Looks good, the critical data seems to remain protected.

dsiganos (approved):

PR #3572, opened 2021-11-24
Unlocking lock in all branch cases.

clemahieu:
Unlocking lock in all branch cases otherwise lock.lock () can double-lock.

### DIFF
--- a/nano/node/election_scheduler.cpp
+++ b/nano/node/election_scheduler.cpp
@@ -140,6 +140,10 @@ void nano::election_scheduler::run ()
 					election->transition_active ();
 				}
 			}
+			else
+			{
+				lock.unlock ();
+			}
 			notify ();
 			lock.lock ();
 		}


thsfs (approved):

dsiganos (approved):

nano/node/election_scheduler.cpp
```diff
@@ -140,6 +140,10 @@ void nano::election_scheduler::run ()
 					election->transition_active ();
```
dsiganos, line 140: Is there an else clause (with a lock.unlock ()) missing here as well?
If we think we do not need an else clause then lets add an else clause containing assert zero.

dsiganos: There is no problem here. I misread the code because the indentation appears wrong in the browser.

dsiganos (approved):

PR #3575, opened 2021-11-26
Add nano::message_header::to_string ()

Airtune:
Added:
std::string nano::message_header::to_string ()
std::string nano::message_type_to_string (nano::message_type)
std::string nano::network::to_string (nano::networks)


keerifox:
this was continued in #3583 and can probably be closed

PR #3577, opened 2021-11-27
exclude latest for live

### LINKED ISSUE #3573: Update workflows to allow `latest` container tag for beta/test
We are currently excluding any tags containing `latest` in their name from being pushed, but we want this included for beta and test only: https://github.com/nanocurrency/nano-node/blob/develop/ci/actions/linux/deploy-docker.sh#L54

There is already an exclusion further above to ensure live doesn't have any latest tags included.

We can remove the `latest-with-rc` tag option as well because that isn't useful on beta/test networks and isn't differentiated from the `latest` tag.

argakiig:
still allow for beta or test network runs
drops latest-including-rc entirely
closes #3573

### DIFF
--- a/ci/actions/linux/deploy-docker.sh
+++ b/ci/actions/linux/deploy-docker.sh
@@ -7,9 +7,6 @@ TRAVIS_BRANCH=$(git branch | cut -f2 -d' ')
 tags=()
 if [ -n "$TRAVIS_TAG" ]; then
     tags+=("$TRAVIS_TAG")
-    if [[ "$GITHUB_WORKFLOW" = "Beta" || "$GITHUB_WORKFLOW" = "Test" ]]; then
-        tags+=(latest)
-    fi
 elif [ -n "$TRAVIS_BRANCH" ]; then
     TRAVIS_TAG=$TRAVIS_BRANCH
     tags+=("$TRAVIS_BRANCH")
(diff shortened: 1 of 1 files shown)


ci/actions/linux/deploy-docker.sh
```diff
@@ -8,7 +8,7 @@ tags=()
 if [ -n "$TRAVIS_TAG" ]; then
     tags+=("$TRAVIS_TAG")
     if [[ "$GITHUB_WORKFLOW" = "Beta" || "$GITHUB_WORKFLOW" = "Test" ]]; then
-        tags+=(latest latest-including-rc)
+        tags+=(latest)
```
zhyatt: @argakiig If we only have `latest` here is this needed? I vaguely recall that tag being added by default, not needing it specifically. If that is the case, then the selective removal of `latest` for live in line 55 below should be enough.

argakiig: Good call, will adjust to just remove this case entirely

zhyatt (approved):

zhyatt:
Removed my review and still considering impacts of this. I realize that if we want to use these workflows for beta builds that aren't being widely distributed to the beta network, having the latest include them is problematic. 🤔

zhyatt (approved):
I think the above mentioned scenario of beta builds that are not widely distributed is very much an edge case we don't need to solve for here. This looks good.

PR #3582, opened 2021-12-04
Include election.confirmed outcome in log

jamescoxon:
Noted by @RickiNano on beta-testing that its not clear the outcome of an election when it is being erased (confirmed or dropped) by reviewing node logs. Added true/false to log string.

Co-authored-by: RickiNano <81099017+RickiNano@users.noreply.github.com>

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -382,7 +382,7 @@ void nano::active_transactions::cleanup_election (nano::unique_lock<nano::mutex>
 			node.network.publish_filter.clear (block);
 		}
 	}
-	node.logger.try_log (boost::str (boost::format ("Election erased for root %1%") % election.qualified_root.to_string ()));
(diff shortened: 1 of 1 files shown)


nano/node/active_transactions.cpp
```diff
@@ -382,7 +382,7 @@ void nano::active_transactions::cleanup_election (nano::unique_lock<nano::mutex>
 			node.network.publish_filter.clear (block);
 		}
 	}
-	node.logger.try_log (boost::str (boost::format ("Election erased for root %1%") % election.qualified_root.to_string ()));
+	node.logger.try_log (boost::str (boost::format ("Election erased for root %1%, confirmed: %2$b") % election.qualified_root.to_string () % election.confirmed ()));
```
dsiganos, line 385: There is a strange $b that looks wrong.

jamescoxon: this turns the 1/0 into true/false (from boost::format)

dsiganos: Yeap, section "printf format specifications" in
https://www.boost.org/doc/libs/1_66_0/libs/format/doc/format.html

dsiganos (approved):

PR #3583, opened 2021-12-05
to_string function for message_header class

dsiganos:
A simple commit adding a to_string() function to message_header.
This was done by Airtune. Hopefully this is the first of a series of commits adding to_string() functionality to many classes.

### DIFF
--- a/nano/node/common.cpp
+++ b/nano/node/common.cpp
@@ -140,10 +140,10 @@ std::string nano::message_header::to_string ()
 
 	std::stringstream stream;
 
-	stream << boost::format ("NetID: %1%, ") % nano::network::to_string (network);
+	stream << boost::format ("NetID: %1%(%2%), ") % nano::to_string_hex (static_cast<uint16_t> (network)) % nano::network::to_string (network);
 	stream << boost::format ("VerMaxUsingMin: %1%/%2%/%3%, ") % version_max_l % version_using_l % version_min_l;
 	stream << boost::format ("MsgType: %1%(%2%), ") % type_l % type_text;
-	stream << boost::format ("Extensions: %1%") % extensions.to_ulong ();
+	stream << boost::format ("Extensions: %1%") % nano::to_string_hex (static_cast<uint16_t> (extensions.to_ulong ()));
 
 	return stream.str ();
 }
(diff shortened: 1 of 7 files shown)


nano/node/common.cpp
```diff
@@ -90,6 +93,61 @@ bool nano::message_header::deserialize (nano::stream & stream_a)
...
+	uint16_t version_using_l = static_cast<uint16_t> (version_using);
+	uint16_t version_min_l = static_cast<uint16_t> (version_min);
+	std::string type_text = nano::message_type_to_string (type);
+
+	std::stringstream stream;
+
+	stream << boost::format ("NetID: %1%, ") % nano::network::to_string (network);
+	stream << boost::format ("VerMaxUsingMin: %1%/%2%/%3%, ") % version_max_l % version_using_l % version_min_l;
+	stream << boost::format ("MsgType: %1%(%2%), ") % type_l % type_text;
+	stream << boost::format ("Extensions: %1%") % extensions.to_ulong ();
```
clemahieu: I think the extensions should be printed as hex since it's a bitfield.

nano/node/common.cpp
```diff
@@ -90,6 +93,61 @@ bool nano::message_header::deserialize (nano::stream & stream_a)
...
+		case nano::message_type::bulk_pull_account:
+			return "bulk_pull_account";
+		case nano::message_type::telemetry_req:
+			return "telemetry_req";
+		case nano::message_type::telemetry_ack:
+			return "telemetry_ack";
+			// default case intentionally omitted to cause warnings for unhandled enums
+	}
+
+	return "n/a";
```
clemahieu, line 129: If the type isn't in the enumeration should we still print out what it is?

nano/node/network.cpp
```diff
@@ -1074,3 +1074,23 @@ std::unique_ptr<nano::container_info_component> nano::syn_cookies::collect_conta
...
+		case nano::networks::nano_dev_network:
+			return "dev";
+		case nano::networks::nano_live_network:
+			return "live";
+		case nano::networks::nano_test_network:
+			return "test";
+			// default case intentionally omitted to cause warnings for unhandled enums
+	}
+
+	return "n/a";
```
clemahieu, line 1095: Seems we should print the unknown value here too.

clemahieu (approved):

thsfs (approved):

PR #3586, opened 2021-12-06
Update images for README and QT wallet to be new symbol

zhyatt:


argakiig (approved):

dsiganos:
I built nano_wallet on linux and looked at it but I do not see any nano logos... what changed and where is it?

zhyatt:
@dsiganos Should be splash screen on load up of the QT wallet and app logo on the OS. @argakiig Can you verify those are the locations that you verified?

argakiig:
@dsiganos Linux associates the image external to the resulting binary, do you see It on the splash. @zhyatt can confirm the built binary on win had it as well as on the splash.

dsiganos:
I do not see a splash image and I do not see any logos anywhere.

clemahieu:
I'm able to see the splash screen quickly on launch with the symbol.

clemahieu (approved):

zhyatt:
> I do not see a splash image and I do not see any logos anywhere.

@dsiganos Are you seeing a splash screen on initial QT wallet startup that has the old logo or no logo?

dsiganos:
I see something that looks like the borders of a splash screen for a second but it has no contents.
I hadn't even noticed it until now. So I think the splash screen doesn't show on my machine for some reason.
I didn't even know there was supposed to be a splash screen until now.

zhyatt:
> I see something that looks like the borders of a splash screen for a second but it has no contents. I hadn't even noticed it until now. So I think the splash screen doesn't show on my machine for some reason. I didn't even know there was supposed to be a splash screen until now.

So you haven't previously seen a splash screen with other builds? Perhaps there is a longer standing issue with the splash screen on Linux not related to these simple image changes.

dsiganos:
The effect I see (no splash screen, just the outline of a border) is not a regression.
The same things happens with V22.1.

zhyatt:
> The effect I see (no splash screen, just the outline of a border) is not a regression. The same things happens with V22.1.

Ok thanks for confirming, I created another issue to tackle that. https://github.com/nanocurrency/nano-node/issues/3599

clemahieu:
It sounds like it's disappearing really quickly. Maybe the startup is just fast? We don't have a minimum time it stays up.

PR #3587, opened 2021-12-06
Fixes the early drops of not yet timed out connections

thsfs:
Both member variables of nano::socket class, last_receive_time and last_completion_time were being initialized to 0. This won't be a problem if these values were not being used in concurrent access context.

now (current clock in seconds) - 0 would always evaluate true compared greater than an expiration time given in (few) seconds. So initializing them with a clock time makes more sense, since there will be still a valid time even if receive or completion events never happen.


dsiganos (approved):
I do not see a need for the member variable init_time_reference and I am not a fan of the very long variable name but I it looks good otherwise.

PR #3600, opened 2021-12-10
Unchecked cleanup

clemahieu:
This is a series of commits in preparation for adding the unchecked_map ADT.

Each commit is best reviewed individually.

### DIFF
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -352,7 +352,7 @@ TEST (bootstrap, simple)
 	nano::logger_mt logger;
 	auto store = nano::make_store (logger, nano::unique_path (), nano::dev::constants);
 	ASSERT_TRUE (!store->init_error ());
-	auto block1 (std::make_shared<nano::send_block> (0, 1, 2, nano::keypair ().prv, 4, 5));
+	std::shared_ptr<nano::block> block1 = std::make_shared<nano::send_block> (0, 1, 2, nano::keypair ().prv, 4, 5);
 	auto transaction (store->tx_begin_write ());
 	auto block2 (store->unchecked.get (transaction, block1->previous ()));
 	ASSERT_TRUE (block2.empty ());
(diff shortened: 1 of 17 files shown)


nano/core_test/block_store.cpp
```diff
@@ -352,7 +352,7 @@ TEST (bootstrap, simple)
 	nano::logger_mt logger;
 	auto store = nano::make_store (logger, nano::unique_path (), nano::dev::constants);
 	ASSERT_TRUE (!store->init_error ());
-	auto block1 (std::make_shared<nano::send_block> (0, 1, 2, nano::keypair ().prv, 4, 5));
+	std::shared_ptr<nano::block> block1 = std::make_shared<nano::send_block> (0, 1, 2, nano::keypair ().prv, 4, 5);
```
theohax, line 355: This clean up looks good, but since you touched the lines anyway, why not `auto block = std::make_shared...`? Not really important but just in the spirit of `if cleaning it up, clean it all the way` :D.

nano/core_test/node.cpp
```diff
@@ -3317,7 +3317,9 @@ TEST (node, block_processor_signatures)
 				 .sign (nano::dev::genesis_key.prv, nano::dev::genesis_key.pub)
 				 .work (*node1.work_generate_blocking (send3->hash ()))
 				 .build_shared ();
-	send5->signature.bytes[31] ^= 0x1;
+	auto signature = send5->block_signature ();
+	signature.bytes[31] ^= 0x1;
+	send5->signature_set (signature);
```
theohax, line 3322: I agree it's more clean to get the signature, touch it, then set it back, but was there a functional problem with changing it in-place?

theohax:
Looks good to me; also, reviewing commits individually turned out to be really helpful, so I recommend that other reviewers do that.

theohax (approved):

nano/core_test/node.cpp
clemahieu: Since we're holding a pointer to nano::block instead of the specific block type, the signature field is no longer directly accessible.

nano/core_test/block_store.cpp
clemahieu: This one specifically is because std::shared_ptr<nano::send_block> won't implicitly convert to unchecked_info when passed to unchecked_store::put. It won't do an upcast and also an implicit conversion.

dsiganos (approved):

PR #3601, opened 2021-12-11
Display node startup time for easy access to up time without RPC access

RickiNano:


dsiganos:
Personally, I think it should be the responsibility of the layer that starts nano_node (e.g. systemd) to record the start time and not nano_node itself but since it is just a single line, I have no objection to adding this.

However, printing localtime without printing timezone info would create as many problems as it solves.
Servers are typically running in different timezones from the operator and often are not setup properly for timezone and printing localtime usually creates too many problems. For this reason, usually timestamps are printed in UTC.

If we are going to print this, I think we should print in UTC and make it obvious in the output that it is UTC.

RickiNano:
That's a fair point. I have updated the PR to use UTC time

dsiganos (dismissed):

dsiganos:
Hi Ricki, please run the script `ci/clang-format-do.sh` to format the code, to commit code into the project, it must be formatted according to the rules of that script. The problem in your case is some extra spaces or tables at the end of one line. Other than that, your commit looks fine.

RickiNano:
I was unable to run the clang-format-do.sh script. It's probably because I'm using windows. I downloaded the clang format plugin for Visual Studio and ran it. Hope it passes the test now

dsiganos:
Yes, it looks good now, thank you for doing the boring bits and pushing it all the way.

dsiganos (approved):

PR #3606, opened 2021-12-14
Better move semantics & fix clang-tidy warnings

theohax:
This is a PR born from the work that I've been doing lately in the networking areas -- things that did not belong to any specific PR, but generally fixing a few clang-tidy warnings and better usage of move semantics.

For a more thorough explanation of why some pass-by-ref is replaced with pass-by-value (and subsequent `std::move`s), check out [this comment](https://github.com/nanocurrency/nano-node/pull/3576#discussion_r768701381), I think it explains it pretty well, but also feel free to bring up any thought you have here, I am open to discuss about it. Thanks!

### DIFF
--- a/nano/lib/ipc_client.cpp
+++ b/nano/lib/ipc_client.cpp
@@ -46,7 +46,7 @@ class socket_client : public nano::ipc::socket_base, public channel, public std:
 {
 public:
 	socket_client (boost::asio::io_context & io_ctx_a, ENDPOINT_TYPE endpoint_a) :
-		socket_base (io_ctx_a), endpoint (endpoint_a), socket (io_ctx_a), resolver (io_ctx_a), strand (io_ctx_a.get_executor ())
+		socket_base (io_ctx_a), endpoint (std::move (endpoint_a)), socket (io_ctx_a), resolver (io_ctx_a), strand (io_ctx_a.get_executor ())
 	{
 	}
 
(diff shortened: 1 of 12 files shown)


thsfs:
There too many changes for a single commit, I'd suggest grouping them per change type.

nano/lib/ipc_client.cpp
```diff
@@ -107,6 +107,8 @@ class socket_client : public nano::ipc::socket_base, public channel, public std:
 		}));
 	}
 
+	// TODO: investigate clang-tidy warning about recursive call chain
```
thsfs, line 110: I wouldn't recommend adding a TODO comment for something that comes from outside the source code (a tool), since the warning will eventually come up when running clang-tidy.

thsfs (approved):

theohax:
> There too many changes for a single commit, I'd suggest grouping them per change type.

They are really (all changes) fixing `cland-tidy` warnings, so the commit message isn't misguiding at all. It could maybe be broken down into more specific warning types, I'll keep that in mind for next time.

nano/lib/ipc_client.cpp
theohax: Yeah that's more a note for myself to be honest, I intend to investigate and fix those two very soon so it's easier to see the todo rather than look up the warning again. But I agree with your advice generally.

PR #3610, opened 2021-12-14
Fixing slow_test compile error.

clemahieu:
Fixes a compile error in slow_test due to cleanup of the unchecked_map table.


theohax:
This is also being done in https://github.com/nanocurrency/nano-node/pull/3606 but yeah sure let's merge this as it has a better naming at least.

theohax (approved):

PR #3616, opened 2021-12-15
Changes the silent connection timeouts to consider full-duplex communication flow

thsfs:
The first implementation wrongly considered the communication to be half-duplex, so talking-only connections were getting shutdown as the checking time was only being set for the `async_read()` function. The new implementation considers now the silent connections to be when there is no communication in any of both directions.

This PR also changes the the silent/drop procedure to be applicable only to the server-side. The `nano::socket` class was changed to get informed on its constructor about the `endpoint_type` that can be either socket or client.

### DIFF
--- a/nano/node/socket.cpp
+++ b/nano/node/socket.cpp
@@ -22,6 +22,7 @@ nano::socket::socket (nano::node & node_a) :
 	next_deadline{ std::numeric_limits<uint64_t>::max () },
 	last_completion_time_or_init{ nano::seconds_since_epoch () },
 	last_receive_time_or_init{ nano::seconds_since_epoch () },
+	last_send_time_or_init{ nano::seconds_since_epoch () },
 	io_timeout{ node_a.config.tcp_io_timeout },
 	silent_connection_tolerance_time{ node_a.network_params.network.silent_connection_tolerance_time }
 {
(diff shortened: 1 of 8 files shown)


clemahieu (dismissed):

theohax (dismissed):
Looks good, it's basically a symmetric thing of what was already being done for the receive, right?

thsfs:
> Looks good, it's basically a symmetric thing of what was already being done for the receive, right?

Yes, that is right.

theohax:
@thsfs is this ready to merge?

dsiganos:
I suspect this is dead. Thiago is working on another solution that will focus on silence on tcp server side only.

thsfs:
> I suspect this is dead. Thiago is working on another solution that will focus on silence on tcp server side only.

I'm intending to append a complement to this. It will check for is_server () instead of type == realtime, plus the related code for ensuring the socket runs on the server-side.

nano/node/socket.cpp
```diff
@@ -140,14 +149,20 @@ void nano::socket::checkup ()
 		{
 			uint64_t now (nano::seconds_since_epoch ());
 			auto condition_to_disconnect{ false };
-			if (this_l->is_realtime_connection () && (now - this_l->last_receive_time_or_init) > this_l->silent_connection_tolerance_time.count ())
+			auto is_silent = [this_l] (auto const & time) {
+				auto const silent_for_receiving = (time - this_l->last_receive_time_or_init) > this_l->silent_connection_tolerance_time.count ();
+				auto const silent_for_sending = (time - this_l->last_send_time_or_init) > this_l->silent_connection_tolerance_time.count ();
+				return silent_for_receiving && silent_for_sending;
```
dsiganos, line 155: This `silent_for_receiving && silent_for_sending` shouldn't be an "AND" relationship.
It should at least be an OR relationship but I am not sure if we have a guaranteed minimum reply gap.

thsfs:
Replaced by: https://github.com/nanocurrency/nano-node/pull/3623

PR #3618, opened 2021-12-16
Update confirmation_info RPC call to include representatives final votes for monitoring purposes

MajorChump:
We're attempting to debug the slow cementing on the long chain on beta, at the moment its not possible to determine whether a vote received from a specific representative is final or not. We're seeing quite a few elections will total > quorum but final < quorum this may help debugging the issue.


dsiganos:
Hi guys, it is great that you are pushing pieces of work to us and we thank you and encourage you to do more of that.
However, it would be nice, if when submitting the push request, you explain in detail what you are doing, what possible effects the change will have to the rest of the system, whether it is backwards compatible, how much you tested it and ideally submit a unit test for it.

Without having a report to give me some idea of the above, I would have to spend at least 2-3 hours to discover the above by myself and I generally do not have the luxury of time to do that and then the PR sits waiting for attention. If you could please provide the above information, you would make our life much easier.

MajorChump:
Tried changing the base to develop as its unlikely to ship for V23 now as the DB5 release seems to resolve the underlying issue and this is just an RPC improvement. This has meant a number of unrelated commits which have been merged into releases/v23 have been included. 

Will the v23 be merged back into develop upon release? If so I will leave PR as it is is, if not I can cherrypick my 2 commits and recreate.

PR #3626, opened 2021-12-29
upgraded Qt to 6 for nano_wallet

jamescoxon:
(originally released in 2020), while the last version of Qt5 will have support for 3 years for commercial licence users increasingly there will be a move to Qt6 - for example for M1 silicon only Qt6 is available (for now https://www.qt.io/blog/qt-on-apple-silicon)

also important to recognise that nano_wallet is rarely used these days due to better community gui wallets.

Main change is making Qt references in CMakeList.txt versionless

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -633,6 +633,10 @@ if(NANO_GUI OR RAIBLOCKS_GUI)
 
   find_package(Qt6 COMPONENTS Core Gui Widgets ${PLATFORM_QT_PACKAGES})
 
+  if (NOT Qt6_FOUND)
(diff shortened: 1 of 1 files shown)


dsiganos:
Abandoning this Pr because we do not need QT6. We were upgrading to Qt6 because we thought Qt6 was necessary to work on Apple M1.

However, the following appears to work on Apple M1.
```
brew install qt@5
brew link qt@5
```
Then adjust PATH in .zrcsh as suggested by brew link command:
```
echo 'export PATH="/opt/homebrew/opt/qt@5/bin:$PATH"' >> ~/.zshrc
```

Issue #3495, opened 2021-10-07
nano_wallet: does not show cemented block count, which is a vital statistic


dsiganos:
nano_wallet does not show the number of cemented blocks. It shows the number of checked and unchecked blocks only. But cemented block count is vital for determining the health of a node and for determining bootstrap success. I propose that we add cemented block count on nano_wallet somehow.

qwahzi:
If it's not too much trouble, I would also ask for all block_count RPC stats to be displayed somewhere in the UI. I.e. this info:

```
{
    "count": "49658059",
    "unchecked": "0",
    "cemented": "24104738",
    "full": "26090306",
    "pruned": "23567753"
}
```

Could maybe be shown here, perhaps by hovering with your mouse?
![image](https://user-images.githubusercontent.com/3682613/136422002-398cb4a8-8f58-404f-b1d4-77fc6c4fa36d.png)

jamescoxon:
<img width="439" alt="screenshot5" src="https://user-images.githubusercontent.com/175479/147711084-997671b7-d02e-4ad5-872d-d9959f8e9b70.png">

As an example

zhyatt:
Resolved by #3627

PR #3627, opened 2021-12-29
added cemented, full and pruned to nano_wallet gui

jamescoxon:
in response to #3495 . Pruning stats should only appear when pruning is enabled in config.


jamescoxon:
screenshot5

As an example

clemahieu (dismissed):

clemahieu (approved):

PR #3628, opened 2022-01-03
Refactoring within networking areas

theohax:
This PR cleans up some redundant/unused members/methods and adds better support for working with ports for various purposes.

Also, within `nano::socket` there are a few changes regarding the remote endpoint, which isn't now always guaranteed to exist, but better conveys semantics for the flow and the lifetime of the socket.

There are a couple changes that I missed to pull out in a different PR regarding shadowing member variables, etc, but they should be minimal.

### DIFF
--- a/nano/node/bootstrap/bootstrap_connections.cpp
+++ b/nano/node/bootstrap/bootstrap_connections.cpp
@@ -217,7 +217,7 @@ void nano::bootstrap_connections::populate_connections (bool repeat)
 			if (auto client = c.lock ())
 			{
 				new_clients.push_back (client);
-				endpoints.insert (client->socket->remote_endpoint ());
+				endpoints.insert (*client->socket->remote_endpoint ());
 				double elapsed_sec = client->elapsed_seconds ();
 				auto blocks_per_sec = client->sample_block_rate ();
 				rate_sum += blocks_per_sec;
--- a/nano/node/bootstrap/bootstrap_server.cpp
+++ b/nano/node/bootstrap/bootstrap_server.cpp
@@ -7,22 +7,21 @@
 #include <boost/format.hpp>
 #include <boost/variant/get.hpp>
 
-nano::bootstrap_listener::bootstrap_listener (uint16_t port_a, nano::node & node_a) :
-	node (node_a),
-	port (port_a)
+nano::bootstrap_listener::bootstrap_listener (nano::node & node_a) :
+	node (node_a)
 {
 }
 
-void nano::bootstrap_listener::start ()
+void nano::bootstrap_listener::start (std::uint16_t port_a)
 {
 	nano::lock_guard<nano::mutex> lock (mutex);
 	on = true;
-	listening_socket = std::make_shared<nano::server_socket> (node, boost::asio::ip::tcp::endpoint (boost::asio::ip::address_v6::any (), port), node.config.tcp_incoming_connections_max);
+	listening_socket = std::make_shared<nano::server_socket> (node, node.config.tcp_incoming_connections_max);
 	boost::system::error_code ec;
-	listening_socket->start (ec);
+	listening_socket->start (nano::tcp_endpoint{ boost::asio::ip::address_v6::any (), port_a }, ec);
 	if (ec)
 	{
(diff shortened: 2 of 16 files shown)


nano/node/bootstrap/bootstrap_connections.cpp
```diff
@@ -217,7 +217,7 @@ void nano::bootstrap_connections::populate_connections (bool repeat)
 			if (auto client = c.lock ())
 			{
 				new_clients.push_back (client);
-				endpoints.insert (client->socket->remote_endpoint ());
+				endpoints.insert (*client->socket->remote_endpoint ());
```
clemahieu, line 220: Is this guaranteed to have a value? Isn't remote_endpoint the side that can switch to ANYADDR?

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -33,33 +32,27 @@ void nano::bootstrap_listener::start ()
 	// (3): UDP disabled, port specified
 	// (4): UDP disabled, port not specified
 	//
-	const auto listening_port = listening_socket->listening_port ();
+	const auto listening_port = listening_socket->local_endpoint ()->port ();
```
clemahieu, line 35: Can local_endpoint become unavailable just like remote_endpoint?

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -95,7 +89,9 @@ std::size_t nano::bootstrap_listener::connection_count ()
 
 void nano::bootstrap_listener::accept_action (boost::system::error_code const & ec, std::shared_ptr<nano::socket> const & socket_a)
 {
-	if (!node.network.excluded_peers.check (socket_a->remote_endpoint ()))
+	auto const remote_endpoint = socket_a->remote_endpoint ();
+	debug_assert (remote_endpoint.has_value ());
```
clemahieu, line 93: Can we guarantees this? If they connect and immediately disconnect before this gets processed?

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -107,22 +103,24 @@ void nano::bootstrap_listener::accept_action (boost::system::error_code const &
...
 {
 	nano::lock_guard<nano::mutex> lock (mutex);
 	if (on && listening_socket)
 	{
-		return boost::asio::ip::tcp::endpoint (boost::asio::ip::address_v6::loopback (), port);
-	}
-	else
-	{
-		return boost::asio::ip::tcp::endpoint (boost::asio::ip::address_v6::loopback (), 0);
+		auto const local_endpoint = listening_socket->local_endpoint ();
```
clemahieu, line 116: This already returns an optional so it seems we can just return that directly.

nano/node/socket.cpp
```diff
@@ -324,28 +342,43 @@ size_t network_prefix)
...
 	{
 		// If the limit is disabled, then it is unreachable.
 		// If the address is IPv4 we don't check for a network limit, since its address space isn't big as IPv6 /64.
 		return false;
 	}
+
 	auto const counted_connections = socket_functions::count_subnetwork_connections (
 	connections_per_address,
-	new_connection->remote.address ().to_v6 (),
+	new_connection->remote_endpoint ()->address ().to_v6 (),
```
clemahieu, line 360: I don't think remote_endpoint () is guaranteed to dereference here. While there is a check remote_endpoint ().has_value () above, it's calling remote_endpoint () again here and it may not have a value during this call.

nano/node/socket.cpp
```diff
@@ -324,28 +342,43 @@ size_t network_prefix)
...
 	debug_assert (strand.running_in_this_thread ());
+
 	if (node.flags.disable_max_peers_per_ip)
 	{
 		// If the limit is disabled, then it is unreachable.
 		return false;
 	}
-	auto const address_connections_range = connections_per_address.equal_range (new_connection->remote.address ());
+
+	if (!new_connection->remote_endpoint ().has_value ())
```
clemahieu, line 376: If we're going to check has_value here and assume that below, this should be a copy of the endpoint.

dsiganos:
Closing because:
* It is too old
* It has too many merge conflicts
* it doesn’t seem to do anything important that we can’t easily redo

Issue #3574, opened 2021-11-26
Very slow confirmation with large amount of single account blocks


RickiNano:
### Summary

Confirming large amount of 1 raw receive blocks  becomes extremely slow
During testing of v23RC1  I created 300k+ single account blocks
At first the blocks were processed fine at around 2 cps. Over time though the confirmation has slowed to a crawl 

![image](https://user-images.githubusercontent.com/81099017/143558522-c2f3691c-43d7-4b03-96ca-7e56dc96e97f.png)

It has been speculated that each node has different blocks in the AEC and not voting on the same blocks.
The rest of the network seems fine and other blocks are confirmed quickly

### Node version

V23RC1

### Build details

Official binaries provided from the beta Discord

### OS and version

Windows 10

### Steps to reproduce the behavior

1. Create 300k spam blocks of 1 raw each to account "X"
2. Go to account "X" and receive the 300k blocks

### Expected behavior

Blocks should confirm steadily at about 2 CPS

### Actual behavior

Blocks are initially confirmed at 2 CPS but then becomes very erratic with long periods of no confirmations at all

### Possible solution

Could be caused by the AEC being filled with blocks that can not be confirmed until all the previous blocks of this chain has been confirmed
Blocks that have no confirmed previous block should not be in the AEC. I don't know if that is the case today

zhyatt:
@RickiNano So you published the 300k send blocks of 1 raw each and those were fully confirmed before you started the 300k receive blocks to match them? Or did you send out the 300k receive blocks before the sends were confirmed?

And to be clear, I am assuming the 300k send blocks were all done from a single account, thus the expected consistent 2-3 CPS.

RickiNano:
@zhyatt 
The 300k send blocks were actually just very old spam from months ago. They were all single account transactions and they were all confirmed.
I then began receiving the transactions to test this bug:
https://github.com/nanocurrency/nano-node/issues/3491
That's when I noticed the problem described in this ticket

zhyatt:
I see, thanks for clarifying @RickiNano

zhyatt:
[...]

PR #3644, opened 2022-01-04
This demonstrates a speedup in confirmation rate during initial boots…

clemahieu:
…trapping due to conservative limiters in the confirmation_solicitor. Rather than limiting with a constant number, it limits based on feedback directly from network channels.

Rather than limiting on a per-channel basis, a lot of complexity can be removed with by limiting the total number of active elections based on feedback from the confirmation rate.

### DIFF
--- a/nano/node/confirmation_solicitor.cpp
+++ b/nano/node/confirmation_solicitor.cpp
@@ -57,7 +57,6 @@ bool nano::confirmation_solicitor::add (nano::election const & election_a)
 	debug_assert (prepared);
 	bool error (true);
 	unsigned count = 0;
-	auto const max_channel_requests (config.confirm_req_batches_max * nano::network::confirm_req_hashes_max);
 	auto const & hash (election_a.status.winner->hash ());
 	for (auto i (representatives_requests.begin ()); i != representatives_requests.end () && count < max_election_requests;)
 	{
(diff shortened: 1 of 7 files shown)


MajorChump:
This looks like the issue I found with announcements being dropped due to max_channel_requests being exceeded. 

Looking at that PR it removes max_channel_requests from the flow, probably could clean the code up a little by removing 

`L60 auto const max_channel_requests (config.confirm_req_batches_max * nano::network::confirm_req_hashes_max);`

And confirm_req_batches_max could be removed from config I think.

Reference: https://github.com/nanocurrency/nano-node/issues/3574#issuecomment-982728204

nano/node/transport/tcp.hpp
```diff
@@ -68,6 +68,16 @@ namespace transport
 			return nano::transport::transport_type::tcp;
 		}
 
+		virtual bool full () override
+		{
+			bool result = true;
+			if (auto socket_l = socket.lock ())
+			{
+				result = socket_l->max ();
```
dsiganos, line 76: This function is called full() but it calls a socket function max().
Confusingly, there is a socket called full(), which the channel function full() does not call.
I think this should be at least called max() to match the socket max().
Although, I think it would be better to call both max functions something else, half_full() maybe?

I wonder if it would be better to have full(), half_full() and quarter_full() consistent across tcp sockets and tcp channels.

nano/node/confirmation_solicitor.cpp
```diff
@@ -70,7 +69,7 @@ bool nano::confirmation_solicitor::add (nano::election const & election_a)
 		if (!exists || !is_final || different)
 		{
 			auto & request_queue (requests[rep.channel]);
-			if (request_queue.size () < max_channel_requests)
+			if (!rep.channel->full ())
```
dsiganos: Completely filling up the buffers is possibly not a good idea. It might starve other processes of network bandwidth. We are going from a value of 2 to thousands. How about half or quarter full?

thsfs:
Looks good to me, but @dsiganos suggestions make sense, it is worth to improve the function names he pointed out.

dsiganos (approved):

thsfs (approved):

PR #3645, opened 2022-01-04
Already-confirmed transactions occupy space in the active_transaction…

clemahieu:
…s list. This was added to collect extra votes for diagnostics, however, the elections should be moved to another container so they don't occupy space that could be used by unconfirmed transactions.


dsiganos:
Does this mean that confirmed elections will not remain in the AEC for some minutes after confirmation?

qwahzi:
Is this issue related to (or resolved by) https://github.com/nanocurrency/nano-node/pull/3995 (immediately remove confirmed elections from AEC)?

dsiganos:
Yes, this is implemented by https://github.com/nanocurrency/nano-node/pull/3995.

PR #3648, opened 2022-01-04
Backlog population is done by a linear scan separated by a 5 minute w…

clemahieu:
…ait period. In this wait period and even while still scanning it's possible for the election scheduler to empty due to everything being confirmed faster than the scan can operate.

This should be rewritten to not scan at all and instead, a new database table can store the bucket/time as a key and when buckets are emptyied, it can be refilled directly from disk without scanning.

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -975,8 +975,12 @@ void nano::node::ongoing_unchecked_cleanup ()
 
 void nano::node::ongoing_backlog_population ()
 {
-	populate_backlog ();
+	auto overflow = populate_backlog ();
 	auto delay = config.network_params.network.is_dev_network () ? std::chrono::seconds{ 1 } : std::chrono::duration_cast<std::chrono::seconds> (std::chrono::minutes{ 5 });
+	if (overflow)
+	{
+		delay = std::chrono::seconds{ 0 };
+	}
 	workers.add_timed_task (std::chrono::steady_clock::now () + delay, [this_l = shared ()] () {
 		this_l->ongoing_backlog_population ();
 	});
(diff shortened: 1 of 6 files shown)


nano/node/node.cpp
```diff
@@ -1745,24 +1749,29 @@ std::pair<uint64_t, decltype (nano::ledger::bootstrap_weights)> nano::node::get_
 	return { max_blocks, weights };
 }
 
-void nano::node::populate_backlog ()
+bool nano::node::populate_backlog ()
 {
+	std::cerr << "<---------------Populating...\n";
```
jamescoxon, line 1754: Is a direct print to console needed here?

jamescoxon (changes requested):

nano/node/node.cpp
RickiNano: I've done some testing on this PR and it looks really good
This is the baseline test (v23 RC3 branch)
BaselineV23
Cycling is obvious

This V23 RC3 bu with included cherry pick of this PR:
V23+3648
No cycling and average CPS is approx three times higher

nano/node/node.cpp
```diff
@@ -975,8 +975,12 @@ void nano::node::ongoing_unchecked_cleanup ()
 
 void nano::node::ongoing_backlog_population ()
 {
-	populate_backlog ();
+	auto overflow = populate_backlog ();
 	auto delay = config.network_params.network.is_dev_network () ? std::chrono::seconds{ 1 } : std::chrono::duration_cast<std::chrono::seconds> (std::chrono::minutes{ 5 });
+	if (overflow)
```
fikumikudev, line 980: Is setting delay to 0 when there is overflown bucket really indented? The improvements @RickiNano wrote about are because it happens that it's always true, so we never wait thus constantly overfilling some buckets and in effect never starving aec. Not starving aec is good but definitely can be done more efficiently by using shorter delay time and doing backlog population when accounts queued in scheduler drop below some threshold.

aecbeforeafter

dsiganos:
Duplicate of https://github.com/nanocurrency/nano-node/pull/3999

dsiganos:
This just does the continuous backlog propagation in a a tight loop.
The new newer version does it better.

PR #3655, opened 2022-01-06
Remove json config code

### LINKED ISSUE #3530: Remove unnecessary json serializing of config files
With previous `.json` config files being replaced in V20.0 with `.toml` files, we kept the related `nano::logging::serialize_json` code to support upgrade paths from V19.0 forward. But now that V19.0 is an [inactive release](https://docs.nano.org/releases/node-releases/#inactive-releases) we should verify this serialization code and other code related to processing `.json` config files is removed and cleaned up.

cryptocode:
This improves compile/test times a bit and removes unnecessary code. 

The json->toml upgrade paths are removed as well. In the unlikely case someone needs to upgrade from v19.x and earlier, the solution is to run a v2x node (prior to this change) to perform the upgrade.

Closes #3530


clemahieu (dismissed):

nano/core_test/work_pool.cpp
```diff
@@ -115,21 +114,6 @@ TEST (work, opencl)
 	}
 }
 
-TEST (work, opencl_config)
```
dsiganos, line 118: Should this test case be removed?
Can we convert it to use toml and keep it?

nano/rpc_test/rpc.cpp
```diff
@@ -5592,40 +5592,6 @@ TEST (rpc, in_process)
 	ASSERT_EQ ("0", pending_text);
 }
 
-TEST (rpc_config, serialization)
```
dsiganos, line 5595: I think this test case should be converted to toml and kept.
I do not see an equivalent one for toml.

dsiganos:
Should we remove `nano::stat_config::deserialize_json` also?

theohax (dismissed):

nano/core_test/work_pool.cpp
cryptocode: Yes should be removed as this is covered by the TOML deserialization tests which have full coverage of all properties, including opencl. The serialization part is irrelevant for TOML.

nano/rpc_test/rpc.cpp
cryptocode: There's never any serialization to TOML (except when generating defaults - unless this has been changed recently).Deserialization is covered by `toml, rpc_config_deserialize_no_defaults` and `rpc_config_no_required`

cryptocode:
> Should we remove `nano::stat_config::deserialize_json` also?

Yeah good point, will push an update.

nano/lib/stats.cpp
```diff
@@ -10,36 +10,6 @@
 #include <fstream>
 #include <sstream>
 
-nano::error nano::stat_config::deserialize_json (nano::jsonconfig & json)
```
dsiganos, line 13: The header file declaration needs removing too.
`nano::error deserialize_json (nano::jsonconfig & json);`

dsiganos:
And I think we can also remove jsonconfig::read_and_update().

dsiganos (approved):

dsiganos:
Looks good to me! Thank you for doing the amendments!

clemahieu (approved):
Very nice cleanup. Thank you for pushing this through!

PR #3656, opened 2022-01-07
Allow overriding initial peering DNS by environment variable

dsiganos:
The initial peering DNS is peering.nano.org.
This commit allows to set NANO_DEFAULT_PEER to override it.
This allows for more flexibility when running private test networks.

### DIFF
--- a/nano/node/nodeconfig.cpp
+++ b/nano/node/nodeconfig.cpp
@@ -15,9 +15,9 @@ namespace
 char const * preconfigured_peers_key = "preconfigured_peers";
 char const * signature_checker_threads_key = "signature_checker_threads";
 char const * pow_sleep_interval_key = "pow_sleep_interval";
-char const * default_live_peer_network = "peering.nano.org";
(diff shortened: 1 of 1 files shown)


zhyatt:
@dsiganos In what scenario is using the beta network with custom peering DNS preferred over the test network which already allows this customization?

dsiganos:
If we wanted to run SPAM attack tests then they are easier on beta network than on test network due to reduced difficulty.
Additionally, being able to override the peering DNS to something that doesn't resolve, is a very easy way to stop a node from contacting any peers (after wiping the peer table).

dsiganos:
After discussion, it was decided to:
* Rename NANO_TEST_PEER_NETWORK to NANO_DEFAULT_PEER
* Apply it to the active network (live, beta, test)

clemahieu (approved):

zhyatt:
Just noting that there are no documentation changes to be considered for this.

PR #3662, opened 2022-01-11
Use the default peering port for daemon start up

theohax:
This PR fixes a small regression in that after #3554 (which correctly considers input port numbers with a value of 0 mean 'let the OS decide'), the default peering port was *not* used anymore by the node not only in unit tests (which is correct), but also by the daemon itself at start up.

Now the daemon checks the value that comes from the configuration and if there isn't a specific port value in there, it will automatically set the default peering port value for the given network type before attempting to create and start the node.


nano/node/node.cpp
```diff
@@ -631,7 +631,10 @@ void nano::node::start ()
 		{
 			network.port = bootstrap.port;
 		}
+
+		logger.always_log (boost::str (boost::format ("Node started with peering port `%1%`.") % network.port));
```
theohax, line 638: Thought it's good to have this logged down so that we can check the used value in the logs.

nano/node/node.cpp
```diff
@@ -702,6 +705,7 @@ void nano::node::keepalive_preconfigured (std::vector<std::string> const & peers
 {
 	for (auto i (peers_a.begin ()), n (peers_a.end ()); i != n; ++i)
 	{
+		// Question: why isn't keepalive being called with `network.port`? looks like it should. Can we make this change?
```
theohax: I want to push another commit to either fix this or at least remove the comment after we've discussed it.

@clemahieu @dsiganos -- does this look ok to you? Why would the default value be used for these keepalives?

dsiganos: The default value is used for the preconfigured keepalives because the preconfigured peers config only takes an address and not address and port pair as it probably should.
I would like to change the preconfigured peers config to take address and port info to make it more consistent and to make it easier to setup private tests networks.

nano/nano_node/daemon.cpp
```diff
@@ -128,6 +128,14 @@ void nano_daemon::daemon::run (boost::filesystem::path const & data_path, nano::
...
 			}
 
+			// a 0-valued port means for the node 'let the OS decide'; however, for the daemon start up, if
+			// the user hasn't specified a port in the config, we must use the default peering port for the network
+			//
+			if (!config.node.peering_port)
+			{
+				config.node.peering_port = network_params.network.default_node_port;
+			}
+
```
dsiganos, line 138: This looks good to me. Personally, I would have the following for maximum flexibility:
* convert peering_port from uint64_t to int.
* Set the default value to "-1" to specify the value is not set
* if the value is -1 then do `config.node.peering_port = network_params.network.default_node_port`

That would allow for 0 to mean, select the port number yourself and be more consistent.
But I am happy with the change as is, as well.

dsiganos (dismissed):

nano/nano_node/daemon.cpp
theohax: It is true that currently `0` means "there hasn't been any explicit choice made" (which is fine, no magic meaning), and which further might mean "let the OS decide" (which is again fine, it's an idiomatic meaning for the port value), or, could also mean "we have to set the default value" if it's the daemon who asks the question. I am not extremely happy about the last part, but it's a very reasonable magic meaning IMO.

I'd rather keep the port an unsigned as it's more semantic for a port value; having yet another magic number (`-1` in this case) to only represent that last case brings some temporary joy but leaves room for problems down the way IMHO.

But yeah, very decent thoughts, thanks for the review. Any thoughts about the question in the other comment? LE: oh, just saw you answered, nevermind.

nano/node/node.cpp
theohax: Ohh okay so we can't use our port because they use the default one anyways...if they haven't been set up with another explicit one which they mustn't in order for this to work... Ok I see. Well, yeah I agree with you there's some room for more flexibility in there, but it's clear for me now so I'll just take out the question and leave a small comment in there.

theohax: Alright, updated the comment.

dsiganos (dismissed):

dsiganos:
Is there a way to test this using a unit test?

theohax:
> Is there a way to test this using a unit test?

I was thinking about this. I don't think we have any daemon level unit tests.

nano/nano_node/daemon.cpp
clemahieu: std::optional would be the canonical way to indicate whether a value was specified or not. Then if it's std::nullopt_t we know it wasn't specified and if it's 0 we know it was specified to 0.

clemahieu:
> Is there a way to test this using a unit test?

We could test argument parsing by putting all that functionality in to a common function which we can call from unit tests.

If we're trying to test int main () functionality, we can move everything inside int main () to a new function with the same signature that does all the command line parsing. int main () would only call this new function and we can make a test that calls this signature with unit test argv/argc arguments to test parsing. For testing success, we could connect a directly created node to the node created through the entry-point and connect it on the expected port.

clemahieu (approved):

theohax:
Merging this -- with a ticket in the backlog for creating a test for it.

PR #3663, opened 2022-01-11
Fixing slow_test store.pruned_load.

clemahieu:
This was crashing on the line store->pruned.del as hashes was empty and querying for .begin (). hashes was never populated even when this test was originally added.

### DIFF
--- a/nano/slow_test/node.cpp
+++ b/nano/slow_test/node.cpp
@@ -16,6 +16,22 @@
 
 using namespace std::chrono_literals;
 
+/**
+ * function to count the block in the pruned store one by one
+ * we manually count the blocks one by one because the rocksdb count feature is not accurate
+ */
+size_t manually_count_pruned_blocks (nano::store & store)
+{
+	size_t count = 0;
+	auto transaction = store.tx_begin_read ();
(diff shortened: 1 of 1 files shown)


nano/slow_test/node.cpp
```diff
@@ -443,12 +443,13 @@ TEST (store, pruned_load)
...
 					nano::random_pool::generate_block (random_hash.bytes.data (), random_hash.bytes.size ());
 					store->pruned.put (transaction, random_hash);
+					hashes.insert (random_hash);
 				}
 			}
 			if (!nano::rocksdb_config::using_rocksdb_in_tests ())
 			{
 				auto transaction (store->tx_begin_write ());
-				for (auto k (0); k < batch_size / 2; ++k)
+				for (auto k (0); !hashes.empty () && k < batch_size / 2; ++k)
```
dsiganos, line 476: Do we need the check `!hashes.empty ()`?
This is a unit test, if anything is wrong, we want to fail loud and clear rather than handle it quietly.
So I would say that we should ASSERT that hashes is not empty.

dsiganos (dismissed):

dsiganos (approved):

Issue #3674, opened 2022-01-17
Move code inspection script to a separated file


thsfs:
Currently in `build-travis.sh`, this should be in its own script file, like we have the `clang-format-check.sh` and `cmake-format-check.sh`:
```
# This is to prevent out of scope access in async_write from asio which is not picked up by static analysers
if [[ $(grep -rl --exclude="*asio.hpp" "asio::async_write" ./nano) ]]; then
    echo "Using boost::asio::async_write directly is not permitted (except in nano/lib/asio.hpp). Use nano::async_write instead"
    exit 1
fi

# prevent unsolicited use of std::lock_guard, std::unique_lock, std::condition_variable & std::mutex outside of allowed areas
if [[ $(grep -rl --exclude={"*random_pool.cpp","*random_pool.hpp","*random_pool_shuffle.hpp","*locks.hpp","*locks.cpp"} "std::unique_lock\|std::lock_guard\|std::condition_variable\|std::mutex" ./nano) ]]; then
    echo "Using std::unique_lock, std::lock_guard, std::condition_variable or std::mutex is not permitted (except in nano/lib/locks.hpp and non-nano dependent libraries). Use the nano::* versions instead"
    exit 1
fi

if [[ $(grep -rlP "^\s*assert \(" ./nano) ]]; then
    echo "Using assert is not permitted. Use debug_assert instead."
    exit 1
fi
```

Issue #3673, opened 2022-01-17
Remove old Travis CI related scripts from the CI directory


thsfs:
We are not using Travis CI anymore. There are a couple of files related to it that should be removed or renamed in case we still use for another reason.

```
ci/deploy-travis.sh
ci/build-travis.sh
```

There are other mentions to Travis in `./actions/windows/build.ps1`, `./actions/linux/deploy-docker.sh`, and in `./deploy-docker.sh`.

PR #3675, opened 2022-01-17
Adds the code-inspector script

thsfs:
Attends the issue https://github.com/nanocurrency/nano-node/issues/3674. This script is part of `build-travis.sh` that may be removed from the CI directory soon (https://github.com/nanocurrency/nano-node/issues/3673). It is also better to have a dedicated script for these code inspections.

Modifies the GitHub - Static Analyzers to run also the `code-inspector-check.sh`.

### DIFF
--- a/.github/workflows/analyzers.yml
+++ b/.github/workflows/analyzers.yml
@@ -30,3 +30,11 @@ jobs:
             cmake-format
       - name: Check cmake-format
         run: ci/cmake-format-check.sh
+
+  code_inspector:
+    runs-on: ubuntu-20.04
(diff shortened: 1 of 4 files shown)


dsiganos:
This only adds a task. But I understtod that this is about moving a task from one place to another, so I would expect to also see a deletion.

thsfs:
> This only adds a task. But I understtod that this is about moving a task from one place to another, so I would expect to also see a deletion.

Made the `build-travis.sh` to reference it, that should be ok to make this PR coherent.

dsiganos (approved):

PR #3680, opened 2022-01-20
Enable and fix `node.unconfirmed_send` unit test

theohax:
I couldn't find an issue for this unit test to link it to the PR, perhaps it was disabled way back.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1681,82 +1681,51 @@ TEST (node, bootstrap_confirm_frontiers)
 	}
 }
 
-// Test that if we create a block that isn't confirmed, we sync.
+// Test that if we create a block that isn't confirmed, the bootstrapping processes sync the missing block.
 TEST (node, unconfirmed_send)
 {
 	nano::system system{};
 
-    nano::node_flags node_flags{};
-    node_flags.disable_bootstrap_listener = true;
-    node_flags.disable_rep_crawler = true;
-
-    nano::node_config node_config1{nano::get_available_port (), system.logging};
-    node_config1.frontiers_confirmation = nano::frontiers_confirmation_mode::disabled;
-
-	auto & node1 = *system.add_node(node_config1, node_flags);
+	auto & node1 = *system.add_node ();
 	auto wallet1 = system.wallet (0);
 	wallet1->insert_adhoc (nano::dev::genesis_key.prv);
 
 	nano::keypair key2{};
-    nano::node_config node_config2{nano::get_available_port (), system.logging};
-    // node_config2.frontiers_confirmation = nano::frontiers_confirmation_mode::disabled;
-
-    auto & node2 = *system.add_node(node_config2, node_flags);
+	auto & node2 = *system.add_node ();
 	auto wallet2 = system.wallet (1);
 	wallet2->insert_adhoc (key2.prv);
 
 	// firstly, send two units from node1 to node2 and expect that both nodes see the block as confirmed
 	// (node1 will start an election for it, vote on it and node2 gets synced up)
 	auto send1 = wallet1->send_action (nano::dev::genesis->account (), key2.pub, 2 * nano::Mxrb_ratio);
-    std::cout << "send1 = " << send1->hash().to_string() << std::endl;
 	ASSERT_TIMELY (5s, node1.block_confirmed (send1->hash ()));
 	ASSERT_TIMELY (5s, node2.block_confirmed (send1->hash ()));
-    // expect the balance on node2 to have increased, but actually this takes into
-    // calculation even non-confirmed blocks (which is the case for our receive block)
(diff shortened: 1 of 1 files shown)


nano/core_test/node.cpp
```diff
@@ -1682,25 +1682,43 @@ TEST (node, bootstrap_confirm_frontiers)
 }
 
 // Test that if we create a block that isn't confirmed, we sync.
-TEST (node, DISABLED_unconfirmed_send)
+TEST (node, unconfirmed_send)
```
dsiganos, line 1685: Could you write an overview paragraph here explaining what the test does overall?

dsiganos: On seconds thoughts,  it is fine, the inline comments are good enough to understand what is happening.

nano/core_test/node.cpp
```diff
@@ -1682,25 +1682,43 @@ TEST (node, bootstrap_confirm_frontiers)
...
+
+	nano::keypair key1{};
+	auto & node1 = *system.nodes[1];
+	auto wallet1 = system.wallet (1);
+	wallet1->insert_adhoc (key1.prv);
+
+	// firstly, send two units from node0 to node1 and expect they're both received (as balance) on the other end
+	//
+	auto send1 = wallet0->send_action (nano::dev::genesis->account (), key1.pub, 2 * nano::Mxrb_ratio);
+	ASSERT_TIMELY (10s, node1.balance (key1.pub) == 2 * nano::Mxrb_ratio && !node1.bootstrap_initiator.in_progress ());
```
dsiganos: Why does it matter if bootstrap initiator is in progress?

nano/core_test/node.cpp
```diff
@@ -1682,25 +1682,43 @@ TEST (node, bootstrap_confirm_frontiers)
...
+				 .previous (latest1)
+				 .representative (nano::dev::genesis_key.pub)
+				 .balance (nano::Mxrb_ratio)
+				 .link (nano::dev::genesis->account ())
+				 .sign (key1.prv, key1.pub)
+				 .work (*system.work.generate (latest1))
+				 .build_shared ();
+	ASSERT_EQ (nano::process_result::progress, node1.ledger.process (node1.store.tx_begin_write (), *send2).code);
+	node1.process_confirmed (nano::election_status{ send2 });
+	ASSERT_TIMELY (5s, node1.block_confirmed (send2->hash ()));
```
dsiganos: Does node0 get synced at this point?
Is there an election taking place at this point?
I think an overall comment describing those details would be good afterall.

theohax:
@reviewers -- addressed CR, please re-review, thanks

nano/core_test/node.cpp
```diff
@@ -1682,26 +1682,40 @@ TEST (node, bootstrap_confirm_frontiers)
 }
 
 // Test that if we create a block that isn't confirmed, we sync.
-TEST (node, DISABLED_unconfirmed_send)
+TEST (node, unconfirmed_send)
```
dsiganos, line 1685: What is the intention of this test? What is it trying to test?
It is called unconfirmed send but I do not see anything unconfirmed in the test...
Which block is unconfirmed?

nano/core_test/node.cpp
```diff
@@ -1682,26 +1682,40 @@ TEST (node, bootstrap_confirm_frontiers)
...
+	auto send1 = wallet1->send_action (nano::dev::genesis->account (), key2.pub, 2 * nano::Mxrb_ratio);
+	ASSERT_TIMELY (5s, node1.block_confirmed (send1->hash ()));
+	ASSERT_TIMELY (5s, node2.block_confirmed (send1->hash ()));
+	ASSERT_TIMELY (10s, node2.balance (key2.pub) == 2 * nano::Mxrb_ratio);
+
+	// then send one of the received units back from node2 to node1;
+	//
+	auto const latest = node2.latest (key2.pub);
+	auto send2 = nano::state_block_builder{}.make_block ().account (key2.pub).previous (latest).representative (nano::dev::genesis_key.pub).balance (nano::Mxrb_ratio).link (nano::dev::genesis->account ()).sign (key2.prv, key2.pub).work (*system.work.generate (latest)).build_shared ();
+	node2.process_active (send2);
```
dsiganos: I think that this should not be a process_active. I think this should just insert a block into the ledger bypassing block processing. That is probably what is meant by "unconfirmed send". The original unit test calls `node1.ledger.process`

theohax: Not sure about that to be honest but I addressed it by using `node.process` now.

theohax:
@dsiganos @clemahieu addressed CR guys, pinging you once more to have a look at this, thanks

dsiganos (approved):

theohax:
LGTM, good to merge

PR #3694, opened 2022-01-28
Do not ignore socket errors

dsiganos:
We were not checking the result of the connect, send, recv operations.
We treated failed operations as successful, which
gave at least wrong stats and possibly other minor problems too.

Added counters for tcp connect, read, write errors.

### DIFF
--- a/nano/node/socket.cpp
+++ b/nano/node/socket.cpp
@@ -62,7 +62,10 @@ void nano::socket::async_connect (nano::tcp_endpoint const & endpoint_a, std::fu
 	this_l->tcp_socket.async_connect (endpoint_a,
 	boost::asio::bind_executor (this_l->strand,
 	[this_l, callback = std::move (callback_a), endpoint_a] (boost::system::error_code const & ec) {
-		this_l->stop_timer ();
+		if (!ec)
+		{
+			this_l->stop_timer ();
+		}
 		this_l->remote = endpoint_a;
 		callback (ec);
 	}));
@@ -80,9 +83,12 @@ void nano::socket::async_read (std::shared_ptr<std::vector<uint8_t>> const & buf
 				boost::asio::async_read (this_l->tcp_socket, boost::asio::buffer (buffer_a->data (), size_a),
 				boost::asio::bind_executor (this_l->strand,
 				[this_l, buffer_a, cbk = std::move (callback)] (boost::system::error_code const & ec, std::size_t size_a) {
-					this_l->node.stats.add (nano::stat::type::traffic_tcp, nano::stat::dir::in, size_a);
-					this_l->stop_timer ();
-					this_l->update_last_receive_time ();
+					if (!ec)
+					{
+						this_l->node.stats.add (nano::stat::type::traffic_tcp, nano::stat::dir::in, size_a);
+						this_l->stop_timer ();
+						this_l->update_last_receive_time ();
+					}
 					cbk (ec, size_a);
 				}));
 			}));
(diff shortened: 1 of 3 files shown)


nano/node/socket.cpp
```diff
@@ -109,8 +123,15 @@ void nano::socket::async_write (nano::shared_const_buffer const & buffer_a, std:
...
-					this_l->node.stats.add (nano::stat::type::traffic_tcp, nano::stat::dir::out, size_a);
-					this_l->stop_timer ();
+					if (ec)
+					{
+						this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_write_error, nano::stat::dir::in);
+					}
+					else
+					{
+						this_l->node.stats.add (nano::stat::type::traffic_tcp, nano::stat::dir::out, size_a);
+						this_l->stop_timer ();
```
thsfs, line 133: `last_completion_time_or_init` won't get updated and the socket will have a timeout condition to get closed. It makes sense to have it closed, but the `tcp_io_timeout_drop` stat will be wrongly incremented as the reason for it to get dropped is `tcp_write_error`, already recorded.

nano/node/socket.cpp
```diff
@@ -80,9 +87,16 @@ void nano::socket::async_read (std::shared_ptr<std::vector<uint8_t>> const & buf
...
-					this_l->update_last_receive_time ();
+					if (ec)
+					{
+						this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_read_error, nano::stat::dir::in);
+					}
+					else
+					{
+						this_l->node.stats.add (nano::stat::type::traffic_tcp, nano::stat::dir::in, size_a);
+						this_l->stop_timer ();
+						this_l->update_last_receive_time ();
```
thsfs, line 98: besides `tcp_read_error`, either `tcp_silent_connection_drop` or `tcp_io_timeout_drop` will get recorded as well

nano/node/socket.cpp
```diff
@@ -62,7 +62,14 @@ void nano::socket::async_connect (nano::tcp_endpoint const & endpoint_a, std::fu
 	this_l->tcp_socket.async_connect (endpoint_a,
 	boost::asio::bind_executor (this_l->strand,
 	[this_l, callback = std::move (callback_a), endpoint_a] (boost::system::error_code const & ec) {
-		this_l->stop_timer ();
+		if (ec)
+		{
+			this_l->node.stats.inc (nano::stat::type::tcp, nano::stat::detail::tcp_connect_error, nano::stat::dir::in);
```
thsfs, line 67: `tcp_io_timeout_drop` will be recorded together with `tcp_connect_error` as the error cause.

thsfs (changes requested):
The fix is ok for detecting the socket error and letting the it to get closed, but it seems the stats are getting wrongly registered in the `nano::socket::checkup()` function.

nano/node/socket.cpp
dsiganos: They are different kind of errors, it makes sense to increment both. One says there was a tcp write error, it doesn't mean that the socket closed. The caller then decides to to do nothing in case of tcp write error and let's the socket timeout. So I think it makes sense to record both because they mean slightly different things and they are both correct.

nano/node/socket.cpp
dsiganos: It is not nano::socket's class fault that the owner of that socket doesn't directly close the socket when receiving a connection error and simply let's it timeout. So in this case, both the connect error and the timeout are correct because that is exactly what is happening, we get a connect error and then the socket is left to timeout. It is not optimal but it is also not a problem.

nano/node/socket.cpp
dsiganos: Yes, that is the intention

nano/node/socket.cpp
dsiganos: Have a look at `nano::bootstrap_connections::connect_client`, when it receives a connection error, it does nothing and let's the socket timeout. But that is not nano::socket's fault and it is not a problem to solve now (just a slight inefficiency) that I am focusing on nano::socket.

thsfs (approved):

PR #3700, opened 2022-01-31
Add optional offset parameter for the receivable RPC

### LINKED ISSUE #3699: Add optional offset parameter for the receivable RPC
### Summary

There is currently no way to request receivable blocks in chunks. Getting all receivable blocks of addresses like the famous burn address `nano_1111111111111111111111111111111111111111111111111111hifc8npp` is hell, because the response body is too huge for some system and/or software environments, which leads to either request timeouts or exceeding of the configured max HTTP body size. A new offset parameter would provide an alternative method to request these receivable blocks in chunks.

Adding the possibility to request receivable blocks in chunks would help to avoid request timeouts and/or exceeding of the configured max HTTP body size.

Not that I'm aware of.

Yes, a new offset parameter for the receivable RPC would allow to request receivable blocks in chunks and solve this issue.

I would

### Possible solution

I will create a PR.

### Supporting files

_No response_

Exxenoz:
Closes #3699

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -2947,6 +2947,7 @@ void nano::json_handler::pending ()
 {
 	auto account (account_impl ());
 	auto count (count_optional_impl ());
+	auto offset (offset_optional_impl (0));
 	auto threshold (threshold_optional_impl ());
 	bool const source = request.get<bool> ("source", false);
 	bool const min_version = request.get<bool> ("min_version", false);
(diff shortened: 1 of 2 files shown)


Exxenoz:
@dsiganos I forgot to add some small tweaks to the sorting case. Build and tests should be working now.

clemahieu (approved):

dsiganos:
This looks good but we should really have a unit test for this too.

zhyatt:
> This looks good but we should really have a unit test for this too.

@dsiganos Is the unit test done here https://github.com/nanocurrency/nano-node/pull/3700/files#diff-b053a0c38b909565d0ba60ffdfec20e375bb648c815b8088b9c51e6ed22cc71cR1862 incomplete or need to be broken out into a separate test?

dsiganos:
Doh! I missed that! We are good.

PR #3702, opened 2022-01-31
Add receive_hash option for the blocks_info RPC

### LINKED ISSUE #3701: Add receive_hash option for the blocks_info RPC
### Summary

There is currently no easy/fast way to find the corresponding receive block of a specific send block by using the available RPCs. The only method I could think of is to request ALL blocks of the receiving account and iterate through them, which sounds like something that should be rather done on node side than on client side; due to performance reasons.

This issue could be solved by adding a receive_hash option to the blocks_info RPC. In contrast to the source_account option (which adds the address of the send block to its corresponding receive block), the receive_hash option would add the hash of the receive block to its corresponding send block.

Complicated additional client processing logic to find receive blocks by their send block hash could/should be avoided as well as unnecessary HTTP traffic and resulting delays.

Not that I'm aware of.

Yes.

I would

### Possible solution

I will create a PR.

### Supporting files

_No response_

Exxenoz:
Closes #3701

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -1239,6 +1239,7 @@ void nano::json_handler::blocks_info ()
 {
 	bool const pending = request.get<bool> ("pending", false);
 	bool const receivable = request.get<bool> ("receivable", pending);
+	bool const receive_hash = request.get<bool> ("receive_hash", false);
 	bool const source = request.get<bool> ("source", false);
 	bool const json_block_l = request.get<bool> ("json_block", false);
 	bool const include_not_found = request.get<bool> ("include_not_found", false);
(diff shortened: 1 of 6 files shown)


Exxenoz:
@dsiganos * Fixed clang format

Exxenoz:
Small consistency update: In case no receive block is found the full zero hash is now returned instead of just "0" (similar to the "successor" field).

nano/node/json_handler.cpp
```diff
@@ -1282,15 +1283,55 @@ void nano::json_handler::blocks_info ()
...
 						{
-							exists = node.store.pending.exists (transaction, nano::pending_key (destination, hash));
+							if (pending)
+							{
+								entry.put ("pending", "0");
+							}
+							if (receive_hash)
+							{
+								nano::block_hash block_source;
+								nano::block_hash previous (node.ledger.latest (transaction, destination));
```
dsiganos: We do not want to return the latest block because that could be unconfirmed.
We want to stop returning unconfirmed blocks in general so any new functionality should default to returning confirmed data only by querying the confirmation height processor.

nano/node/json_handler.cpp
```diff
@@ -1282,15 +1283,55 @@ void nano::json_handler::blocks_info ()
...
+								auto previous_block (node.store.block.get (transaction, previous));
+								while (previous_block != nullptr)
+								{
+									block_source = node.ledger.block_source (transaction, *previous_block);
```
dsiganos: Can we call this function in a pruned node?
It seems to assert that the previous block to the block given exists in the ledger.

dsiganos:
@Exxenoz I made some improvements. PLease review, if possible.

Exxenoz:
> @Exxenoz I made some improvements. PLease review, if possible.

Looks good to me. 👍

clemahieu (dismissed):

nano/node/node.cpp
```diff
@@ -1774,6 +1774,54 @@ void nano::node::populate_backlog ()
 	}
 }
 
+/** Given the block hash of a send block, find the associated receive block that receives that send.
```
thsfs: IMO, this function should be packed into the ledger.cpp file, or in another more specific file, if there is a better one. This is more related to the ledger logic and the node is a more general class.

thsfs (changes requested):

nano/node/node.cpp
dsiganos: I agree. But I do not yet understand how the database classes are structured and I don't know how to do that.
So I have taken the easy and slightly less accurate route of implementing in the node class.
We can work on this together to improve it, if you like.

thsfs: My preference is to send the function to `json_handler.cpp` as it isn't used anywhere else, but as you @dsiganos foresee other usages, so let's keep the function in the ledger class, as it is also better for unit test access.

thsfs (dismissed):

dsiganos:
This PR couldn't be merged because it was conflicting.
Merging would be too messy because this Pr has been alive for too long.
So I squashed and rebased and then resolved the conflicts.
Please recheck.

dsiganos (approved):

PR #3703, opened 2022-02-01
Fix unit tests within `active_transactions`

theohax:
Fixed unit tests:
- `active_transactions.keep_local`
- `active_transactions.fork_filter_cleanup`
- `active_transactions.fifo`

No issues to link to this PR (tests weren't disabled before).

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -147,41 +147,47 @@ TEST (active_transactions, keep_local)
 	auto const send5 = wallet.send_action (nano::dev::genesis_key.pub, key5.pub, node.config.receive_minimum.number ());
 	auto const send6 = wallet.send_action (nano::dev::genesis_key.pub, key6.pub, node.config.receive_minimum.number ());
 
-	// force-confirm elections one after another
-	ASSERT_TIMELY (5s, node.active.size () == 1);
+	// force-confirm blocks
 	for (auto const & block : { send1, send2, send3, send4, send5, send6 })
 	{
 		std::shared_ptr<nano::election> election{};
 		ASSERT_TIMELY (5s, (election = node.active.election (block->qualified_root ())) != nullptr);
+		node.process_confirmed (nano::election_status{ block });
 		election->force_confirm ();
+		ASSERT_TIMELY (5s, node.block_confirmed (block->hash ()));
 	}
-	ASSERT_TIMELY (5s, node.active.empty ());
 
-	nano::open_block_builder builder{};
-	auto open1 = builder.make_block ()
-				 .account (key1.pub)
-				 .representative (key1.pub)
-				 .source (send1->hash ())
-				 .sign (key1.prv, key1.pub)
-				 .work (*system.work.generate (key1.pub))
-				 .build_shared ();
-	auto open2 = builder.make_block ()
-				 .account (key2.pub)
-				 .representative (key2.pub)
-				 .source (send2->hash ())
-				 .sign (key2.prv, key2.pub)
-				 .work (*system.work.generate (key2.pub))
-				 .build_shared ();
-	auto open3 = builder.make_block ()
-				 .account (key3.pub)
-				 .representative (key3.pub)
-				 .source (send3->hash ())
-				 .sign (key3.prv, key3.pub)
-				 .work (*system.work.generate (key3.pub))
-				 .build_shared ();
-	node.process_active (open1);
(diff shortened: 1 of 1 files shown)


nano/core_test/active_transactions.cpp
```diff
@@ -120,67 +120,71 @@ TEST (active_transactions, confirm_frontier)
...
-		auto election = node.active.election (block->qualified_root ());
-		ASSERT_NE (nullptr, election);
+		std::shared_ptr<nano::election> election{};
+		ASSERT_TIMELY (5s, (election = node.active.election (block->qualified_root ())) != nullptr);
 		election->force_confirm ();
 	}
 	ASSERT_TIMELY (5s, node.active.empty ());
-	nano::state_block_builder builder;
+
+	nano::open_block_builder builder{};
```
clemahieu: We're trying to move tests to state blocks only. Is there a reason to change this?

nano/core_test/active_transactions.cpp
```diff
@@ -120,67 +120,71 @@ TEST (active_transactions, confirm_frontier)
...
+	// force-confirm elections one after another
 	ASSERT_TIMELY (5s, node.active.size () == 1);
 	for (auto const & block : { send1, send2, send3, send4, send5, send6 })
 	{
-		ASSERT_TIMELY (1s, node.active.election (block->qualified_root ()));
-		auto election = node.active.election (block->qualified_root ());
-		ASSERT_NE (nullptr, election);
+		std::shared_ptr<nano::election> election{};
+		ASSERT_TIMELY (5s, (election = node.active.election (block->qualified_root ())) != nullptr);
 		election->force_confirm ();
```
clemahieu, line 156: It looks like these are setup blocks and we're using node.process_confirmed for that now.

nano/core_test/active_transactions.cpp
```diff
@@ -707,60 +711,68 @@ TEST (active_transactions, republish_winner)
...
+
+	nano::node_config node_config{ nano::get_available_port (), system.logging };
 	node_config.frontiers_confirmation = nano::frontiers_confirmation_mode::disabled;
-	auto & node1 (*system.add_node (node_config));
 
-	nano::keypair key;
-	nano::state_block_builder builder;
+	auto & node1 = *system.add_node (node_config);
+	nano::keypair key{};
+	nano::send_block_builder builder{};
```
clemahieu: This also reverts to legacy blocks.

nano/core_test/active_transactions.cpp
```diff
@@ -1403,71 +1415,74 @@ TEST (active_transactions, vacancy)
...
 	nano::node_config config{ nano::get_available_port (), system.logging };
 	config.active_elections_size = 1;
+
 	auto & node = *system.add_node (config);
-	nano::keypair key0;
-	nano::keypair key1;
-	nano::state_block_builder builder;
+	auto latest_hash = nano::dev::genesis->hash ();
+	nano::keypair key0{};
+	nano::send_block_builder send_builder{};
```
clemahieu: Reverted to legacy blocks.

nano/core_test/active_transactions.cpp
```diff
@@ -707,60 +711,68 @@ TEST (active_transactions, republish_winner)
...
-	std::vector<uint8_t> block_bytes;
+
+	std::vector<uint8_t> send_block_bytes{};
 	{
-		nano::vectorstream stream (block_bytes);
+		nano::vectorstream stream{ send_block_bytes };
 		send1->serialize (stream);
 	}
 
 	// Generate 10 forks to prevent new block insertion to election
```
clemahieu, line 746: Can you determine why we need to prevent block insertion? It isn't immediately apparent how/why this is done.

nano/core_test/active_transactions.cpp
keerifox: perhaps it should be called `legacy_send_block_builder` to prevent others from making that mistake

theohax: Yeah it is a bit unfortunate -- I actually thought we're better off using send_, open_ etc builders since they're more specific, unbeknownst they refer to the legacy block types. Addressed, renaming might be a good idea too, any input on this @clemahieu ?

nano/core_test/active_transactions.cpp
theohax: I can't say either, let's bring it up

thsfs (approved):

PR #3704, opened 2022-02-02
Socket timeout tidyup

dsiganos:
Tidy up of how timeouts are handled in socket class. 

* Use guard ifs to make nano::socket::async_write more readable
* Removed needless code
* Added comments to nano::socket class
* Renamed the nano::socket concept of deadline to timeout  
    A deadline makes people imagine a fixed point in time, a timestamp.
    However, socket::next_deadline is a timeout value and not a deadline.
    It signifies the seconds of inactivity that would cause a socket timeout.
* Remove network constant socket_dev_idle_timeout
* Rename socket::stop_timer to socket:set_last_completion    
    All stop timer does is to set a variable. It is a simply setter function.
    It is confusing to call it stop_timer because it implies that it does
    something more than just set a timestamp.
* Added some socket_timeout unit tests

### DIFF
--- a/nano/core_test/socket.cpp
+++ b/nano/core_test/socket.cpp
@@ -7,6 +7,8 @@
 
 #include <gtest/gtest.h>
 
+#include <boost/asio/read.hpp>
+
 #include <map>
 #include <memory>
 #include <utility>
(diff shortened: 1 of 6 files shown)


thsfs (approved):

thsfs:
It seems there is no major change other than tidying it up. They seem to be ok and the comments will be very helpful. @dsiganos, if you want to remove one more thing I could point the function `socket::set_silent_connection_tolerance_time (std::chrono::seconds tolerance_time_a);`. I added it to facilitate injecting a tolerance time for testing, also because it'd rather read if from the `network_params`, but it ended up unused.

clemahieu (approved):

PR #3710, opened 2022-02-03
Apply the timestamp mask to timestamp field to remove the duration bits

### LINKED ISSUE #3705: Mask off the low 4 bits in timestamp field for vote duration
Although the timestamp field in votes already contains the vote duration in the low 4 bits, the node needs to properly mask this off.

dsiganos:
closes #3705

### DIFF
--- a/nano/secure/common.cpp
+++ b/nano/secure/common.cpp
@@ -519,9 +519,15 @@ std::string nano::vote::to_json () const
 	return stream.str ();
 }
 
+/**
+ * Returns the timestamp of the vote (with the duration bits masked, set to zero)
+ * If it is a final vote, all the bits including duration bits are returned as they are, all FF
+ */
 uint64_t nano::vote::timestamp () const
 {
-	return timestamp_m;
+	return (timestamp_m == std::numeric_limits<uint64_t>::max ())
+	? timestamp_m // final vote
+	: (timestamp_m & timestamp_mask);
 }
 
 uint8_t nano::vote::duration_bits () const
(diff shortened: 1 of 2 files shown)


clemahieu (approved):

nano/secure/common.cpp
```diff
@@ -519,9 +519,15 @@ std::string nano::vote::to_json () const
...
 }
 
+/**
+ * Returns the timestamp of the vote (with the duration bits masked, set to zero)
+ * If it is a final vote, all the bits including duration bits are returned as they are, all FF
+ */
 uint64_t nano::vote::timestamp () const
 {
-	return timestamp_m;
+	return (timestamp_m == std::numeric_limits<uint64_t>::max ())
```
cryptocode, line 528: It's not entirely clear to me what uint64_t's `::max` here signifies, because I see code like:

`auto timestamp = this->is_final ? nano::vote::timestamp_max : nano::milliseconds_since_epoch ();`

where `timestamp_max` is not ::max, but the "masked max" constant `...fff0`

If `::max` really is a sentinel for final votes in some circumstance, maybe putting the `::max` value here into a descriptive constant could be useful?

dsiganos: Yeap, I've thinking along the same lines. I would like to introduce a final_vote sentinel value to compare against instead of comparing against std::numeric_limits<uint64_t>::max () which is not very good at showing the intention.

dsiganos: So that it is obvious and explicit we are checking for a final vote.

dsiganos: You are referring to this code:
```
auto timestamp = this->is_final ? nano::vote::timestamp_max : nano::milliseconds_since_epoch ();
uint8_t duration = this->is_final ? nano::vote::duration_max : /*8192ms*/ 0x9;
votes_l.emplace_back (std::make_shared<nano::vote> (pub_a, prv_a, timestamp, duration, hashes_a));
```
Yes, that is slightly confusing... it is like that because when we construct a vote, we give it the bits represantation and that causes the internal implementation details to leak out of the vote class.
Maybe introduce a final vote constructor? Or final vote builder.

dsiganos: Created issue #3711 to track the final vote improvements.

cryptocode (approved):

PR #3714, opened 2022-02-07
Fix network.construction_with_specified_port (uses test network port)

dsiganos:
The test case network.construction_with_specified_port uses the test
network port as the test port but that means this test fails if there
is a test network node running on that machine.

Convert the test to use nano::get_available_port() to avoid that
problem.

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -56,7 +56,7 @@ TEST (network, tcp_connection)
 TEST (network, construction_with_specified_port)
 {
 	nano::system system{};
-	auto const port = nano::test_node_port ();
+	auto const port = nano::get_available_port ();
 	auto const node = system.add_node (nano::node_config{ port, system.logging });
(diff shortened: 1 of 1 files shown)


theohax:
`nano::get_available_port()` is about to be gotten rid of completely, so perhaps use another static fixed port number?

dsiganos:
Why are we removing nano::get_available_port? It seems very useful.
What will it be replaced by?

clemahieu (approved):

clemahieu:
I don't think we should totally get rid of get_available_port, though most usages should be replaced where we don't need to specify a port; this case would be an exception I think where we do want to specify a port.

theohax:
> I don't think we should totally get rid of get_available_port, though most usages should be replaced where we don't need to specify a port; this case would be an exception I think where we do want to specify a port.

This might just leave room for bad usages of it IMHO, but I can see your point. Well so far it's being used everywhere nonetheless so I guess another one usage won't hurt too much.

PR #3718, opened 2022-02-08
Sort results of the receivable RPC by block hash in case of equal amounts

Exxenoz:
Original issue #3700 

`std::partial_sort` turned out to be nondeterministic in case of equal send block amounts, so I updated the sort condition to sort by block hash if the amount of two blocks is equal. I also added more detailled test cases for the offset parameter. @dsiganos

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -3029,9 +3029,7 @@ void nano::json_handler::pending ()
 			{
 				auto mid = hash_ptree_pairs.size () <= (offset + count) ? hash_ptree_pairs.end () : hash_ptree_pairs.begin () + offset + count;
 				std::partial_sort (hash_ptree_pairs.begin (), mid, hash_ptree_pairs.end (), [] (auto const & lhs, auto const & rhs) {
-					auto amount_l = lhs.second.template get<nano::uint128_t> ("amount");
-					auto amount_r = rhs.second.template get<nano::uint128_t> ("amount");
-					return amount_l == amount_r ? lhs.first.compare (rhs.first) < 0 : amount_l > amount_r;
+					return lhs.second.template get<nano::uint128_t> ("amount") > rhs.second.template get<nano::uint128_t> ("amount");
 				});
 				for (auto i = offset, j = offset + count; i < hash_ptree_pairs.size () && i < j; ++i)
 				{
@@ -3042,9 +3040,7 @@ void nano::json_handler::pending ()
 			{
 				auto mid = hash_amount_pairs.size () <= (offset + count) ? hash_amount_pairs.end () : hash_amount_pairs.begin () + offset + count;
 				std::partial_sort (hash_amount_pairs.begin (), mid, hash_amount_pairs.end (), [] (auto const & lhs, auto const & rhs) {
-					auto amount_l = lhs.second;
-					auto amount_r = rhs.second;
-					return amount_l == amount_r ? lhs.first.compare (rhs.first) < 0 : amount_l > amount_r;
+					return lhs.second > rhs.second;
 				});
 
 				for (auto i = offset, j = offset + count; i < hash_amount_pairs.size () && i < j; ++i)
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -3029,7 +3029,9 @@ void nano::json_handler::pending ()
 			{
 				auto mid = hash_ptree_pairs.size () <= (offset + count) ? hash_ptree_pairs.end () : hash_ptree_pairs.begin () + offset + count;
 				std::partial_sort (hash_ptree_pairs.begin (), mid, hash_ptree_pairs.end (), [] (auto const & lhs, auto const & rhs) {
-					return lhs.second.template get<nano::uint128_t> ("amount") > rhs.second.template get<nano::uint128_t> ("amount");
+					auto amount_l = lhs.second.template get<nano::uint128_t> ("amount");
+					auto amount_r = rhs.second.template get<nano::uint128_t> ("amount");
+					return amount_l == amount_r ? lhs.first.compare (rhs.first) < 0 : amount_l > amount_r;
 				});
 				for (auto i = offset, j = offset + count; i < hash_ptree_pairs.size () && i < j; ++i)
 				{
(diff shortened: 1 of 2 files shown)


nano/rpc_test/rpc.cpp
```diff
@@ -1858,16 +1858,148 @@ TEST (rpc, pending)
 		ASSERT_EQ (block4->hash (), hash);
 		ASSERT_EQ (block3->hash (), hash1);
 	}
+}
+
+TEST (rpc, receivable_offset)
```
dsiganos: It would be nice to have a comment explaining what this unit test intends to test.

nano/rpc_test/rpc.cpp
```diff
@@ -1858,16 +1858,148 @@ TEST (rpc, pending)
...
+	ASSERT_TIMELY (5s, !node->active.active (*block6));
+	ASSERT_TIMELY (5s, node->block_confirmed (block6->hash ()));
+	ASSERT_TIMELY (5s, node->ledger.cache.cemented_count == 7 && node->confirmation_height_processor.current ().is_zero () && node->confirmation_height_processor.awaiting_processing_size () == 0);
+
+	auto const rpc_ctx = add_rpc (system, node);
+	boost::property_tree::ptree request;
+	request.put ("action", "receivable");
+	request.put ("account", key1.pub.to_account ());
+
+	// Default: Sorted by hash
```
dsiganos: The documentation does not say that the results of the "receivable" rpc commands are sorted by hash.
The fact that the current implementation appears to return them sorted by hash (this is probably because they are sorted in the database) does not mean that this a feature and therefore we should not check that results are sorted by hash when sorting is false.

dsiganos:
Related docs issue:
https://github.com/nanocurrency/nano-docs/issues/622

nano/node/json_handler.cpp
```diff
@@ -3029,7 +3029,9 @@ void nano::json_handler::pending ()
 			{
 				auto mid = hash_ptree_pairs.size () <= (offset + count) ? hash_ptree_pairs.end () : hash_ptree_pairs.begin () + offset + count;
 				std::partial_sort (hash_ptree_pairs.begin (), mid, hash_ptree_pairs.end (), [] (auto const & lhs, auto const & rhs) {
-					return lhs.second.template get<nano::uint128_t> ("amount") > rhs.second.template get<nano::uint128_t> ("amount");
+					auto amount_l = lhs.second.template get<nano::uint128_t> ("amount");
```
dsiganos: This change seems to sort the blocks first by amount and then by hash.
However, the documentation does not mention anything about sorting by hash.
And I do not see much value in sorting by hash, we are just making our life more difficult for no obvious value gain.
So, I think we should not make this change unless it fixes something else that I have missed.

nano/node/json_handler.cpp
```diff
@@ -3040,7 +3042,9 @@ void nano::json_handler::pending ()
 			{
 				auto mid = hash_amount_pairs.size () <= (offset + count) ? hash_amount_pairs.end () : hash_amount_pairs.begin () + offset + count;
 				std::partial_sort (hash_amount_pairs.begin (), mid, hash_amount_pairs.end (), [] (auto const & lhs, auto const & rhs) {
-					return lhs.second > rhs.second;
+					auto amount_l = lhs.second;
```
dsiganos: This change seems to sort the blocks first by amount and then by hash.
However, the documentation does not mention anything about sorting by hash.
And I do not see much value in sorting by hash, we are just making our life more difficult for no obvious value gain.
So, I think we should not make this change unless it fixes something else that I have missed.

dsiganos:
I suggest we do the following:
* not sort by hash because there is not much value in doing it
* keep the new test case written by @Exxenoz after fixing it to expect sorting by hash
* fix the documentation of receivable as explained

@zhyatt , @clemahieu , do you agree?

zhyatt:
> do you agree?

From the original description it sounds like if one were to use the `offset` to paginate the receivable responses and the returned blocks had equal amounts in them that bridged across the pagination threshold, you could potentially get the same block on two different offsets and miss a block. If that is the case then it seems sorting by hash would prevent that. Am I reading that correcly @dsiganos ?

dsiganos:
So it seems that we have 2 choices:
1) Use stable_sort instead of the faster partial_sort. stable_sort preserves the order of equal elements and then the problem you describe @zhyatt goes away.
2) Sort the equal elements by hash but then that means that we are saying that elements are always ordered by hash when sorting is not specified, which possibly wasn't intended.
@clemahieu?

Exxenoz:
> > do you agree?
> 
> From the original description it sounds like if one were to use the `offset` to paginate the receivable responses and the returned blocks had equal amounts in them that bridged across the pagination threshold, you could potentially get the same block on two different offsets and miss a block. If that is the case then it seems sorting by hash would prevent that. Am I reading that correcly @dsiganos ?

That's exactly the problem. 👍

clemahieu:
We should be able to pass std::partial_sort a different comparison operator depending on if ordered is set. Overload #3 https://en.cppreference.com/w/cpp/algorithm/partial_sort

nano/rpc_test/rpc.cpp
```diff
@@ -1858,16 +1858,148 @@ TEST (rpc, pending)
...
 	}
+}
+
+TEST (rpc, receivable_offset)
+{
+	nano::system system;
+	auto node = add_ipc_enabled_node (system);
+	nano::keypair key1;
+	system.wallet (0)->insert_adhoc (nano::dev::genesis_key.prv);
+	auto block1 (system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, 200));
```
clemahieu: We're moving to assignment syntax since it's more common:
auto block1 = system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, 200);

nano/rpc_test/rpc.cpp
```diff
@@ -1858,16 +1858,148 @@ TEST (rpc, pending)
...
+	auto block2 (system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, 100));
+	auto block3 (system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, 400));
+	auto block4 (system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, 300));
+	auto block5 (system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, 300));
+	auto block6 (system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, 300));
+	node->scheduler.flush ();
```
clemahieu: This line is unnecessary since you're waiting for all of them to be in the ledger on the next line. We're trying to remove reliance on flush() functions in unit tests as they cause race conditions when the implementation changes.

clemahieu:
I think the stable_sort sounds like the better method. This way we don't need to strengthen the specification on the API.

Then, the only way a stable_sort could be messed up for pagination purposes would be if the container iterated items in a different order across successive calls. But we've already committed to that stability in the RPC API when the offset field was added.

dsiganos (approved):

nano/rpc_test/rpc.cpp
dsiganos: done

clemahieu (approved):

PR #3719, opened 2022-02-08
Removes the nano::unchecked_info::confirmed member. 

clemahieu:
This was not serialized to the unchecked table. Additionally, the value was assigned from a comparison of retry counts so the naming of "confirmed" was confusing. Finally, its removal doesn't break any tests so it's not a tested branch of code.


theohax (approved):

theohax:
```
[ RUN      ] active_transactions.fork_replacement_tally
/Users/runner/work/nano-node/nano-node/nano/core_test/active_transactions.cpp:849: Failure
  max_blocks
    Which is: 10
  election->blocks ().size ()
    Which is: 1
```

Already on the to-fix list, so CI looks green-ish otherwise

dsiganos (approved):

PR #3723, opened 2022-02-08
Fix unchecked_map destructor.

clemahieu:
Fix an issue with unchecked_map::~unchecked_map where the condition variable signal can be missed since the object mutex isn't acquired while setting stopped=true.

### DIFF
--- a/nano/node/unchecked_map.cpp
+++ b/nano/node/unchecked_map.cpp
@@ -63,8 +63,10 @@ size_t nano::unchecked_map::count (nano::transaction const & transaction) const
 
 void nano::unchecked_map::stop ()
 {
-	if (!stopped.exchange (true))
+	nano::unique_lock<nano::mutex> lock{ mutex };
+	if (!stopped)
 	{
+		stopped = true;
 		condition.notify_all (); // Notify flush (), run ()
 	}
 }
--- a/nano/node/unchecked_map.hpp
+++ b/nano/node/unchecked_map.hpp
@@ -56,7 +56,7 @@ private:
 	std::deque<boost::variant<insert, query>> buffer;
 	std::deque<boost::variant<insert, query>> back_buffer;
 	bool writing_back_buffer{ false };
-	std::atomic<bool> stopped{ false };
+	bool stopped{ false };
 	nano::condition_variable condition;
 	nano::mutex mutex;
 	std::thread thread;


theohax:
In practice I doubt this can actually bite us, but yeah, the proper way of announcing changes via CVs is under the mutex indeed, even if the change is atomic like this one. Just curious, have you seen this cause issues?

theohax (dismissed):

nano/node/unchecked_map.cpp
```diff
@@ -63,6 +63,7 @@ size_t nano::unchecked_map::count (nano::transaction const & transaction) const
 
 void nano::unchecked_map::stop ()
 {
+	nano::unique_lock<nano::mutex> lock{ mutex };
```
dsiganos, line 66: If we are protecting `stopped` with a mutex then why do we also make it atomic?
I do not think `stopped` needs to be atomic.

dsiganos:
@theohax In practice, it would bite us on node shutdown, the node could deadlock on shutdown.

thsfs:
> In practice I doubt this can actually bite us, but yeah, the proper way of announcing changes via CVs is under the mutex indeed, even if the change is atomic like this one. Just curious, have you seen this cause issues?



> In practice I doubt this can actually bite us, but yeah, the proper way of announcing changes via CVs is under the mutex indeed, even if the change is atomic like this one. Just curious, have you seen this cause issues?

I also can't see where this bit us.

nano/node/unchecked_map.cpp
clemahieu: We don't really.

clemahieu:
If the working thread is between these two lines when stopped is set to true and the condition variable in notified, it will neither detect the stopped flag nor detect the condition variable notification.

```
	while (!stopped)
	{
		if (!buffer.empty ())
```

dsiganos (dismissed):

thsfs (dismissed):

thsfs (approved):

nano/node/unchecked_map.cpp
```diff
@@ -63,8 +63,10 @@ size_t nano::unchecked_map::count (nano::transaction const & transaction) const
 
 void nano::unchecked_map::stop ()
 {
-	if (!stopped.exchange (true))
+	nano::unique_lock<nano::mutex> lock{ mutex };
+	if (!stopped)
 	{
+		stopped = true;
 		condition.notify_all (); // Notify flush (), run ()
```
thsfs, line 70: shouldn't `condition.notify_all` be moved out from the if guard? So it would always be called.

dsiganos: Once it is stopped, it is stopped, there should be no need to notify anything...

PR #3726, opened 2022-02-08
Updates the doxygen.config

thsfs:
Adding a Doxyfile that will facilitate generating doc info from the node's source-code. The Doxygen docs may look a bit overwhelming, as it tries to doc all the smaller pieces of code, but it may also be handy to have some HTML files or SVG diagrams through which it can be easily navigated.

Takes advantage of: https://github.com/nanocurrency/nano-node/issues/3724


dsiganos (dismissed):

cryptocode (commented):
There's already a `doxygen.config` in the root directory. I added that a long time ago, so it may need an update or get removed by this PR.

thsfs:
> There's already a `doxygen.config` in the root directory. I added that a long time ago, so it may need an update or get removed by this PR.

Nice, didn't realize it before. Now just merged the two files into yours. Excluded some third-party and testing code, so this will only generate docs for the core classes. Also added a main page with a brief explanation of what is generated.

cryptocode (approved):

zhyatt (approved):

PR #3727, opened 2022-02-09
Block pipeline

clemahieu:
This is a refactor of the block processor which breaks the block processing steps in to small components focused on a specific part of processing blocks.


nano/core_test/early_verify_filter.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include <nano/lib/blockbuilders.hpp>
+#include <nano/lib/stats.hpp>
+#include <nano/secure/common.hpp>
+
+#include <gtest/gtest.h>
+
+namespace
+{
+std::shared_ptr<nano::block> early_block ()
```
dsiganos: Can you add small comment explaining the concept "early block"?

nano/node/block_pipeline/early_verify_filter.hpp
```diff
@@ -0,0 +1,30 @@
...
+{
+/**
+ *  This class filters blocks by when their signatures can be verified.
+ *  Blocks can be early which means they can be verified without a ledger access, or they can be late which means the ledger must be accessed to determine the signer
+ */
+class early_verify_filter
+{
+public:
+	early_verify_filter (nano::epochs & epochs, nano::stat & stats);
+	void sink (std::shared_ptr<nano::block> block) const;
```
dsiganos: Is sink the input of the filter? Do you call sink to provide blocks to the filter?

nano/lib/stats.hpp
```diff
@@ -387,7 +387,11 @@ class stat final
 		generator_broadcasts,
 		generator_replies,
 		generator_replies_discarded,
-		generator_spacing
+		generator_spacing,
+		
+		// Block pipeline details
+		early_verify_filter_early,
```
dsiganos: Should these be primary counters? If we add such detail at this level, this list will get very very long.

clemahieu: Moving them to primary counters doesn't seem like it would reduce the count. We could do pass/reject as in/out for direction which would cut these in half. I think the nice part about having them under the same primary counter is that it could be filtered by that counter to get just the block pipeline stats.

nano/node/block_pipeline/early_verify_filter.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <memory>
+
+namespace nano
+{
+class block;
+class epochs;
+class stat;
+namespace block_pipeline
+{
+/**
```
dsiganos: Is this better maybe?

```
This class filters blocks by when their signatures can be verified. Legacy and epoch blocks do not contain all the info required to have their signature checked. One problem is finding the public key to verify the signature, in the case of non-open legacy blocks, the account cannot be easily found without analysing and connecting previous blocks. Another problem is finding the balance before the block being verified.
Blocks can be early which means they can be verified without a ledger access, or they can be late which means the ledger must be accessed to determine the signer or type of block or both.

File: nano/node/blockprocessor.cpp [hunk 5]
```diff
@@ -180,34 +179,25 @@ bool nano::block_processor::have_blocks ()
 	return have_blocks_ready () || state_block_signature_verification.size () != 0;
 }
 
-void nano::block_processor::process_verified_state_blocks (std::deque<nano::unchecked_info> & items, std::vector<int> const & verifications, std::vector<nano::block_hash> const & hashes, std::vector<nano::signature> const & blocks_signatures)
```
dsiganos: This function seems unused.

nano/node/blockprocessor.cpp
```diff
@@ -26,13 +26,60 @@ nano::block_post_events::~block_post_events ()
...
+		node.stats.inc (nano::stat::type::block_pipeline, nano::stat::detail::block_position_filter_reject);
+	};
+	determinate.pass = [this] (std::shared_ptr<nano::block> block_a, nano::account const & account) {
+		node.stats.inc (nano::stat::type::block_pipeline, nano::stat::detail::determinate_signer_filter_pass);
+		state_block_signature_verification.add ({ block_a, account, nano::signature_verification::unknown });
+	};
+	determinate.reject = [this] (std::shared_ptr<nano::block> block_a) {
+		node.stats.inc (nano::stat::type::block_pipeline, nano::stat::detail::determinate_signer_filter_reject);
+		node.workers.push_task ([node = node.shared (), block_a] () {
+			node->unchecked.put (block_a->previous (), block_a);
```
dsiganos: Why is this is done by background task when a very similar operation in `gap_previous.reject` is done in the foreground?

dsiganos: What if it is a receive state block? Then previous will be 0 and the real dependency is the associated send block.

nano/node/blockprocessor.cpp
```diff
@@ -26,13 +26,60 @@ nano::block_post_events::~block_post_events ()
...
+		gap_previous.sink (block_a);
+	};
+	gap_previous.pass = [this] (std::shared_ptr<nano::block> block_a, std::shared_ptr<nano::block> previous) {
+		node.stats.inc (nano::stat::type::block_pipeline, nano::stat::detail::gap_previous_filter_pass);
+		position.sink (block_a, previous);
+	};
+	gap_previous.reject = [this] (std::shared_ptr<nano::block> block_a) {
+		node.stats.inc (nano::stat::type::block_pipeline, nano::stat::detail::gap_previous_filter_reject);
+		node.gap_cache.add (block_a->hash ());
+		node.unchecked.put (block_a->previous (), block_a);
```
dsiganos: What if it is a receive state block? Then previous will be 0 and the real dependency is the associated send block.

nano/node/blockprocessor.cpp
```diff
@@ -187,34 +216,25 @@ bool nano::block_processor::have_blocks ()
 	return have_blocks_ready () || state_block_signature_verification.size () != 0;
 }
 
-void nano::block_processor::process_verified_state_blocks (std::deque<nano::unchecked_info> & items, std::vector<int> const & verifications, std::vector<nano::block_hash> const & hashes, std::vector<nano::signature> const & blocks_signatures)
+void nano::block_processor::process_verified_state_blocks (std::deque<nano::state_block_signature_verification::value_type> & items, std::vector<int> const & verifications, std::vector<nano::block_hash> const & hashes, std::vector<nano::signature> const & blocks_signatures)
```
dsiganos: This function is not used. If it is not needed, maybe delete it?

nano/node/block_pipeline/stateless_signer_filter.hpp
```diff
@@ -0,0 +1,29 @@
...
+{
+class block;
+class epochs;
+class public_key;
+class stat;
+using account = public_key;
+namespace block_pipeline
+{
+/**
+ *  This class filters blocks whos signer can be determined statelessly based on block contents
```
dsiganos: Some more info:
Legacy and epoch blocks do not contain all the info required to have their signature checked. One problem is finding the public key to verify the signature. In the case of non-open legacy blocks, the account cannot be easily found without analysing previous blocks. Another problem is finding the balance before the block being verified.

nano/node/block_pipeline/stateless_signer_filter.cpp
```diff
@@ -0,0 +1,45 @@
...
+		case nano::block_type::state:
+			// If it's not an epoch link, it's signed by the value in the account field
+			if (!epochs.is_epoch_link (block->link ()))
+			{
+				pass (block, block->account ());
+			}
+			else
+			{
+				// If there is no previous block, the balance must already be 0 and therefore cannot be a send
+				if (block->previous ().is_zero ())
```
dsiganos: I do not know if this check is complete or makes sense.
At this point, the block could be even a receive, since at this point we wouldn't know that a receive doesn't have a corresponding send block.
It coudl also be a change block. I am not sure what this check gives us.

nano/node/block_pipeline/block_position_filter.cpp
```diff
@@ -0,0 +1,31 @@
+#include <nano/node/block_pipeline/block_position_filter.hpp>
+#include <nano/secure/ledger.hpp>
+#include <nano/secure/store.hpp>
+
+void nano::block_pipeline::block_position_filter::sink (std::shared_ptr<nano::block> block, std::shared_ptr<nano::block> previous)
```
dsiganos: This filter mentions that it checks something to do with epoch going backwards.
I do not understand what the means.
However, I see no reference to any epoch blocks, is that a sign that something is missing?
Also I see no reference to work difficulties and epoch v2 is about work difficulty.

This simply seems to check that legacy blocks never follows a state block.

nano/lib/stats.cpp
```diff
@@ -889,10 +888,40 @@ std::string nano::stat::detail_to_string (uint32_t key)
...
+			break;
+		case nano::stat::detail::stateless_signer_filter_reject:
+			res = "stateless_signer_filter_reject";
+			break;
+		case nano::stat::detail::determinate_signer_filter_pass:
+			res = "determinate_signer_filter_pass";
+			break;
+		case nano::stat::detail::determinate_signer_filter_reject:
+			res = "determinate_signer_filter_reject";
+			break;
```
dsiganos: We could add a default case with an debug_assert(0) to catch any missed cases.

nano/node/block_pipeline/determinate_signer_filter.cpp
```diff
@@ -0,0 +1,52 @@
...
+void nano::block_pipeline::determinate_signer_filter::sink (std::shared_ptr<nano::block> block, std::shared_ptr<nano::block> previous) const
+{
+	assert (block != nullptr);
+	nano::account account;
+	switch (block->type ())
+	{
+		case nano::block_type::send:
+		case nano::block_type::receive:
+		case nano::block_type::change:
+			switch (previous->type ())
```
dsiganos, line 25: We assume previous cannot be null here, is not risky to make such an assumption here?
What if it is a legacy receive block opening an account?

nano/node/block_pipeline/determinate_signer_filter.cpp
```diff
@@ -0,0 +1,52 @@
...
+			break;
+		}
+		case nano::block_type::invalid:
+		case nano::block_type::not_a_block:
+		case nano::block_type::open: // Open block is early-verify
+			debug_assert (false);
+			break;
+	}
+	debug_assert (!account.is_zero ());
+	pass (block, account);
```
dsiganos: This filter never calls reject!

nano/lib/stats.cpp
clemahieu: If we turned on warning to error conversion for missing case statements the CI could catch it at compile time.

nano/node/block_pipeline/determinate_signer_filter.cpp
clemahieu: Legacy receive blocks can't open an account, only legacy open blocks can. I added a debug_assert and several comments to clarify.

nano/node/block_pipeline/determinate_signer_filter.cpp
clemahieu: True, this is an artifact of a refactor I did. I'll have to redesign this a bit.

nano/node/blockprocessor.cpp
clemahieu: This filter passes if previous is zero. It fails iff previous is not zero and the previous block does not exist in the ledger.

PR #3728, opened 2022-02-10
Remove coupling of block processor and unchecked_map on unchecked_info::modified.

clemahieu:


nano/node/blockprocessor.cpp
```diff
@@ -376,10 +376,6 @@ nano::process_return nano::block_processor::process_one (nano::write_transaction
 				node.logger.try_log (boost::str (boost::format ("Gap previous for: %1%") % hash.to_string ()));
 			}
 			info_a.verified = result.verified;
-			if (info_a.modified == 0)
```
thsfs, line 379: I think there should be debug_asserts so `(info_a.modified != 0)`. Otherwise this may get uncovered and it is prone to insert a bug.

thsfs (changes requested):

nano/node/blockprocessor.cpp
clemahieu: This is moving the assignment of this field to right before it's inserted. Maybe "modified" should be made private and only unchecked_map can access it.

thsfs: It won't guarantee, the `modified` is already initialized as `0`. Not getting updated by a malformed construction on unchecked_map on any further change could lead to a missed case. I still think an assert wouldn't hurt in this case.

theohax: +1 for a private `modified` member.

Can't really tell if the debug assert makes sense though, but @thsfs might have a point. Maybe we can keep the check for `release` and add the `debug` assert to catch eventual issues in debug builds?

theohax (dismissed):
Looks good, but the open discussion is worth taking into account for another addressing commit.

thsfs (dismissed):

nano/node/blockprocessor.cpp
```diff
@@ -99,7 +99,7 @@ bool nano::block_processor::half_full ()
 
 void nano::block_processor::add (std::shared_ptr<nano::block> const & block_a, uint64_t origination)
```
thsfs: it seems origination won't be needed anymore.

thsfs (approved):

PR #3729, opened 2022-02-10
Remove coupling between nano::state_block_signature_verification and nano::unchecked_info.

clemahieu:


nano/node/blockprocessor.cpp
```diff
@@ -31,7 +31,7 @@ nano::block_processor::block_processor (nano::node & node_a, nano::write_databas
 	write_database_queue (write_database_queue_a),
 	state_block_signature_verification (node.checker, node.ledger.constants.epochs, node.config, node.logger, node.flags.block_processor_verification_size)
 {
-	state_block_signature_verification.blocks_verified_callback = [this] (std::deque<nano::unchecked_info> & items, std::vector<int> const & verifications, std::vector<nano::block_hash> const & hashes, std::vector<nano::signature> const & blocks_signatures) {
+	state_block_signature_verification.blocks_verified_callback = [this] (std::deque<nano::state_block_signature_verification::value_type> & items, std::vector<int> const & verifications, std::vector<nano::block_hash> const & hashes, std::vector<nano::signature> const & blocks_signatures) {
```
thsfs, line 34: ~~`nano::unchecked_info` is a core class, from my perspective, I'd say it seems more tied to the node's domain than the `state_block_signature_verification` class, so I think that using it to pack data is a good practice. The point about uncoupling it from the this last class would make sense if the signature verification is going to be used on different data types.~~

nano/node/blockprocessor.cpp
```diff
@@ -392,10 +388,6 @@ nano::process_return nano::block_processor::process_one (nano::write_transaction
 				node.logger.try_log (boost::str (boost::format ("Gap source for: %1%") % hash.to_string ()));
 			}
 			info_a.verified = result.verified;
-			if (info_a.modified == 0)
```
thsfs: Same as stated on the another PR (https://github.com/nanocurrency/nano-node/pull/3728). It seems we need an assert here so we ensure the `modified` field won't pass without being set.

thsfs (changes requested):

nano/node/blockprocessor.cpp
thsfs: Fixed in the other PR.

nano/node/blockprocessor.cpp
thsfs: Changed my mind, perhaps just packing the tuples as a context class `{ block, account, verified }` if it makes sense.

thsfs (approved):

PR #3735, opened 2022-02-12
Fix unit test rpc.confirmation_info (remove scheduler flush)

dsiganos:
Removed the need for scheduler flush, which is broken.

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -6246,7 +6246,7 @@ TEST (rpc, confirmation_info)
 
 	auto send (std::make_shared<nano::send_block> (nano::dev::genesis->hash (), nano::public_key (), nano::dev::constants.genesis_amount - 100, nano::dev::genesis_key.prv, nano::dev::genesis_key.pub, *system.work.generate (nano::dev::genesis->hash ())));
 	node1->process_active (send);
-	ASSERT_TIMELY (10s, node1->active.empty () == false);
+	ASSERT_TIMELY (5s, !node1->active.empty ());
 
 	boost::property_tree::ptree request;
 	request.put ("action", "confirmation_info");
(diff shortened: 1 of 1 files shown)


thsfs (dismissed):

nano/rpc_test/rpc.cpp
```diff
@@ -6246,9 +6246,7 @@ TEST (rpc, confirmation_info)
 
 	auto send (std::make_shared<nano::send_block> (nano::dev::genesis->hash (), nano::public_key (), nano::dev::constants.genesis_amount - 100, nano::dev::genesis_key.prv, nano::dev::genesis_key.pub, *system.work.generate (nano::dev::genesis->hash ())));
 	node1->process_active (send);
-	node1->block_processor.flush ();
-	node1->scheduler.flush ();
-	ASSERT_FALSE (node1->active.empty ());
+	ASSERT_TIMELY (10s, node1->active.empty () == false);
```
theohax: We've been using `5s` as a timeout for changes to AEC that we expect to see. Not that it matters in its absolute value (the timeout), but since we are kind of reworking these things now, unless there's a reason not to stick to the same value, it's probably better to keep it consistent at least in unit tests that we change if not in old ones as well.

Also, an even less important nitpick, but just thinking out loud in case you decide to do any addressing (otherwise please don't bother changing just this) -- I haven't seen `boolVariable == false` or `boolVariable == true` usages across the codebase, so again for consistency reasons we might be better off with the existing `if (boolVariable)` / `if (!boolVariable)` style.

theohax:
Fix itself looks good, couple nitpicks from my side.

Other than that, looking at the CI failure and seeing something that I haven't before (not a "usual" test failure):

```
2022-02-12T18:34:12.9954965Z /workspace/nano/core_test/block_store.cpp:468: Failure
2022-02-12T18:34:12.9955382Z Expected equality of these values:
2022-02-12T18:34:12.9955635Z   unchecked1_blocks.size ()
2022-02-12T18:34:12.9955875Z     Which is: 0
2022-02-12T18:34:12.9956080Z   3
2022-02-12T18:34:13.0750638Z [  FAILED  ] unchecked.multiple_get (499 ms)
```

@clemahieu / @thsfs -- is this something we already know about?

nano/rpc_test/rpc.cpp
dsiganos: Personally I find 5 seconds to be a little too low for comfort. But I am OK to follow that pattern if that is what you guys decided. It should probably be OK.

If find this to be more easily readable
```
ASSERT_TIMELY (10s, node1->active.empty () == false);
```
than this:
```
ASSERT_TIMELY (10s, !node1->active.empty ());
```

The ! character is too easy miss in the middle of a statement.
Although I have no objection to using `!` instead of `== false`.

clemahieu (approved):

PR #3738, opened 2022-02-15
Fix for unit test unchecked.multiple_get, do not use store.count()

### LINKED ISSUE #3737: unchecked.multiple_get failed in CI
[ RUN      ] unchecked.multiple_get
/workspace/nano/core_test/block_store.cpp:468: Failure
  unchecked1_blocks.size ()
    Which is: 0
  3
[  FAILED  ] unchecked.multiple_get (499 ms)

https://github.com/nanocurrency/nano-node/runs/5168846382?check_suite_focus=true#step:6:316

Linux test with ROCKSDB

dsiganos:
When using rocksdb as the backend database, the count() method of the
store cannot be trusted, it does not give accurate results.

This problem is fixed by counting the blocks in the unchecked table
one by one, by iterating through all of them, which should work for all
db backends.

resolves #3737

### DIFF
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -459,8 +459,22 @@ TEST (unchecked, multiple_get)
 	unchecked.put (block3->previous (), block3);
 	unchecked.put (block3->hash (), block3); // unchecked4
 	unchecked.put (block1->previous (), block3); // unchecked1
+
+	// count the number of blocks in the unchecked table by counting them one by one
+	// we cannot trust the count() method if the backend is rocksdb
+	auto count_unchecked_blocks_one_by_one = [&store, &unchecked] () {
+		size_t count = 0;
+		auto transaction = store->tx_begin_read ();
+		for (auto [i, end] = unchecked.full_range (transaction); i != end; ++i)
+		{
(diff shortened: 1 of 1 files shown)


theohax:
This looks good, but is this something we need to consider in a general case or can we get around it by this change only, done in this unit test only?

theohax (approved):

theohax:
Additionally, but not related to this PR itself:

```
2022-02-15T01:30:27.5395493Z [  FAILED  ] 1 test, listed below:
2022-02-15T01:30:27.5395811Z [  FAILED  ] telemetry.remove_peer_different_genesis_udp
2022-02-15T01:30:27.5396018Z 
2022-02-15T01:30:27.5396096Z  1 FAILED TEST
```

This failure isn't in my list and I don't remember seeing it before. Also, from the sound of it it is clearly not related to scheduler flushes. An issue for it would be good to have as well as a temporary disablement.

dsiganos:
We possibly have more bugs like this, after a quick look, there too many candidates to consider.
So for now, I will just focus on failing test cases only.

It is less likely that this bug exists in the nano node because this issue was well understood by the previous developers.

PR #3741, opened 2022-02-15
Fix confirmation_height.election_winner_details_clearing

dsiganos:
There is a race condition between a block getting confirmed and setting
of statistics counters. In this case, the counter is incremented in an
observer callback that is executed after the block is confirmed.

One could say that the problem is that the stat counters should updated
before the block is confirmed but I think that is unrealistic to achieve
and possibly not even the right approach.

Converting the check to an ASSERT_TIMELY fixes the problem.
It seems likely that other test cases will have this problem too.


theohax (approved):

theohax:
`quorum_minimum_update_weight_before_quorum_checks` failed in the CI but it looks like a scheduler flush thing, I'll have a look at it -- I see it's not yet on the list

PR #3746, opened 2022-03-02
Introduce `ledger.rollback_failed` stat counter

theohax:
Also, use that counter in `confirmation_height.conflict_rollback_cemented` unit test.

Based off of #3745


clemahieu:
This is missing the string name for the enumeration:

/Users/clemahieu/nano-node.src/nano/lib/stats.cpp:544:10: warning: enumeration value 'rollback_failed' not handled in switch [-Wswitch]
        switch (detail)
                ^
/Users/clemahieu/nano-node.src/nano/lib/stats.cpp:544:10: note: add missing switch cases
        switch (detail)

clemahieu (approved):

PR #3749, opened 2022-03-10
Fix Issue #3748

### LINKED ISSUE #3748: Nano's CLI --help should display options in alphabetic order
### Summary

Explained in the title

The current order has poor usability making it harder to find what one is looking for. 

I believe not.

Sort the options in alphabetic order.

I would not

### Possible solution

_No response_

### Supporting files

_No response_

Justin-Randall:
Sorted command line options output

### DIFF
--- a/nano/lib/CMakeLists.txt
+++ b/nano/lib/CMakeLists.txt
@@ -90,7 +90,8 @@ target_link_libraries(
   ipc_flatbuffers_lib
   ${CRYPTOPP_LIBRARY}
   ${CMAKE_DL_LIBS}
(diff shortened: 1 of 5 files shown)


clemahieu:
Looks like this is failing because lib utility doesn’t link to boost program options. I tried moving it to cli.hpp/cli.cpp though nano_rpc doesn’t link to the node library.

Justin-Randall:
I updated the CMakeLists.txt to reflect the dependency.

clemahieu (approved):

Issue #3709, opened 2022-02-03
slow_test node.fork_storm:telemetry.under_load fails every time


dsiganos:
Node version: v23.0

Note: incrase the file limit before running the test: `ulimit -n 10000`

```
ds@tux:~/nano/releases/V23.0/build$ ulimit -n 10000
ds@tux:~/nano/releases/V23.0/build$ ./slow_test --gtest_filter=node.fork_storm:telemetry.under_load
Note: Google Test filter = node.fork_storm:telemetry.under_load
[==========] Running 2 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 1 test from node
[ RUN      ] node.fork_storm
Assertion (iterations1 < 10000) failed
std::shared_ptr<nano::node> nano::system::add_node(const nano::node_config&, nano::node_flags, nano::transport::transport_type)
/home/ds/nano/releases/V23.0/nano-node/nano/test_common/system.cpp:93

 0# nano::generate_stacktrace[abi:cxx11]() at /home/ds/nano/releases/V23.0/nano-node/nano/lib/utility.cpp:129
 1# assert_internal(char const*, char const*, char const*, unsigned int, bool, std::basic_string_view<char, std::char_traits<char> >) at /home/ds/nano/releases/V23.0/nano-node/nano/lib/utility.cpp:175
 2# nano::system::add_node(nano::node_config const&, nano::node_flags, nano::transport::transport_type) at /home/ds/nano/releases/V23.0/nano-node/nano/test_common/system.cpp:93
 3# nano::system::system(unsigned short, nano::transport::transport_type, nano::node_flags) at /home/ds/nano/releases/V23.0/nano-node/nano/test_common/system.cpp:127
 4# node_fork_storm_Test::TestBody() at /home/ds/nano/releases/V23.0/nano-node/nano/slow_test/node.cpp:195
 5# void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) at /home/ds/nano/releases/V23.0/nano-node/gtest/googletest/src/gtest.cc:2607
 6# void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) at /home/ds/nano/releases/V23.0/nano-node/gtest/googletest/src/gtest.cc:2643
[...]

PR #3750, opened 2022-03-12
Fix_fork_storm failure #3709 on node allocation and catch stalled vote/fork resolution errorrs in the test

Justin-Randall:
This does not make the test actually pass, it merely keeps if from failing over administrative interventions so developers can see what conditions the test reveals (either in the implementation or assumptions in the test itself).

Some additional diagnostics streaming to GTEST are included as a both a reference for better practices and to illustrate the failure condition.

Addressing issue #3709


clemahieu (approved):

zhyatt:
Thanks for your help @Justin-Randall !

PR #3753, opened 2022-03-21
Pending to receivable renames

### LINKED ISSUE #3689: Finish pending to receivable changes for RPCs
The following RPC commands need an alias added and the previous RPC command to be marked as deprecated:

- [x] `accounts_pending` deprecated, add `accounts_receivable`
- [x] `search_pending` deprecated, add `search_receivable`
- [x] `search_pending_all` deprecated, add `search_receivable_all`
- [x] `wallet_pending` deprecated, add `wallet_receivable`

- [x] `account_info` and `blocks_info` RPC command needs the optional `pending` parameter to be deprecated and `receivable` parameter added
- [x] `account_info` and `blocks_info` RPC response for both should include both `pending` and `receivable` for backwards compatibility
- [x] `ledger` and `wallet_ledger` RPCs need `receivable` option support, deprecate `pending` option
- [x] `pending` and `pending_exists` RPCs need to have `deprecation` = `1` returned

These details were incorrectly noted in the documentation as having been completed in V23 (via PR #3411) but those notes will be removed by https://github.com/nanocurrency/nano-docs/pull/613.

dsiganos:
* Add deprecated field in RPC response of pending and pending_exists  
    The RPC commands did not return "deprecated=true", now they do.
    Also renamed underlying handling functions from pending to receivable.

* Add support for receivable option to wallet_ledger RPC 

* Add support for receivable option to ledger RPC command  
    The ledger RPC command only supported pending option.
    Now, it supports both with receivable having priority, if both are set.

* Convert RPC blocks_info to use receivable rather than pending 

* Rename `nano::ledger::account_pending` to `nano::ledger::account_receivable`

* account_info rpc: introduce receivable option and confirmed_receivable reply  
    Introduce new option called receivable to replace pending option.
    Currently both are supported and do the same job.
    However, if both are set, receivable has priority.
    Also added confirmed_receivable reply when include_confirmed is set.

* Copy search_pending to search_receivable and deprecate the former

* Copy search_pending_all to search_receivable_all and deprecate the former

* Copy wallet_pending to wallet_receivable and deprecate the former

* Copy accounts_pending to accounts_receivable and deprecate the former

resolves #3689

### DIFF
--- a/nano/core_test/ledger.cpp
+++ b/nano/core_test/ledger.cpp
@@ -107,7 +107,7 @@ TEST (ledger, process_send)
 	ASSERT_EQ (nano::process_result::progress, return1.code);
 	ASSERT_EQ (nano::dev::genesis_key.pub, store->block.account_calculated (send));
 	ASSERT_EQ (50, ledger.account_balance (transaction, nano::dev::genesis_key.pub));
-	ASSERT_EQ (nano::dev::constants.genesis_amount - 50, ledger.account_pending (transaction, key2.pub));
+	ASSERT_EQ (nano::dev::constants.genesis_amount - 50, ledger.account_receivable (transaction, key2.pub));
 	nano::account_info info2;
 	ASSERT_FALSE (store->account.get (transaction, nano::dev::genesis_key.pub, info2));
 	ASSERT_EQ (2, info2.block_count);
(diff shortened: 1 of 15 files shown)


theohax (approved):
There's probably quite a bunch of `pending`s left, but this definitely looks good towards our new terminology and hopefully will make things less confusing (Y)

clemahieu (approved):

Issue #3581, opened 2021-12-04
nano node .deb package dependency mismatch on debian testing (bullseye)


oiime:
### Summary

Hello, `qt5-default` no longer comes with debian 11, that causes installation difficultly as it either needs to be installed manually or using --force-all on the package which can cause issues later.

### Node version

22.1

### Build details

dpkg --force-all -i nano-node-V22.1-Linux.deb

from log
[2021-Dec-04 09:17:09.713984]: Node starting, version: V22.1
[2021-Dec-04 09:17:09.714030]: Build information: d91016b "GNU C++ version " "7.5.0" "BOOST 107000" BUILT "Jun 11 2021"
[2021-Dec-04 09:17:09.714046]: Database backend: LMDB 0.9.25

### OS and version

5.10.0-9-amd64 #1 SMP Debian 5.10.70-1 (2021-09-30) x86_64 GNU/Linux

### Steps to reproduce the behavior

1. install debian 11
2. download nano-node-V23.0RC1-Linux.deb
3. dpkg -i nano-node-V23.0RC1-Linux.deb

### Expected behavior

debian package managing to install on what would eventually be debian stable

### Actual behavior

```
dpkg: dependency problems prevent configuration of nano-node:
 nano-node depends on qt5-default; however:
  Package qt5-default is not installed.
```

### Possible solution

1. have a version specific .deb with different dependencies 
2. break installation into several packages as qt is not needed for just running a node (as far as i understood from the code)

PR #3760, opened 2022-03-25
Fixes the missing of qt5-default package on newer Debian/Ubuntu versions

thsfs:
Fixes the missing `qt5-default` package as described in https://github.com/nanocurrency/nano-node/issues/3581.

Since Debian 11 (bullseye), this package isn't used anymore, and it will probably not be available on Ubuntu 22.04, as it isn't on 21.04. The following packages can be used as an alternative: `qtbase5-dev`, `qtchooser`, `qt5-qmake`, `qtbase5-dev-tools`,
so the .deb package now instructs in its dependency either options.


dsiganos:
Why is `qtchooser` needed?

clemahieu (approved):

thsfs:
> Why is `qtchooser` needed?

@dsiganos . Not sure where it is used, but it seems to be a default replacement for the whole `qt5-default` package, like mentioned here:
https://askubuntu.com/questions/1335184/qt5-default-not-in-ubuntu-21-04

PR #3763, opened 2022-03-28
Add reverse link store for fast receive block lookups by only their send block hashes

Exxenoz:
- Added reverse_link_store to persist the [send block hash] -> [receive block hash] pairs.
- Implemented LMDB v21 -> v22 migration to create the new reverse_links table on startup.
- Implemented node.enable_reverse_links option (default disabled)
- Hooked into block_cemented_callback to store the reverse links of new cemented blocks.
    - I'm still not sure if this is the best place to do that, but it seems to work.
- Implemented initial reverse link pair loading into the node constructor.
    - I'm still not sure if this is the best place to start this relatively long loading process (it took around an hour on my system to process all cemented blocks from the live ledger), but it works.
- Added some reverse link test cases
    - Test case for the initial reverse link loading in the nano::node constructor is missing, because I was not yet able to find a proper solution to modify/inject the nano::store before nano::node creation.

This PR will drastically improve the performance of #3702

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -967,6 +967,53 @@ TEST (active_transactions, confirmation_consistency)
 }
 }
 
+/**
+ * This test case tests the creation of reverse links for incoming confirmed blocks.
+ */
+TEST (active_transactions, reverse_link)
+{
+	// Initialise node environment and enable reverse links
+	nano::system system;
+	nano::node_config node_config = nano::node_config (nano::get_available_port (), system.logging);
+	node_config.enable_reverse_links = true;
+	auto & node = *system.add_node (node_config);
+
+	// Create and process two blocks
+	nano::keypair key1;
+	system.wallet (0)->insert_adhoc (nano::dev::genesis_key.prv);
+	auto const send1 = system.wallet (0)->send_action (nano::dev::genesis_key.pub, key1.pub, node.config.receive_minimum.number ());
+	ASSERT_NE (send1, nullptr);
+	system.wallet (0)->insert_adhoc (key1.prv);
+	auto const open1 = system.wallet (0)->receive_action (send1->hash (), key1.pub, node.config.receive_minimum.number (), send1->link ().as_account ());
+	ASSERT_NE (open1, nullptr);
+
+	ASSERT_TIMELY (5s, node.block_confirmed (send1->hash ()));
+	ASSERT_TIMELY (5s, node.block_confirmed (open1->hash ()));
+
+	// Check if reverse link was created
+	ASSERT_EQ (1, node.store.reverse_link.count_accurate (node.store.tx_begin_read ()));
+	ASSERT_EQ (open1->hash (), node.store.reverse_link.get (node.store.tx_begin_read (), send1->hash ()));
+
+	// Now disable reverse links
+	node.config.enable_reverse_links = false;
+
+	// Create and process two more blocks
+	nano::keypair key2;
+	system.wallet (0)->insert_adhoc (nano::dev::genesis_key.prv);
+	auto const send2 = system.wallet (0)->send_action (nano::dev::genesis_key.pub, key2.pub, node.config.receive_minimum.number ());
+	ASSERT_NE (send2, nullptr);
+	system.wallet (0)->insert_adhoc (key2.prv);
+	auto const open2 = system.wallet (0)->receive_action (send2->hash (), key2.pub, node.config.receive_minimum.number (), send2->link ().as_account ());
+	ASSERT_NE (open2, nullptr);
+
+	ASSERT_TIMELY (5s, node.block_confirmed (send2->hash ()));
+	ASSERT_TIMELY (5s, node.block_confirmed (open2->hash ()));
+
+	// The second reverse link should not be created this time
+	ASSERT_EQ (1, node.store.reverse_link.count_accurate (node.store.tx_begin_read ()));
+	ASSERT_EQ (nano::block_hash (0), node.store.reverse_link.get (node.store.tx_begin_read (), send2->hash ()));
+}
+
 // Test disabled because it's failing intermittently.
 // PR in which it got disabled: https://github.com/nanocurrency/nano-node/pull/3629
 // Issue for investigating it: https://github.com/nanocurrency/nano-node/issues/3634
--- a/nano/node/lmdb/lmdb.cpp
+++ b/nano/node/lmdb/lmdb.cpp
@@ -53,6 +53,7 @@ nano::lmdb::store::store (nano::logger_mt & logger_a, boost::filesystem::path co
 		peer_store,
 		confirmation_height_store,
 		final_vote_store,
+		reverse_link_store,
 		version_store
 	},
 	// clang-format on
@@ -65,6 +66,7 @@ nano::lmdb::store::store (nano::logger_mt & logger_a, boost::filesystem::path co
 	peer_store{ *this },
 	confirmation_height_store{ *this },
 	final_vote_store{ *this },
+	reverse_link_store{ *this },
 	unchecked_store{ *this },
 	version_store{ *this },
 	logger (logger_a),
@@ -222,6 +224,7 @@ void nano::lmdb::store::open_databases (bool & error_a, nano::transaction const 
 	error_a |= mdb_dbi_open (env.tx (transaction_a), "pending", flags, &pending_store.pending_v0_handle) != 0;
 	pending_store.pending_handle = pending_store.pending_v0_handle;
 	error_a |= mdb_dbi_open (env.tx (transaction_a), "final_votes", flags, &final_vote_store.final_votes_handle) != 0;
+	error_a |= mdb_dbi_open (env.tx (transaction_a), "reverse_links", flags, &reverse_link_store.reverse_links_handle) != 0;
 
 	auto version_l = version.get (transaction_a);
 	if (version_l < 19)
@@ -306,6 +309,9 @@ bool nano::lmdb::store::do_upgrades (nano::write_transaction & transaction_a, na
 			upgrade_v20_to_v21 (transaction_a);
 			[[fallthrough]];
 		case 21:
+			upgrade_v21_to_v22 (transaction_a);
+			[[fallthrough]];
+		case 22:
 			break;
 		default:
 			logger.always_log (boost::str (boost::format ("The version of the ledger (%1%) is too high for this node") % version_l));
@@ -779,6 +785,14 @@ void nano::lmdb::store::upgrade_v20_to_v21 (nano::write_transaction const & tran
 	logger.always_log ("Finished creating new final_vote table");
 }
 
+void nano::lmdb::store::upgrade_v21_to_v22 (nano::write_transaction const & transaction_a)
+{
+	logger.always_log ("Preparing v21 to v22 database upgrade...");
+	mdb_dbi_open (env.tx (transaction_a), "reverse_links", MDB_CREATE, &reverse_link_store.reverse_links_handle);
+	version.put (transaction_a, 22);
+	logger.always_log ("Finished creating new reverse_links table");
+}
+
 /** Takes a filepath, appends '_backup_<timestamp>' to the end (but before any extension) and saves that file in the same directory */
 void nano::lmdb::store::create_backup_file (nano::mdb_env & env_a, boost::filesystem::path const & filepath_a, nano::logger_mt & logger_a)
 {
@@ -882,6 +896,8 @@ MDB_dbi nano::lmdb::store::table_to_dbi (tables table_a) const
 			return confirmation_height_store.confirmation_height_handle;
 		case tables::final_votes:
 			return final_vote_store.final_votes_handle;
+		case tables::reverse_links:
+			return reverse_link_store.reverse_links_handle;
 		default:
 			release_assert (false);
 			return peer_store.peers_handle;
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -2014,6 +2014,41 @@ TEST (block_store, final_vote)
 	}
 }
 
+/**
+ * This test case tests insertion, finding and clearing of reverse links.
+ */
+TEST (block_store, reverse_link)
+{
+	nano::logger_mt logger;
+	auto store = nano::make_store (logger, nano::unique_path (), nano::dev::constants);
+	ASSERT_TRUE (!store->init_error ());
+
+	// Insert some reverse links and check if count got increased
+	{
+		auto transaction = store->tx_begin_write ();
+		store->reverse_link.put (transaction, nano::block_hash (100), nano::block_hash (1));
+		ASSERT_EQ (store->reverse_link.count_accurate (transaction), 1);
+		store->reverse_link.put (transaction, nano::block_hash (200), nano::block_hash (2));
+		ASSERT_EQ (store->reverse_link.count_accurate (transaction), 2);
+		store->reverse_link.put (transaction, nano::block_hash (300), nano::block_hash (3));
+		ASSERT_EQ (store->reverse_link.count_accurate (transaction), 3);
+	}
+
+	// Request recently added reverse links and check if results are matching
+	{
+		auto transaction = store->tx_begin_read ();
+		ASSERT_EQ (store->reverse_link.get (transaction, nano::block_hash (300)), nano::block_hash (3));
+		ASSERT_EQ (store->reverse_link.get (transaction, nano::block_hash (200)), nano::block_hash (2));
+		ASSERT_EQ (store->reverse_link.get (transaction, nano::block_hash (100)), nano::block_hash (1));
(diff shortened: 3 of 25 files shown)


nano/secure/store_partial.hpp
```diff
@@ -135,7 +139,7 @@ class store_partial : public store
 
 protected:
 	nano::ledger_constants & constants;
-	int const version_number{ 21 };
+	int const version_number{ 22 };
```
thsfs: As this table doesn't affect the way the ledger is organized and as it'll be disabled by default, do we want to increase the database version @clemahieu ?

nano/node/lmdb/lmdb.cpp
```diff
@@ -307,6 +310,10 @@ bool nano::mdb_store::do_upgrades (nano::write_transaction & transaction_a, bool
 			upgrade_v20_to_v21 (transaction_a);
 			[[fallthrough]];
 		case 21:
+			upgrade_v21_to_v22 (transaction_a);
```
thsfs, line 312: The are no upgrade operations to be performed, but this will be necessary in case we're going to increase the database version number.

nano/core_test/active_transactions.cpp
```diff
@@ -967,6 +967,91 @@ TEST (active_transactions, confirmation_consistency)
...
+				.link (send1->hash ())
+				.sign (key.prv, key.pub)
+				.work (*system.work.generate (key.pub))
+				.build_shared ();
+	// Process and confirm both blocks
+	node.process_active (send1);
+	node.process_active (open1);
+	auto vote = std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, nano::vote::timestamp_max, nano::vote::duration_max, std::vector<nano::block_hash>{ send1->hash (), open1->hash () });
+	node.vote_processor.vote (vote, std::make_shared<nano::transport::channel_loopback> (node));
+	node.block_processor.flush ();
```
thsfs: We are now avoiding using this flush function as it doesn't guarantee everything is synchronized and it usually leads to errors on our CI. We should ensure the awaited processes are complete by using ASSERT_TIMELY. Would this work if we remove the flush and relying on the assert from line 1009?

nano/core_test/active_transactions.cpp
```diff
@@ -967,6 +967,91 @@ TEST (active_transactions, confirmation_consistency)
...
+				.link (send2->hash ())
+				.sign (key2.prv, key2.pub)
+				.work (*system.work.generate (key2.pub))
+				.build_shared ();
+	// Process and confirm both blocks
+	node.process_active (send2);
+	node.process_active (open2);
+	vote = std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, nano::vote::timestamp_max, nano::vote::duration_max, std::vector<nano::block_hash>{ send2->hash (), open2->hash () });
+	node.vote_processor.vote (vote, std::make_shared<nano::transport::channel_loopback> (node));
+	node.block_processor.flush ();
```
thsfs: Same here, we can't rely on this flush.

nano/node/lmdb/lmdb.cpp
```diff
@@ -780,6 +786,14 @@ void nano::mdb_store::upgrade_v20_to_v21 (nano::write_transaction const & transa
 	logger.always_log ("Finished creating new final_vote table");
 }
 
+void nano::mdb_store::upgrade_v21_to_v22 (nano::write_transaction const & transaction_a)
```
thsfs: Should be applicable only if the table is enabled.

nano/node/lmdb/lmdb.cpp
Exxenoz: > The are no upgrade operations to be performed, but this will be necessary in case we're going to increase the database version number.

Using LMDB the reverse_links table must be explicitly created, which is the upgrade operation I implemented in upgrade_v21_to_v22. RocksDB is more dynamic and works just fine.

nano/core_test/active_transactions.cpp
Exxenoz: > We are now avoiding using this flush function as it doesn't guarantee everything is synchronized and it usually leads to errors on our CI. We should ensure the awaited processes are complete by using ASSERT_TIMELY. Would this work if we remove the flush and relying on the assert from line 1009?

I just removed the flushes and it seems to work without them.

nano/node/lmdb/lmdb.cpp
Exxenoz: > Should be applicable only if the table is enabled.

It is just the empty table creation for LMDB, which should be finished rather fast. I don't really know where I could move the table creation otherwise if I may not use the version counter as well. How do I check if the table already exists? 😅

Exxenoz: Btw. thanks for the fast feedback! 😄

thsfs: Np, there is an alternative you can use, `mdb_dbi_open` returns an error when the handle doesn't exist.
Look at this: https://github.com/nanocurrency/nano-node/blob/a37a1480d8a13821dc6b2345ecc15517d64b0e35/nano/node/lmdb/lmdb.cpp#L83

nano/core_test/active_transactions.cpp
```diff
@@ -967,6 +967,89 @@ TEST (active_transactions, confirmation_consistency)
...
+	// Process and confirm both blocks
+	node.process_active (send1);
+	node.process_active (open1);
+	auto vote = std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, nano::vote::timestamp_max, nano::vote::duration_max, std::vector<nano::block_hash>{ send1->hash (), open1->hash () });
+	node.vote_processor.vote (vote, std::make_shared<nano::transport::channel_loopback> (node));
+	ASSERT_TIMELY (5s, node.ledger.cache.cemented_count == 3);
+	// Check if reverse link was created
+	{
+		auto transaction = node.store.tx_begin_read ();
+		ASSERT_EQ (1, node.store.reverse_link.count (transaction));
```
dsiganos: You cannot depend on count being accurate when using rocksdb.
I tend to do a manual count one by one by iterating through all the elements.
See function `manually_count_pruned_blocks`.

Exxenoz: > You cannot depend on count being accurate when using rocksdb. I tend to do a manual count one by one by iterating through all the elements. See function `manually_count_pruned_blocks`.

I guess that depends on the count implementation in `nano::rocksdb_store::count`? Some tables use `db->GetIntProperty (table_to_column_family (table_a), "rocksdb.estimate-num-keys", &sum);` as an approximation, because it is faster. But since the count of reverse links is only used in test cases with a rather small amount of elements, I decided to implement the exact counting:

```
else if (table_a == tables::reverse_links)
{
	for (auto i (reverse_link.begin (transaction_a)), n (reverse_link.end ()); i != n; ++i)
	{
		++sum;
	}
}
```

nano/core_test/active_transactions.cpp
```diff
@@ -967,6 +967,89 @@ TEST (active_transactions, confirmation_consistency)
...
+				 .link (send1->hash ())
+				 .sign (key.prv, key.pub)
+				 .work (*system.work.generate (key.pub))
+				 .build_shared ();
+	// Process and confirm both blocks
+	node.process_active (send1);
+	node.process_active (open1);
+	auto vote = std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, nano::vote::timestamp_max, nano::vote::duration_max, std::vector<nano::block_hash>{ send1->hash (), open1->hash () });
+	node.vote_processor.vote (vote, std::make_shared<nano::transport::channel_loopback> (node));
+	ASSERT_TIMELY (5s, node.ledger.cache.cemented_count == 3);
```
dsiganos: I am not convinced that this is a good way to check that the blocks are cemented.
I know we use this pattern in other places but I think this is a race condition.
The cache counter is not synchronized with the database transactions.
The counter and the database might not be aligned.
I think using `nano::node::get_confirmation_height`, which takes a transaction object, would be better.

nano/core_test/active_transactions.cpp
```diff
@@ -967,6 +967,89 @@ TEST (active_transactions, confirmation_consistency)
...
+ */
+TEST (active_transactions, reverse_link)
+{
+	// Initialise node environment and enable reverse links
+	nano::system system;
+	nano::node_config node_config = nano::node_config (nano::get_available_port (), system.logging);
+	node_config.enable_reverse_links = true;
+	auto & node = *system.add_node (node_config);
+
+	// Generate a random key pair
```
dsiganos: Are you aware that you can use something like:
```
	wallet.insert_adhoc (nano::dev::genesis_key.prv);
	auto const send1 = wallet.send_action (nano::dev::genesis_key.pub, key1.pub, node.config.receive_minimum.number ());
```
to do transaction pairs (send/recv) easily?

nano/core_test/active_transactions.cpp
dsiganos: I think it would be better to follow the existing pattern we already have for consistency.
The count of a store is unreliable. And unit tests use a different count function that counts one by one and is reliable.

nano/core_test/active_transactions.cpp
Exxenoz: > Are you aware that you can use something like:
> 
> ```
> 	wallet.insert_adhoc (nano::dev::genesis_key.prv);
> 	auto const send1 = wallet.send_action (nano::dev::genesis_key.pub, key1.pub, node.config.receive_minimum.number ());
> ```
> 
> to do transaction pairs (send/recv) easily?

Oh, this looks indeed cleaner if I can drop the vote stuff then. I will give it a try. Thanks 😄

dsiganos: Just remember that wallet auto-receive only works for receivables larger or equal to: `node.config.receive_minimum.number ()`

nano/node/lmdb/lmdb.cpp
Exxenoz: @thsfs Looks like lmdb tables are already created when nano::mdb_store::open_databases is called with MDB_CREATE before upgrades are processed. So there is no extra need for a database upgrade. The last upgrade v20 to v21 just got me confused, because the table for final votes was created there explicitly. The current design makes this obsolete as well.

nano/node/lmdb/lmdb.cpp
Exxenoz: > Np, there is an alternative you can use, `mdb_dbi_open` returns an error when the handle doesn't exist. Look at this:
> 
> https://github.com/nanocurrency/nano-node/blob/a37a1480d8a13821dc6b2345ecc15517d64b0e35/nano/node/lmdb/lmdb.cpp#L83

@thsfs I removed the upgrade method completely and reset the version number back to 21. Is it fine if the reverse_links table is created in `nano::mdb_store::open_databases` with the other tables, even if the reverse link option is disabled? We don't have direct node config access there, so I would have to either add an extra parameter for this option or add it to `nano::lmdb_config`.

//EDIT: I had to readd the upgrade method, because `nano::mdb_store::open_databases` with the "create" flag is only called on version mismatch.

nano/core_test/active_transactions.cpp
Exxenoz: > I think it would be better to follow the existing pattern we already have for consistency. The count of a store is unreliable. And unit tests use a different count function that counts one by one and is reliable.

@dsiganos I implemented a generic count_accurate method for stores: https://github.com/Exxenoz/nano-node/commit/d58386ac33bc0fe632f2283e74c33b30c8763c4c

Will update your code later; If you don't mind I will include it here into this PR. Alternately, I could move the iterable_store implementation into another PR, if you think this PR is already too huge.

nano/core_test/active_transactions.cpp
Exxenoz: > I am not convinced that this is a good way to check that the blocks are cemented. I know we use this pattern in other places but I think this is a race condition. The cache counter is not synchronized with the database transactions. The counter and the database might not be aligned. I think using `nano::node::get_confirmation_height`, which takes a transaction object, would be better.

@dsiganos I improved the reverse link test cases after your feedback: https://github.com/Exxenoz/nano-node/commit/0a047014f824b047306aa97f5d150510723e6900

nano/core_test/active_transactions.cpp
Exxenoz: > Just remember that wallet auto-receive only works for receivables larger or equal to: `node.config.receive_minimum.number ()`

Thanks, `send_action` and `receive_action` worked great. 😊

Exxenoz:
@dsiganos @thsfs Commit squash & rebase done (PR includes the iterable_store implementation from #3766).

thsfs:
@dsiganos , @clemahieu , how the reverse links table could get affected by the confirmation height?

nano/node/lmdb/lmdb.cpp
Exxenoz: > @thsfs Looks like lmdb tables are already created when nano::mdb_store::open_databases is called with MDB_CREATE before upgrades are processed. So there is no extra need for a database upgrade. The last upgrade v20 to v21 just got me confused, because the table for final votes was created there explicitly. The current design makes this obsolete as well.

@thsfs @dsiganos Ok, I was wrong. `open_databases (error, transaction, MDB_CREATE);` gets called only if `is_fully_upgraded == false`, which is not the case if we don't increase the version number. Will think about this tomorrow.

dsiganos:
I think the reverse_link table should be outside of the core database and in a separate auxiliary database. That will:
* avoid the core database versioning problem
* allow flexibility
* separation of core data and auxiliary data

Exxenoz:
> I think the reverse_link table should be outside of the core database and in a separate auxiliary database. That will:
> 
> * avoid the core database versioning problem
> * allow flexibility
> * separation of core data and auxiliary data

Oof, this would make the persistence layer even more complicated than it already is. In addition these changes would have to be made for both lmdb and rocksdb. I'm not really convinced it's worth the effort, especially if there are plans to move to an SQL backend.

Exxenoz:
I will update the PR within the next two days. Have to finish another task first, but should be done soon.

Exxenoz:
> I will update the PR within the next two days. Have to finish another task first, but should be done soon.

Done. I also readded the LMDB reverse_links table creation and increased the store version.

Exxenoz:
I'll maybe take another look at this soon. A user reached out to me and requested that blocklattice.io should display the corresponding receive block hash for every send block. The background was as follows: The user accidentally sent Nano to Binance's hot wallet, and the Binance support team apparently refused to process a refund without knowing the associated receive block. By that time, however, the receive block had already been buried under several hundred other blocks, making it difficult to locate.

Are there any further comments regarding this PR, apart from the suggestion to move the table to a separate LMDB database? (If that's acceptable, I'll skip implementing it with RocksDB.)

pwojcikdev:
Honestly, I personally won’t be reviewing this PR because I don’t like the idea that much and don’t like additional complexity; I don’t want to maintain this. I don't think original reviewers actively work on this project anymore, so not sure how they envisioned this.
It’s pretty simple to create a background process for mirroring ledger into some kind of external relational database and doing this lookup having sql db is trivial.

Exxenoz:
> Honestly, I personally won’t be reviewing this PR because I don’t like the idea that much and don’t like additional complexity; I don’t want to maintain this. I don't think original reviewers actively work on this project anymore, so not sure how they envisioned this. It’s pretty simple to create a background process for mirroring ledger into some kind of external relational database and doing this lookup having sql db is trivial.

Yes, that's the alternative. For my original use case, though, the cache needed to be integrated directly into the node, which is why I created this extension in the first place. I'll go ahead and close the PR for now, as it seems to be addressing a non-issue for most users. Appreciate your input, thanks!

qwahzi:
@Exxenoz Would `blocks_info` (not `block_info`) with the `receive_hash` flag accomplish what you're looking for?

https://docs.nano.org/commands/rpc-protocol/#blocks_info

Exxenoz:
> @Exxenoz Would `blocks_info` (not `block_info`) with the `receive_hash` flag accomplish what you're looking for?
> 
> https://docs.nano.org/commands/rpc-protocol/#blocks_info

Hi qwahzi! Yes, my concern is that the `receive_hash` option could take a long time for some requests. In the worst-case scenario, the entire account chain must be iterated to find the receive block. You can test it yourself:

```
{  
  "action": "blocks_info",
  "json_block": "true",
  "hashes": ["82402F0572B9398E3C3BD3B889078B4B139053FE131CC677A89569A474F5AB33"],
  "receive_hash": true
}
```

```
{  
  "action": "blocks_info",
  "json_block": "true",
  "hashes": ["82402F0572B9398E3C3BD3B889078B4B139053FE131CC677A89569A474F5AB33", "82402F0572B9398E3C3BD3B889078B4B139053FE131CC677A89569A474F5AB33", "82402F0572B9398E3C3BD3B889078B4B139053FE131CC677A89569A474F5AB33"],
  "receive_hash": true
}
```

This is unusable on most machines.

PR #3766, opened 2022-03-29
Let stores inherit from new iterable_store class and implement count_accurate method

Exxenoz:
Optimisation part of #3763


dsiganos (dismissed):

dsiganos:
It makes sense to me but @clemahieu is a better person to review since he has plans for these classes.

Exxenoz:
Updated to latest develop commit.

PR #3767, opened 2022-03-30
Delay function for unit tests

dsiganos:
Sometimes, it is useful it useful to be able to delay a unit test and have
the back async processing running too. This function continuously calls
io_ctx.run_one_for in system class for the specified amount of time.

### DIFF
--- a/nano/test_common/system.hpp
+++ b/nano/test_common/system.hpp
@@ -43,6 +43,7 @@ public:
 	 */
 	std::error_code poll (std::chrono::nanoseconds const & sleep_time = std::chrono::milliseconds (50));
 	std::error_code poll_until_true (std::chrono::nanoseconds deadline, std::function<bool ()>);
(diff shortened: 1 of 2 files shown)


nano/test_common/system.hpp
```diff
@@ -54,8 +55,6 @@ class system final
 	std::chrono::time_point<std::chrono::steady_clock, std::chrono::duration<double>> deadline{ std::chrono::steady_clock::time_point::max () };
 	double deadline_scaling_factor{ 1.0 };
 	unsigned node_sequence{ 0 };
-
-private:
```
clemahieu: This doesn't seem intended.

clemahieu (changes requested):
Looks good save one item.

nano/test_common/system.hpp
dsiganos: Yeap, commit mistake. Wrong commit.

clemahieu (approved):

PR #3773, opened 2022-03-30
Improve/fix test rpc.account_representative_set

dsiganos:
The test wasn't testing properly that the RPC commands worked.
It just checked that it got a reply.
I improved the test to check that the representative really changed.


zhyatt (approved):

zhyatt:
I approved this as it matches the discussion we had about it, but another review of more of the code details is needed before merging.

clemahieu (approved):

Issue #3777, opened 2022-04-03
Allow RPC 'nano_to_raw' to accept decimals.


nano2dev:
### Summary

RPC: 

```
{
  "action": "nano_to_raw",
  "amount": "10.50"
}
```
```
{ error: 'Invalid amount number' }
```

### What problem would be solved by this feature?

Would be nice if the Node handle decimals. 

> Why make an rpc call for rather basic arithmetic?

 When creating new blocks, it's expected that you provide the 'new balance'... a lot of math that can be solved by the Node.

I'd also like to propose a new RPC call: 'calculate_new_balance' -- or something like that.

The use case is to avoid having to code my own arithmetic. 

I'm not lazy, I just don't trust my own code, and prefer to depend on Node when working with such small decimals. 

### Are there any previous requests for this feature?

Could not find any. 

### Do you have a suggested solution?

My apologies. Don't speak C++. Only JS.

### If this feature is approved, would you be willing to submit a pull request with the solution?

I am willing to collaborate

### Possible solution

Accept decimals. 

Additionally, (and I can make a separate Issue) add a RPC call(s) to handle 'subtraction' & 'addition' to be used in the 'block_create' RPC and anywhere a 'new balance' is expected.

### Supporting files

I was pointe to this file in the source code: 

https://github.com/nanocurrency/nano-node/blob/4752c66c80a6543ae0c1840e9036377561b85e43/nano/node/json_handler.cpp#L272-L284

clemahieu:
We force the API caller to do the conversion because of the general problem of using floating point math in finance.

clemahieu:
I do wonder if we could interpret it as a fixed-fraction so it can be assured precision won't be lost.

shryder:
Btw you don't have to write your own code, just use one of the existing Nano libraries to do the conversion. This is one of them if you're writing code in JS: 
https://github.com/marvinroger/nanocurrency-js/tree/master/packages/nanocurrency

I agree it would be nice to have a `create_send_block` RPC where you pass the amount you want to send instead of the updated balance though.

nano2dev:
Thank you, gentlemen.

Wish I could respond to Colin, intelligently. Math is not my strong suit. 

Indeed, a single RPC call that returns the calculated new balance of a "new block" would be lovely. 

I feel bad not being able to provide a PR, so for future readers:

[...]

PR #3778, opened 2022-04-04
Feature/raw amount arithmetic

RickiNano:
RPC functions to add and subtract two raw values. This is useful for calculating the new balance when creating a block with block_create
This was created as a possible solution to #3777


nano2dev:
Please note this PR is outside the scope of #3777 and not a desired outcome.

However, if the core team believes that this solution is better or more feasible than modifying "nano_to_raw" - then I'll will accept it.

Thank you @RickiNano for taking the time, nonetheless.

PR #3782, opened 2022-04-07
Segregates the CPack setup from the principal CMakeLists.txt

thsfs:
- Put the packing setup in a different cmake module
- Put the platform setup into specific files
- Updates the cmake-format script to find the cmake modules


CMakeLists.txt
```diff
@@ -161,54 +154,6 @@ option(
   ON)
 option(ENABLE_AVX2 "Enable AVX2 optimizations" OFF)
 
-set(ACTIVE_NETWORK
```
clemahieu: This is passed as a #define to the compiler so I think this needs to stay in the main cmake file.

PR #3783, opened 2022-04-08
Remove db templates

clemahieu:
This branch removes templates from the lmdb/rocksdb backend.

Templates were intended to reduce the amount of boilerplate code shared between different backends, however, backends infrequently had matching implementations making the template code confusing and unmaintainable.

This removes templates from the various block store classes and replaces it with simple inheritance.


thsfs:
Seems to be a good path to reduce complexity and enable further changes.

thsfs (approved):

PR #3784, opened 2022-04-09
Add db backend namespaces

clemahieu:
This change forces classes under the nano namespace to use a global scope when referencing rocksdb library functions. This allows rocksdb to be used as a namespace under the nano namespace.


thsfs (approved):
Looks correct.

PR #3785, opened 2022-04-10
Further removal of templates in db classes

clemahieu:
Moves methods off of store_partial and eliminates it.
Moves mdb_store/rocksdb_store in to their respective namespaces and renamed to store.


nano/node/lmdb/account_store.cpp
```diff
@@ -27,7 +28,7 @@ bool nano::lmdb::account_store::get (nano::transaction const & transaction, nano
 void nano::lmdb::account_store::del (nano::write_transaction const & transaction_a, nano::account const & account_a)
 {
 	auto status = store.del (transaction_a, tables::accounts, account_a);
-	release_assert_success (store, status);
+	store.release_assert_success (status);
```
thsfs, line 31: I'd let this as a free function.

thsfs (approved):
Looks good.

Issue #3752, opened 2022-03-18
`accounts_representatives` is failing when one of the account is not opened


tombertrand:
### Summary

Original PR: #3409

When using the `accounts_representatives` RPC if one of the account in the list is not opened, it returns an error message instead of setting the representative response value to `""` or `null` or `undefined` for that account.

```
curl -d '{
  "action": "accounts_representatives",
  "accounts": ["nano_1openedaccount", "nano_3unopenedaccount"]
}' http://127.0.0.1:7076
```

### Node version

V23

### Build details

Production build

### OS and version

unrelated

### Steps to reproduce the behavior

Query the `accounts_representative` with opened and unopened accounts.

### Expected behavior

```
{
  "representatives": {
    "nano_1openedaccount": "nano_1repaccount",
    "nano_3unopenedaccount": "" or undefined or null
  }
}
```

### Actual behavior

```
{
    "error": "Account not found"
}
```

### Possible solution

include the unopened account in the response but mark the rep as an empty string `""` (preferred solution since it would be easier to read the response with missing reps than trying to diff the accounts sent in parameters vs the response) 
or
exclude the unopened account from the response

dsiganos:
This is the code that handles the RPC:
```
void nano::json_handler::accounts_representatives ()
{
        boost::property_tree::ptree representatives;
        for (auto & accounts : request.get_child ("accounts"))
        {
                auto account (account_impl (accounts.second.data ()));
                auto transaction (node.store.tx_begin_read ());
                auto info (account_info_impl (transaction, account));

                if (!ec)
                {
                        boost::property_tree::ptree entry;
                        entry.put ("", info.representative.to_account ());
                        representatives.push_back (std::make_pair (accounts.second.data (), entry));
                }
        }
        response_l.add_child ("representatives", representatives);
        response_errors ();
}
```

The only way I can explain what is happening is, the variable `ec` must be getting set as a side effect by something (probably by account_impl) and then somehow that causes the entire response to be ignore and a general error to be returned.

dsiganos:
Yeap, just as I thought. If `ec` is set then `response_l` is ignored and `response_error` is returned instead.
```
void nano::json_handler::response_errors ()
{
	if (!ec && response_l.empty ())
	{
		// Return an error code if no response data was given
		ec = nano::error_rpc::empty_response;
	}
	if (ec)
	{
		boost::property_tree::ptree response_error;
		response_error.put ("error", ec.message ());
		std::stringstream ostream;
[...]

PR #3789, opened 2022-04-13
Changes the accounts_representatives RPC to return per account response

thsfs:
The previous implementation returned only an error result in case any of the provided accounts were invalid or not found. This implementation handles each account separately returning either a successful or an error response.

Example:
```
    "representatives": {
        "nano_3wfddg7a1paogrcwi3yhwnaerboukbr7rs3z3ino5toyq3yyhimo6f6egij6": "nano_3wfddg7a1paogrcwi3yhwnaerboukbr7rs3z3ino5toyq3yyhimo6f6egij6",
        "nano_36uccgpjzhjsdbj44wm1y5hyz8gefx3wjpp1jircxt84nopxkxti5bzq1rnz": "error: Bad account number",
        "nano_1hrts7hcoozxccnffoq9hqhngnn9jz783usapejm57ejtqcyz9dpso1bibuy": "error: Account not found"
    }
```

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -14,6 +14,7 @@
 
 #include <algorithm>
 #include <chrono>
+#include <vector>
 
 namespace
 {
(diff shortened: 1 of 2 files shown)


nano/node/json_handler.cpp
```diff
@@ -911,18 +912,24 @@ void nano::json_handler::accounts_balances ()
...
-			representatives.push_back (std::make_pair (accounts.second.data (), entry));
+			auto info = account_info_impl (node.store.tx_begin_read (), account);
+			if (!ec)
+			{
+				entry.put ("", info.representative.to_account ());
+				representatives.push_back (std::make_pair (account_from_request.second.data (), entry));
+				continue;
+			}
 		}
+		// Returns only the first error.
```
dsiganos: What does this comment mean?
I cannot understand its meaning.

thsfs: there might be an error for the provided account number (invalid number), and also an error for when it is not found "Account not found". Instead of retrieving a list of errors this implementation only returns the first it finds.

dsiganos: There cannot be a list of errors, there are only 2 errors and they cannot happen together

thsfs: Yes, returning a list was just I thought I had, it doesn't make sense for this function. My other explanation is: there is a stack where the account number is verified first and a new object is constructed for it, if there is an error on it, it won't look for its representatives and will return "Bad account number". But if it passes the first check and construct an account object, this account may still not be found. The first or the "only" error possible to be found is going to be added to the response.

dsiganos: I must say, I still do not understand what it means... but I don't want to focus on it too much.
Can we add some unit tests instead to clearly show what the expected behaviour is?

clemahieu:
> @zhyatt , @clemahieu : for the case the account is not found should we return an empty string instead?
> 
> `"nano_37rmg9h1iyyaqfdpmms3cm4fd5ewqdjcnj71i9ow14h3mx8dd5pn6ogqb1af": ""` instead of `"nano_37rmg9h1iyyaqfdpmms3cm4fd5ewqdjcnj71i9ow14h3mx8dd5pn6ogqb1af": "error: Account not found"`

It's probably more helpful to list the error for each entry rather than not giving no feedback.

nano/node/json_handler.cpp
thsfs: Removed the comment.

dsiganos (approved):

PR #3790, opened 2022-04-13
Changes the accounts_balances RPC to return per account results

thsfs:
The previous implementation returned only an error result in case any of the accounts were invalid. This implementation now handles each account separately returning per account results.

```
    "balances": {
        "nano_3wfddg7a1paogrcwi3yhwnaerboukbr7rs3z3ino5toyq3yyhimo6f6egij6": {
            "balance": "442000000000000000000000000000",
            "pending": "0",
            "receivable": "0"
        },
        "nano_36uccgpjzhjsdbj44wm1y5hyz8gefx3wjpp1jircxt84nopxkxti5bzq1rnz": {
            "error": "Bad account number"
        },
        "nano_1hrts7hcoozxccnffoq9hqhngnn9jz783usapejm57ejtqcyz9dpso1bibuy": {
            "error": "Account not found"
        }
    }
```

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -891,18 +891,23 @@ void nano::json_handler::account_weight ()
 void nano::json_handler::accounts_balances ()
 {
 	boost::property_tree::ptree balances;
-	for (auto & accounts : request.get_child ("accounts"))
+	for (auto & account_from_request : request.get_child ("accounts"))
 	{
-		auto account (account_impl (accounts.second.data ()));
+		boost::property_tree::ptree entry;
+		auto account = account_impl (account_from_request.second.data ());
 		if (!ec)
 		{
-			boost::property_tree::ptree entry;
-			auto balance (node.balance_pending (account, false));
+			auto balance = node.balance_pending (account, false);
 			entry.put ("balance", balance.first.convert_to<std::string> ());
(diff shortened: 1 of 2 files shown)


clemahieu (dismissed):

zhyatt:
Since we are doing `Account not found` error in some of the other responses when the account hasn't been opened, such as in #3789 , should we also do that here instead of `0` values? @clemahieu @thsfs @dsiganos 
```
        "nano_1hrts7hcoozxccnffoq9hqhngnn9jz783usapejm57ejtqcyz9dpso1bibuy": {
            "error": "Account not found"
        },
 ```

thsfs:
> Since we are doing `Account not found` error in some of the other responses when the account hasn't been opened, such as in #3789 , should we also do that here instead of `0` values? @clemahieu @thsfs @dsiganos
> 
> ```
>         "nano_1hrts7hcoozxccnffoq9hqhngnn9jz783usapejm57ejtqcyz9dpso1bibuy": {
>             "error": "Account not found"
>         },
> ```

It makes sense to me to have the same type of error response when the account is not found.

dsiganos:
This RPC should also be taking `include_only_confirmed` as option.
Like `account_balance` does. But that could be a different PR.

dsiganos (approved):

PR #3791, opened 2022-04-13
Changes the accounts_frontiers RPC to return per account results

thsfs:
The previous implementation returned only an error result in case any of the accounts were invalid. This implementation now handles each account separately returning per account results. It returns an empty string when the account's frontier is zero, and returns an error message for when the provided account is invalid.

Example:
```
    "frontiers": {
        "nano_3wfddg7a1paogrcwi3yhwnaerboukbr7rs3z3ino5toyq3yyhimo6f6egij6": "75BD65296241EB871918EBE3E99E9A191970A2724B3214B27F8AB205FF4FC30A",
        "nano_36uccgpjzhjsdbj44wm1y5hyz8gefx3wjpp1jircxt84nopxkxti5bzq1rnz": "error: Bad account number",
        "nano_1hrts7hcoozxccnffoq9hqhngnn9jz783usapejm57ejtqcyz9dpso1bibuy": "error: Account not found"
    }
```

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -956,18 +956,25 @@ void nano::json_handler::accounts_create ()
 void nano::json_handler::accounts_frontiers ()
 {
 	boost::property_tree::ptree frontiers;
-	auto transaction (node.store.tx_begin_read ());
-	for (auto & accounts : request.get_child ("accounts"))
+	auto transaction = node.store.tx_begin_read ();
+	for (auto & account_from_request : request.get_child ("accounts"))
 	{
-		auto account (account_impl (accounts.second.data ()));
+		auto account = account_impl (account_from_request.second.data ());
 		if (!ec)
 		{
-			auto latest (node.ledger.latest (transaction, account));
+			auto latest = node.ledger.latest (transaction, account);
 			if (!latest.is_zero ())
 			{
 				frontiers.put (account.to_account (), latest.to_string ());
+				continue;
+			}
+			else
+			{
+				ec = nano::error_common::account_not_found;
 			}
 		}
+		frontiers.put (account_from_request.second.data (), boost::str (boost::format ("error: %1%") % ec.message ()));
+		ec = {};
 	}
 	response_l.add_child ("frontiers", frontiers);
 	response_errors ();
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -3174,27 +3174,59 @@ TEST (rpc, accounts_representatives)
 	ASSERT_EQ (response_representative, nano::dev::genesis->account ().to_account ());
 }
 
+/**
+ * Test the RPC accounts_frontiers with 3 accounts, one good one, one with an invalid account ID and one with an account that does not exist.
+ */
 TEST (rpc, accounts_frontiers)
 {
 	nano::system system;
 	auto node = add_ipc_enabled_node (system);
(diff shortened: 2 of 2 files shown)


nano/node/json_handler.cpp
```diff
@@ -956,17 +956,26 @@ void nano::json_handler::accounts_create ()
 void nano::json_handler::accounts_frontiers ()
 {
 	boost::property_tree::ptree frontiers;
-	auto transaction (node.store.tx_begin_read ());
-	for (auto & accounts : request.get_child ("accounts"))
+	auto transaction = node.store.tx_begin_read ();
+	for (auto & account_from_request : request.get_child ("accounts"))
```
dsiganos, line 964: A plural name (as it was) makes more sense here than a singlular name. This is a list of accounts.

nano/node/json_handler.cpp
```diff
@@ -956,17 +956,26 @@ void nano::json_handler::accounts_create ()
...
 		{
-			auto latest (node.ledger.latest (transaction, account));
+			auto latest = node.ledger.latest (transaction, account);
 			if (!latest.is_zero ())
 			{
 				frontiers.put (account.to_account (), latest.to_string ());
 			}
+			else
+			{
+				frontiers.put (account.to_account (), "");
```
dsiganos: Shouldn't we return an error message here like "account not found or something"?

nano/node/json_handler.cpp
thsfs: this is just one account got from the list of accounts `request.get_child ("accounts")`. The prefix `_from_request` is to state this is the account number got from the request. Notice that the request account number may differ from the decoded account. Example: if you decode `"nano_3e3j5tkog48pnny9dmfzj1r16pg8t1e76dz5tmac6iq689wyjfpiij4txtdo"` it will return `"nano_3e3j5tkog48pnny9dmfzj1r16pg8t1e76dz5tmac6iq689wyjfpiij4txtd1"`. The last characters are corrected to `b1af`. To the response I think it is important to address what was requested, so it keeps the requested number in the response.

nano/node/json_handler.cpp
thsfs: You are right, it should be consistent now.

nano/node/json_handler.cpp
dsiganos: Yes, you are right, sorry. Ignore what I said.

dsiganos:
The "feature" where we correct accounts, sounds like an unnecessary complication that we should remove.
If we get a bad account, we shouldn't try to fix it, we should just return an error.
Where does that autofixing happen?

thsfs:
> The "feature" where we correct accounts, sounds like an unnecessary complication that we should remove. If we get a bad account, we shouldn't try to fix it, we should just return an error. Where does that autofixing happen?

In the `account_impl (...)` function, if you look at the code segment:
```
...
		if (result.decode_account (account_text))
		{
			ec = ec_a;
		}
...
```
The `decode_account` function returns a corrected number.

But, in anyway, this PR is handling it correctly by using the requested account number (`account_from_request`) instead of the corrected one.

clemahieu (dismissed):

dsiganos (dismissed):

thsfs:
Added a new unit test for it.

clemahieu (dismissed):

dsiganos (dismissed):

dsiganos (approved):

PR #3794, opened 2022-04-14
Improve public_key::decode_account to not write result on error

### LINKED ISSUE #3793: Parsing account can write a result, if the result can be autocorrected, even when it signals an error
The function nano::public_key::decode_account has a surprising behavior which is exposed by this little test:
```
TEST (uint256_union, error_correcting)
{
    nano::account key;
    bool result = key.decode_account ("nano_3e3j5tkog48pnny9dmfzj1r16pg8t1e76dz5tmac6iq689wyjfpiij4txtd1");
    std::cerr << "result = " << result << "\n";
    std::cerr << "key = " << key.to_account() << "\n";
}
```

The output of running this test is:
```
result = 1
key = nano_3e3j5tkog48pnny9dmfzj1r16pg8t1e76dz5tmac6iq689wyjfpiij4txtdo
```

Notice that the account parsed finishes in "d1" whereas the variable key finishes in "do".  
Technically, you could say that there is no bug here, because the result of the operation is 1, which is error.  
But this behavior is subtle and surprising and I imagine it is not intended.  
It would be safer for the object to not change value unless the function returns no error.

dsiganos:
public_key::decode_account() would overwrite the object it was called
on error, if the error was with the checksum part of the address.

This commit changes the function to work on a temporary object and
store the result only on success.

resolves  #3793

### DIFF
--- a/nano/core_test/uint256_union.cpp
+++ b/nano/core_test/uint256_union.cpp
@@ -375,6 +375,20 @@ TEST (uint256_union, decode_nano_variant)
 	ASSERT_FALSE (key.decode_account ("nano_1111111111111111111111111111111111111111111111111111hifc8npp"));
 }
 
+/**
+ * It used to be the case that when the address was wrong only in the checksum part
+ * then the decode_account would return error and it would also write the address with
+ * fixed checksum into 'key', which is not desirable.
+ */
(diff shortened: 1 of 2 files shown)


clemahieu (approved):

thsfs:
I think the most important is not changing the account's value.

PR #3808, opened 2022-05-05
Bulk pull ascending

clemahieu:


dsiganos:
The protocol docs and kaitai systax will also need updating.
And possibly the protocol version too.

dsiganos:
This has been superseded by Piotr's version of the ascending bootstrap.

PR #3809, opened 2022-05-05
Bootstrap cleanup

clemahieu:
Removing a number of abstraction holes on nano::bootstrap_attempt that expose implementation details of concrete subtypes.

### DIFF
--- a/nano/node/bootstrap/bootstrap_bulk_pull.hpp
+++ b/nano/node/bootstrap/bootstrap_bulk_pull.hpp
@@ -49,15 +49,16 @@ public:
 	uint64_t unexpected_count;
 	bool network_error{ false };
 };
+class bootstrap_attempt_wallet;
 class bulk_pull_account_client final : public std::enable_shared_from_this<nano::bulk_pull_account_client>
 {
 public:
-	bulk_pull_account_client (std::shared_ptr<nano::bootstrap_client> const &, std::shared_ptr<nano::bootstrap_attempt> const &, nano::account const &);
+	bulk_pull_account_client (std::shared_ptr<nano::bootstrap_client> const &, std::shared_ptr<nano::bootstrap_attempt_wallet> const &, nano::account const &);
 	~bulk_pull_account_client ();
 	void request ();
 	void receive_pending ();
 	std::shared_ptr<nano::bootstrap_client> connection;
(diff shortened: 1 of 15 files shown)


nano/node/bootstrap/bootstrap_bulk_pull.hpp
```diff
@@ -49,15 +49,16 @@ class bulk_pull_client final : public std::enable_shared_from_this<nano::bulk_pu
 	uint64_t unexpected_count;
 	bool network_error{ false };
 };
+class bootstrap_attempt_wallet;
 class bulk_pull_account_client final : public std::enable_shared_from_this<nano::bulk_pull_account_client>
 {
 public:
-	bulk_pull_account_client (std::shared_ptr<nano::bootstrap_client> const &, std::shared_ptr<nano::bootstrap_attempt> const &, nano::account const &);
```
dsiganos, line 55: I may be misunderstanding this but it seems strange to me to limit the bulk pull to a wallet bootstrap attempt only.

nano/node/bootstrap/bootstrap_lazy.cpp
```diff
@@ -489,7 +488,7 @@ void nano::bootstrap_attempt_wallet::request_pending (nano::unique_lock<nano::mu
 		auto account (wallet_accounts.front ());
 		wallet_accounts.pop_front ();
 		++pulling;
-		auto this_l (shared_from_this ());
```
dsiganos, line 492: Is this correct? Why is bootstrap_attempt_lazy converting to bootstrap_attempt_wallet, it seems strange.

thsfs (approved):

nano/node/bootstrap/bootstrap_lazy.cpp
clemahieu: It's in the bootstrap_lazy file but this is inside a bootstrap_attempt_wallet method: nano::bootstrap_attempt_wallet::request_pending

nano/node/bootstrap/bootstrap_bulk_pull.hpp
clemahieu: This is a bulk_pull_account call which is a different network messages from bulk_pull. It is only used by bootstrap_attempt_wallet.

Issue #12, opened 2016-05-01
Add Linux repositories


otta88sun:
To install wallet with somethink like: sudo apt-get install raiblocks-core

rkeene:
We should support CentOS/RHEL 7, Ubuntu LTS, and Debian with our repositories.  These should mirror the artifacts we produce for Docker.

rkeene:
To start with, and to satisfy #1091 we should create a CentOS 7 repo first and support V15.2 on it.

systemd unit files and packaging guides for rai_node already exist that we can adopt.

argakiig:
V16.0 will have a RPM package for RHEL systemd systems.
A new user `nanocurrency` is created with a home directory of /var/nanocurrency
Daemon will be installed to /usr/bin and a unit file for nanocurrency will be installed but not enabled or started.
More information as things progress

argakiig:
- [ ] Source tarball
- [ ] RPM build
- [ ] DEB build
- [ ] Repositories created.
- [ ] Automation of package creation

Ethunxxx:
Would it make more sense to create a Nano flatpak instead? This could be used on almost all Linux systems. See #1129.

DarrenAlex:
Since the recent mention of a plug-and-play node, I think adding an apt installation method would be really easy.

argakiig:
the deb artifact is included in releases now as well, its just a matter of backing a repo for it now

argakiig:
> Since the recent mention of a plug-and-play node, I think adding an apt installation method would be really easy.

are you meaning something modular as in separating the wallet and node from each other? like with the external rpc server thats bundled with release artifacts? I should update the documentation to reflect the steps to install via deb though as well for example 
```bash
curl -sLO https://github.com/nanocurrency/nano-node/releases/download/V22.1/nano-node-V22.1-Linux.deb && \
sudo apt install ./nano-node-V22.1-Linux.deb && \
sudo systemctl status nanocurrency
```
this install adds a user nanocurrency and creates a group nanocurrency and service for nanocurrency running as nanocurrency in the nanocurrency home path, and then starts the service

zhyatt:
@argakiig What is left here? Just the docs update for apt install mentioned above?

argakiig:
deb repo still needs to be stood up

PR #3810, opened 2022-05-05
Adds RPM artifact generation to GitHub CI

thsfs:
Adds RPM artifacts (source and binary) generation. Reference issue: https://github.com/nanocurrency/nano-node/issues/12

**Note:** needs to be tested in the main repository.

### DIFF
--- a/.github/workflows/live_artifacts.yml
+++ b/.github/workflows/live_artifacts.yml
@@ -70,9 +70,10 @@ jobs:
     runs-on: ubuntu-20.04
     timeout-minutes: 90
     steps:
-      - name: Checkout the Tag
+      - name: Sets the tag and repo variables (to build)
         run: |
           echo "TAG=${{ github.event.inputs.ref }}" >> $GITHUB_ENV
+          echo "REPO_TO_BUILD=${{ github.event.inputs.repo }}" >> $GITHUB_ENV
       - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
         with:
           submodules: "recursive"
@@ -84,7 +85,8 @@ jobs:
         run: |
           mkdir -p ${GITHUB_WORKSPACE}/artifacts
           docker run -e LIVE=1 -e PAT=${{ secrets.GITHUB_TOKEN }} \
-          -e TAG=$TAG -v ${GITHUB_WORKSPACE}:/workspace \
+          -e TAG=$TAG -e REPO_TO_BUILD=$REPO_TO_BUILD
+          -v ${GITHUB_WORKSPACE}:/workspace \
           -v ${GITHUB_WORKSPACE}/artifacts:/root/rpmbuild \
           nanocurrency/nano-env:centos /bin/bash -c "ci/build-centos.sh"
       - name: Deploy Artifacts
--- a/.github/workflows/live_artifacts.yml
+++ b/.github/workflows/live_artifacts.yml
@@ -92,6 +92,7 @@ jobs:
         run: ci/actions/deploy.sh
         env:
           LINUX_RPM: 1
+          S3_BUILD_DIRECTORY: ${{ secrets.S3_BUILD_DIRECTORY }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: us-east-2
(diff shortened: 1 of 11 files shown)


util/changelog.py
```diff
@@ -1,18 +1,35 @@
...
+from typing import Tuple
 
 """
 Changelog generation script, requires PAT with public_repo access, 
 see https://github.com/settings/tokens
 
-Caveats V20 and prior release tags are tips on their respective release branches
-If you try to use a start tag with one of these a full changelog will be generated
-since the commit wont appear in your iterations
+usage: changelog [-h] [-e END] [-m {full,incremental}] -p PAT [-r REPO] [-s START] [-t TAG]
```
thsfs, line 11: @zhyatt , needs your careful review on these changes. This script seemed to be broken in one of my runs. It didn't work for getting the changelogs that would be added to the RPMs. Where else is this script used on?

zhyatt: As we discussed on our call, testing this command locally with `./util/changelog.py --start [hash] --end [hash] --pat [personal access token]` should help determine if there is an issue with the script or the way the CI is calling it. Determining the start hash is a bit tricky due to the way we branch for a release from develop.

To get the start hash, from the previous release branch (so for V23, this would be `git checkout releases/v22`), run this command: `diff -u <(git rev-list --first-parent "${1:-develop}") <(git rev-list --first-parent "${2:-HEAD}") | sed -ne "s/^ //p" | head -1`.

thsfs: https://git-scm.com/docs/git-merge-base
using with --octopus seems to do the same for finding the common ancestor

thsfs: BTW, I'll remove the change log generation from this PR and create new workflow for it.

util/makesrc
```diff
@@ -1,25 +1,26 @@
...
 
-#    makesrc <tag>
-#    requires ruby gem github_changelog_generator
-#    API_KEY generated here https://github.com/settings/tokens/new?description=GitHub%20Changelog%20Generator%20token
+#    makesrc <tag> <pat> 
+#    personal access token allows around rate limit
 
-if [ -e $1 ]; then
-	echo "makesrc <tag>" >&2
+if [ -e $2 ]; then
```
zhyatt: If we remove the changelog section below, we should probably clean up all the $PAT related items too

.github/workflows/live_artifacts.yml
```diff
@@ -64,6 +64,36 @@ jobs:
...
+          repository: ${{ github.repository }}
+      - name: CentOS Base
+        run: ci/build-docker-image.sh docker/ci/Dockerfile-centos nanocurrency/nano-env:centos
+      - name: Build Artifact
+        run: |
+          mkdir -p ${GITHUB_WORKSPACE}/artifacts
+          docker run -e LIVE=1 -e TAG=$TAG -e REPO_TO_BUILD=$REPO_TO_BUILD -e RPM_RELEASE=1 \
+          -v ${GITHUB_WORKSPACE}:/workspace \
+          -v ${GITHUB_WORKSPACE}/artifacts:/root/rpmbuild \
+          nanocurrency/nano-env:centos /bin/bash -c "ci/build-centos.sh"
```
zhyatt, line 90: @thsfs Just noting that the error I hit is from this `ci/build-centos.sh` script I believe and can be seen here: https://github.com/zhyatt/nano-node/runs/6675034868?check_suite_focus=true#step:5:6085

thsfs: Discovered that using dash is a limitation of the version number regarding the RPM naming convention.
> The version number is normally taken verbatim from the package's version. The only restriction placed on the version is that it cannot contain a dash "-". 

http://ftp.rpm.org/max-rpm/ch-rpm-file-format.html

Worked around this by force converting the dashes to underscores.

.github/workflows/live_artifacts.yml
```diff
@@ -66,6 +66,36 @@ jobs:
...
+      - name: Deploy Artifacts
+        run: ci/actions/deploy.sh
+        env:
+          LINUX_RPM: 1
```
zhyatt, line 94: Due to the changes I made to allow deploying to internal, this needs to be added to the `env:` section:
`S3_BUILD_DIRECTORY: ${{ secrets.S3_BUILD_DIRECTORY }}`
so
```
...
env:
  S3_BUILD_DIRECTORY: ${{ secrets.S3_BUILD_DIRECTORY }}
  LINUX_RPM: 1
...
```

zhyatt (dismissed):

.github/workflows/live_artifacts.yml
thsfs: Added.

dsiganos (approved):

PR #3811, opened 2022-05-06
Adding unhandled case for message_visitor

clemahieu:
Add unhandled case for message_visitor which all messages route to by default. This simplifies implementing new message handlers.

### DIFF
--- a/nano/node/transport/inproc.cpp
+++ b/nano/node/transport/inproc.cpp
@@ -41,47 +41,7 @@ public:
 	// the channel to reply to, if a reply is generated
 	std::shared_ptr<nano::transport::inproc::channel> channel;
 
-	void keepalive (nano::keepalive const & message)
-	{
-		inbound (message, channel);
-	}
-	void publish (nano::publish const & message)
-	{
-		inbound (message, channel);
-	}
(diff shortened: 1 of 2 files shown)


nano/node/transport/inproc.cpp
```diff
@@ -41,47 +41,7 @@ class message_visitor_inbound : public nano::message_visitor
...
-	void node_id_handshake (nano::node_id_handshake const & message)
-	{
-		inbound (message, channel);
-	}
-	void telemetry_req (nano::telemetry_req const & message)
-	{
-		inbound (message, channel);
-	}
-	void telemetry_ack (nano::telemetry_ack const & message)
+	void unhandled (nano::message const & message) override
```
dsiganos: I would call it default handler rather than unhandled.
But other than that, it looks good to me.

dsiganos (dismissed):

thsfs (dismissed):

nano/node/transport/inproc.cpp
clemahieu: Fixed.

dsiganos (approved):

PR #3812, opened 2022-05-06
Extracting nano::bootstrap::block_deserializer class

clemahieu:
The code to deserialize type-prefixed blocks from a stream was tightly coupled to the legacy bootstrap process. This prevented direct testing or reuse of block deserialization code.

Rewriting existing bulk_pull_client in terms of the new block deserializer.

### DIFF
--- a/nano/node/bootstrap/block_deserializer.cpp
+++ b/nano/node/bootstrap/block_deserializer.cpp
@@ -1,0 +1,65 @@
+#include <nano/lib/blocks.hpp>
+#include <nano/node/bootstrap/block_deserializer.hpp>
+#include <nano/node/socket.hpp>
+#include <nano/secure/buffer.hpp>
+
+nano::bootstrap::block_deserializer::block_deserializer () :
+	read_buffer{ std::make_shared<std::vector<uint8_t>> () }
+{
+}
+
+void nano::bootstrap::block_deserializer::read (nano::socket & socket, callback_type const && callback)
+{
+	debug_assert (callback);
+	read_buffer->resize (1);
+	socket.async_read (read_buffer, 1, [this_l = shared_from_this (), &socket, callback = std::move (callback)] (boost::system::error_code const & ec, std::size_t size_a) {
+		if (ec)
+		{
+			callback (ec, nullptr);
+			return;
+		}
+		if (size_a != 1)
+		{
+			callback (boost::asio::error::fault, nullptr);
+			return;
+		}
+		this_l->received_type (socket, std::move (callback));
+	});
+}
+
+void nano::bootstrap::block_deserializer::received_type (nano::socket & socket, callback_type const && callback)
+{
+	nano::block_type type = static_cast<nano::block_type> (read_buffer->data ()[0]);
+	if (type == nano::block_type::not_a_block)
+	{
+		callback (boost::system::error_code{}, nullptr);
+		return;
+	}
+	auto size = nano::block::size (type);
+	if (size == 0)
+	{
+		callback (boost::asio::error::fault, nullptr);
+		return;
+	}
+	read_buffer->resize (size);
+	socket.async_read (read_buffer, size, [this_l = shared_from_this (), size, type, callback = std::move (callback)] (boost::system::error_code const & ec, std::size_t size_a) {
+		if (ec)
+		{
+			callback (ec, nullptr);
+			return;
+		}
+		if (size_a != size)
+		{
+			callback (boost::asio::error::fault, nullptr);
+			return;
+		}
+		this_l->received_block (type, std::move (callback));
+	});
+}
+
+void nano::bootstrap::block_deserializer::received_block (nano::block_type type, callback_type const && callback)
+{
+	nano::bufferstream stream{ read_buffer->data (), read_buffer->size () };
+	auto block = nano::deserialize_block (stream, type);
+	callback (boost::system::error_code{}, block);
+}
--- a/nano/node/bootstrap/block_deserializer.hpp
+++ b/nano/node/bootstrap/block_deserializer.hpp
@@ -1,0 +1,47 @@
+#pragma once
+
+#include <nano/lib/blocks.hpp>
+
+#include <boost/system/error_code.hpp>
+
+#include <memory>
+#include <vector>
+
+namespace nano
+{
+class block;
+class socket;
+namespace bootstrap
+{
+	/**
+	 * Class to read a block-type byte followed by a serialised block from a stream.
+	 * It is typically used to used to read a series of block-types and blocks terminated by a not-a-block type.
+	 */
+	class block_deserializer : public std::enable_shared_from_this<nano::bootstrap::block_deserializer>
+	{
+	public:
+		using callback_type = std::function<void (boost::system::error_code, std::shared_ptr<nano::block>)>;
+
+		block_deserializer ();
+		/**
+		 * Read a type-prefixed block from 'socket' and pass the result, or an error, to 'callback'
+		 * A normal end to series of blocks is a marked by return no error and a nullptr for block.
+		 */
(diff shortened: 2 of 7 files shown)


nano/node/bootstrap/block_deserializer.hpp
```diff
@@ -0,0 +1,32 @@
...
+public:
+	block_deserializer ();
+	void read (std::shared_ptr<nano::socket> socket);
+	void reset ();
+	std::shared_ptr<nano::block> block;
+	boost::system::error_code ec;
+private:
+	void received_type ();
+	void received_block (nano::block_type type);
+	size_t block_size (nano::block_type type);
```
thsfs: We prefer `std::size_t`, right?

nano/node/bootstrap/block_deserializer.cpp
```diff
@@ -11,48 +11,56 @@ nano::bootstrap::block_deserializer::block_deserializer () :
 
 }
 
-void nano::bootstrap::block_deserializer::read (std::shared_ptr<nano::socket> socket)
+void nano::bootstrap::block_deserializer::read (std::shared_ptr<nano::socket> socket, callback_type callback)
 {
+	debug_assert (socket != nullptr);
```
thsfs: if we had the reference `nano::socket & socket` we won't need to assert `socket != nullptr`. The same works for the `callback`, but the function couldn't be placed inline to the `read` function.

nano/node/bootstrap/block_deserializer.cpp
```diff
@@ -0,0 +1,84 @@
...
+	read_buffer{ std::make_shared<std::vector<uint8_t>> () }
+{
+}
+
+void nano::bootstrap::block_deserializer::read (nano::socket & socket, callback_type const && callback)
+{
+	debug_assert (callback);
+	read_buffer->resize (1);
+	socket.async_read (read_buffer, 1, [this_l = shared_from_this (), &socket, callback = std::move (callback)] (boost::system::error_code const & ec, std::size_t size_a) {
+		if (size_a != 1)
```
dsiganos: Shouldn't we check the error code first before checking anything else?
check ec
check size_a

nano/node/bootstrap/block_deserializer.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+void nano::bootstrap::block_deserializer::received_block (nano::block_type type, callback_type const && callback)
+{
+	nano::bufferstream stream{ read_buffer->data (), read_buffer->size () };
+	auto block = nano::deserialize_block (stream, type);
+	callback (boost::system::error_code{}, block);
+}
+
+std::size_t nano::bootstrap::block_deserializer::block_size (nano::block_type type)
```
dsiganos: We already have the function nano::block::size().
We just have to ensure that the block passed to that is not not_a_block nor invalid.

nano/node/bootstrap/block_deserializer.cpp
```diff
@@ -0,0 +1,84 @@
...
+	}
+	auto size = block_size (type);
+	if (size == 0)
+	{
+		callback (boost::asio::error::fault, nullptr);
+		return;
+	}
+	read_buffer->resize (size);
+	socket.async_read (read_buffer, size, [this_l = shared_from_this (), size, type, callback = std::move (callback)] (boost::system::error_code const & ec, std::size_t size_a) {
+		if (size_a != size)
```
dsiganos: Shouldn't we check the error code first before checking anything else?
check ec
check size_a

nano/node/bootstrap/bootstrap_bulk_pull.cpp
```diff
@@ -1,3 +1,4 @@
+#include <nano/node/bootstrap/block_deserializer.hpp>
```
dsiganos, line 1: It is not easy to review the changes in this file because it is not easily obvious to me how the internal operation bulk_pull_client. This is a good opportunity to add some comments to explain what it does. For exmaple, it would be good to comment the following variables:
```
	nano::block_hash expected;
	nano::account known_account;
	nano::pull_info pull;
	uint64_t pull_blocks;
	uint64_t unexpected_count;
```

nano/node/bootstrap/bootstrap_bulk_pull.cpp
```diff
@@ -121,174 +120,98 @@ void nano::bulk_pull_client::throttled_receive_block ()
...
-			if (connection->node->config.logging.bulk_pull_logging ())
-			{
-				connection->node->logger.try_log ("Error deserializing block received from pull request");
-			}
-			connection->node->stats.inc (nano::stat::type::bootstrap, nano::stat::detail::bulk_pull_deserialize_receive_block, nano::stat::dir::in);
+			connection->connections.pool_connection (connection);
 		}
-		else // Work invalid
+	}
+	else if (block == nullptr)
```
thsfs: This seems to be an unreachable code. Please check the line 135 drives the function to return.

nano/node/bootstrap/bootstrap_bulk_pull.cpp
```diff
@@ -121,174 +120,98 @@ void nano::bulk_pull_client::throttled_receive_block ()
...
-			{
-				connection->node->logger.try_log (boost::str (boost::format ("Insufficient work for bulk pull block: %1%") % block->hash ().to_string ()));
-			}
-			connection->node->stats.inc_detail_only (nano::stat::type::error, nano::stat::detail::insufficient_work);
+			connection->node->logger.try_log ("Error deserializing block received from pull request");
 		}
+		connection->node->stats.inc (nano::stat::type::bootstrap, nano::stat::detail::bulk_pull_deserialize_receive_block, nano::stat::dir::in);
 	}
-	else
+	else // Work invalid
```
thsfs: If you move the statement block from lines 194-200 to the if at line 135, this else can be implemented as a guard clause, with a return to its end. The remaining code would be the case for the condition `if (!connection->node->network_params.work.validate_entry (*block))`.

thsfs: If you move the statement block from lines 194-200 to the if at line 135, this else can be implemented as a guard clause, with a return to its end. The remaining code would be the case for the condition `if (!connection->node->network_params.work.validate_entry (*block))`.

nano/node/bootstrap/bootstrap_bulk_pull.cpp
```diff
@@ -121,174 +120,98 @@ void nano::bulk_pull_client::throttled_receive_block ()
...
-			break;
+			connection->connections.pool_connection (connection);
 		}
-		case nano::block_type::receive:
+		return;
+	}
+	if (!connection->node->network_params.work.validate_entry (*block))
+	{
+		auto hash = block->hash ();
+		//std::cerr << "Deserialized: " << hash.to_string () << std::endl;
```
thsfs: Can be removed.

thsfs: Can be removed.

thsfs (dismissed):

nano/node/bootstrap/bootstrap_bulk_pull.cpp
```diff
@@ -276,8 +200,7 @@ void nano::bulk_pull_client::received_block (boost::system::error_code const & e
 nano::bulk_pull_account_client::bulk_pull_account_client (std::shared_ptr<nano::bootstrap_client> const & connection_a, std::shared_ptr<nano::bootstrap_attempt_wallet> const & attempt_a, nano::account const & account_a) :
 	connection (connection_a),
 	attempt (attempt_a),
-	account (account_a),
-	pull_blocks (0)
+	account (account_a)
```
pwojcikdev: This line removal is causing problems 😬

pwojcikdev (approved):

PR #3813, opened 2022-05-09
Remove vote-by-block support from vote class

clemahieu:
The existing implementation was problematic, modeled as a vector-of-variants rather than a variant of a block and hash vector. It also disallowed behavior that can be correctly encoded, such as empty vote-by-hash votes.

### DIFF
--- a/nano/secure/common.cpp
+++ b/nano/secure/common.cpp
@@ -451,38 +451,7 @@ nano::block_info::block_info (nano::account const & account_a, nano::amount cons
 
 bool nano::vote::operator== (nano::vote const & other_a) const
 {
-	auto blocks_equal (true);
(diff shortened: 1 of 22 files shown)


nano/secure/common.cpp
```diff
@@ -694,74 +656,39 @@ void nano::vote::serialize (nano::stream & stream_a, nano::block_type type) cons
...
 	}
 	catch (std::runtime_error const &)
 	{
-		error = true;
-	}
-
-	if (blocks.empty ())
-	{
-		error = true;
+		result = true;
```
thsfs: if the result means that there was an error, IMO it is easier to understand if it remains named as error.

thsfs (approved):

PR #3816, opened 2022-05-12
Node ID persistence

dsiganos:
After this change, node IDs will persist on the disk file system and
reused. The node ID will not be changing after every restart.
This will help to more reliably track PRs.

The node can be forced to generate a new ID by deleting the file:
node_id_private.key

Note that this change does not prevent the man in the middle attack.

However, it makes it possible to detect telemetry spoofing. Sites like
nanolooker and nanoticker can maintain associations between node ID and PR
and only accept a telemetry packet, if the telemetry message is signed
by the known Node ID.

### DIFF
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <cstdlib>
+#include <fstream>
 #include <future>
 #include <sstream>
 
--- a/nano/node/node.cpp
+++ b/nano/node/node.cpp
@@ -74,6 +74,33 @@ std::unique_ptr<nano::container_info_component> nano::collect_container_info (re
 	return composite;
 }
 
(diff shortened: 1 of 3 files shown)


nano/node/node.cpp
```diff
@@ -74,6 +75,33 @@ std::unique_ptr<nano::container_info_component> nano::collect_container_info (re
...
+		ifs >> node_private_key;
+		release_assert (node_private_key.size () == 64);
+		nano::keypair kp = nano::keypair (node_private_key);
+		return kp;
+	}
+	else
+	{
+		// no node_id found, generate new one
+		logger.always_log (boost::str (boost::format ("%1% does not exist, creating a new node_id") % node_private_key_path.string ()));
+		nano::keypair kp = nano::keypair ();
```
clemahieu: This can be just `nano::keypair kp;`

clemahieu (dismissed):

thsfs (approved):

PR #3819, opened 2022-05-16
Adds a changelog workflow

thsfs:
- Brings part of the updates made by Russel (https://github.com/argakiig/raiblocks/tree/rpm_building) in the `changelog.py` script, but removed the tag option, as it can be passed to the end option (-e).
- Adds bash scripts to generate and to deploy the changelog file.
- Adds a GitHub workflow.


ci/actions/deploy-changelog.sh
```diff
@@ -0,0 +1,16 @@
...
+set -o xtrace
+
+TAG="$1"
+
+if [[ -z "$TAG" ]]; then
+  echo "The TAG argument is required"
+  exit 1
+fi
+
+sha256sum $GITHUB_WORKSPACE/artifacts/CHANGELOG.md > $GITHUB_WORKSPACE/artifacts/CHANGELOG.md.sha256
```
zhyatt: I don't think a sha256 is useful here. What was the reason this was included?

zhyatt (changes requested):

ci/actions/deploy-changelog.sh
thsfs: Not really, it just replicates the data we already have at GitHub, easily verifiable, so you are right, there is no need for it.

util/changelog.py
```diff
@@ -80,114 +90,88 @@
...
+            if commit.sha == self.start:
                 break
-            else:
-                message = commit.commit.message.partition('\n')[0]
-                try:
-                    pr_number = int(
-                        message[message.rfind('#')+1:message.rfind(')')])
+            m = commit.commit.message.partition('\n')[0]
+            try:
+                pr_number = int(m[m.rfind('#')+1:m.rfind(')')])
```
zhyatt: I ran this for tag `V23.1` and made sure `V23.0` was also present as a tag on my own fork. Other tags may not exist on my repository, but I'm not sure they should be relevant. An error was hit on this line, see here: https://github.com/zhyatt/nano-node/runs/6822477580?check_suite_focus=true#step:4:332

zhyatt (dismissed):

util/changelog.py
```diff
@@ -80,114 +90,88 @@
...
+                pull = self.repo.get_pull(pr_number)
+            except (ValueError, UnknownObjectException):
+                p = commit.get_pulls()
+                if p.totalCount > 0:
+                    pr_number = p[0].number
                     pull = self.repo.get_pull(pr_number)
```
zhyatt, line 244: @thsfs I found an edge case where if a commit was added directly to the branch and then later included in an open PR somewhere, that PR would still get added into the changelog when it shouldn't. 

To fix this, ensuring whatever pull is found is not in an open state works:

```
pull = self.repo.get_pull(pr_number)
if pull.state == 'open':
                        print(
                            f"Commit is in tree but only associated with open PR {pr_number}: \"{pull.title}\"")
                        self.other_commits.append((commit.sha, m))
                        continue
                        ```

util/changelog.py
thsfs: This happened because the built repository doesn't contain the PRs the scripts asks for.

util/changelog.py
thsfs: Will open this as an issue to improve the `changelog.py`.

PR #3820, opened 2022-05-16
Splits the deploy-docker.sh to have build/deploy parts

thsfs:


.github/workflows/beta_artifacts.yml
```diff
@@ -81,8 +81,12 @@ jobs:
         env:
           COMPILER: gcc
         run: ci/actions/linux/install_deps.sh
+      - name: Build Docker (nanocurrency/nano-beta)
+        run: TRAVIS_TAG=${TAG} ci/actions/linux/docker-build.sh
+        env:
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
zhyatt: The build process doesn't need the DOCKER_PASSWORD

zhyatt (changes requested):
@thsfs  If you can make the requested changes, I will pull this for further testing

.github/workflows/beta_artifacts.yml
thsfs: It should be ok now.

.github/workflows/develop.yml
```diff
@@ -22,6 +22,6 @@ jobs:
       - name: Deploy Docker (ghcr.io)
         run: ci/actions/linux/ghcr_push.sh
       - name: Deploy Docker (nanocurrency/nano-env)
-        run: ci/actions/linux/deploy-docker.sh
+        run: ci/actions/linux/docker-deploy.sh
```
zhyatt, line 25: @thsfs There is no action here on this PR, but this is a note that in order to provide more repeatable builds for Docker, we may want to revisit this setup in the future. This is because the "Develop" workflow runs every time there is a push to the `develop` branch, and as part of that workflow the `-env` Docker images are updated. But this will continue after we've branched off for a release (`releases/v23` for example) so by the time we go to do a release (and in the future after the release) the base `-env` builds will keep changing and won't be static.

zhyatt (approved):

.github/workflows/develop.yml
thsfs: Did a review on the workflows before opening another issue for this. The `-env` docker images are generated always from the docker files, saved locally in the CI VM, then pushed with different tags for develop or the other release runs. The `docker pull` command from the `docker-build-image.sh` seems to be just a mechanism to enable speedups when there is something already cached.

PR #3821, opened 2022-05-17
Fix request loop unblocking during continuous inflood of votes

Exxenoz:
In case there is an continuous inflood of votes, is_active often does not equal false long enough to unblock the request loop thread which is then stuck in flush_active for multiple seconds or even minutes. This can be solved by only checking once for is_active in flush_active and unblocking the request loop thread by a separate flush notification.


KaffinPX:
Should be merged asap.

dsiganos:
Fixed by
https://github.com/nanocurrency/nano-node/commit/5b9eb588e65911fda4cd17719bc3d81f9ef2ac7b

PR #3826, opened 2022-05-19
Adding tests and explanations demonstrating rejection of zero-account votes.

clemahieu:


nano/core_test/vote_processor.cpp
```diff
@@ -343,3 +343,47 @@ TEST (vote, empty_hashes)
...
+{
+	nano::raw_key junk_key{ 0 }; // A signature is made by nano::vote from this key but is overwritten by direct signature deserialization
+	uint64_t junk_timestamp = 0;
+	std::vector<nano::block_hash> junk_hashes;
+
+	// Changing this value will cause the vote to never validate
+	nano::account account{ 0 };
+	// Chosen to make the vote validate
+	// Changing this will make the vote sometimes validate
+	uint8_t chosen_duration = 3;
```
dsiganos, line 357: What is the significance of this value, 3?
Why not set it to max timestamp?

dsiganos (approved):

PR #3835, opened 2022-06-08
Unchecked memory

clemahieu:
This change puts unchecked blocks in to a memory container instead of disk when the initial bootstrap threshold is reached. Blocks are removed from the container in fifo order once the maximum of 256,000 blocks are reached.

### DIFF
--- a/nano/secure/store/unchecked_store_partial.hpp
+++ b/nano/secure/store/unchecked_store_partial.hpp
@@ -87,19 +87,9 @@ public:
 
 	void put (nano::write_transaction const & transaction_a, nano::block_hash const & hash_a, std::shared_ptr<nano::block> const & block_a) override
 	{
-		nano::lock_guard<std::recursive_mutex> lock{ mutex };
-		if (entries == nullptr)
-		{
-			nano::unchecked_key key (hash_a, block_a->hash ());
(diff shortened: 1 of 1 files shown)


pwojcikdev:
Fails build on linux, needs `#include <boost/multi_index_container.hpp>` in `nano/secure/store/unchecked_store_partial.hpp`. Other than that it looks good to me.

nano/secure/store/unchecked_store_partial.hpp
```diff
@@ -19,81 +26,188 @@ void release_assert_success (store_partial<Val, Derived_Store> const &, int cons
...
+		}
 	}
 
 	void put (nano::write_transaction const & transaction_a, nano::block_hash const & hash_a, std::shared_ptr<nano::block> const & block_a) override
 	{
-		nano::unchecked_key key (hash_a, block_a->hash ());
-		nano::unchecked_info info (block_a, block_a->account (), nano::seconds_since_epoch (), nano::signature_verification::unknown);
-		put (transaction_a, key, info);
+		nano::lock_guard<std::recursive_mutex> lock{ mutex };
+		if (entries == nullptr)
```
dsiganos: Both sides of this if statement appear to be identical.

dsiganos (approved):

dsiganos (approved):

Issue #3659, opened 2022-01-07
Make silent the overflow report for xorshift1024star calculation


thsfs:
Got an UBSan error report from Address Sanitizer when running it upon core_test.

```
/ws/nano-node/nano/node/xorshift.hpp:18:12: runtime error: left shift of 3860899979500397099 by 31 places cannot be represented in type 'uint64_t' (aka 'unsigned long')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /ws/nano-node/nano/node/xorshift.hpp:18:12 in 
```
The overflow is expected as part of the calculation, but this should ideally not trigger an error.

Code portion from which the complain comes from:
```
class xorshift1024star final
{
public:
	std::array<uint64_t, 16> s;
	unsigned p{ 0 };
	uint64_t next ()
	{
		auto p_l (p);
		auto pn ((p_l + 1) & 15);
		p = pn;
		uint64_t s0 = s[p_l];
		uint64_t s1 = s[pn];
		s1 ^= s1 << 31; // a
		s1 ^= s1 >> 11; // b
		s0 ^= s0 >> 30; // c
		return (s[pn] = s0 ^ s1) * 1181783497276652981LL;
	}
};
```
OS: Ubuntu 20.04

Issue #3658, opened 2022-01-07
Make silent the overflow report for uint256_t hash calculator (UBSan report)


thsfs:
Got an UBSan error report from Address Sanitizer when running it upon core_test.

```
/ws/nano-node/nano/lib/numbers.hpp:278:46: runtime error: unsigned integer overflow: 18363490017020192492 + 15047902047331995846 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /ws/nano-node/nano/lib/numbers.hpp:278:46 in 
```

The overflow is expected as part of the hash calculation, but this should ideally not trigger an error.

Code portion from which the complain comes from:
```
template <>
struct hash<::nano::uint256_union>
{
	size_t operator() (::nano::uint256_union const & data_a) const
	{
		return data_a.qwords[0] + data_a.qwords[1] + data_a.qwords[2] + data_a.qwords[3];
	}
};
```
OS: Ubuntu 20.04

PR #3844, opened 2022-06-20
Ignore asan warnings (clang)

ctummon:
Wanted to test out my WSL dev environment with Visual Code so took a crack at some of the address sanitizer warnings

1.) **Added  function attribute to ignore false unsigned int overflow warnings which fixes:**
https://github.com/nanocurrency/nano-node/issues/3658
https://github.com/nanocurrency/nano-node/issues/3659

2.) **Added blacklist file for Clang Address Sanitizer.**

Also when running Address Sanitizer with core_test aborts in a couple of places:

```
config_a.network_params.network.is_dev_network () 
nano::confirmation_solicitor::{ctor}
nano::active_transactions::request_confirm()
```

```
config.network_params.network.is_dev_network ()
void nano::node::ongoing_backlog_population ()
```

It doesn't like that network_params is a reference and it is being accessed on different threads.

Is this a known issue or should I create a new issue?


asan_blacklist
```diff
@@ -1 +0,0 @@
-src:*ed25519*
```
ctummon, line 1: I don't think this file was used, correct me if I'm wrong

CMakeLists.txt
```diff
@@ -247,11 +247,25 @@ else()
   endif()
 
   if(${USING_ASAN} OR ${USING_ASAN_INT})
+
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
ctummon, line 251: -fsanitize-blacklist and -fsanitize=integer only seem to be available with Clang so I added some checks

ctummon:
OSX (RocksDB) is failing on core_test: error code '124' after "active_transactions.fork_filter_cleanup" looks like a false positive?

Same error for the code of conduct PR:
https://github.com/nanocurrency/nano-node/runs/6762524136?check_suite_focus=true

PR #3850, opened 2022-07-09
Fix rocksdb_block_store.tombstone_count

clemahieu:
With the addition of storing unchecked entries in memory, the unchecked_map class doesn't give guarantees that entries are written to disk which causes the rocksdb-specific tombstone checking to fail.

This rewrites the test directly in terms of the unchecked_store rather than unchecked_map.

### DIFF
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -2645,34 +2645,36 @@ namespace nano
 // logic bound to the way RocksDB is used by the node.
 TEST (rocksdb_block_store, tombstone_count)
 {
-	if (nano::rocksdb_config::using_rocksdb_in_tests ())
+	if (!nano::rocksdb_config::using_rocksdb_in_tests ())
 	{
-		nano::system system{};
-		nano::logger_mt logger{};
-		auto store = std::make_unique<nano::rocksdb::store> (logger, nano::unique_path (), nano::dev::constants);
-		nano::unchecked_map unchecked{ *store, false };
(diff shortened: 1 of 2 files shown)


nano/core_test/block_store.cpp
```diff
@@ -2645,34 +2645,36 @@ namespace nano
...
-		auto check_block_is_listed = [&] (nano::transaction const & transaction_a, nano::block_hash const & block_hash_a) {
-			return unchecked.get (transaction_a, block_hash_a).size () > 0;
-		};
-		// Waits for the block to get saved
-		ASSERT_TIMELY (5s, check_block_is_listed (store->tx_begin_read (), block->previous ()));
-		ASSERT_EQ (store->tombstone_map.at (nano::tables::unchecked).num_since_last_flush.load (), 0);
-		// Perorms a delete and checks for the tombstone counter
-		unchecked.del (store->tx_begin_write (), nano::unchecked_key (block->previous (), block->hash ()));
-		ASSERT_EQ (store->tombstone_map.at (nano::tables::unchecked).num_since_last_flush.load (), 1);
+		return;
```
theohax, line 2650: How about GTEST_SKIP instead to have the report show this as skipped instead of passed?

theohax (approved):

nano/core_test/block_store.cpp
clemahieu: Tracked with https://github.com/nanocurrency/nano-node/issues/3854

PR #3852, opened 2022-07-09
Expands ability for ledger::is_send to accept any block type

clemahieu:
Expands ability for ledger::is_send to accept any block type, not just state blocks.

Two new ledger contexts generate a send/receive pair on the genesis chain with state and legacy blocks.

### DIFF
--- a/nano/core_test/ledger.cpp
+++ b/nano/core_test/ledger.cpp
@@ -5646,3 +5646,32 @@ TEST (ledger, unconfirmed_frontiers)
 	ASSERT_EQ (uncemented_info1.cemented_frontier, uncemented_info2.cemented_frontier);
 	ASSERT_EQ (uncemented_info1.frontier, uncemented_info2.frontier);
 }
+
+TEST (ledger, is_send_genesis)
+{
+	auto ctx = nano::test::context::ledger_empty ();
+	auto & ledger = ctx.ledger ();
+	auto & store = ctx.store ();
+	auto tx = store.tx_begin_read ();
+	ASSERT_FALSE (ledger.is_send (tx, *nano::dev::genesis));
+}
+
+TEST (ledger, is_send_state)
+{
+	auto ctx = nano::test::context::ledger_send_receive ();
+	auto & ledger = ctx.ledger ();
+	auto & store = ctx.store ();
+	auto tx = store.tx_begin_read ();
+	ASSERT_TRUE (ledger.is_send (tx, *ctx.blocks ()[0]));
+	ASSERT_FALSE (ledger.is_send (tx, *ctx.blocks ()[1]));
+}
+
+TEST (ledger, is_send_legacy)
+{
+	auto ctx = nano::test::context::ledger_send_receive_legacy ();
+	auto & ledger = ctx.ledger ();
+	auto & store = ctx.store ();
+	auto tx = store.tx_begin_read ();
+	ASSERT_TRUE (ledger.is_send (tx, *ctx.blocks ()[0]));
+	ASSERT_FALSE (ledger.is_send (tx, *ctx.blocks ()[1]));
+}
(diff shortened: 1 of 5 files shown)


clemahieu:
Breaking this in to smaller commits

nano/secure/ledger.cpp
```diff
@@ -939,10 +944,9 @@ bool nano::ledger::is_send (nano::transaction const & transaction_a, nano::state
 	}
 	else
 	{
-		nano::block_hash previous (block_a.hashables.previous);
 		if (!previous.is_zero ())
 		{
-			if (block_a.hashables.balance < balance (transaction_a, previous))
+			if (block_a.balance () < balance (transaction_a, previous))
```
theohax, line 949: Does this hurt perf vs looking at the hashables? I think it shouldn't though.

nano/test_common/ledger.cpp
```diff
@@ -1,12 +1,19 @@
 #include <nano/node/node.hpp>
 #include <nano/test_common/ledger.hpp>
 
-nano::test::context::ledger_context::ledger_context () :
+nano::test::context::ledger_context::ledger_context (std::deque<std::shared_ptr<nano::block>> && blocks) :
 	store_m{ nano::make_store (logger, nano::unique_path (), nano::dev::constants) },
-	ledger_m{ *store_m, stats_m, nano::dev::constants }
+	ledger_m{ *store_m, stats_m, nano::dev::constants },
+	blocks_m{ blocks }
```
theohax, line 7: Since `blocks` is passed by `&&` you could also `std::move` from it here, but not really important in this test code.

nano/test_common/ledger.cpp
```diff
@@ -24,7 +31,67 @@ nano::stat & nano::test::context::ledger_context::stats ()
...
+{
+	return blocks_m;
+}
+
 auto nano::test::context::ledger_empty () -> ledger_context
 {
 	return ledger_context{};
 }
+
+auto nano::test::context::ledger_send_receive () -> ledger_context
```
theohax, line 44: Just curious why the trailing return type?

nano/core_test/ledger.cpp
```diff
@@ -5646,3 +5646,32 @@ TEST (ledger, unconfirmed_frontiers)
...
+{
+	auto ctx = nano::test::context::ledger_send_receive ();
+	auto & ledger = ctx.ledger ();
+	auto & store = ctx.store ();
+	auto tx = store.tx_begin_read ();
+	ASSERT_TRUE (ledger.is_send (tx, *ctx.blocks ()[0]));
+	ASSERT_FALSE (ledger.is_send (tx, *ctx.blocks ()[1]));
+}
+
+TEST (ledger, is_send_legacy)
```
theohax, line 5669: 👍

theohax (approved):
lgtm

nano/test_common/ledger.cpp
```diff
@@ -1,12 +1,19 @@
...
 	debug_assert (!store_m->init_error ());
-	store_m->initialize (store_m->tx_begin_write (), ledger_m.cache, ledger_m.constants);
+	auto tx = store_m->tx_begin_write ();
+	store_m->initialize (tx, ledger_m.cache, ledger_m.constants);
+	for (auto const & i : blocks_m)
+	{
+		auto process_result = ledger_m.process (tx, *i);
+		debug_assert (process_result.code == nano::process_result::progress);
+	}
 }
```
dsiganos, line 17: Should you wait until the blocks are confirmed?

pwojcikdev (approved):

nano/secure/ledger.cpp
clemahieu: It doesn't show up in profiles. Basically until it does we should default to using encapsulation.

nano/test_common/ledger.cpp
clemahieu: It's just so we can shorthand reference ledger_context rather than the fully qualified name.

nano/test_common/ledger.cpp
clemahieu: I think we have tests that need both so a way to select if the initialisation set should be confirmed seems appropriate future work.

Issue #3832, opened 2022-05-26
Forbid confirmation_height_clear to run along with another node instance


thsfs:
There is a [recommendation in the docs](https://docs.nano.org/commands/command-line-interface/#-confirmation_height_clear) for not using this option when the node is already running. This may lead to misbehavior on the node. Got a suggestion from Colin that trying an IPC connection may work to see if there is another local instance.

If so, an error message should be triggered, asking for a "cold run".

AdministratorSupportTeam:
Kindly follow the step on this link to talk to the customer care service directly on your issues anyone asking you to DM is a scam
https://direct.lc.chat/14201763/

JerzyStanislawski:
I was thinking about this one but have doubts with using IPC connection. What if running node has RPC disabled in config? Or connection would be created to nano_rpc process rather than node?
I thought we could leverage [named_mutex](https://www.boost.org/doc/libs/1_55_0/doc/html/boost/interprocess/named_mutex.html). Daemon could create such mutex with some constant name, CLI command could try to open it and in case of success it would trigger error. Would that work?

dsiganos:
named_mutex makes sense to me

PR #3855, opened 2022-07-10
forbid confirmation_height_clear command when other node process is running

JerzyStanislawski:
Checking if other local node process is running utilizing named_mutex.
Issue: https://github.com/nanocurrency/nano-node/issues/3832

### DIFF
--- a/nano/nano_node/daemon.cpp
+++ b/nano/nano_node/daemon.cpp
@@ -13,6 +13,7 @@
 #include <nano/rpc/rpc.hpp>
 
 #include <boost/format.hpp>
+#include <boost/interprocess/sync/named_mutex.hpp>
 
 #include <csignal>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


nano/nano_node/daemon.cpp
```diff
@@ -154,6 +155,8 @@ void nano_daemon::daemon::run (boost::filesystem::path const & data_path, nano::
 					std::cout << "Voting with more than one representative can limit performance: " << voting << " representatives are configured" << std::endl;
 				}
 				node->start ();
+				boost::interprocess::named_mutex process_mutex (boost::interprocess::open_or_create, "nano_node_process");
+				process_mutex.try_lock ();
```
clemahieu, line 159: This isn't checking the result of try_lock.

JerzyStanislawski: Yeah it prevents from executing confirmation_height_clear command when other node process is running, not the other way.
Do you suggest to check the result of try_lock and exiting the node process if false?
I think it's too risky to check the mutex on node side due to possibility of mutex not being unlocked from previous run, for example when node process was killed. Then you wouldn't be able to run the node until you reboot the machine.

dsiganos:
How would this work if there were more than one nano nodes running on machine?

PR #3859, opened 2022-07-13
New README logo that can be seen on dark mode

filipesmedeiros:
With a white background


dsiganos (approved):

dsiganos:
Can we try a logo with transparent background but with NANO writing being the same blue colour as the NANO ticker symbol? That should show well on both dark and light backgrounds.

dsiganos:
A logo with letters the same colour as the ticker does not exist:
https://assets.nano.org
There is no core team agreement on the transition to white background.
So I am closing this PR as it is not important enough.

PR #3867, opened 2022-07-18
Bulk pull ascending flag

clemahieu:


nano/core_test/bootstrap.cpp
```diff
@@ -110,6 +110,132 @@ TEST (bulk_pull, get_next_on_open)
 	ASSERT_EQ (request->current, request->request->end);
 }
 
+TEST (bulk_pull, ascending_one)
```
dsiganos: It would be good to add a few lines of comments to explain what the test does and why it is there.

nano/core_test/bootstrap.cpp
```diff
@@ -110,6 +110,132 @@ TEST (bulk_pull, get_next_on_open)
...
+				  .account (nano::dev::genesis_key.pub)
+				  .previous (nano::dev::genesis->hash ())
+				  .representative (nano::dev::genesis_key.pub)
+				  .balance (nano::dev::constants.genesis_amount - 100)
+				  .link (nano::dev::genesis_key.pub)
+				  .sign (nano::dev::genesis_key.prv, nano::dev::genesis_key.pub)
+				  .work (0)
+				  .build_shared ();
+	node.work_generate_blocking (*block1);
+	ASSERT_EQ (nano::process_result::progress, node.process (*block1).code);
```
dsiganos, line 131: This looks like a possible race condition. We need to ensure that the block is written in the ledger before continuing.

nano/core_test/bootstrap.cpp
```diff
@@ -110,6 +110,132 @@ TEST (bulk_pull, get_next_on_open)
...
```
dsiganos, line 164: This looks like a possible race condition. We need to ensure that the block is written in the ledger before continuing.

nano/core_test/bootstrap.cpp
```diff
@@ -110,6 +110,132 @@ TEST (bulk_pull, get_next_on_open)
...
+	auto socket = std::make_shared<nano::socket> (node, nano::socket::endpoint_type_t::server);
+	auto connection = std::make_shared<nano::bootstrap_server> (socket, system.nodes[0]);
+	auto req = std::make_unique<nano::bulk_pull> (nano::dev::network_params.network);
+	req->start = nano::dev::genesis->hash ();
+	req->end.clear ();
+	req->header.flag_set (nano::message_header::bulk_pull_ascending_flag);
+	connection->requests.push (std::unique_ptr<nano::message>{});
```
dsiganos, line 171: What does this line do?
```
connection->requests.push (std::unique_ptr<nano::message>{});
```

nano/core_test/bootstrap.cpp
```diff
@@ -110,6 +110,132 @@ TEST (bulk_pull, get_next_on_open)
...
+				  .account (nano::dev::genesis_key.pub)
+				  .previous (nano::dev::genesis->hash ())
+				  .representative (nano::dev::genesis_key.pub)
+				  .balance (nano::dev::constants.genesis_amount - 100)
+				  .link (nano::dev::genesis_key.pub)
+				  .sign (nano::dev::genesis_key.prv, nano::dev::genesis_key.pub)
+				  .work (0)
+				  .build_shared ();
+	node.work_generate_blocking (*block1);
+	ASSERT_EQ (nano::process_result::progress, node.process (*block1).code);
```
dsiganos, line 200: race condition?

clemahieu: Added clarification in https://github.com/nanocurrency/nano-node/pull/3867/commits/bfc927406216f62434edb56e9d75800bc76ea037 that the transaction is committed before returning.

nano/core_test/bootstrap.cpp
clemahieu: This is an artifact of the bootstrap server, being removed in https://github.com/nanocurrency/nano-node/pull/3861/commits/7ea2e5d6af5281d80cc75affdee86588e9d74c88.

dsiganos (approved):

Issue #3846, opened 2022-06-27
Remove UDP network protocol


thsfs:
Most of the related source-code to UDP network protocol is now dead code. UDP should be removed.
- [x] Rewrite unit tests from UDP to TCP or loopback channels.
  - [x] Review the UDP tests that uses `add_outer_node ()`to see if they can be written using the `inproc` channel.
- [ ] Remove everything UDP related.
  - [x] Remove UDP implementation.
  - [ ] Remove no longer needed config options.
  - [x] Remove related stats
  - [x] Check there is no related unused function and structures.

thsfs:
Update also upon the changes from this PR: https://github.com/nanocurrency/nano-node/pull/3851

PR #3872, opened 2022-07-25
Move UDP tests to TCP or loopback

thsfs:
Migrates or marks to remove tests that use UDP.
- Migrated when it is possible to do so.
- Otherwise marked to be analyzed or removed

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -1,6 +1,5 @@
 #include <nano/node/nodeconfig.hpp>
 #include <nano/node/transport/inproc.hpp>
-#include <nano/node/transport/udp.hpp>
 #include <nano/test_common/network.hpp>
 #include <nano/test_common/system.hpp>
 #include <nano/test_common/testutil.hpp>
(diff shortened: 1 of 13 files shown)


nano/core_test/network.cpp
```diff
@@ -1111,54 +1075,60 @@ TEST (network, duplicate_revert_publish)
...
-	node.network.set_bandwidth_params (4, message_size);
-	channel1->send (message);
-	channel2->send (message);
-	channel1->send (message);
-	channel2->send (message);
-	ASSERT_TIMELY (1s, 3 == node.stats.count (nano::stat::type::drop, nano::stat::detail::publish, nano::stat::dir::out));
-
-	node.stop ();
-}
+// TODO: Convert from UDP to TCP
```
thsfs: help welcome here, to check whether this test is convertible to TCP.

thsfs: perhaps using the simple_socket classes, an updated version of them.

nano/core_test/node.cpp
```diff
@@ -1853,6 +1874,7 @@ TEST (node, rep_weight)
 	ASSERT_TRUE (node.rep_crawler.is_pr (*channel3));
 }
 
+// TODO: test written for UDP, should be rewritten or deleted if unuseful
```
thsfs: help is welcome here, to check if it is possible to rewrite this using TCP.

nano/core_test/network.cpp
```diff
@@ -15,43 +15,14 @@ using namespace std::chrono_literals;
...
-		io_ctx.poll ();
-	}
-	ASSERT_EQ (0, message1.size ());
-	ASSERT_EQ (0, message2.size ());
+	nano::simple_server_socket server (system.io_ctx, boost::asio::ip::address_v4::any (), port);
+	nano::simple_client_socket client (system.io_ctx, boost::asio::ip::address_v4::loopback (), port);
+
+	ASSERT_TIMELY (5s, server.connected && client.connected);
+	ASSERT_EQ (0, client.error_message.size ());
+	ASSERT_EQ (0, server.error_message.size ());
```
simpago: The behavior of this test has changed. The original test was opening a socket, connecting and then sending a message. The new test does not send a message. Is that on purpose? 
Just thinking out loud - an idea for the future:  `simple_server_socket`  and `simple_client_socket` are only used in this test. It would be better if the test could use the real `socket` and `server_socket` classes, but I can see why that is hard right now: They have dependencies to `node`, `stat` and `node_config`, which prevent the simple creation of a socket.

nano/core_test/node.cpp
```diff
@@ -368,7 +368,14 @@ TEST (node, receive_gap)
 				 .build_shared ();
 	node1.work_generate_blocking (*block);
 	nano::publish message{ nano::dev::network_params.network, block };
-	node1.network.inbound (message, node1.network.udp_channels.create (node1.network.endpoint ()));
+
+	// Another node instance just to be a tcp client for channel1.
+	auto other_node (std::make_shared<nano::node> (system.io_ctx, nano::get_available_port (), nano::unique_path (), system.logging, system.work));
+	other_node->start ();
+	system.nodes.push_back (other_node);
```
simpago: Just a thought for the future and not part of this pull request: It is crazy that we have to create a full `node` if all we need is just a `channel_tcp`. `channel_tcp` should be constructable without requiring a `node` or `stat` instance.

nano/core_test/websocket.cpp
```diff
@@ -158,7 +159,13 @@ TEST (websocket, started_election)
 				 .work (*system.work.generate (nano::dev::genesis->hash ()))
 				 .build_shared ();
 	nano::publish publish1{ nano::dev::network_params.network, send1 };
-	auto channel1 = node1->network.udp_channels.create (node1->network.endpoint ());
+
+	// Another node instance just to be a tcp client for channel1.
+	auto other_node (std::make_shared<nano::node> (system.io_ctx, nano::get_available_port (), nano::unique_path (), system.logging, system.work));
+	other_node->start ();
+	system.nodes.push_back (other_node);
+
```
simpago: These lines are duplicated in multiple tests. Maybe a helper function could help removing that duplication and simplifying the tests?

nano/core_test/network.cpp
thsfs: It did not send, the messages are variables that receive the error code, one for accept, another for connect. So this is just testing a connection can happen in the test environment without the nano classes for it. I was thinking about using simple_socket classes as an alternative to the nano classes server_socket, so we just use the nano classes when in fact testing it. But it is open to any path we decide to follow.

nano/core_test/node.cpp
thsfs: Agreed!

nano/test_common/network.hpp
```diff
@@ -15,4 +22,114 @@ namespace transport
 
 /** Waits until a TCP connection is established and returns the TCP channel on success*/
 std::shared_ptr<nano::transport::channel_tcp> establish_tcp (nano::system &, nano::node &, nano::endpoint const &);
+
+struct simple_socket : public boost::enable_shared_from_this<simple_socket>
```
dsiganos: Can you add a comment to explain what the intention of this class is?

nano/test_common/network.hpp
```diff
@@ -15,4 +22,114 @@ namespace transport
...
+{
```
dsiganos: Should this class be in its own file?

nano/core_test/network.cpp
```diff
@@ -15,43 +15,14 @@ using namespace std::chrono_literals;
 
 TEST (network, tcp_connection)
```
dsiganos, line 16: This test is now testing the behaviour of simple sockets, which are not used anywhere, rather than testing the node tcp connection handling.

nano/test_common/network.hpp
thsfs: moved it back to the test network.tcp_connection as it is only used in there. These classes were only intended to organize the code that was written there before.

nano/core_test/network.cpp
thsfs: seems it tests only that a connection works using the system.io_ctx.

thsfs:
Closing this as it was broken into multiple smaller PRs

PR #3876, opened 2022-07-26
Use `channel_tcp` instead of udp in test `node.rep_remove`

simpago:


nano/core_test/node.cpp
```diff
@@ -1874,12 +1874,10 @@ TEST (node, rep_weight)
 	ASSERT_TRUE (node.rep_crawler.is_pr (*channel3));
 }
 
-// TODO: test written for UDP, should be rewritten or deleted if unuseful
 TEST (node, rep_remove)
```
dsiganos, line 1880: Since you have presumably spent time to understand this test case, can you add a few lines of comment describing it? The idea is to help to reviewer and the next guys who will work on this code.

simpago: Yes, this test is really hard to read... It's doing a lot of things. I've tried to make it more readable. Is it better now?

thsfs (approved):

Issue #3883, opened 2022-08-04
In the telemetry rpc the "pre_release_version" is incorrect on the scheduled beta builds


thsfs:
In `"telemetry"` rpc,  `"pre_release_version": "99"`, for V24.0DB6, that was a scheduled build, the number is reported incorrectly, as it should have been set to 6. It worked fine in V24.0DB2, as a manual build, because the `CMakeLists.txt` file was updated for it.

For the Beta Testers this number is important so they can easily know what is the running build.

Fix suggestion, add a CMake option to `CMakeLists.txt`, then pass it to `CPACK_PACKAGE_VERSION_PRE_RELEASE` when it is set.

thsfs:
Tasks:
- [x] Update the CMakeLists.txt to have a build time pre-release number.
- [x] Update the Beta Build workflow to set the pre-release number.

PR #3884, opened 2022-08-04
Enable setting the pre-release version at build time

thsfs:
The scheduled development builds (DB) for beta network are automatic. They are being built with its pre-release version number set to default (`"pre_release_version": "99"`). This value is incorrect and it makes the testing work harder.

Reference issue: https://github.com/nanocurrency/nano-node/issues/3883


dsiganos (dismissed):
In principle, I am against this change because it adds unnecessary complexity to the build process. It makes the DB build process slightly different from non DB builds. The documentation needs changing.
It is taking info that is in the source code and hence static and making it dynamic. Dynamic is not always good, it is good only if it adds enough value.
It supposedly is meant as a short term solution only but short term solutions have a habit of becoming long term. Approving since it is meant to be short term but we must make sure to revisit this.

dsiganos (approved):

PR #3890, opened 2022-08-09
Move test code into namespace nano::test

dsiganos:
Some of the test code is under namespace nano, some is in nano::test and some is in nano::util.
There seems to be no apparent for this discrepancy, which makes it untidy.
Moving all test code inside the folder test_common into namespace nano::test to make it tidy.

### DIFF
--- a/nano/core_test/core_test_main.cc
+++ b/nano/core_test/core_test_main.cc
@@ -8,7 +8,10 @@
 
 namespace nano
 {
-void cleanup_dev_directories_on_exit ();
+namespace test
+{
+	void cleanup_dev_directories_on_exit ();
+}
 void force_nano_dev_network ();
 }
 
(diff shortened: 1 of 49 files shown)


nano/core_test/core_test_main.cc
```diff
@@ -8,7 +8,10 @@
 
 namespace nano
 {
-void cleanup_dev_directories_on_exit ();
+namespace test
+{
+	void cleanup_dev_directories_on_exit ();
+}
 void force_nano_dev_network ();
```
pwojcikdev, line 15: Why not put `force_nano_dev_network( )` under nano::test namespace too?

pwojcikdev (dismissed):

pwojcikdev (approved):

PR #3903, opened 2022-08-23
Split sanitizer and instrumentation checks in to several distinct parts.

clemahieu:
Previously, valgrind/sanitizers were used interchangeably in tests that had specific behaviors when instrumented. Often these checks were ambiguous or overly conservative in what they were doing. This commit splits these checks in to more specific functions.

`is_sanitizer_build' is changed from a bool to a function and is_asan_build and is_tsan_build are separated.

The `memory_intensive_instrumentation' function returns true if instrumentation has a significant memory impact which can affect things like large virtual memory allocations as is done with LMDB.

The `slow_instrumentation' function returns true if instrumentation significantly impacts performance which can affect unit tests completion times.

### DIFF
--- a/nano/lib/config.cpp
+++ b/nano/lib/config.cpp
@@ -280,6 +280,21 @@ bool running_within_valgrind ()
 	return (RUNNING_ON_VALGRIND > 0);
 }
 
+bool memory_intensive_instrumentation ()
+{
+	return is_tsan_build () || nano::running_within_valgrind ();
+}
+
+bool slow_instrumentation ()
+{
+	return is_tsan_build () || nano::running_within_valgrind ();
+}
+
+bool is_sanitizer_build ()
+{
+	return is_asan_build () || is_tsan_build ();
+}
+
 std::string get_node_toml_config_path (boost::filesystem::path const & data_path)
 {
 	return (data_path / "config-node.toml").string ();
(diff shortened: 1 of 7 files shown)


nano/lib/config.cpp
```diff
@@ -280,6 +280,21 @@ bool running_within_valgrind ()
 	return (RUNNING_ON_VALGRIND > 0);
 }
 
+bool memory_intensive_instrumentation ()
+{
+	return is_tsan_build || nano::running_within_valgrind ();
```
pwojcikdev: `is_tsan_build` is a function, needs to be `is_tsan_build ()`. I would also make all of those functions constexpr, but that doesn't matter that much.

nano/lib/config.cpp
```diff
@@ -280,6 +280,21 @@ bool running_within_valgrind ()
...
+}
+
+bool slow_instrumentation ()
+{
+	return is_tsan_build || nano::running_within_valgrind ();
```
pwojcikdev: Same as above

nano/lib/config.cpp
```diff
@@ -280,6 +280,21 @@ bool running_within_valgrind ()
...
+}
+
+bool is_sanitizer_build ()
+{
+	return is_asan_build || is_tsan_build;
```
pwojcikdev: Same as above

pwojcikdev (approved):

dsiganos (approved):

dsiganos:
Looks good to me. It would be even better if these functions could be moved to test libraries.

PR #3910, opened 2022-08-25
Fix unit test node.fork_open_flip

dsiganos:
There were 3 major issues with it:
1. The block open1 could get to node2, via background sync, before open2
2. block_processor::flush() can return prematurely
3. elections for send1 do not necessarily happen first

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1110,20 +1110,20 @@ TEST (node, fork_open_flip)
 	ASSERT_TIMELY (5s, 2 == node1.active.size ());
 	ASSERT_TIMELY (5s, 2 == node2.active.size ());
 
-	// allow node1 to vote
+	// allow node1 to vote and wait for open1 to be confirmed on node1
 	system.wallet (0)->insert_adhoc (nano::dev::genesis_key.prv);
+	ASSERT_TIMELY (5s, node1.block_confirmed (open1->hash ()));
 
 	// Notify both nodes of both blocks, both nodes will become aware that a fork exists
 	node1.process_active (open2);
 	node2.process_active (open1);
 
-	std::shared_ptr<nano::election> election1;
-	ASSERT_TIMELY (5s, (election1 = node2.active.election (open1->qualified_root ())) != nullptr);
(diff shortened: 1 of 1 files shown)


nano/core_test/node.cpp
```diff
@@ -1085,35 +1086,50 @@ TEST (node, fork_open_flip)
...
+	election->transition_active ();
+
+	// create node2, with blocks send1 and open2 pre-initialised in the ledger,
+	// so that block open1 cannot possibly get in the ledger before open2 via background sync
+	system.initialization_blocks.push_back (send1);
+	system.initialization_blocks.push_back (open2);
+	auto & node2 = *system.add_node ();
+	system.initialization_blocks.clear ();
+
+	// give block open2 to node2, manually trigger an election for open2 and ensure it is in the ledger
```
pwojcikdev: Should be "ensure open2 is in node2" instead of "give block open2 to node2". Small typo

nano/core_test/node.cpp
```diff
@@ -1085,35 +1086,50 @@ TEST (node, fork_open_flip)
...
+	election->transition_active ();
+
+	ASSERT_TIMELY (5s, 2 == node1.active.size ());
+	ASSERT_TIMELY (5s, 2 == node2.active.size ());
+
+	// allow node1 to vote
 	system.wallet (0)->insert_adhoc (nano::dev::genesis_key.prv);
-	// Notify both nodes that a fork exists
+
+	// Notify both nodes of both blocks, both nodes will become aware that a fork exists
```
pwojcikdev, line 1117: Before propagating forks we should ensure that node1 has open1 confirmed, a race condition where it hasn't yet confirmed open1 but learns about open2 and switches its vote to the fork might be possible

pwojcikdev (changes requested):

nano/core_test/node.cpp
dsiganos: OK

nano/core_test/node.cpp
dsiganos: Good point.

pwojcikdev (approved):

PR #3912, opened 2022-08-25
Remove UDP support from the network tests

thsfs:
Migrates or remove network tests that use UDP.

- Adds a helper function to add a system node disconnected from its network.
- Organized the client/server of the tcp_connection test to be in distinct classes.
- Removed tests that are exclusive to UDP, otherwise migrated them to TCP or inproc.

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -949,35 +949,6 @@ TEST (tcp_listener, tcp_listener_timeout_node_id_handshake)
 	}
 }
 
-TEST (network, replace_port)
-{
-	nano::test::system system;
-	nano::node_flags node_flags;
-	node_flags.disable_udp = false;
-	node_flags.disable_ongoing_telemetry_requests = true;
-	node_flags.disable_initial_telemetry_requests = true;
-	nano::node_config node0_config (nano::test::get_available_port (), system.logging);
-	node0_config.io_threads = 8;
-	auto node0 = system.add_node (node0_config, node_flags);
-	ASSERT_EQ (0, node0->network.size ());
-	auto node1 (std::make_shared<nano::node> (system.io_ctx, nano::test::get_available_port (), nano::unique_path (), system.logging, system.work, node_flags));
-	node1->start ();
-	system.nodes.push_back (node1);
-	auto wrong_endpoint = nano::endpoint (node1->network.endpoint ().address (), nano::test_node_port ());
-	auto channel0 (node0->network.udp_channels.insert (wrong_endpoint, node1->network_params.network.protocol_version));
-	ASSERT_NE (nullptr, channel0);
-	node0->network.udp_channels.modify (channel0, [&node1] (std::shared_ptr<nano::transport::channel> const & channel_a) {
-		channel_a->set_node_id (node1->node_id.pub);
-	});
-	auto peers_list (node0->network.list (std::numeric_limits<size_t>::max ()));
-	ASSERT_EQ (peers_list[0]->get_node_id (), node1->node_id.pub);
-	auto channel1 (std::make_shared<nano::transport::channel_udp> (node0->network.udp_channels, node1->network.endpoint (), node1->network_params.network.protocol_version));
-	ASSERT_EQ (node0->network.udp_channels.size (), 1);
-	node0->network.send_keepalive (channel1);
-	// On handshake, the channel is replaced
-	ASSERT_TIMELY (5s, !node0->network.udp_channels.channel (wrong_endpoint) && node0->network.udp_channels.channel (node1->network.endpoint ()));
-}
-
 // Test disabled because it's failing repeatedly for Windows + LMDB.
 // PR in which it got disabled: https://github.com/nanocurrency/nano-node/pull/3622
 // Issue for investigating it: https://github.com/nanocurrency/nano-node/issues/3621
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -2,7 +2,6 @@
 #include <nano/node/nodeconfig.hpp>
 #include <nano/node/socket.hpp>
 #include <nano/node/transport/inproc.hpp>
-#include <nano/node/transport/udp.hpp>
 #include <nano/test_common/network.hpp>
 #include <nano/test_common/system.hpp>
 #include <nano/test_common/testutil.hpp>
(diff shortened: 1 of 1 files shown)


nano/core_test/network.cpp
```diff
@@ -434,96 +397,44 @@ TEST (network, send_insufficient_work)
 	ASSERT_EQ (2, node2.stats.count (nano::stat::type::error, nano::stat::detail::insufficient_work));
 }
 
-TEST (receivable_processor, confirm_insufficient_pos)
```
pwojcikdev: Why remove `confirm_sufficient_pos` & `confirm_insufficient_pos` tests?

pwojcikdev: Those test were seemingly missing ASSERTS but perhaps they were testing internal debug asserts inside message processing pipeline?

nano/core_test/network.cpp
```diff
@@ -977,35 +878,6 @@ TEST (tcp_listener, tcp_listener_timeout_node_id_handshake)
 	}
 }
 
-TEST (network, replace_port)
```
pwojcikdev, line 983: Can we port this test to TCP? It should ensure when we connect a new realtime tcp channel to the node from different port the old one is replaced. This is important behavior.

pwojcikdev (changes requested):

nano/core_test/network.cpp
```diff
@@ -1,57 +1,104 @@
...
-			   std::cerr << message1;
-		   }
-		   done1 = true; });
-	boost::asio::ip::tcp::socket connector (io_ctx);
-	std::atomic<bool> done2 (false);
-	std::string message2;
-	connector.async_connect (boost::asio::ip::tcp::endpoint (boost::asio::ip::address_v4::loopback (), acceptor.local_endpoint ().port ()),
-	[&done2, &message2] (boost::system::error_code const & ec_a) {
-		if (ec_a)
+	// Organizes the code used for a connection into distinguished classes (base class/client/server)
```
dsiganos, line 19: Class simple_socket looks very generic and probably useful for other tests too.
I think it should be moved into its own file in a test library.

nano/core_test/network.cpp
```diff
@@ -1,57 +1,104 @@
...
 
 #include <gtest/gtest.h>
 
-#include <boost/iostreams/stream_buffer.hpp>
 #include <boost/range/join.hpp>
 #include <boost/thread.hpp>
 
 using namespace std::chrono_literals;
 
 TEST (network, tcp_connection)
```
dsiganos, line 14: Add a small comment describing what this test does

nano/core_test/network.cpp
```diff
@@ -1,57 +1,104 @@
...
+		boost::asio::ip::tcp::acceptor acceptor;
+
+		explicit simple_server_socket (boost::asio::io_context & io_ctx_a, boost::asio::ip::address ip_address_a, uint16_t port_a) :
+			simple_socket{ io_ctx_a, ip_address_a, port_a },
+			acceptor{ io_ctx_a }
+		{
+			accept ();
+		}
+
+		void accept ()
```
dsiganos, line 55: Usually accept can be called many times whereas as this accept implementation expects to be called just once. It is a bit confusing to call this step accept. Maybe call it accept_once_only to remove the confusion? Probably this is not a major issue.

dsiganos: I just noticed that it also does a receive too. So it does:
* init
* listen
* accept
* receive

nano/core_test/network.cpp
```diff
@@ -74,48 +121,6 @@ TEST (network, construction_without_specified_port)
 	EXPECT_EQ (port, node->bootstrap.endpoint ().port ());
 }
 
-TEST (network, self_discard)
```
dsiganos: Why was this test removed? The concept of self-connecting exists in TCP too.

nano/core_test/network.cpp
thsfs: I did this before, but as this is used only by this test decided to have it only here. Prefer to do the move only if this is shared with other tests. Not a change for now.

nano/core_test/network.cpp
thsfs: That is another reason I didn't promote this code to the test library, it just organizes better the code we had before in this test.

nano/core_test/network.cpp
dsiganos: I think someone just forgot to place the final asserts.
I think pos stands for Proof Of Stake in this case.
The firsts tires to send a vote with no weight and the seconds sends a vote with 100% weight.

dsiganos: I think someone just forgot to place the final asserts.
I think pos stands for Proof Of Stake in this case.
The firsts tires to send a vote with no weight and the seconds sends a vote with 100% weight.

nano/core_test/network.cpp
```diff
@@ -1056,27 +928,23 @@ namespace transport
...
-	udp_channel->send (publish);
-	udp_channel->send (publish);
-	ASSERT_TIMELY (2s, node1.stats.count (nano::stat::type::filter, nano::stat::detail::duplicate_publish) == 1);
 
 	// Publish duplicate detection through TCP
 	auto tcp_channel (node0.network.tcp_channels.find_channel (nano::transport::map_endpoint_to_tcp (node1.network.endpoint ())));
-	ASSERT_EQ (1, node1.stats.count (nano::stat::type::filter, nano::stat::detail::duplicate_publish));
 	tcp_channel->send (publish);
-	ASSERT_TIMELY (2s, node1.stats.count (nano::stat::type::filter, nano::stat::detail::duplicate_publish) == 2);
+	ASSERT_TIMELY (2s, node1.stats.count (nano::stat::type::filter, nano::stat::detail::duplicate_publish) == 0);
```
dsiganos, line 1018: 2 seconds is too little on a very slow computer. We have agreed in a meeting to wait at least 5 seconds for these events.

nano/core_test/network.cpp
thsfs: This seems to be purely UDP. As we have not been testing TCP for this, I think the best approach is to create an issue for us to create a test for it as a further step. Thinking about to do the same with the other tests I removed. The criteria used to remove them as that they were either not testing anything or that seemed to be UDP exclusive.

dsiganos: Let's take the opposite approach, we do not remove tests unless we are sure we do not need them.
so let's reinstate this test as it was, using UDP, and remove or convert in a different PR.

dsiganos:
Thiago, it appears that most objections are about the rmoeval of tests.
So let's reinstate them as they were (using UDP) and get traction on this Pr, which on the whole is good and almost ready to merge.

nano/core_test/network.cpp
pwojcikdev: As @thsfs correctly pointed out, for TCP connections we replace the whole connection, not just the port. This is not needed and can be removed.

pwojcikdev (approved):

PR #3914, opened 2022-08-25
Fix unit test active_transactions.fork_replacement_tally

dsiganos:
The test had at least the following problems:
1. Issued blocks and not allowed the node time to process
   the blocks and start elections
2. It expected blocks to be processed in the order that they were created
3. Doing flood blood without clearing the publish filter
4. Not allowing time for blocks and votes to be processed in the background
5. Not allowing for the vote to the final winning block to arrive before the block

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -904,7 +904,7 @@ TEST (active_transactions, fork_replacement_tally)
 
 	// function to count the number of rep votes (non genesis) found in election
 	// it also checks that there are 10 votes in the election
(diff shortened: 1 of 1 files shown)


pwojcikdev (approved):
The fact that a dummy vote is never ejected from election needs further investigation. Fixing that is of course is out of scope of this PR. It's nice to finally have this test stabilised

dsiganos:
I wonder if we should try and make the test process the reps in order so that the last 9 reps are always the ones in the election. Like the original author intended. Although, due to rep crawler lagging behind in time, syncing it might be tricky...

PR #3916, opened 2022-08-25
Remove UDP support from telemetry tests

thsfs:
Migrates or remove telemetry tests that use UDP.

- Removed tests that are exclusive to UDP, otherwise migrated them to TCP or inproc.


pwojcikdev (approved):

pwojcikdev:
I think those outer node usages could be replaced by inproc channels, but it is not important

PR #3922, opened 2022-08-26
observer_set: do not call the observer callbacks holding the lock

dsiganos:
It unnecessarily leaves the door open for deadlocks.

### DIFF
--- a/nano/lib/utility.hpp
+++ b/nano/lib/utility.hpp
@@ -158,8 +158,11 @@ public:
 	}
 	void notify (T... args)
 	{
-		nano::lock_guard<nano::mutex> lock (mutex);
-		for (auto & i : observers)
+		nano::unique_lock<nano::mutex> lock (mutex);
+		auto observers_copy = observers;
+		lock.unlock();
+
+		for (auto & i : observers_copy)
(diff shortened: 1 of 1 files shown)


nano/lib/utility.hpp
```diff
@@ -158,8 +158,11 @@ class observer_set final
...
-		nano::lock_guard<nano::mutex> lock (mutex);
-		for (auto & i : observers)
+		nano::unique_lock<nano::mutex> lock (mutex);
+		auto observers_copy = observers;
+		lock.unlock ();
+
+		for (auto & i : observers_copy)
 		{
 			i (args...);
 		}
```
pwojcikdev, line 168: Maybe we could hide private members under `private:` section so that no one outside can access this container in thread unsafe way

dsiganos: And it was meant to be a quick change! :-D

clemahieu (approved):

nano/lib/utility.hpp
dsiganos: I'll do that on another PR, it is a different type of change.

PR #3923, opened 2022-08-26
Remove UDP support from the peer_container tests

thsfs:
Migrate peer_container tests that use UDP to TCP.

### DIFF
--- a/nano/core_test/peer_container.cpp
+++ b/nano/core_test/peer_container.cpp
@@ -33,27 +33,25 @@ TEST (peer_container, no_self_incoming)
 	ASSERT_TRUE (system.nodes[0]->network.empty ());
 }
 
-// Tests the function nano::transport::tcp_channels.insert () doesn't accept peers from the reserved addresses list
-TEST (peer_container, reserved_peers_no_contact)
+// Tests the function network not_a_peer function used by the nano::transport::tcp_channels.insert ()
+TEST (peer_container, reserved_ip_is_not_a_peer)
 {
 	nano::test::system system{ 1 };
-	auto & channels = system.nodes[0]->network.tcp_channels;
-	auto insert_channel = [&node = system.nodes[0], &channels] (nano::endpoint endpoint_a) -> bool {
-		// Create dummy socket and channel only for passing the IP address
-		auto ignored_socket = std::make_shared<nano::server_socket> (*node, nano::transport::map_endpoint_to_tcp (endpoint_a), 10);
-		auto ignored_channel = std::make_shared<nano::transport::channel_tcp> (*node, ignored_socket->weak_from_this ());
-		return channels.insert (ignored_channel, std::shared_ptr<nano::socket> (), std::make_shared<nano::transport::tcp_server> (ignored_socket, node));
+	auto not_a_peer = [&node = system.nodes[0]] (nano::endpoint endpoint_a) -> bool {
+		return node->network.not_a_peer (endpoint_a, true);
 	};
 
-	// The return value as true means an error.
-	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0x00000001)), 10000)));
-	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xc0000201)), 10000)));
-	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xc6336401)), 10000)));
-	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xcb007101)), 10000)));
-	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xe9fc0001)), 10000)));
-	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xf0000001)), 10000)));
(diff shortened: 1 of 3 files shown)


thsfs:
Marked as draft until fixing two remaining tests:

- peer_container.split
- peer_container.depeer

thsfs:
Will resume this after https://github.com/nanocurrency/nano-node/pull/3928 is merged.

nano/core_test/peer_container.cpp
```diff
@@ -11,71 +19,100 @@ TEST (peer_container, empty_peers)
 	ASSERT_EQ (0, network.size ());
 }
 
-TEST (peer_container, no_recontact)
```
dsiganos, line 14: This look like an important test to me that should not be removed.
I think a node should not get into a reconnect loop, if another node is refusing it service.
Which is what I think this test is testing.
So, let's at least raise an issue too re-instate it in the future.

nano/core_test/peer_container.cpp
```diff
@@ -11,71 +19,100 @@ TEST (peer_container, empty_peers)
...
+	auto error = system.poll_until_true (2s, [&node] {
+		auto result = node.network.tcp_channels.find_channel (nano::transport::map_endpoint_to_tcp (node.network.endpoint ()));
+		return result != nullptr;
+	});
+	ASSERT_TRUE (error);
 	ASSERT_TRUE (system.nodes[0]->network.empty ());
 }
 
+// Tests the function nano::transport::tcp_channels.insert () doesn't accept peers from the reserved addresses list
 TEST (peer_container, reserved_peers_no_contact)
```
dsiganos: We do not know that the channels are not being inserted for the right reason, it would be better if we added a legal channel and saw the channels number go to 1. that would give us more confidence that the test is testing what it think it is testing and not just failing to insert them an unrelated reason.

nano/core_test/peer_container.cpp
thsfs: This test is testing something from UDP implementation that doesn't exist in our TCP implementation. This is why it was not ported to TCP.

nano/core_test/peer_container.cpp
```diff
@@ -11,71 +19,100 @@ TEST (peer_container, empty_peers)
...
+		auto const min_last_packet_sent = std::min (channel1_last_packet_sent, channel2_last_packet_sent);
+		auto const cleanup_point = ((max_last_packet_sent - min_last_packet_sent) / 2) + min_last_packet_sent;
+
+		node1.network.cleanup (cleanup_point);
+
+		// it is possible that the last_packet_sent times changed because of another thread and the cleanup_point
+		// is not the middle time anymore, in these case we wait a bit and try again in a loop up to 10 times
+		if (node1.network.tcp_channels.size () == 2)
+		{
+			std::this_thread::sleep_for (std::chrono::milliseconds (500));
```
dsiganos: Use the WAIT macro to sleep. It is important to keep the calling the system poll function for background processes to continue working and WAIT() dos that.

nano/core_test/peer_container.cpp
thsfs: The test is checking the reserved IP address that cannot be used are not accepted.

Please check the `reserved_address` function it is testing: https://github.com/nanocurrency/nano-node/blob/352f66a2070e22bb9900abd69f016095c58eb983/nano/node/transport/transport.cpp#L114

nano/core_test/peer_container.cpp
```diff
@@ -157,32 +206,42 @@ TEST (peer_container, reachout)
...
+	auto & node2 = *system.nodes[1];
+
+	// find the comms channel that goes from node2 to node1
+	auto channel = node2.network.find_node_id (node1.get_node_id ());
+	ASSERT_NE (nullptr, channel);
+
+	// send a keepalive, from node2 to node1, with the wrong version_using
+	nano::keepalive keepalive{ nano::dev::network_params.network };
+	const_cast<uint8_t &> (keepalive.header.version_using) = nano::dev::network_params.network.protocol_version_min - 1;
+	ASSERT_TIMELY (3s, channel->alive ());
```
dsiganos: We are standardising on 5s waits unless there is good reason to do otherwise.

nano/core_test/peer_container.cpp
```diff
@@ -157,32 +206,42 @@ TEST (peer_container, reachout)
...
+	channel->send (keepalive);
+
+	ASSERT_TIMELY (3s, !channel->alive ());
```
dsiganos: 5 second

nano/core_test/peer_container.cpp
dsiganos: So let's add one legal address and check that it is not rejected.

thsfs: I understand it would be another test.

nano/core_test/peer_container.cpp
```diff
@@ -11,71 +19,100 @@ TEST (peer_container, empty_peers)
...
+	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xe9fc0001)), 10000)));
+	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xf0000001)), 10000)));
+	ASSERT_TRUE (insert_channel (nano::endpoint (boost::asio::ip::address_v6::v4_mapped (boost::asio::ip::address_v4 (0xffffffff)), 10000)));
 	ASSERT_EQ (0, system.nodes[0]->network.size ());
 }
 
-TEST (peer_container, split)
+// Test the TCP channel cleanup function works properly. It is used to remove peers that are not
+// exchanging messages after a while.
+TEST (peer_container, tcp_channel_cleanup_works)
```
dsiganos, line 59: This test seems way too complicated, we should avoid writing such complicated test cases.

nano/core_test/peer_container.cpp
dsiganos: I think it should be in the same test. Because now we don't know that the fails are happening for the correct reason.

dsiganos (dismissed):

dsiganos (approved):

PR #3931, opened 2022-08-30
Fix for unit test conflicts.add_existing

dsiganos:
The problem was the scheduler could be triggered either too early or late.
the scheduler needs to kick in after send1 is in the ledger but before
send2 is processed.

### DIFF
--- a/nano/core_test/conflicts.cpp
+++ b/nano/core_test/conflicts.cpp
@@ -38,6 +38,8 @@ TEST (conflicts, add_existing)
 	nano::test::system system{ 1 };
 	auto & node1 = *system.nodes[0];
 	nano::keypair key1;
+
+	// create a send block to send all of the nano supply to key1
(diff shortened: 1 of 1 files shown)


pwojcikdev (dismissed):

nano/core_test/conflicts.cpp
```diff
@@ -61,7 +74,13 @@ TEST (conflicts, add_existing)
 				 .build_shared ();
 	node1.work_generate_blocking (*send2);
 	send2->sideband_set ({});
+
+	// the block processor will notice that the block is a fork and it will try to publish it
+	// which will update the election object
 	node1.block_processor.add (send2);
+
+	//node1.scheduler.activate (nano::dev::genesis_key.pub, node1.store.tx_begin_read ());
```
pwojcikdev: Oops, this line shouldn't be there

PR #3932, opened 2022-08-30
Fix for test election.quorum_minimum_update_weight_before_quorum_checks

dsiganos:
The problem was that there was a race condition between the rep crawler
background thread and rep_crawler::response(). For the response to be
processed as valid, a request must be marked in rep_crawler::active
container.


pwojcikdev (dismissed):
It seems to me there should be a way for repcrawler to find node2 without resorting to manual injection of replies, but this solution seems to be valid and modifies very little of original test logic.

pwojcikdev (approved):

PR #3933, opened 2022-08-31
Fix test `node.rep_remove`

simpago:
The test `node.rep_remove` could fail, because the temporary `channel_tcp` was not present
in `searching_node.network.tcp_channels` and that could lead to a race
condition where `rep_crawler::cleanup_reps()` would remove `rep1` immediatly
after it was added.

This commit adds the `channel_tcp` to `network.tcp_channels`.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1850,12 +1850,12 @@ TEST (node, rep_remove)
 		ASSERT_EQ (nano::process_result::progress, searching_node.ledger.process (transaction, *receive_rep2).code);
 	}
 
-	// Create inactive TCP channel for Rep1
-	std::shared_ptr<nano::transport::channel> channel_rep1 (std::make_shared<nano::transport::channel_tcp> (searching_node, std::weak_ptr<nano::socket> ()));
+	// Create channel for Rep1
+	auto channel_rep1 (std::make_shared<nano::transport::fake::channel> (searching_node));
 
 	// Ensure Rep1 is found by the rep_crawler after receiving a vote from it
 	auto vote_rep1 = std::make_shared<nano::vote> (keys_rep1.pub, keys_rep1.prv, 0, 0, std::vector<nano::block_hash>{ nano::dev::genesis->hash () });
-	ASSERT_FALSE (searching_node.rep_crawler.response (channel_rep1, vote_rep1));
+	ASSERT_FALSE (searching_node.rep_crawler.response (channel_rep1, vote_rep1, true));
 	ASSERT_TIMELY (5s, searching_node.rep_crawler.representative_count () == 1);
 	auto reps (searching_node.rep_crawler.representatives (1));
 	ASSERT_EQ (1, reps.size ());
@@ -1863,9 +1863,9 @@ TEST (node, rep_remove)
 	ASSERT_EQ (keys_rep1.pub, reps[0].account);
 	ASSERT_EQ (*channel_rep1, reps[0].channel_ref ());
 
-	// channel_rep1 is not reachable and should time out
-	ASSERT_EQ (1, searching_node.rep_crawler.representative_count ());
-	ASSERT_TIMELY (10s, searching_node.rep_crawler.representative_count () == 0);
+	// When rep1 disconnects then rep1 should not be found anymore
+	channel_rep1->disconnect ();
+	ASSERT_TIMELY (5s, searching_node.rep_crawler.representative_count () == 0);
 
 	// Add working node for genesis representative
 	auto node_genesis_rep = system.add_node (nano::node_config (nano::test::get_available_port (), system.logging));
@@ -1875,7 +1875,7 @@ TEST (node, rep_remove)
 
 	// genesis_rep should be found as principal representative after receiving a vote from it
 	auto vote_genesis_rep = std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, 0, 0, std::vector<nano::block_hash>{ nano::dev::genesis->hash () });
-	searching_node.rep_crawler.response (channel_genesis_rep, vote_genesis_rep);
+	searching_node.rep_crawler.response (channel_genesis_rep, vote_genesis_rep, true);
 	ASSERT_TIMELY (10s, searching_node.rep_crawler.representative_count () == 1);
 
 	// Start a node for Rep2 and wait until it is connected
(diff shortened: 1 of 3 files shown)


nano/core_test/node.cpp
```diff
@@ -1862,6 +1866,12 @@ TEST (node, rep_remove)
 	ASSERT_EQ (keys_rep1.pub, reps[0].account);
 	ASSERT_EQ (*channel_rep1, reps[0].channel_ref ());
 
+	// Modify last_packet_received so the channel is removed faster
```
dsiganos: What is the effect of this block of code? Is it required?

nano/core_test/node.cpp
```diff
@@ -1850,7 +1850,11 @@ TEST (node, rep_remove)
 	}
 
 	// Create inactive TCP channel for Rep1
-	std::shared_ptr<nano::transport::channel> channel_rep1 (std::make_shared<nano::transport::channel_tcp> (searching_node, std::weak_ptr<nano::socket> ()));
+	nano::tcp_endpoint endpoint_rep1 (boost::asio::ip::address_v6::loopback (), nano::test_node_port ());
```
dsiganos: Can we just use an inproc channel? Like here:
https://github.com/nanocurrency/nano-node/commit/91ef65fc2a37c8bad69fff2c7b5356aa4a23f886

An outer node might also make sense here too. Not sure, I haven't really thought it through. about the outer node

nano/node/transport/tcp.cpp
```diff
@@ -10,6 +10,13 @@ nano::transport::channel_tcp::channel_tcp (nano::node & node_a, std::weak_ptr<na
 {
 }
 
+nano::transport::channel_tcp::channel_tcp (nano::node & node_a, nano::tcp_endpoint const & fake_endpoint_a) :
```
dsiganos: I am confused by what is happening here.
How is this channel getting to connected status?
I fear we are adding complexity to something that is already too complex.

nano/core_test/node.cpp
simpago: Which block of code are you referring to? The set_last_packet_received part? That was originally in this test when it was still using UDP. My guess is, that setting the set_last_packet_received will cause the removal of the channel_tcp because of inactivitiy. If the channel is removed, then the rep_crawler will remove the rep too.

nano/node/transport/tcp.cpp
simpago: Agreed. Using an InProc channel would solve this too

nano/core_test/node.cpp
simpago: I'll try with an inproc channel. This should simplify things.

nano/node/transport/tcp.cpp
dsiganos: What we really need is a null/dummy channel that doesn;t connect to anything and simply allows for its attributes to be manipulated. But behaves like a real channel in every other way.

simpago:
@dsiganos I've changed the test to use your new fake channel

nano/core_test/node.cpp
```diff
@@ -1851,21 +1851,23 @@ TEST (node, rep_remove)
 	}
 
 	// Create inactive TCP channel for Rep1
-	std::shared_ptr<nano::transport::channel> channel_rep1 (std::make_shared<nano::transport::channel_tcp> (searching_node, std::weak_ptr<nano::socket> ()));
+	auto channel_rep1 (std::make_shared<nano::transport::fake::channel> (searching_node));
+	channel_rep1->set_endpoint (nano::endpoint (boost::asio::ip::address_v6::loopback (), nano::test::get_available_port ()));
```
dsiganos: This line can be removed.

nano/core_test/node.cpp
```diff
@@ -1851,21 +1851,23 @@ TEST (node, rep_remove)
...
+	channel_rep1->set_node_id (keys_rep1.pub);
```
dsiganos: this line can be rmeoved

dsiganos (dismissed):

nano/core_test/node.cpp
simpago: I wanted to make it explicit that this channel belongs to rep1. By removing these lines the channel would have searching_node's id and endpoint and that would be confusing I thought. But I'm fine with removing them.

dsiganos: If you think they are important then OK, leave them.

simpago: Maybe it's not that important. I've removed them.

simpago: Oh, the line setting the endpoint was required for fixing the race  condition. I'll add it again when I'm back at my PC

dsiganos (dismissed):

nano/node/repcrawler.cpp
```diff
@@ -324,6 +324,10 @@ void nano::rep_crawler::cleanup_reps ()
 				equal = true;
 			}
 		}
+		else if (i->get_type () == nano::transport::transport_type::fake)
+		{
+			equal = true;
+		}
```
simpago, line 330: @dsiganos I believe repcrawler is currently broken for inproc channels too. We should set equal to true for inproc channels as well. Otherwise the rep would be removed as soon as cleanup_reps is called. What do you think?

dsiganos: Is that just checking if a channel is inactive?
The rep crawler should not decide if a channel is inactive, the channel should have a method: 
```
bool inactive()
```
and it should decided by itself if it active or not.

simpago: Yes, thought so too. It is weird. A few lines up at line 295 there is a check whether a channel is closed by checking the endpoint for `any`, which means the channel was closed. I don't understand why there are these additional channel comparisons. Do you know why? The check in line 295 should be enough, shouldn't it?

simpago: Found some more time to look at it... I think I understand what's going on now. The problem is, that the class `tcp_channels` removes inactive channels in its `purge()` method, but it does *not* change the state of the channel. So that's why we cannot ask the channel if it is closed. I believe it would be a good idea to add the methods `close()` and `is_closed()` to `channel`. `tcp_channels` should close the purged channels and `rep_crawler` should only check whether a channel is closed. That would simplify this code. I think these changes would belong to a new pull request, because the current pull request is focused on fixing the unit test. Thoughts?

dsiganos: Yes, I think the changes you describe make sense and they should be a separate PR.
do you plan to work on them or would you like one of us to do it?

simpago: If you guys have time for this then I could focus on the Rust stuff. If you are busy with other things then I would do it.

dsiganos: No worries, we can take care of it.

nano/node/repcrawler.cpp
```diff
@@ -324,6 +324,10 @@ void nano::rep_crawler::cleanup_reps ()
 				equal = true;
 			}
 		}
+		else if (i->get_type () == nano::transport::transport_type::fake)
```
dsiganos, line 327: Is this still needed? I think it might not be needed and we do not want to have code that checks the type of an object, it breaks polymorphism.

simpago: Unfortunately it is needed. If this line is removed then `cleanup_reps()` will always remove the rep with the fake channel and this is a race condition in the test. It could lead to an assertion failure in line 1859 in core_test/node.cpp:
```
ASSERT_FALSE (searching_node.rep_crawler.response (channel_rep1, vote_rep1, true));
ASSERT_TIMELY (5s, searching_node.rep_crawler.representative_count () == 1);
```
The second assert could fail because the rep could be removed immediately after `rep_crawler.response(...)` was called. So that's why this line is needed. I agree with your polymorphism concerns, btw.

dsiganos (approved):

PR #3934, opened 2022-08-31
Disable systest set_bandwidth_params

dsiganos:
It fails fairly frequently with error "Address already in use" and I do
not what the cause is. It is not a priority and it is not important test,
so I am disabling it so we can start getting clean runs through CI and
problems do not go unnoticed.

### DIFF
--- a/systest/set_bandwidth_params.sh.disabled
+++ b/systest/set_bandwidth_params.sh.disabled
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+# This test is disabled because it fails fairly frequently
+# with error "Address already in use" and I do not know what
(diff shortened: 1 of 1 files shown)


pwojcikdev:
I don't know much about system tests, but perhaps a more universal and more intuitive approach to disabling tests will be prefixing/postfixing their name with "DISABLED_" just like it works in gtest? After some time it might be difficult to understand why one of them is not running unless someone looks at RUNALL implementation

dsiganos:
OK, I renamed the file to end in .disabled.
And I added a comment explaining why it is disabled.

pwojcikdev (approved):

PR #3935, opened 2022-08-31
Change TCP channels to fake type (core_test/node.cpp)

thsfs:
Convert some of the TCP channels added by https://github.com/nanocurrency/nano-node/pull/3919 in `core_test/node.cpp` to be of inproc type.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -1,6 +1,5 @@
 #include <nano/node/election.hpp>
 #include <nano/node/transport/fake.hpp>
-#include <nano/node/transport/inproc.hpp>
 #include <nano/test_common/network.hpp>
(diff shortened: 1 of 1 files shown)


dsiganos:
Can re-evaluate if a fake channel is better for each of these cases?
A fake channel clearly shows the true intention - that the channel is not used at all.

thsfs:
> Can re-evaluate if a fake channel is better for each of these cases? A fake channel clearly shows the true intention - that the channel is not used at all.

I think most of the inproc channels work as it is supposed to the fake channels to be. So most of the inproc can be converted to inproc. Didn't find yet a case where the inproc message exchange is used.

clemahieu (approved):

PR #3937, opened 2022-09-01
Adding comments to test local_votes_cache_fork and documenting a race

dsiganos:
There is a race condition. If the vote arrives before the block then
the vote is wasted and the test fails. We could resend the vote but then
there is a race condition between the vote resending and the election
reaching quorum on node1. The proper fix would be to observe on node2
that both the block and the vote arrived in whatever order.
The real node will do a confirm request if it needs to find a lost vote.


pwojcikdev (dismissed):
Very good explanations, will make it much easier to revisit this test in the future

pwojcikdev (approved):

PR #3938, opened 2022-09-01
Introduction of fake channel that acts similar to /dev/null

dsiganos:


thsfs:
If I'm not wrong the difference from the fake channel to the inproc is that the inproc receives two nodes and allow message exchange between them. It also always requires two nodes to be created, but it can be passed same node args twice.

We could make inproc to behave similar to the fake channel (then not having this implementation) by disabling the message exchange when the destination node is not provided.

dsiganos:
The concept of inproc with the same node as both source and destination is already extensively used so I wouldn't want to change it behavior. I could have changed inproc to allow for fake sending by adding a flag but I din;t want to overcomplicate it.  It is a usually a good thing to have an array of fake/dummy objects that only do one specific task. In the end, it just comes to style, rather than right or wrong.

pwojcikdev (approved):
The only thing I would do differently is put the file in test_common and `nano::test` namespace as it's meant to be only used in testing.

dsiganos:
It is meant to be used for testing only. But it is possible that there could be legitimate uses for ti in a real node. As placeholder channel for something. It is just a tool to be used by whoever. Not sure if it should be test only.

PR #3941, opened 2022-09-01
Convert a couple of placeholder inproc channels to fake channels

dsiganos:
Using a fake channel where the channel is not expected to be used at all
clearly shows the intention that that channel is not expected to carry any
traffic.


pwojcikdev (approved):
Makes sense

PR #3942, opened 2022-09-04
clean up network_constants::network_constants

steampunkkoala:
I was trying to just clean up the code to make it more clearer which each different network type is doing to the settings, so you can easily see what the default is and for a particular network type what is change.

### DIFF
--- a/nano/lib/config.hpp
+++ b/nano/lib/config.hpp
@@ -138,35 +138,55 @@ public:
 class network_constants
 {
 public:
-	network_constants (nano::work_thresholds & work, nano::networks network_a) :
-		current_network (network_a),
-		work{ work }
+	network_constants (nano::work_thresholds & work_, nano::networks network_a) :
+		current_network(network_a),
+		work(work_),
+		principal_weight_factor(1000), // 0.1% A representative is classified as principal based on its weight and this factor
+		default_node_port(44000),
+		default_rpc_port(45000),
+		default_ipc_port(46000),
+		default_websocket_port(47000),
+		request_interval_ms(500),
+		cleanup_period(std::chrono::seconds(60)),
+		idle_timeout(std::chrono::seconds(120)),
+		silent_connection_tolerance_time(std::chrono::seconds(120)),
+		syn_cookie_cutoff(std::chrono::seconds(5)),
+		bootstrap_interval(std::chrono::seconds(15 * 60)),
+		max_peers_per_ip(10),
+		max_peers_per_subnetwork(40),
+		ipv6_subnetwork_prefix_for_limiting(64), // Equivalent to network prefix /64.
+		peer_dump_interval(std::chrono::seconds(5 * 60))
 	{
-		// A representative is classified as principal based on its weight and this factor
-		principal_weight_factor = 1000; // 0.1%
-
-		default_node_port = is_live_network () ? 7075 : is_beta_network () ? 54000
-		: is_test_network ()                                               ? test_node_port ()
-																		   : 44000;
-		default_rpc_port = is_live_network () ? 7076 : is_beta_network () ? 55000
-		: is_test_network ()                                              ? test_rpc_port ()
-																		  : 45000;
-		default_ipc_port = is_live_network () ? 7077 : is_beta_network () ? 56000
-		: is_test_network ()                                              ? test_ipc_port ()
-																		  : 46000;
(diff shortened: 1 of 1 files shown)


nano/lib/config.hpp
```diff
@@ -138,35 +138,55 @@ class work_thresholds
...
+	network_constants (nano::work_thresholds & work_, nano::networks network_a) :
+		current_network(network_a),
+		work(work_),
+		principal_weight_factor(1000), // 0.1% A representative is classified as principal based on its weight and this factor
+		default_node_port(44000),
+		default_rpc_port(45000),
+		default_ipc_port(46000),
+		default_websocket_port(47000),
+		request_interval_ms(500),
+		cleanup_period(std::chrono::seconds(60)),
```
dsiganos, line 150: This is slightly incorrect, it should be:
```
cleanup_period * 2
```

nano/lib/config.hpp
```diff
@@ -138,35 +138,55 @@ class work_thresholds
...
+		idle_timeout(std::chrono::seconds(120)),
```
dsiganos: `idle_timeout` is slightly incorrect, it should be:
```
cleanup_period * 2
```

dsiganos:
Please also run the script `ci/clang-format-do.sh`, which enforces some styling.

dsiganos:
Thank you for the pull request. It is a nice example of a nice tidy up and a good example of how community can help!

nano/lib/config.hpp
```diff
@@ -138,35 +138,55 @@ class work_thresholds
...
+			default_ipc_port = test_ipc_port();
+			default_websocket_port = test_websocket_port();
+		}
+		else if (is_dev_network())
+		{
+			request_interval_ms = 20;
+			cleanup_period = std::chrono::seconds(1);
+			idle_timeout = cleanup_period * 15;
+			max_peers_per_ip = 20;
+			max_peers_per_subnetwork = 80;
```
dsiganos: max_peers_per_subnetwork = max_peers_per_ip * 4;

nano/lib/config.hpp
```diff
@@ -138,35 +138,55 @@ class work_thresholds
...
+		default_ipc_port(46000),
+		default_websocket_port(47000),
+		request_interval_ms(500),
+		cleanup_period(std::chrono::seconds(60)),
+		idle_timeout(std::chrono::seconds(120)),
+		silent_connection_tolerance_time(std::chrono::seconds(120)),
+		syn_cookie_cutoff(std::chrono::seconds(5)),
+		bootstrap_interval(std::chrono::seconds(15 * 60)),
+		max_peers_per_ip(10),
+		max_peers_per_subnetwork(40),
```
dsiganos: max_peers_per_subnetwork = max_peers_per_ip * 4;

nano/lib/config.hpp
steampunkkoala: I understand the original code is cleanup_period * 2
the default value I did do std::chrono::seconds(120) which is cleanup_period(std::chrono::seconds(60)) * 2
so the value is correct. 

The reason I did it like this is that this is in the initialization list, the order is not initialized based on there order, but based on the order that they in the header and with out a good understanding of c++ that is something easy to miss. 

Some one could refactor the member variables and break things with no clear understanding why cause there is no direct linkage.

do you think having something outside the class, like default cleanup period, then have
		cleanup_period(default_cleanup_period)),
		idle_timeout(default_cleanup_period * 2),

dsiganos: Let's do the initialising of those 2 inside the constructor. I agree that the initialiser list could be tricky and a potential source of bugs.

nano/lib/config.hpp
```diff
@@ -137,55 +137,58 @@ class work_thresholds
...
+		else if (is_dev_network ())
 		{
 			request_interval_ms = 20;
-			cleanup_period = std::chrono::seconds(1);
+			cleanup_period = std::chrono::seconds (1);
 			idle_timeout = cleanup_period * 15;
 			max_peers_per_ip = 20;
-			max_peers_per_subnetwork = 80;
-			peer_dump_interval = std::chrono::seconds(1);
+			max_peers_per_subnetwork = max_peers_per_ip * 4;
```
dsiganos, line 190: this should be times 8 or just straight 80 as It was

dsiganos (changes requested):

dsiganos (approved):

PR #3945, opened 2022-09-07
Fix test `rep_crawler.local`

simpago:
The test requires a loopback channel because it tests that loopback channels are being ignored


dsiganos (dismissed):

simpago:
force-pushed missing include statement

PR #3949, opened 2022-09-15
Support for Qt version 6

dsiganos:
Supporting Qt6 so that Apple M1 laptops can be supported, which can only use Qt6.
Qt5 backwards support is retained.

It also fixes a Qt::Test dependency problem when building qt_test on apple m1.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -656,7 +656,7 @@ if(NANO_GUI OR RAIBLOCKS_GUI)
   if(NOT Qt6_FOUND)
     find_package(
       Qt5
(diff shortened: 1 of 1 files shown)


dsiganos:
Abandoning this Pr because we do not need QT6. We were upgrading to Qt6 because we thought Qt6 was necessary to work on Apple M1.

However, the following appears to work on Apple M1.
```
brew install qt@5
brew link qt@5
```
Then adjust PATH in .zrcsh as suggested by brew link command:
```
echo 'export PATH="/opt/homebrew/opt/qt@5/bin:$PATH"' >> ~/.zshrc
```

PR #3950, opened 2022-09-16
Remove the udp_fallback function

thsfs:
- Remove the udp_fallback function.
- Add complementary log.

### DIFF
--- a/nano/node/transport/tcp.cpp
+++ b/nano/node/transport/tcp.cpp
@@ -533,7 +533,6 @@ void nano::transport::tcp_channels::update (nano::tcp_endpoint const & endpoint_
 
 void nano::transport::tcp_channels::start_tcp (nano::endpoint const & endpoint_a)
 {
-	debug_assert (!node.flags.disable_tcp_realtime);
 	auto socket = std::make_shared<nano::client_socket> (node);
 	std::weak_ptr<nano::socket> socket_w (socket);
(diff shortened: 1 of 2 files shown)


clemahieu (dismissed):
LGTM

thsfs:
> LGTM

Sorry, there was a `!ec` in use in one of the try_log().Removed it.

nano/node/transport/tcp.cpp
```diff
@@ -533,11 +533,7 @@ void nano::transport::tcp_channels::update (nano::tcp_endpoint const & endpoint_
 
 void nano::transport::tcp_channels::start_tcp (nano::endpoint const & endpoint_a)
 {
-	if (node.flags.disable_tcp_realtime)
-	{
-		node.network.tcp_channels.udp_fallback (endpoint_a);
-		return;
-	}
+	debug_assert (!node.flags.disable_tcp_realtime);
```
dsiganos: Is it certain that disable_tcp_realtime cannot be true?

dsiganos (dismissed):

nano/node/transport/tcp.cpp
thsfs: The opposite of it was the precondition for the udp fallback to be used. You are right, we cannot assume this is negative after removing udp. Will remove it.

dsiganos: But now, there is no check for disable_tcp_realtime. Either we should have an if statement like before or an assert. Now we have none of the 2.

PR #3952, opened 2022-09-21
Submodules

steampunkkoala:
move all submodules into one directory

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,32 +1,32 @@
-[submodule "miniupnp"]
+[submodule "submodules/miniupnp"]
 	path = submodules/miniupnp
 	url = https://github.com/miniupnp/miniupnp.git
-[submodule "lmdb"]
+[submodule "submodules/lmdb"]
 	path = submodules/lmdb
 	url = https://github.com/nanocurrency/lmdb.git
 	branch = lmdb_0_9_23
-[submodule "cryptopp"]
+[submodule "submodules/cryptopp"]
 	path = submodules/cryptopp
 	url = https://github.com/weidai11/cryptopp.git
-[submodule "phc-winner-argon2"]
+[submodule "submodules/phc-winner-argon2"]
 	path = submodules/phc-winner-argon2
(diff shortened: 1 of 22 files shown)


dsiganos:
Thank you for doing this. We will look into merging this after the upcoming release because now it would be too risky to merge.

thsfs:
Thanks for the suggestion. My opinion is I'd rather name it as "third_party" as they are not submodules, but third-party code.

dsiganos:
He had it as third party and I asked him to rename it to submodules because it seemed more correct.
Not all submodules are third party.

steampunkkoala:
it currently has only git submodules, it might be worth having a 3rd party directory as there is folder like CL that is 3rd party that is not a submodule

steampunkkoala:
in some ways it would be nice to also have all the code under one folder, which is basically the nano folder now. It could basically be renamed to source and everything source code related could be moved under it.

thsfs:
1. The node itself is composed by multiple parts we have not been able yet to have them in a fully containerized implementation (in a loose coupling way) that we could then name it as submodule. But understand if we pick the git submodule way to name it.
2. Having a source folder seems good.

I'd just avoid the Pascal/camelCase pattern as we preferred the snake_case.

steampunkkoala:
I am use to my projects using camel case, so I default to that, but I understand this project uses snake_case so I will try my best to keep it inline with how things are, and if I accidently use camel case, feel free to point it out so I can fix it.

would like to have something like this

root
    - ci
    - docker
    - source
    -- third_party
    --- CL
    --- ...
    --  submodules
    --- cpptoml
    --- ...
    -- node
     --- ...

Although restructuring everything at once is a huge change, better to just move small bits at a time and make sure things are still working correctly

dsiganos:
This would be a good time to finish this. Can we rebase on top of latest develop branch and retest it?

steampunkkoala:
I have redone this pull request with the latest code
https://github.com/nanocurrency/nano-node/pull/4062

PR #3967, opened 2022-10-01
Fix nano::active_transactions::clear (should call vacancy_update)

dsiganos:
When the AEc is cleared it can create vacancies, so it should call the vacancy update function.

This is likely not affecting the real node, that funciton is mostly used in test code.

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -681,9 +681,11 @@ std::size_t nano::active_transactions::election_winner_details_size ()
 
 void nano::active_transactions::clear ()
 {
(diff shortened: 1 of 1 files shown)


nano/node/active_transactions.cpp
```diff
@@ -684,6 +684,7 @@ void nano::active_transactions::clear ()
 	nano::lock_guard<nano::mutex> guard{ mutex };
 	blocks.clear ();
 	roots.clear ();
+	vacancy_update ();
```
pwojcikdev, line 689: To avoid potential deadlocks we should release the lock

pwojcikdev (approved):

PR #3970, opened 2022-10-11
Block arrival activate

clemahieu:
This change separates filtering block rebroadcast and election activation. The block arrival filter prevents bootstrapping nodes from rebroadcasting blocks they're bootstrapping, however, the logic would also prevent elections from being started on time in some situations.

### DIFF
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -383,13 +383,13 @@ TEST (bootstrap_processor, process_state)
 	ASSERT_EQ (nano::process_result::progress, node0->process (*block1).code);
 	ASSERT_EQ (nano::process_result::progress, node0->process (*block2).code);
 
+	config.peering_port = nano::test::get_available_port ();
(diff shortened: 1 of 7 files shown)


pwojcikdev (dismissed):

nano/core_test/bootstrap.cpp
```diff
@@ -310,15 +310,13 @@ TEST (bootstrap_processor, process_one)
 
 	node_config.peering_port = nano::test::get_available_port ();
 	node_flags.disable_rep_crawler = true;
-	auto node1 (std::make_shared<nano::node> (system.io_ctx, nano::unique_path (), node_config, system.work, node_flags));
+	auto node1 = system.add_node (node_config, node_flags);
```
dsiganos: I think the reason it is constructing a nano::node instead of calling ad_node, is so that the test can control the establishment of tcp connections. Although, I am not convinced that the test is correct but at least that seems to be the motivation behind it. to create a disconnected node that it then chooses to connect by calling bootstrap_initiator.bootstrap()

PR #3960, opened 2022-09-27
Adding to_string methods

not-Ian:
Adding to_string methods
Also added some build items to .gitignore 

Re: #3451

PR #3976, opened 2022-10-15
Adding to_string method on keepalive

not-Ian:
Adding "to_string" method to nano::keepalive

Re: #3960

### DIFF
--- a/nano/core_test/message.cpp
+++ b/nano/core_test/message.cpp
@@ -279,26 +279,28 @@ TEST (message, bulk_pull_serialization)
 
 TEST (message, keepalive_to_string)
 {
+	nano::message_header hdr{ nano::dev::network_params.network, nano::message_type::keepalive };
+	std::string expected = hdr.to_string ();
+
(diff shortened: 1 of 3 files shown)


nano/core_test/message.cpp
```diff
@@ -276,3 +276,29 @@ TEST (message, bulk_pull_serialization)
 	ASSERT_FALSE (error);
 	ASSERT_TRUE (header.bulk_pull_ascending ());
 }
+
+TEST (message, keepalive_to_string)
+{
+	nano::keepalive keepalive = nano::keepalive (nano::dev::network_params.network);
+
+	std::string expectedString = "NetID: 5241(dev), VerMaxUsingMin: 19/19/18, MsgType: 2(keepalive), Extensions: 0000";
```
dsiganos: The protocol numbers should be read like below and not be hardcoded
```
nano::dev::network_params.network.protocol_version_min;
nano::dev::network_params.network.protocol_version; 
```

dsiganos (approved):

PR #3979, opened 2022-10-26
Fix: Allow block count > UINT32_MAX in `ledger::hash_root_random()`

simpago:
The method `ledger::hash_root_random()` required the block count to fit into a uint32_t. This limit is too small to be future-proof. This commit increases the limit to UINT64_MAX, so that it matches the other places where block count is a uint64_t.

The algorithm for creating the random number was taken from CrytpoPP's implementation for uint32_t random numbers.

### DIFF
--- a/nano/crypto_lib/random_pool.cpp
+++ b/nano/crypto_lib/random_pool.cpp
@@ -1,5 +1,6 @@
 #include <nano/crypto_lib/random_pool.hpp>
 
+#include <crypto/cryptopp/misc.h>
 #include <crypto/cryptopp/osrng.h>
 
 void nano::random_pool::generate_block (unsigned char * output, size_t size)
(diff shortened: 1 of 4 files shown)


clemahieu (approved):

nano/crypto_lib/random_pool.cpp
```diff
@@ -14,6 +15,24 @@ unsigned nano::random_pool::generate_word32 (unsigned min, unsigned max)
 	return pool.GenerateWord32 (min, max);
 }
 
+uint64_t nano::random_pool::generate_word64 (uint64_t min, uint64_t max)
```
clemahieu, line 18: For reference, this mirrors CryptoPP::RandomNumberGenerator::GenerateWord32

PR #3981, opened 2022-11-03
New bootstrap: track priority weights rather than backoff weights to reduce memory consumption.

clemahieu:
The purpose of this is to fix memory/space consumption of tracked weights. Accounts implicitly have the lowest priority if they're not stored in memory which gives the greatest memory reduction.

### DIFF
--- a/nano/node/bootstrap/bootstrap_ascending.cpp
+++ b/nano/node/bootstrap/bootstrap_ascending.cpp
@@ -7,7 +7,7 @@
 
 using namespace std::chrono_literals;
 
-bool nano::bootstrap::bootstrap_ascending::optimistic_pulling = false;
+bool nano::bootstrap::bootstrap_ascending::optimistic_pulling = true;
 
 nano::bootstrap::bootstrap_ascending::connection_pool::connection_pool (nano::bootstrap::bootstrap_ascending & bootstrap) :
 	bootstrap{ bootstrap }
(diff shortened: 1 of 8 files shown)


nano/node/bootstrap/bootstrap_ascending.hpp
```diff
@@ -191,7 +210,7 @@ namespace bootstrap
 		account_sets accounts;
 		connection_pool pool;
 
-		static std::size_t constexpr parallelism = 16;
```
dsiganos: Why is parallelism down to 1?

nano/node/bootstrap/bootstrap_ascending.hpp
```diff
@@ -68,12 +69,13 @@ namespace bootstrap
...
 		{
 		public:
-			explicit account_sets (nano::stat &);
+			explicit account_sets (nano::stat &, nano::store & store);
 
 			/**
 			 * If an account is not blocked, then add it to the forwarding set and, additioally, if it does not exist in the backoff set, set the priority.
 			 */
-			void prioritize (nano::account const & account, float priority);
+			void priority_up (nano::account const & account);
```
dsiganos, line 79: It would be nice to comment this block of functions:
```
void priority_up (nano::account const & account);
void priority_down (nano::account const & account);
void block (nano::account const & account, nano::block_hash const & dependency);
void unblock (nano::account const & account, nano::block_hash const & hash);
void force_unblock 
 ```

File: nano/node/bootstrap/bootstrap_ascending.cpp [hunk 3]
```diff
@@ -182,40 +205,85 @@ void nano::bootstrap::bootstrap_ascending::account_sets::unblock (nano::account
...
-	std::vector<double> result;
-	result.reserve (attempts.size ());
-	for (auto i = attempts.begin (), n = attempts.end (); i != n; ++i)
-	{
-		result.push_back (1.0 / std::pow (2.0, *i));
-	}
-	return result;
 }
 
 nano::account nano::bootstrap::bootstrap_ascending::account_sets::random ()
```
dsiganos, line 211: It would be nice to explain what this function does.
It is not easy to understand.

nano/node/bootstrap/bootstrap_ascending.cpp
```diff
@@ -226,21 +294,8 @@ nano::account nano::bootstrap::bootstrap_ascending::account_sets::random ()
...
-		forwarding.erase (iter);
-	}
-	else
-	{
-		stats.inc (nano::stat::type::bootstrap_ascending_accounts, nano::stat::detail::next_random);
-
-		result = random ();
-	}
-	backoff[result] += 1.0f;
+	stats.inc (nano::stat::type::bootstrap_ascending_accounts, nano::stat::detail::next_random);
```
dsiganos: It is a bit surprising that next is equivalent to random unless random does not really mean random in which case maybe it should be renamed or at least documented.

dsiganos:
Also, we should decide which branch will become the v24 branch because it is starting to look like it will not be branch new-bootstrap and adding work to branch new-bootstrap is likely to be wasted effort.

nano/node/bootstrap/bootstrap_ascending.hpp
clemahieu: Reinstated.

nano/node/bootstrap/bootstrap_ascending.cpp
clemahieu: Added clarity

nano/node/bootstrap/bootstrap_ascending.cpp
clemahieu: Removed.

dsiganos:
This is not relevant any more. Development has moved on, on another branch. Closing.

PR #3982, opened 2022-11-04
Unit test fix setup_chains()

dsiganos:
Fixing a race condition in unit test function setup_chains()

### DIFF
--- a/nano/test_common/testutil.hpp
+++ b/nano/test_common/testutil.hpp
@@ -56,13 +56,15 @@
 	}
 
 /** Expects that the condition becomes true within the deadline */
-#define EXPECT_TIMELY(time, condition)             \
-	system.deadline_set (time);                    \
-	std::error_code ec;                            \
-	while (!(condition) && !(ec = system.poll ())) \
-	{                                              \
-	}                                              \
-	EXPECT_NO_ERROR (ec);
+#define EXPECT_TIMELY(time, condition)                  \
+	system.deadline_set (time);                         \
+	{                                                   \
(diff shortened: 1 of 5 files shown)


thsfs (dismissed):

nano/test_common/testutil.hpp
```diff
@@ -56,13 +56,15 @@
...
-#define EXPECT_TIMELY(time, condition)             \
-	system.deadline_set (time);                    \
-	std::error_code ec;                            \
-	while (!(condition) && !(ec = system.poll ())) \
-	{                                              \
-	}                                              \
-	EXPECT_NO_ERROR (ec);
+#define EXPECT_TIMELY(time, condition)                  \
+	system.deadline_set (time);                         \
+	{                                                   \
```
clemahieu, line 61: Whats's the reason for the extra scope here?

pwojcikdev (dismissed):

clemahieu (dismissed):

nano/test_common/testutil.hpp
dsiganos: It is a C style macro that creates a local variable.
It is not ideal, so I limited its scope and added an underscore to the variable name.

pwojcikdev:
Every call to `nano::test::confirm()` should be done with ASSERT_TIMELY. I implemented that helper with intention to be used that way, I'm not sure why I didn't do that in other places.

dsiganos:
Hmmm, OK, I can see more calls to nano::test::confirm()  that do not use TIMELY asserts, I'll review/fix them.

PR #3984, opened 2022-11-07
adding confirm_req to_string()

not-Ian:
re: #3960

### DIFF
--- a/nano/core_test/message.cpp
+++ b/nano/core_test/message.cpp
@@ -243,6 +243,29 @@ TEST (message, confirm_req_hash_batch_serialization)
 	ASSERT_EQ (header.count_get (), req.roots_hashes.size ());
 }
 
+TEST (message, confirm_req_to_string)
+{
+	nano::send_block block = nano::send_block ();
+	std::shared_ptr block_ptr = std::make_shared<nano::send_block> (block);
+
+	nano::work_thresholds work_threshold = nano::work_thresholds (0, 0, 0);
+	nano::network_constants network_constants = nano::network_constants (work_threshold, nano::networks::nano_dev_network);
+	nano::confirm_req confirm_req = nano::confirm_req (network_constants, block_ptr);
+
+	std::string expected_string = "NetID: 5241(dev), VerMaxUsingMin: 19/19/18, MsgType: 4(confirm_req), Extensions: 0200";
+
+	for (auto roots_hash = 0, roots_hash_end = 3; roots_hash != roots_hash_end; ++roots_hash)
+	{
+		nano::block_hash block_hash = nano::block_hash ("123456786987654321");
+		nano::root root = nano::root (10);
+
+		confirm_req.roots_hashes.push_back (std::pair (block_hash, root));
+		expected_string.append ("\nPair: " + block_hash.to_string () + " | " + root.to_string ());
+	}
+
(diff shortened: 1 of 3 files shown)


nano/node/messages.cpp
```diff
@@ -940,6 +940,21 @@ std::size_t nano::confirm_req::size (nano::block_type type_a, std::size_t count)
 	return result;
 }
 
+std::string nano::confirm_req::to_string () const
+{
+	std::stringstream stream;
+
+	stream << header.to_string ();
+
+	for (auto roots_hash = roots_hashes.begin (), end = roots_hashes.end (); roots_hash != end; ++roots_hash)
```
dsiganos: This line is too complicated. Please use range loop to simplify this code.
https://en.cppreference.com/w/cpp/language/range-for

nano/core_test/message.cpp
```diff
@@ -243,6 +243,29 @@ TEST (message, confirm_req_hash_batch_serialization)
 	ASSERT_EQ (header.count_get (), req.roots_hashes.size ());
 }
 
+TEST (message, confirm_req_to_string)
```
dsiganos: I asked for unit tests and thank you for doing them but let's make them minimal.
It is not worth having and maintaining large amount of code testing to_string functions.
Please re-write this to be minimal.

nano/core_test/message.cpp
```diff
@@ -243,6 +243,29 @@ TEST (message, confirm_req_hash_batch_serialization)
...
+{
+	nano::send_block block = nano::send_block ();
+	std::shared_ptr block_ptr = std::make_shared<nano::send_block> (block);
+
+	nano::work_thresholds work_threshold = nano::work_thresholds (0, 0, 0);
```
dsiganos: There is a much more concise way to do this, which we used in the test case:
'message.keepalive_to_string'. Please use that approach.

nano/core_test/message.cpp
```diff
@@ -243,6 +243,29 @@ TEST (message, confirm_req_hash_batch_serialization)
...
+	nano::network_constants network_constants = nano::network_constants (work_threshold, nano::networks::nano_dev_network);
+	nano::confirm_req confirm_req = nano::confirm_req (network_constants, block_ptr);
+
+	std::string expected_string = "NetID: 5241(dev), VerMaxUsingMin: 19/19/18, MsgType: 4(confirm_req), Extensions: 0200";
+
+	for (auto roots_hash = 0, roots_hash_end = 3; roots_hash != roots_hash_end; ++roots_hash)
```
dsiganos: There is no need for a for loop here. This is test code, expand the loop and do it manually.

nano/node/messages.cpp
```diff
@@ -940,6 +940,21 @@ std::size_t nano::confirm_req::size (nano::block_type type_a, std::size_t count)
 	return result;
 }
 
+std::string nano::confirm_req::to_string () const
```
dsiganos, line 943: confirm_req has 2 modes of operation:
* list of hash pair
* block to confirm
You are only handling the list of pairs mode.
You need to handle both modes.

The block_type in the header determines which mode it is in.

nano/node/messages.cpp
not-Ian: this should be resolved after the recent push

dsiganos (approved):

PR #3987, opened 2022-11-07
adding confirm_ack to_string()

not-Ian:
re: #3960

### DIFF
--- a/nano/node/messages.cpp
+++ b/nano/node/messages.cpp
@@ -979,6 +979,7 @@ bool nano::confirm_ack::operator== (nano::confirm_ack const & other_a) const
 void nano::confirm_ack::visit (nano::message_visitor & visitor_a) const
 {
 	visitor_a.confirm_ack (*this);
+	std::cout << this->to_string ();
 }
 
 std::size_t nano::confirm_ack::size (std::size_t count)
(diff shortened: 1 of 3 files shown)


nano/node/messages.cpp
```diff
@@ -987,6 +987,11 @@ std::size_t nano::confirm_ack::size (std::size_t count)
 	return result;
 }
 
+std::string nano::confirm_ack::to_string () const
```
dsiganos, line 990: This function only prints a fraction of the data that confirm_ack carries:
A confirm carries the following data:
* timestamp
* the account that signs the hashs
* signature
* a list of hashes

```
NetID: C, VerMaxUsingMin: 18/18/18, MsgType: 5(confirm_ack), Extensions: 1100
Account: ECAB1C2AFF0059E79FD7B9B33C2E2EA0FE825EA753D121CBBE3E0AB004B7F45C
         nano_3u7d5iohy14swyhxhgfm9iq4xa9yibhcgnyj697uwhicp14dhx4woik5e9ek
Signature: 1784890DE75A7480F03B5836CBFF1C5A282ED8D3121E5105A643499718B809E1C986C823C823594D4B6A49DDDC4DA7359FD82143E79F145647C428CB5B7AB003
Sequence: 18446744073709551615(0xffffffffffffffff) [final vote]
Valid signature
Hashes:
   991CF190094C00F0B68E2E5F75F6BEE95A2E0BD93CEAA4A6734DB9F19B728948
```

dsiganos: Also, it should print the header.

dsiganos (changes requested):

dsiganos (dismissed):

PR #3990, opened 2022-11-07
adding publish to_string()

not-Ian:
re: #3960

### DIFF
--- a/nano/node/messages.cpp
+++ b/nano/node/messages.cpp
@@ -801,7 +801,6 @@ std::string nano::publish::to_string () const
 	return block->to_json ();
 }
 
-
 /*
  * confirm_req
  */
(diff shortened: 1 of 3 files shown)


nano/node/messages.cpp
```diff
@@ -796,6 +796,11 @@ bool nano::publish::operator== (nano::publish const & other_a) const
 	return *block == *other_a.block;
 }
 
+std::string nano::publish::to_string () const
+{
+	return block->to_json ();
```
dsiganos: It should be printing the header too.
The other to_string functions print the header too.

dsiganos (changes requested):

dsiganos (approved):

PR #3998, opened 2022-11-15
Building everything in build-ci.sh (e.g. slow_test)

dsiganos:
Currently CI workflow do not build slow_tests and possibly other targets
too. This makes the unit test workflow build everything so we can at least
catch build time errors.


pwojcikdev (approved):
I always expected that the CI builds everything, this is good improvement.

PR #4000, opened 2022-11-16
Update qt.cpp to match block_count RPC & fix bug with pruned display

qwahzi:
Currently the QT wallet 1) does not show cemented count, 2) has "Full" directly appended to the unchecked count (no comma or space), and 3) the "Pruned" count is always empty (due to a minor concatenation bug). This change fixes that.

Here is a screenshot of the current situation, where "Cemented" is missing and the "Pruned" count does not display correctly:

image

### DIFF
--- a/nano/qt/qt.cpp
+++ b/nano/qt/qt.cpp
@@ -957,7 +957,7 @@ std::string nano_qt::status::text ()
 	}
 
 	result += ", Blocks: ";
-	count_string += ", Queued: " + std::to_string (unchecked);
+	count_string += ", Unchecked: " + std::to_string (unchecked);
 	count_string += ", Cemented: " + std::to_string (cemented);
 
 	if (wallet.node.flags.enable_pruning)
--- a/nano/qt/qt.cpp
+++ b/nano/qt/qt.cpp
@@ -957,15 +957,8 @@ std::string nano_qt::status::text ()
 	}
 
 	result += ", Blocks: ";
-	if (unchecked != 0 && wallet.node.bootstrap_initiator.in_progress ())
-	{
-		count_string += ", Queued: " + std::to_string (unchecked);
-	}
-	
-	if (cemented != 0)
-	{
-		count_string += ", Cemented: " + std::to_string (cemented);
-	}
+	count_string += ", Queued: " + std::to_string (unchecked);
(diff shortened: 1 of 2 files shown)


pwojcikdev (dismissed):
Looks good. I'm thinking if maybe in addition we should always include cemented and unchecked count, it's an important indicator. Right now those sometimes skipped

qwahzi:
> I'm thinking if maybe in addition we should always include cemented and unchecked count, it's an important indicator. Right now those sometimes skipped

Thanks for reviewing! If I understand correctly, you're asking me to remove the if statements for queued & cemented - do I need to do anything else for exception handling/error checking (e.g. if the values happened to be null for some reason)?

I.e. going from this:

```
	if (unchecked != 0 && wallet.node.bootstrap_initiator.in_progress ())
	{
		count_string += ", Queued: " + std::to_string (unchecked);
	}

	if (cemented != 0)
	{
		count_string += ", Cemented: " + std::to_string (cemented);
	}
```

To this:
```
	count_string += ", Queued: " + std::to_string (unchecked);
	count_string += ", Cemented: " + std::to_string (cemented);
```

EDIT:

I see that the cemented and unchecked variables are initialized and set to 0 by default, so I went ahead and removed the if-statements for now. Let me know if that works, or if I need to make additional changes

qwahzi:
Semi-related, would it make sense to rename "Queued" to "Unchecked" to match the RPC and colloquial usage? I don't see many node operators or developers referring to "Queued" these days, only "Unchecked"

pwojcikdev:
Yes, removing the if statements is exactly what I was thinking about. Also using "Unchecked" makes more sense to me, I have never seen anyone using "Queued".

qwahzi:
Perfect, thanks! The requested changes have been made

dsiganos:
Nice removal of complexity, checking for bootstrap in progress doesn't make much sense, we are always in bootstrapping mode, bootstrapping never stops and it is misleading to pretend otherwise.

pwojcikdev (dismissed):

dsiganos (dismissed):

dsiganos:
Look like unit test 'wallet.enter_password' broke by this change because it is looking for a specific status line.

pwojcikdev (approved):

PR #4006, opened 2022-11-26
Add GitHub workflow for Google Code Sanitizers

thsfs:
The code sanitizers ASAN, ASAN_INT and TSAN were configured to run in a matrix with clang compiler and both DB types (LDMB and RocksDB), composing six combinations.

The code set to be exercised are the binaries `core_test` and `rpc_test`. The workflow doesn't stop if any of these tests doesn't complete. The final result as success means that all steps have ran, regardless errors happening on the test steps.

### DIFF
--- a/.github/workflows/code_sanitizers.yml
+++ b/.github/workflows/code_sanitizers.yml
@@ -1,0 +1,74 @@
+name: Code Sanitizers
+
+on: [push, pull_request, workflow_dispatch]
+
+env:
+  RELEASE: 0
+  artifact: 0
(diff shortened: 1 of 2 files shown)


clemahieu (dismissed):

thsfs:
Run example: https://github.com/nanocurrency/nano-node/actions/runs/3587174261/jobs/6037219738

Changed the workflow to only fail if the report cannot be generated, so this won't affect the "green checks" we have for code pushes and pull requests.

PR #4011, opened 2022-11-30
Update CPACK_PACKAGE_VERSION_* for V24.0RC1

thsfs:
Update the package versions, part of the RC release process:
- Increment `CPACK_PACKAGE_VERSION_PATCH` version on release branch.
- As it is RC1, set  the `CPACK_PACKAGE_VERSION_PRE_RELEASE` version to 0.


CMakeLists.txt
```diff
@@ -49,11 +49,11 @@ endif()
 
 set(CPACK_PACKAGE_VERSION_MAJOR "24")
 set(CPACK_PACKAGE_VERSION_MINOR "0")
-set(CPACK_PACKAGE_VERSION_PATCH "0")
```
dsiganos, line 52: Why are we going to 24.0.1 instead of 24.0.0?
It I not a problem but I wonder why.

dsiganos: I can answer my own question. Because that number is specifically for marking release candidates.

dsiganos (approved):

CMakeLists.txt
thsfs: Yeah, it also matches the RC number, so 0 is not the first version.

PR #4014, opened 2022-11-30
Update GH docker registry login action and credentials

thsfs:
Similar to https://github.com/nanocurrency/nano-node/pull/4013, but this updates also the docker/login-action to a newer version.

### DIFF
--- a/.github/workflows/beta_artifacts.yml
+++ b/.github/workflows/beta_artifacts.yml
@@ -24,7 +24,7 @@ jobs:
       build_tag: ${{ steps.tag_gen.outputs.build_tag }}
       build_number: ${{ steps.tag_gen.outputs.build_number }}
     steps:
-      - uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
+      - uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 #v3.1.0
         with:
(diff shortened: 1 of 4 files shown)


dsiganos:
In the other workflows I did this:
```
-        uses: docker/login-action@adb73476b6e06caddec5db0bc1deacbec8cdd947
-        with:
-          registry: ghcr.io
-          username: ${{ github.repository_owner }}
-          password: ${{ secrets.GHCR_PAT }}
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
```

which is different from what you are doing. Do we want them to be the same and do we understand the difference?

dsiganos (dismissed):

thsfs:
> In the other workflows I did this:
> 
> ```
> -        uses: docker/login-action@adb73476b6e06caddec5db0bc1deacbec8cdd947
> -        with:
> -          registry: ghcr.io
> -          username: ${{ github.repository_owner }}
> -          password: ${{ secrets.GHCR_PAT }}
> +        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u $ --password-stdin
> ```
> 
> which is different from what you are doing. Do we want them to be the same and do we understand the difference?

The only difference is that by using a GH action it is portable and you don't need write your own command line.

dsiganos (approved):

PR #4017, opened 2022-12-02
Updating to use c++20

clemahieu:
Fixing various type ambiguities flagged in c++20 mode.


nano/lib/numbers.cpp
```diff
@@ -144,6 +144,11 @@ bool nano::public_key::decode_account (std::string const & source_a)
 	return error;
 }
 
+bool nano::public_key::operator== (nano::public_key const &) const
+{
+	return true;
```
dsiganos: This can't be right.

dsiganos (changes requested):

dsiganos (dismissed):

clemahieu:
Looks like boost will need a new version.

Boost asio coroutine support: https://www.boost.org/users/history/version_1_74_0.html
c++20 compiler testing: https://www.boost.org/users/history/version_1_77_0.html

guilhermelawless:
Stumbled upon here and remembered this one :)

https://github.com/nanocurrency/nano-node/blob/9bb89ab3232eb603911ae1125a78921ec2d4f591/nano/node/distributed_work_factory.cpp#L57

thsfs:
This change requires updating the documentation website, switching the compiler version from C++17 to C++20.

PR #4019, opened 2022-12-05
Locking around read/write to peering_endpoint.

clemahieu:


thsfs (approved):

nano/node/transport/transport.cpp
```diff
@@ -81,17 +81,20 @@ void nano::transport::channel::send (nano::message & message_a, std::function<vo
...
 nano::endpoint nano::transport::channel::get_peering_endpoint () const
 {
+	nano::unique_lock<nano::mutex> lock{ channel_mutex };
 	if (peering_endpoint)
 	{
 		return *peering_endpoint;
 	}
 	else
 	{
+		lock.unlock ();
```
dsiganos, line 97: It seems a little messy but it is not wrong.

dsiganos (approved):

pwojcikdev (approved):

PR #4021, opened 2022-12-08
Remove unchecked_info::account and unchecked_info::verified

clemahieu:
This PR removes two fields in unchecked_info related to an optimisation that will not be relevant when the ascending bootstrap client is used and with the block processor pipeline improvements.

This reduces complexity around signature checking code and reduces the amount of data tracked for unchecked entries.

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -99,18 +99,17 @@ bool nano::block_processor::half_full ()
 
 void nano::block_processor::add (std::shared_ptr<nano::block> const & block_a)
 {
-	nano::unchecked_info info (block_a, 0);
+	nano::unchecked_info info (block_a);
 	add (info);
 }
 
 void nano::block_processor::add (nano::unchecked_info const & info_a)
 {
 	auto const & block = info_a.block;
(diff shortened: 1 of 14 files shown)


nano/node/blockprocessor.cpp
```diff
@@ -196,28 +195,28 @@ void nano::block_processor::process_verified_state_blocks (std::deque<nano::stat
 		{
 			debug_assert (verifications[i] == 1 || verifications[i] == 0);
 			auto & item = items.front ();
-			auto & [block, account] = item;
+			auto & [block] = item;
```
thsfs, line 198: No need for the brackets, `block = items.front ()`

nano/node/state_block_signature_verification.cpp
```diff
@@ -131,7 +131,7 @@ void nano::state_block_signature_verification::verify_state_blocks (std::deque<v
 		signatures.reserve (size);
 		std::vector<int> verifications;
 		verifications.resize (size, 0);
-		for (auto const & [block, account] : items)
+		for (auto const & [block] : items)
```
thsfs, line 134: Same here, the brackets can be removed.

PR #4025, opened 2022-12-12
Add beta build workflow for the release branch

thsfs:
It looks for the current major version in the root `CMakeLists.txt` in the `develop` branch, then runs the beta build for the latest release branch. So, if current major version in the develop branch is `25`, as example, then the latest release branch must be `releases/v24`.

This is scheduled to run in the same time as the usual beta build for develop, but this runs on top of the release branch (latest available) resuming the build numbering of it.

Also fixes the `dev-build-tag-gen.sh` script to export the variables to a file, so it won't work sourcing anymore. This will avoid possible sourcing errors.


.github/workflows/beta_artifacts_release_branch.yml
```diff
@@ -0,0 +1,161 @@
+name: Beta (Release Branch)
```
dsiganos: It would be nice to state that it builds the last release branch only.
Maybe even put last in the workflow filename.

dsiganos (dismissed):

.github/workflows/beta_artifacts_release_branch.yml
thsfs: Could it be?
- file name changed to beta_artifacts_previous_release
- workflow name "Beta (Previous Release)"

.github/workflows/beta_artifacts_previous_release.yml
```diff
@@ -1,4 +1,4 @@
-name: Beta (Release Branch)
```
dsiganos: It doesn't;t build the previous release, it should build the last release branch.

dsiganos (approved):

PR #4030, opened 2023-01-05
Adding boost as a submodule rather than depending on external references.

clemahieu:
Updating to boost 1.81 and removing CMake target links to Boost::boost which are no longer needed

Added boost_checkout_lite.sh which can be used to only clone required dependencies rather than every boost submodule in the /libs folder. The boost submodule can be initialized with: git submodule update --init boost && ./boost_checkout_lite.sh


clemahieu:
Related to https://github.com/nanocurrency/nano-node/pull/4024

clemahieu:
Disk usage stats:

nano-node repo without boost submodule initialised
nano-node.src % du -sh .
1.5G	.

Size of full boost repo:
boost.src % du -sh .
1.9G

Boost source extracted from archive:
boost_1_74_0$ du -sh .
798M

With boost_checkout_lite.sh
nano-node.src % du -sh .
2.6G

With full recursive checkout
nano-node.src % du -sh .
3.5G

nano/lib/logger_mt.hpp
```diff
@@ -134,6 +134,6 @@ class logger_mt
 private:
 	nano::mutex last_log_time_mutex;
 	std::chrono::steady_clock::time_point last_log_time;
-	boost::log::sources::severity_logger_mt<severity_level> boost_logger_mt;
+	boost::log::sources::severity_logger<severity_level> boost_logger_mt;
```
pwojcikdev: This version of logger (without _mt) is not thread safe and we're using it from multiple threads.

pwojcikdev (changes requested):

CMakeLists.txt
```diff
@@ -591,7 +663,7 @@ if(NANO_FUZZER_TEST)
 endif()
 
 if(NANO_TEST OR RAIBLOCKS_TEST)
-  find_package(Boost 1.70.0 REQUIRED COMPONENTS coroutine context)
+  # find_package(Boost 1.70.0 REQUIRED COMPONENTS coroutine context)
```
dsiganos: This can be removed

CMakeLists.txt
```diff
@@ -363,35 +363,106 @@ else()
...
-endif()
-set(Boost_USE_MULTITHREADED ON)
+add_definitions(-DBOOST_ALL_NO_LIB)
+add_definitions(-DBOOST_LOG_WITHOUT_EVENT_LOG)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/Modules")
 
-find_package(Boost 1.70.0 REQUIRED COMPONENTS filesystem log log_setup thread
-                                              program_options system)
+# find_package(Boost 1.70.0 REQUIRED COMPONENTS filesystem log log_setup thread
```
dsiganos: the version v1.70 n the comment makes no sense anymore

CMakeLists.txt
```diff
@@ -485,6 +556,7 @@ add_definitions(-DCRYPTOPP_DISABLE_SHANI)
 # Fix failing builds after commit
 # https://github.com/weidai11/cryptopp/commit/df9fa62205f2d341e2b1b26595a3a1b6377c60c5
 add_definitions(-DCRYPTOPP_DISABLE_CLMUL)
+add_definitions(-DCRYPTOPP_DISABLE_ASM)
```
dsiganos: If we are going disable functionality that otherwise looks good to have then we need explain why we are disabling it.

ci/actions/windows/build.ps1
```diff
@@ -41,7 +39,7 @@ mkdir build
 Push-Location build
 
 #accessibility of Boost dlls for generating config samples
-$ENV:PATH = "$ENV:PATH;$ENV:BOOST_ROOT\lib64-msvc-14.2"
+$ENV:PATH = "$ENV:PATH"
```
dsiganos: Is this line needed at all?

nano/node/logging.cpp
```diff
@@ -37,20 +36,6 @@ void nano::logging::init (boost::filesystem::path const & application_path_a)
 		}
 
 #ifdef BOOST_WINDOWS
-		if (nano::event_log_reg_entry_exists () || nano::is_windows_elevated ())
```
dsiganos, line 40: Does the removal of this code mean that on windows we do not log to file?

clemahieu: This means we can't log to the Windows Event Log program, a kind of Windows syslog. I'll raise an issue to reimplement this if we can get the GitHub builders to support the required tool, the Windows Message Compiler mc.ex.. Short of this I don't know anyone that currently uses Windows + Event Log so I wouldn't hold up the boost submodule conversion on this point.

CMakeLists.txt
clemahieu: This was a workaround for cross-compiling, I'll remove it.

nano/node/logging.cpp
dsiganos: Windows is our least supported platform, I do not think we should have extra features on the platform we support the least.
As long as logging to file exists on windows, I don't think we should do anymore than that.

pwojcikdev (dismissed):
It's good to see the conversion removed a lot of unnecessary complexity from cmakefiles and source

PR #4045, opened 2023-01-13
Disambiguity of all the implemented locks

thsfs:
Found a warning when building the node for Windows.
```
...
warning C4930: 'std::unique_lock<nano::mutex> lock(nano::mutex)': prototyped function not called (was a variable definition intended?)
...
```
This is related to an ambiguity described by [the wiki](https://en.wikipedia.org/wiki/Most_vexing_parse), where the C++ compiler tries to declare a function as `lock` instead of creating an object for it.

The best approach is to use the uniform initialization `{}` when there is any argument. The warning was issued only for `vote_processor.cpp`, but decided to change all the mutex lock object creations.

### DIFF
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -507,7 +507,7 @@ TEST (bootstrap_processor, DISABLED_pull_requeue_network_error)
 	ASSERT_TIMELY (2s, attempt->frontiers_received);
 	// Add non-existing pull & stop remote peer
 	{
-		nano::unique_lock<nano::mutex> lock (node1->bootstrap_initiator.connections->mutex);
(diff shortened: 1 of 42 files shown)


dsiganos:
I find it annoying that our formatter treat {} differently from ().
It remove a space before { and adds spaces inside the brackets.
```
-                       nano::lock_guard<nano::mutex> lk (connections_mutex);
+                       nano::lock_guard<nano::mutex> lk{ connections_mutex };
```

thsfs:
> I find it annoying that our formatter treat {} differently from (). It remove a space before { and adds spaces inside the brackets.
> 
> ```
> -                       nano::lock_guard<nano::mutex> lk (connections_mutex);
> +                       nano::lock_guard<nano::mutex> lk{ connections_mutex };
> ```

Yes, but in another perspective it is easier to spot the difference.

clemahieu (approved):

PR #4050, opened 2023-01-14
Add GitHub workflow for Valgrind

thsfs:
This workflow is set to run for every created tag that starts with `V`. So, it is intended to generate reports for the automatic beta builds and also for release candidates and official releases.

This is set up to run upon `core_test` and `rpc_test` on self-hosted runners by default, as these tests are heavy and require higher use of CPU. It is also possible to trigger it to run manually.

The self-hosted label for it is `self-hosted-standard` that is intended to run on machines that won't be used for benchmarks.

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -1,0 +1,120 @@
+name: Valgrind
+
+on:
+  push:
+    tags:
+      - V*
+  workflow_dispatch:
+    inputs:
+      runner_type:
+        type: choice
+        options:
+        - self-hosted-standard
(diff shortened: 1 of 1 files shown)


thsfs:
Will mark this as ready for review when the runner setup is finished.

dsiganos:
This PR depends on this other PR on another repo:
"Add a workflow to automatic respawn the GH self-hosted runners"
https://github.com/nanocurrency/infrastructure/pull/102

However, that other PR needs reworking in my opinion.
And ideally, this workflow will not depend on a workflow of another repo.

PR #4055, opened 2023-01-17
Added latest boost version

MyEcoria:
I added the last version of boost for the automatic compilation of this one


clemahieu:
With this recent change https://github.com/nanocurrency/nano-node/pull/4030 it's no longer necessary to download boost externally so I think we should try to remove bootstrap_boost.sh and dependencies.

MyEcoria:
Ok  🙃

PR #4058, opened 2023-01-18
Fix test case backlog.population

clemahieu:


nano/core_test/backlog.cpp
```diff
@@ -72,10 +72,10 @@ block_list_t setup_independent_blocks (nano::test::system & system, nano::node &
  */
 TEST (backlog, population)
 {
```
dsiganos, line 74: Moving the mutex upwards does not fix the problem. The mutex is still destroyed before the node. The node is destroyed much after the test case comes to an end.

dsiganos:
Duplicate of #4059

nano/core_test/backlog.cpp
pwojcikdev: I don't think this is true. `system` destructor does indeed stop the nodes, which means no node threads should be running after that

pwojcikdev (dismissed):

nano/node/backlog_population.cpp
```diff
@@ -13,6 +13,7 @@ nano::backlog_population::backlog_population (const config & config_a, nano::sto
 
 nano::backlog_population::~backlog_population ()
 {
+	stop ();
```
pwojcikdev: This shouldn't be there, while it's not an error, we want to ensure that when node destruction happens all threads have already been stopped.

pwojcikdev (approved):

dsiganos (approved):

PR #4059, opened 2023-01-18
Fix test case backlog.population

dsiganos:
The problem is that node destruction is not managed and any callback we register must not assume that the resources it accesses are alive

### DIFF
--- a/nano/core_test/backlog.cpp
+++ b/nano/core_test/backlog.cpp
@@ -75,32 +75,44 @@ TEST (backlog, population)
 	nano::test::system system{};
 	auto & node = *system.add_node ();
 
-	nano::mutex mutex;
-	std::unordered_set<nano::account> activated;
+	// the node destruction happens in the background and parts of the node object
+	// are still alive after the test case has exited its scope and therefore
+	// the callback we register, must check that the resources it needs are still
+	// alive before using them
 
-	node.backlog.activate_callback.add ([&] (nano::transaction const & transaction, nano::account const & account, nano::account_info const & account_info, nano::confirmation_height_info const & conf_info) {
-		nano::lock_guard<nano::mutex> lock{ mutex };
+	auto mutex = std::make_shared<nano::mutex> ();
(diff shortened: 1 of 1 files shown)


pwojcikdev:
Don't do it like that, this makes code meesy. Moving mutex & set before system ensures the correct destruction order.

nano/core_test/backlog.cpp
```diff
@@ -75,32 +75,44 @@ TEST (backlog, population)
...
 	auto & node = *system.add_node ();
 
-	nano::mutex mutex;
-	std::unordered_set<nano::account> activated;
-
-	node.backlog.activate_callback.add ([&] (nano::transaction const & transaction, nano::account const & account, nano::account_info const & account_info, nano::confirmation_height_info const & conf_info) {
-		nano::lock_guard<nano::mutex> lock{ mutex };
-
-		activated.insert (account);
+	// the node destruction happens in the background and parts of the node object
```
pwojcikdev, line 78: Well, this comment is not true. `system` destructor ensures node will be stopped and therefore all relevant threads will also be stopped.

dsiganos: I am certain that parts of the node survive even after system destruction. I have spent a month last year due to that problem and in the end I resorted to this pattern of using weak pointers in callbacks to make them safe.

dsiganos:
@pwojcikdev if you can stop it failing just by moving the mutex before the system initialisation, then be guest, you do it.
I know for certain that that is not a fix.

nano/core_test/backlog.cpp
pwojcikdev: You are confusing IO threads with component threads.

pwojcikdev: Resorting to weak_ptr is a sign of not having a well managed lifetime of components and usually not necessary. It's possible to have that for IO threads too, the PR for that was mean to land after V24 release.

PR #4060, opened 2023-01-19
V24 release tag version

clemahieu:


dsiganos:
Should you set CPACK_PACKAGE_VERSION_PATCH to 0 too?

Issue #4064, opened 2023-01-21
accounts_balances returns error on unopened blocks


fossephate:
### Summary

Seemingly starting in v24, on multiple nodes, querying an unopened block with `accounts_balances` returns:
`"error": "Account not found"`
The same block queried with `account_balance` returns as expected

### Node version

v24RC2

### Build details

N/A

### OS and version

Windows, Linux

### Steps to reproduce the behavior

1. Send a block to a new, unopened address
2. Call `accounts_balances` on the address

### Expected behavior

https://docs.nano.org/commands/rpc-protocol/#accounts_balances

### Actual behavior

```
{
    "balances": {
        "nano_37rc7hmpbhaxiiuycptn9shusdiefzwbmh6qj4mcfzex98idmqd4rtm9pkum": {
            "error": "Account not found"
        }
    }
}
```

### Possible solution

This appears to be a bug related to changes introduced in v24, as the issue doesn't appear to be present in v23.3 nodes

### Supporting files

Example account / RPC calls:

```
{
  "action": "account_balance",
  "account": "nano_37rc7hmpbhaxiiuycptn9shusdiefzwbmh6qj4mcfzex98idmqd4rtm9pkum"
}
```

```
{
  "action": "accounts_balances",
  "accounts": ["nano_37rc7hmpbhaxiiuycptn9shusdiefzwbmh6qj4mcfzex98idmqd4rtm9pkum"]
}
```

RickiNano:
It is most likely this commit that caused the problem:
https://github.com/nanocurrency/nano-node/commit/580c3c1af562cd72bc50afbe9637bbab60dd9894

qwahzi:
Fix merged and planned for V25:

https://github.com/nanocurrency/nano-node/pull/4075

PR #4066, opened 2023-01-21
Return unopened accounts

RickiNano:
Fix for issue https://github.com/nanocurrency/nano-node/issues/4064

`accounts_balances` RPC was changed in https://github.com/nanocurrency/nano-node/commit/580c3c1af562cd72bc50afbe9637bbab60dd9894
That also changed the response of unopened accounts into an error message
The response of `accounts_balances` should mimic `account_balance` and this PR makes `accounts_balances` return the actual balance values of unopened accounts the way it did before https://github.com/nanocurrency/nano-node/commit/580c3c1af562cd72bc50afbe9637bbab60dd9894
Invalid accounts will still return an error

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -900,19 +900,13 @@ void nano::json_handler::accounts_balances ()
 		if (!ec)
 		{
 			nano::account_info info;
-			if (!node.store.account.get (transaction, account, info))
-			{
-				auto balance = node.balance_pending (account, false);
-				entry.put ("balance", balance.first.convert_to<std::string> ());
-				entry.put ("pending", balance.second.convert_to<std::string> ());
-				entry.put ("receivable", balance.second.convert_to<std::string> ());
-				balances.put_child (account_from_request.second.data (), entry);
-				continue;
-			}
-			else
-			{
-				ec = nano::error_common::account_not_found;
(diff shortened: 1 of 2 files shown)


pwojcikdev (dismissed):
It doesn't make sense to return an error if one of the accounts is not opened yet, it's unfortunate it wasn't caught during testing. Thanks for the PR!

RickiNano:
I targeted this PR against the develop branch but maybe v24 should get a patched version as well?

pwojcikdev:
When it gets merged into develop it can be cherry picked into v24 branch, so it's fine.

nano/node/json_handler.cpp
```diff
@@ -900,19 +900,13 @@ void nano::json_handler::accounts_balances ()
 		if (!ec)
 		{
 			nano::account_info info;
```
Exxenoz: This variable is not needed anymore.

RickiNano: Good catch

pwojcikdev (approved):
Looks good

Exxenoz (approved):
👍

thsfs:
Documentation changes for V24/V25: https://github.com/nanocurrency/nano-docs/pull/670

PR #4073, opened 2023-01-25
Add comment for NANO_DEFAULT_PEER environment variable

qwahzi:
Adding documentation to the config-node.toml generator for the NANO_DEFAULT_PEER environment variable override, per https://github.com/nanocurrency/nano-node/pull/3656

### DIFF
--- a/nano/node/nodeconfig.cpp
+++ b/nano/node/nodeconfig.cpp
@@ -137,7 +137,7 @@ nano::error nano::node_config::serialize_toml (nano::tomlconfig & toml) const
 		work_peers_l->push_back (boost::str (boost::format ("%1%:%2%") % i->first % i->second));
 	}
 
-	auto preconfigured_peers_l (toml.create_array ("preconfigured_peers", "A list of \"address\" (hostname or ipv6 notation ip address) entries to identify preconfigured peers."));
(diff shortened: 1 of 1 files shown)


nano/node/nodeconfig.cpp
```diff
@@ -137,7 +137,7 @@ nano::error nano::node_config::serialize_toml (nano::tomlconfig & toml) const
 		work_peers_l->push_back (boost::str (boost::format ("%1%:%2%") % i->first % i->second));
 	}
 
-	auto preconfigured_peers_l (toml.create_array ("preconfigured_peers", "A list of \"address\" (hostname or ipv6 notation ip address) entries to identify preconfigured peers."));
```
dsiganos, line 140: I do not think that that statement is correct.
NANO_DEFAULT_PEER overrides the peering address.
I think, but I am not sure, that the contents of NANO_DEFAULT_PEER are added to preconfigured_peers.

qwahzi: Oh you're right, `emplace_back`. Let me fix that

pwojcikdev (approved):
Looks good. We still have a few of those environment variables not documented yet, some have the `NANO_TEST_*` prefix, should probably be renamed to simply `NANO_*` for consistency. Of course this is just an idea for the future

dsiganos:
I believe that the NANO_TEST_ variables are a special subgroup of variables used for creating custom networks. Terrible naming though.

dsiganos (approved):

PR #4074, opened 2023-01-25
Add include_only_confirmed option to rpc accounts_balances

### LINKED ISSUE #3797: Add include_only_confirmed option to accounts_balances RPC
The accounts balances RPC returns possibly unconfined data.
It should be changed to return only confirmed data by default, by use of include_only_confirmed option.
Like the RPC account_balance does.

dsiganos:
The RPC accounts_balances returned possibly unconfirmed data. Now it returns confirmed data only, by default.
Unconfirmed data can still be got by setting include_only_confirmed=false.

resolves #3797

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -899,7 +899,8 @@ void nano::json_handler::accounts_balances ()
 		auto account = account_impl (account_from_request.second.data ());
 		if (!ec)
 		{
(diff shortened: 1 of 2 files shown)


clemahieu (approved):

pwojcikdev (approved):

thsfs:
The doc is going to be added for V25.0 by: https://github.com/nanocurrency/nano-docs/pull/670

PR #4081, opened 2023-01-26
Add ledger::account_info as interface to access account information

clemahieu:
Rather than directly accessing the block store, users should use the ledger::account_info interface to query for account information.

Result is returned as an std::optional instead of the less-obvious way it's currently reported through an error code.

Most though not all of the account_store::get references have been updated to use ledger::account_info.

### DIFF
--- a/nano/secure/ledger.cpp
+++ b/nano/secure/ledger.cpp
@@ -583,9 +583,8 @@ void ledger_processor::receive_block (nano::receive_block & block_a)
 #ifdef NDEBUG
 											if (ledger.store.block.exists (transaction, block_a.hashables.source))
 											{
-												nano::account_info source_info;
-												[[maybe_unused]] auto error (ledger.store.account.get (transaction, pending.source, source_info));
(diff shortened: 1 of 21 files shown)


dsiganos (dismissed):

nano/secure/ledger.cpp
```diff
@@ -34,13 +34,12 @@ class rollback_visitor : public nano::block_visitor
 		}
 		if (!error)
 		{
-			nano::account_info info;
-			[[maybe_unused]] auto error (ledger.store.account.get (transaction, pending.source, info));
-			debug_assert (!error);
+			auto info = ledger.account_info (transaction, pending.source);
+			debug_assert (info);
```
pwojcikdev, line 38: I'm wondering whether those asserts inside `rollback_visitor` & `ledger_processor` should become release asserts. The check itself is very cheap and the logic it guards is very critical.

pwojcikdev (dismissed):

PR #4086, opened 2023-01-27
Increase the timeout of Windows unit test workflow

dsiganos:
The windows workflow regurarly times-out.
Windows seem to be slower than other O/Ses.

Ideally, we would reduce the time it takes to build and test but for now I am simply increasing to 2 hours so that we start to get green ticks again.


pwojcikdev (approved):
Might be something to do with the recently disabled ASM compiler for crypto related stuff.

PR #4091, opened 2023-01-31
Fix fork publish inactive

dsiganos:
In test case node.fork_publish_inactive, there is a race condition.
The election and the processing of block send2 happen in parallel.
Usually the election happens first and the send2 block is added to the election.
However, if the send2 block is processed before the election is started then the election does not notice the send2 block.

The test case can be made to pass by ensuring the election is started before the send2 is processed.
However, is this a problem with the test case or this is a problem with the node handling of forks?

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -648,10 +648,13 @@ TEST (node, DISABLED_fork_publish_inactive)
 	node.process_active (send1);
 	ASSERT_TIMELY (5s, node.block (send1->hash ()));
 
+	std::shared_ptr<nano::election> election;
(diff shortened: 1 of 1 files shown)


dsiganos:
NOTES:
* If I ensure the election starts before the send2 is processed then the test case works reliably.
* The block processor tries to add send2 to an election by doing nano::active_transactions::publish() when it resolves to nano::process_result::fork.

clemahieu (dismissed):

LeandroVCastro (dismissed):

pwojcikdev (dismissed):
If you can leave a TODO comment explaining the concerns regarding this behavior it will be great (same as in this PR description)

PR #4092, opened 2023-01-31
Add to_string function to nano::stats class for convenience

dsiganos:
It is useful to be able to print the stats easily from inside a debugging session.

### DIFF
--- a/nano/lib/stats.cpp
+++ b/nano/lib/stats.cpp
@@ -456,6 +456,25 @@ void nano::stats::stop ()
 	stopped = true;
 }
 
+std::string nano::stats::to_string (std::string type)
+{
+	auto sink = log_sink_json ();
+	if (type == "counters")
+	{
+		log_counters (*sink);
+		return sink->to_string ();
+	}
+	else if (type == "samples")
+	{
+		log_samples (*sink);
+		return sink->to_string ();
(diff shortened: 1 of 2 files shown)


nano/lib/stats.cpp
```diff
@@ -456,6 +456,25 @@ void nano::stats::stop ()
 	stopped = true;
 }
 
+std::string nano::stats::to_string (std::string type)
```
pwojcikdev, line 459: Wouldn't it be easier to use and less error prone if the parameter was an enum?

dsiganos: In theory yes, in practice, I don't know. everywhere the stats class uses strings for these things.
So I just followed the theme started by the stats class.

pwojcikdev: Okay, it's good to keep consistency in that case.

pwojcikdev (approved):

nano/lib/stats.cpp
dsiganos: I should clarify, the idea is to use this function from inside a debugger.
And I thought it made more sense to match the way you would use the RPC calls to get the stats.
It is easier to type "samples" in a debugger than to lookup the value of an enum.

PR #4098, opened 2023-02-02
Do not use expose election dummy vote in non election related unit tests

dsiganos:
Election objects always have one dummy vote in them, they never have zero votes. However that is an implementation detail of the election class, which should not be leaked outside.

By using votes_with_weights, we filter out that dummy votes, which always has zero weight and we do not need to leak out its existence.

It also makes those unit tests easier to reason about. The dummy vote is often a confusing complication.


pwojcikdev (approved):
It should be OK to modify and use the `election::votes ()` method to not expose dummy vote too.

PR #4100, opened 2023-02-03
Introduce nano::test::start_election method

dsiganos:
In many tests, we currently use the method node::block_confirm(), which uses scheduler flush and often block processor flush is called just before it too. The flush functions are unreliable and we plan to delete them.
    
I introduce the method start_election with the aim of replacing node::block_confirm in unit tests.
    
I converted a several unit tests to test it. I will do the rest too if there are no objections to this approach.

### DIFF
--- a/nano/test_common/system.cpp
+++ b/nano/test_common/system.cpp
@@ -640,7 +640,7 @@ std::shared_ptr<nano::election> nano::test::system::start_election (nano::node &
 	{
 		if (poll ())
 		{
-			return {};
+			return nullptr;
 		}
 	}
 
@@ -652,7 +652,7 @@ std::shared_ptr<nano::election> nano::test::system::start_election (nano::node &
 	{
 		if (poll ())
 		{
-			return {};
+			return nullptr;
 		}
 		election = node_a.active.election (block_a->qualified_root ());
 	}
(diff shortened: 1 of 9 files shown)


nano/test_common/system.cpp
```diff
@@ -630,3 +630,33 @@ void nano::test::cleanup_dev_directories_on_exit ()
...
+std::shared_ptr<nano::election> nano::test::system::start_election (nano::node & node_a, const std::shared_ptr<nano::block> & block_a)
+{
+	deadline_set (5s);
+
+	// wait until and ensure that the block is in the ledger
+	while (!node_a.block (block_a->hash ()))
+	{
+		if (poll ())
+		{
+			return {};
```
clemahieu: I think this should return nullptr to be more specific.

nano/test_common/system.cpp
```diff
@@ -630,3 +630,33 @@ void nano::test::cleanup_dev_directories_on_exit ()
...
+		}
+	}
+
+	node_a.scheduler.manual (block_a);
```
clemahieu: This seems like appropriate use of the manual queue.

clemahieu (dismissed):

nano/test_common/system.cpp
dsiganos: OK

clemahieu (approved):

PR #4103, opened 2023-02-06
Fixes an issue where token_bucket::last_refill would update even if there were no tokens to add.

clemahieu:
Fixes an issue where token_bucket::last_refill would update even if there were no tokens to add.

Adds a constant for the unlimited rate sentinel.

### DIFF
--- a/nano/core_test/utility.cpp
+++ b/nano/core_test/utility.cpp
@@ -99,6 +99,25 @@ TEST (rate, unlimited)
 	ASSERT_EQ (bucket.largest_burst (), static_cast<size_t> (1e9));
 }
 
+TEST (rate, busy_spin)
+{
+	// Bucket should refill at a rate of 1 token per second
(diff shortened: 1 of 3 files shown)


nano/core_test/utility.cpp
```diff
@@ -159,7 +178,7 @@ TEST (thread_pool_alarm, one)
 	nano::condition_variable condition;
 	workers.add_timed_task (std::chrono::steady_clock::now (), [&] () {
 		{
-			nano::lock_guard<nano::mutex> lock{ mutex };
+			nano::lock_guard<nano::mutex> lock (mutex);
```
dsiganos: There is a bunch of lock mutex constructions here that look like they were made by mistake.

clemahieu: Fixed.

pwojcikdev (approved):

PR #4105, opened 2023-02-06
Remove the UDP channel implementation from the network

thsfs:
Removes the UDP channel implementation from node's network and some other related code.

Please notice:
- After this change the only implemented transport layer is TCP.
- This doesn't remove the UDP option from port mapping.
- There are still some UDP related code for the endpoints that use `boost::asio::ip::udp::endpoint`.
- There are some remaining enumerations related to UDP that should be removed in a further step.
- This PR depends on https://github.com/nanocurrency/nano-node/pull/3923 to be merged first.
- Removes CLI arguments (requires updating the documentation):
  - --disable_udp and --enable_udp
  - --disable_tcp_realtime


nano/node/portmapping.cpp
```diff
@@ -17,7 +17,9 @@ std::string nano::mapping_protocol::to_string ()
 
 nano::port_mapping::port_mapping (nano::node & node_a) :
 	node (node_a),
-	protocols ({ { { "TCP", boost::asio::ip::address_v4::any (), 0, true }, { "UDP", boost::asio::ip::address_v4::any (), 0, !node_a.flags.disable_udp } } })
+	// Kept UDP in the array (set disabled) so the port mapping is still
+	// implemented in case other transport protocols that rely on it is added.
+	protocols ({ { { "TCP", boost::asio::ip::address_v4::any (), 0, true }, { "UDP", boost::asio::ip::address_v4::any (), 0, false } } })
```
thsfs, line 22: Kept the UDP (disabled) here because we don't need to change the port mapping now, it was already disabled, just kept it disabled. UDP port mapping may still be necessary for another transport protocol, like QUIC.

nano/node/cli.cpp
```diff
@@ -30,10 +30,6 @@ std::string nano::error_cli_messages::message (int ev) const
 			return "Database write error";
 		case nano::error_cli::reading_config:
 			return "Config file read error";
-		case nano::error_cli::disable_all_network:
-			return "Flags --disable_tcp_realtime and --disable_udp cannot be used together";
```
thsfs, line 34: It seems --disable_tcp_realtime makes no sense anymore.

nano/node/node.cpp
```diff
@@ -682,8 +675,8 @@ void nano::node::start ()
 			this_l->bootstrap_wallet ();
 		});
 	}
-	// Start port mapping if external address is not defined and TCP or UDP ports are enabled
-	if (config.external_address == boost::asio::ip::address_v6{}.any ().to_string () && (tcp_enabled || !flags.disable_udp))
+	// Start port mapping if external address is not defined and TCP ports are enabled
+	if (config.external_address == boost::asio::ip::address_v6{}.any ().to_string () && tcp_enabled)
```
clemahieu, line 679: any() is a static function so we don't need to instantiate an address.

`boost::asio::ip::address_v6::any ()`

clemahieu (approved):

nano/node/node.cpp
thsfs: Addressing it on:
- https://github.com/nanocurrency/nano-node/pull/4132

Issue #4108, opened 2023-02-08
Re-write nano::election::vote in a more readable way


dsiganos:
Re-write the method `nano::election::vote` in a cleaner more readable style using guard if statements with early exits to avoid the deeply nested if blocks. Also, convert the expression 
`last_vote_l.timestamp < timestamp_a || (last_vote_l.timestamp == timestamp_a && last_vote_l.hash < block_hash_a)`
to be more easily readable by using multiple if statements.

JerzyStanislawski:
I'll take a look at this.

PR #4119, opened 2023-02-12
Refactor election vote

JerzyStanislawski:
Refactoring nano::election::vote method according to https://github.com/nanocurrency/nano-node/issues/4108.
Main goals:

Reduce nesting of if statements
Use early returns
Simplify longer if statement.

### DIFF
--- a/nano/node/election.cpp
+++ b/nano/node/election.cpp
@@ -409,22 +409,23 @@ nano::election_vote_result nano::election::vote (nano::account const & rep, uint
 
 		should_process = max_vote || past_cooldown;
 	}
-	if (should_process)
+	if (!should_process)
 	{
-		last_votes[rep] = { std::chrono::steady_clock::now (), timestamp_a, block_hash_a };
-		if (vote_source_a == vote_source::live)
-		{
-			live_vote_action (rep);
-		}
-
-		node.stats.inc (nano::stat::type::election, vote_source_a == vote_source::live ? nano::stat::detail::vote_new : nano::stat::detail::vote_cached);
-
-		if (!confirmed ())
-		{
-			confirm_if_quorum (lock);
-		}
+		return nano::election_vote_result (false, false);
 	}
-	return nano::election_vote_result (false, should_process);
+	last_votes[rep] = { std::chrono::steady_clock::now (), timestamp_a, block_hash_a };
+	if (vote_source_a == vote_source::live)
+	{
+		live_vote_action (rep);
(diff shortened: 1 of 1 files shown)


clemahieu (approved):

nano/node/election.cpp
```diff
@@ -388,24 +388,26 @@ nano::election_vote_result nano::election::vote (nano::account const & rep, uint
 	else
 	{
 		auto last_vote_l (last_vote_it->second);
```
dsiganos, line 385: This does not look right to me, the two new if statements, do not do exactly the same thing as the one old if statement.

There re 2 problems:
1) a missing equal sign in `last_vote_l.timestamp > timestamp_a`
2) the logic is different even ignoring the missing equal sign

For 2) We had:
```
if (a && b)
    return;
```

And we replaced it with:
```
if (a)
    return;
if (b)
    return;
```

But it should instead be:
```
if (a) {
  if (b) {
    return;
  }
}
```

dsiganos:
Also, that if statement looks like an important one. And it looks like we are missing a test case for it.
Let's add a test case that checks that if statement.

nano/node/election.cpp
JerzyStanislawski: In general you would be right, but in this case timestamps are compared in both `a` and `b`.
So there are 3 possibilities:

1. `last_vote_l.timestamp < timestamp_a` - always process further
2. `last_vote_l.timestamp > timestamp_a` - always return, because `b` is false - `!(last_vote_l.timestamp == timestamp_a)`. Equal sign is omitted on purpose here.
3. `last_vote_l.timestamp == timestamp_a` - depending on hash, that's what `if (b)` checks.

dsiganos: Nice explanation. I agree.
 One final improvement would be to to convert '!(last_vote_l.hash < block_hash_a)' to 'last_vote_l.hash >= block_hash_a'.

dsiganos (approved):

nano/node/election.cpp
JerzyStanislawski: Yeah, but it seems block_hash has only operator< defined - https://github.com/nanocurrency/nano-node/blob/a0dd56cc8ad07832d4f494ab627dae6eb1472089/nano/lib/numbers.hpp#L105.

dsiganos: OK, no big deal then.

PR #4124, opened 2023-02-14
Adding election::behavior getter method, updating references, and adding direct unit test.

clemahieu:


nano/core_test/election.cpp
```diff
@@ -16,6 +16,16 @@ TEST (election, construction)
 	election->transition_active ();
 }
 
+TEST (election, behavior)
+{
+	nano::test::system system (1);
+	auto & node = *system.nodes[0];
+	node.block_confirm (nano::dev::genesis);
```
dsiganos: It would be better to use nano::test::start_election()

dsiganos (approved):

PR #4131, opened 2023-02-15
Add LMDB and RocksDB upgrades to v22: Remove unchecked table

simpago:
In pull request #4021 the binary representation of `unchecked_info` was changed. We have to clear the unchecked table, because the new binary format is not compatible with the old one.

This commit only clears the unchecked table for the LMDB implementation. The RocksDB implementation doesn't have an upgrade mechanism yet.


nano/node/lmdb/lmdb.cpp
thsfs, line 1: Shouldn't we add an `if (verson_l < 22)` in this line? Is it required when running the `upgrade_v21_to_v22` function?

thsfs (changes requested):
For binary compatibility the respective unchecked table handle should be removed as well. This will require some clean up work in the the unchecked_map.?pp and secure/store.?pp files, because they are currently relying on this table to exist even not using it.

Suggest sending another PR for the clean up work before resuming with this.

nano/core_test/block_store.cpp
```diff
@@ -2177,6 +2177,40 @@ namespace lmdb
 		auto transaction (store.tx_begin_read ());
 		ASSERT_LT (19, store.version.get (transaction));
 	}
+
+	TEST (mdb_block_store, upgrade_v21_v22)
+	{
+		if (nano::rocksdb_config::using_rocksdb_in_tests ())
+		{
+			// Don't test this in rocksdb mode
+			return;
```
clemahieu: This should be GTEST_SKIP ();

nano/node/lmdb/block_store.hpp
```diff
@@ -82,6 +82,12 @@ namespace lmdb
 		 */
 		MDB_dbi state_blocks_handle{ 0 };
 
+		/**
+		 * Unchecked bootstrap blocks info. (Removed)
+		 * nano::block_hash -> nano::unchecked_info
+		 */
+		MDB_dbi unchecked_handle{ 0 };
```
clemahieu: Since this is remove we don't need it in the class definition anymore, it can be moved in to the upgrade path.

nano/node/lmdb/lmdb.cpp
```diff
@@ -221,6 +219,11 @@ void nano::lmdb::store::open_databases (bool & error_a, nano::transaction const
 	error_a |= mdb_dbi_open (env.tx (transaction_a), "final_votes", flags, &final_vote_store.final_votes_handle) != 0;
 
 	auto version_l = version.get (transaction_a);
+	if (version_l < 22)
+	{
+		error_a |= mdb_dbi_open (env.tx (transaction_a), "unchecked", flags, &block_store.unchecked_handle) != 0;
```
clemahieu: unchecked_handle can go here

nano/node/lmdb/lmdb.cpp
```diff
@@ -866,7 +880,7 @@ MDB_dbi nano::lmdb::store::table_to_dbi (tables table_a) const
 		case tables::pending:
 			return pending_store.pending_handle;
 		case tables::unchecked:
```
clemahieu: This enum should be removed since the table is removed.

thsfs:
*Notice:* if merged now, this breaks the RocksDB "binary compatibility" since this doesn't add an upgrade code for it and this bumps the database version to 22.

nano/node/lmdb/lmdb.cpp
thsfs: Removed it from the case only. Will remove the remaining together with the unchecked in RocksDB.

thsfs:
Notice: if merged now, this breaks the RocksDB "binary compatibility" since this doesn't add an upgrade code for it and this bumps the database version to 22.

nano/node/rocksdb/rocksdb.cpp
```diff
@@ -96,15 +96,86 @@ nano::rocksdb::store::store (nano::logger_mt & logger_a, boost::filesystem::path
...
-	if (!error)
+	if (error)
+	{
+		return;
+	}
+
+	generate_tombstone_map ();
+	small_table_factory.reset (::rocksdb::NewBlockBasedTableFactory (get_small_table_options ()));
+
+	// TODO: get_db_options () registers a listener for resetting tombstones, needs to check if it is a problem calling it more than once.
```
thsfs, line 107: Good thing to be reviewed.

clemahieu (approved):
I was able to verify the table was dropped and version number increased.

nano/node/rocksdb/rocksdb.cpp
clemahieu: Encapsulating this to only be called once looks correct.

PR #4133, opened 2023-02-15
This change creates an election_occupancy class which tracks and limits how many elections can be spawned.

clemahieu:
The class is a facade around the active_transactions class and exposes a query election_occupancy::available to see if there is capacity available and a method election_occupancy::insert which checks for availability before inserting an election. Elections are inserted and tracked by the election_occupancy class. Coherency is maintained in a lazy manor by holding weak_ptrs to nano::elections and removing them if the pointer has expired or the election is no longer active as reported by active_transactions.


nano/node/election_occupancy.cpp
```diff
@@ -0,0 +1,53 @@
...
+	return result;
+}
+
+nano::election_insertion_result nano::election_occupancy::insert (std::shared_ptr<nano::block> const & block)
+{
+	if (!available ())
+	{
+		return { nullptr, false };
+	}
+	nano::unique_lock<nano::mutex> lock{ active.mutex };
```
pwojcikdev: Locking mutexes of outside classes is something I went out of my way to eliminate. It's a very bad pattern, makes reasoning about multithreaded code unnecessarily difficult.

nano/node/hinted_scheduler.cpp
```diff
@@ -72,7 +81,7 @@ bool nano::hinted_scheduler::run_one (nano::uint128_t const & minimum_tally)
 			{
 				// Try to insert it into AEC as hinted election
 				// We check for AEC vacancy inside our predicate
-				auto result = node.active.insert_hinted (block);
+				auto result = elections.insert (block);
```
pwojcikdev: Won't the election behaviour type be lost here? If we submit hinted elections as normal ones then they will have a default 5 minute timeout, which is needlessly long, plus it will be impossible to track their number.

pwojcikdev (commented):
I like the facade idea, it's really neat, however I'm concerned about the computational complexity of this solution. Doing a O(n) scan on every update is a no-go from me, it just won't scale well with a large number of elections and high cps.

nano/node/hinted_scheduler.cpp
clemahieu: Made a modification so behaviour is still passed in.

clemahieu:
> I like the facade idea, it's really neat, however I'm concerned about the computational complexity of this solution. Doing a O(n) scan on every update is a no-go from me, it just won't scale well with a large number of elections and high cps.

Replaced the std::list with std::unordered_set.

PR #4134, opened 2023-02-15
Simplify and reduce the nodes needed for test peer_container.list_fanout

dsiganos:
The test was creating 20 nodes which is too much for windows due to lmdb map size issues. Reduced to 10 to see if it works fine on windows.

Also simplified the test a bit.


thsfs:
It could be only 2 peers added so the `fanout` math is not equal to 1 and it is the minimal number of nodes for `fanout` > 1.

thsfs (dismissed):

Issue #3066, opened 2020-12-10
Suggestion: allow port to be defined in preconfigured_peers


Srayman:
Currently preconfigured_peers in config-node.toml only supports adding IP's or DNS entries.  The default node port is used to establish initial connections.  

When setting up a local network, if the default port is changed via "peering_port" in config-node.toml on the primary node then other nodes will not be able to establish a connection since they only try the node port that is set either through ENV settings or coded into the node.

dsiganos:
Currently preconfigured_peers config only accepts ip addresses and dns names. It does not allow for the specification of ports.
When I tried yesterday to setup a private network of 2 nodes, I really needed that feature, it would have made my life so much easier. And it needs to be easy, for productivity’s sake, to setup a local network of 2 nodes without having to do tricks.
We could trivially add support for address and port in preconfigured peers by allowing a port number as an optional postfix parameter. That means it will be backwards compatible and fairly easy to use and clear.
For example:
Peer is at address 1.2.3.4 and port 1075
```
preconfigured_peers = ["::ffff:1.2.3.4,1075"]
```
The use of a comma to separate the address from the port would make the parsing trivial.
However, the use of a comma is controversial. the core team would prefer to use the more standard ":" notation.
But the colon notation does not work well with IPv6 addresses.
A solution for the use of a colon, is to apply the same workaround people use for using IPv6 addresses with ports.
To wrap the address part inside square brackets when specifying the port: `[::ffff:1.2.3.4]:1075`.

This is an example of some of the different ways the address and port could be specified using the colon plus square brackets rule:
```
preconfigured_peers = [
    "1.2.3.4",
    "::ffff:1.2.3.4",
    "google.com",
    "[1.2.3.4]:1075",
    "[::ffff:1.2.3.4]:1075",
    "[google.com]:1075"
]
```

A different configuration entry, the `node.work_peers` uses address and pair notation separated by a colon.
However, that always requires the port part to exist. The port and port separator colon are not optional.
So that one is easier to parse and not applicable in preconfigured_peers case where we also want to preserve backwards compatibility.
However, the work_peers could be extended to operate in a more flexible manner like the preconfigured_peers and reuse the parse functionality of preconfigured_peers.

clemahieu:
Relevant RFC https://www.ietf.org/rfc/rfc2732.txt I think we should only use the square brackets in the same way it would be used for an ipv6 address in the browser i.e. they wouldn't be used for ipv4 or DNS names.

[...]

PR #4145, opened 2023-02-21
Support address:port format for preconfigured_peers entries.

JerzyStanislawski:
Enable defining preconfigured_peers with port according to #3066 including IPv6 support.
The same method for parsing entries is used for work_peers with the difference that port is not required for preconfigured_peers to preserve backward compatibility.


dsiganos:
Thank you for doing this Jerzy, can you write documentation about the format accepted?

JerzyStanislawski:
Described in PR - https://github.com/nanocurrency/nano-node/issues/3066#issuecomment-1439087058
If it's approved, should we put it somewhere in nano-docs?

PR #4160, opened 2023-02-27
Decouple message deserializer from the TCP socket

thsfs:
The message deserializer should not know about any socket class. If so, it would be bound to a specific channel implementation. Replaced this dependency by a function that reads and provides the required data. Modified the inproc channel to use the message deserializer instead of the old message parser, that will be removed.

### DIFF
--- a/nano/node/transport/message_deserializer.cpp
+++ b/nano/node/transport/message_deserializer.cpp
@@ -1,12 +1,14 @@
 #include <nano/node/node.hpp>
 #include <nano/node/transport/message_deserializer.hpp>
 
-nano::transport::message_deserializer::message_deserializer (nano::network_constants const & network_constants_a, nano::network_filter & publish_filter_a, nano::block_uniquer & block_uniquer_a, nano::vote_uniquer & vote_uniquer_a) :
+nano::transport::message_deserializer::message_deserializer (nano::network_constants const & network_constants_a, nano::network_filter & publish_filter_a, nano::block_uniquer & block_uniquer_a, nano::vote_uniquer & vote_uniquer_a,
+channel_read_fn_type channel_read_fn_a) :
 	read_buffer{ std::make_shared<std::vector<uint8_t>> () },
 	network_constants_m{ network_constants_a },
 	publish_filter_m{ publish_filter_a },
 	block_uniquer_m{ block_uniquer_a },
-	vote_uniquer_m{ vote_uniquer_a }
+	vote_uniquer_m{ vote_uniquer_a },
+	channel_read_fn{ std::move (channel_read_fn_a) }
 {
 	read_buffer->resize (MAX_MESSAGE_SIZE);
 }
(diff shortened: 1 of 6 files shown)


nano/node/transport/message_deserializer.cpp
```diff
@@ -1,27 +1,29 @@
...
 	debug_assert (callback);
 
 	status = parse_status::none;
 
-	// Increase timeout to receive TCP header (idle server socket)
-	auto prev_timeout = socket->get_default_timeout_value ();
-	socket->set_default_timeout_value (network_constants_m.idle_timeout);
-
-	socket->async_read (read_buffer, HEADER_SIZE, [this_l = shared_from_this (), socket, callback = std::move (callback), prev_timeout] (boost::system::error_code const & ec, std::size_t size_a) {
+	if (!channel_read_fn)
```
clemahieu: It seems this should be an assert. By comparison if the callback is not set this function already asserts.

clemahieu: This also brings the behaviour in line with received_header.

nano/node/transport/message_deserializer.cpp
```diff
@@ -1,27 +1,29 @@
...
-nano::transport::message_deserializer::message_deserializer (nano::network_constants const & network_constants_a, nano::network_filter & publish_filter_a, nano::block_uniquer & block_uniquer_a, nano::vote_uniquer & vote_uniquer_a) :
+nano::transport::message_deserializer::message_deserializer (nano::network_constants const & network_constants_a, nano::network_filter & publish_filter_a, nano::block_uniquer & block_uniquer_a, nano::vote_uniquer & vote_uniquer_a,
+channel_read_fn_type channel_read_fn_a) :
 	read_buffer{ std::make_shared<std::vector<uint8_t>> () },
 	network_constants_m{ network_constants_a },
 	publish_filter_m{ publish_filter_a },
 	block_uniquer_m{ block_uniquer_a },
-	vote_uniquer_m{ vote_uniquer_a }
+	vote_uniquer_m{ vote_uniquer_a },
+	channel_read_fn{ std::move (channel_read_fn_a) }
```
clemahieu: We should probably just assert here if the channel is nullptr.

nano/node/transport/message_deserializer.hpp
```diff
@@ -45,19 +42,20 @@ namespace transport
 
 		parse_status status;
 
-		message_deserializer (network_constants const &, network_filter &, block_uniquer &, vote_uniquer &);
+		using channel_read_fn_type = std::function<void (std::shared_ptr<std::vector<uint8_t>> const &, size_t, std::function<void (boost::system::error_code const &, std::size_t)>)>;
```
pwojcikdev, line 45: The only thing I would change is name this better, it has nothing to do with channel, imo a short and suiting name is just `read_query_t`

nano/node/transport/message_deserializer.hpp
```diff
@@ -90,6 +88,7 @@ namespace transport
 		nano::network_filter & publish_filter_m;
 		nano::block_uniquer & block_uniquer_m;
 		nano::vote_uniquer & vote_uniquer_m;
+		channel_read_fn_type channel_read_fn;
```
pwojcikdev: The only thing I would change is name this better, it has nothing to do with channel, imo a short and precise name is just `read_query_t read_query;`

pwojcikdev (dismissed):

clemahieu (approved):

PR #4161, opened 2023-02-27
Internally check if the block processor needs to drop blocks.

clemahieu:
Internally check if the block processor needs to drop blocks, rather than making it a requirement of the caller.

Adds a stat counter when the block_processor drops a block as this is new behavior.

### DIFF
--- a/nano/node/blockprocessor.cpp
+++ b/nano/node/blockprocessor.cpp
@@ -97,6 +97,11 @@ void nano::block_processor::add (std::shared_ptr<nano::block> const & block_a)
 
 void nano::block_processor::add (nano::unchecked_info const & info_a)
 {
+	if (full ())
+	{
+		node.stats.inc (nano::stat::type::blockprocessor, nano::stat::detail::drop);
+		return;
+	}
 	auto const & block = info_a.block;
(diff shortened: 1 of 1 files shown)


thsfs (commented):
Noticed that the `add_local` function doesn't check for `full ()`. Isn't it necessary there?

pwojcikdev:
I'm not 100% sure about limiting the add_local, if the source of those blocks is RPC then overriding user action without any feedback might be unexpected.

pwojcikdev (approved):

thsfs (approved):

thsfs:
> I'm not 100% sure about limiting the add_local, if the source of those blocks is RPC then overriding user action without any feedback might be unexpected.

Agree on adding a feedback to the RPC for block processor is `full`.

clemahieu:
Looking through call sites, the add_local is only called by process_local_async, when the "process" RPC is called with "async".

In our documentation we already mention observing when a block is inserted actually, via the websocket stream. Either we can document here or somewhere else that the block processing queue can internally drop and the stat to check for.
https://docs.nano.org/commands/rpc-protocol/#process

PR #4167, opened 2023-03-03
Add batch block processing result observer to block_processor

clemahieu:
- Implements the batch block processing observer from https://github.com/nanocurrency/nano-node/pull/4158.
- Unifies notifications between block_processor observers 'processed' and 'batch_processed'.
- New observer template which puts functionality in to its own file with minimal coupling.
- This will be merge-committed to preserve functional change boundaries.

### node::process_local cleanup
- Can now block without needing to directly call ::process_one. Now it will be placed in to the block_processor queue.
- Doesn't need to pass block_origin through process_one to process_live.
- Return an optional since it may not be possible to return a result if the block_processor stops.

### Block processor cleanup
- Remove block_processor::wait_write workaround which was in effect trying to schedule database batches for node::process_local
- Removes nano::block_origin as unneeded when replaced with observer
- ::process_one is now fully encapsulated
  
### block_processor::process_live cleanup
- Removes block_origin as a parameter and uses a block_processor observer instead to do block publishing.
- Removes unused parameters
- Connect takes an enable flag which can be directly pulled from the node flags.

### DIFF
--- a/nano/node/blocking_observer.cpp
+++ b/nano/node/blocking_observer.cpp
@@ -3,11 +3,8 @@
 
 void nano::blocking_observer::connect (nano::block_processor & block_processor)
 {
-	block_processor.batch_processed.add ([this] (auto const & items) {
-		for (auto const & [result, block] : items)
-		{
-			observe (result, block);
-		}
+	block_processor.processed.add ([this] (auto const & result, auto const & block) {
+		observe (result, block);
 	});
 }
 
--- a/nano/node/blocking_observer.cpp
+++ b/nano/node/blocking_observer.cpp
@@ -1,0 +1,55 @@
+#include <nano/node/blocking_observer.hpp>
+#include <nano/node/blockprocessor.hpp>
+
+void nano::blocking_observer::connect (nano::block_processor & block_processor)
+{
+	block_processor.batch_processed.add ([this] (auto const & items) {
+		for (auto const & [result, block] : items)
+		{
+			observe (result, block);
+		}
+	});
+}
+
+void nano::blocking_observer::stop ()
+{
+	nano::unique_lock<nano::mutex> lock{ mutex };
+	stopped = true;
+	auto discard = std::move (blocking);
+	// Signal broken promises outside lock
+	lock.unlock ();
+	discard.clear (); // ~promise future_error
+}
+
+void nano::blocking_observer::observe (nano::process_return const & result, std::shared_ptr<nano::block> block)
+{
+	nano::unique_lock<nano::mutex> lock{ mutex };
+	auto existing = blocking.find (block);
+	if (existing != blocking.end ())
+	{
+		auto promise = std::move (existing->second);
+		blocking.erase (existing);
+		// Signal promise outside of lock
+		lock.unlock ();
+		promise.set_value (result);
+	}
+}
+
+std::future<nano::process_return> nano::blocking_observer::insert (std::shared_ptr<nano::block> block)
+{
+	nano::lock_guard<nano::mutex> lock{ mutex };
+	if (stopped)
+	{
+		std::promise<nano::process_return> promise;
+		return promise.get_future (); // ~promise future_error
+	}
+	auto iterator = blocking.emplace (block, std::promise<nano::process_return>{});
+	return iterator->second.get_future ();
+}
+
+bool nano::blocking_observer::exists (std::shared_ptr<nano::block> block)
+{
+	nano::lock_guard<nano::mutex> lock{ mutex };
+	auto existing = blocking.find (block);
+	return existing != blocking.end ();
+}
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -1064,14 +1064,14 @@ TEST (active_transactions, conflicting_block_vote_existing_election)
 				.build_shared ();
 	auto vote_fork (std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, nano::vote::timestamp_max, nano::vote::duration_max, std::vector<nano::block_hash>{ fork->hash () }));
 
-	ASSERT_EQ (nano::process_result::progress, node.process_local (send).code);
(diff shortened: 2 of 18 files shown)


nano/node/blocking_observer.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+std::future<nano::process_return> nano::blocking_observer::insert (std::shared_ptr<nano::block> block)
+{
+	nano::lock_guard<nano::mutex> lock{ mutex };
+	if (stopped)
+	{
+		std::promise<nano::process_return> promise;
+		return promise.get_future (); // ~promise future_error
+	}
+	auto iterator = blocking.emplace (block, std::promise<nano::process_return>{});
```
thsfs, line 43: progress is the default enum value, but wouldn't be better to have it written in the constructor?

thsfs (approved):

nano/node/block_broadcast_strategy.cpp
```diff
@@ -0,0 +1,51 @@
...
+				break;
+		}
+	});
+}
+
+void nano::block_broadcast_strategy::observe (std::shared_ptr<nano::block> block)
+{
+	nano::lock_guard<nano::mutex> lock{ mutex };
+	auto existing = local.find (block);
+	if (existing != local.end ())
```
thsfs: is the lock necessary after `local.find (block)`?

nano/node/block_broadcast_strategy.cpp
```diff
@@ -0,0 +1,51 @@
...
+	else
+	{
+		if (block_arrival.recent (block->hash ()))
+		{
+			network.flood_block (block, nano::transport::buffer_drop_policy::limiter);
+		}
+	}
+}
+
+void nano::block_broadcast_strategy::set_local (std::shared_ptr<nano::block> block)
```
thsfs: Where is this called?

thsfs (changes requested):

nano/node/blocking_observer.cpp
clemahieu: This is a promise for an enum value rather than the value itself.

nano/node/block_broadcast_strategy.cpp
clemahieu: This was missing, pushed a new version.

nano/node/block_broadcast_strategy.cpp
```diff
@@ -0,0 +1,51 @@
...
+}
+
+void nano::block_broadcast_strategy::connect (nano::block_processor & block_processor, bool enabled)
+{
+	if (!enabled)
+	{
+		return;
+	}
+	block_processor.processed.add ([this] (auto const & result, auto const & block) {
+		switch (result.code)
```
clemahieu: No items are removed from local.

nano/node/block_broadcast_strategy.cpp
clemahieu: Modified it to release after checking existence.

nano/node/block_broadcast_strategy.cpp
thsfs: Will you add a cleanup routine?

thsfs (dismissed):

pwojcikdev:
Not sure about the way block origin is removed. You are trying to split data into multiple containers that naturally wants to be wrapped in a structure and move through the pipeline. It is more error prone and unnecessary, and in few places that it happens (election related code) it's a pain to work with. It's more or less the same workaround as with block arrival class, instead of tracking the source of each block (`local/live/bootsrap`) next to the block and then choosing broadcasting strategy based on that, we try to synchronise state across multiple components.

nano/node/block_broadcast_strategy.cpp
```diff
@@ -0,0 +1,53 @@
+#include <nano/node/block_arrival.hpp>
+#include <nano/node/block_broadcast_strategy.hpp>
+#include <nano/node/blockprocessor.hpp>
+#include <nano/node/network.hpp>
+
+nano::block_broadcast_strategy::block_broadcast_strategy (nano::network & network, nano::block_arrival & block_arrival) :
```
pwojcikdev: Is a `_stategy` a suitable name for that? Seems to me it is a `block_broadcast` class that selects a broadcasting strategy for each block.

clemahieu: Renamed.

nano/node/block_broadcast_strategy.cpp
clemahieu: Added cleanup.

clemahieu:
> Not sure about the way block origin is removed. You are trying to split data into multiple containers that naturally wants to be wrapped in a structure and move through the pipeline. It is more error prone and unnecessary, and in few places that it happens (election related code) it's a pain to work with. It's more or less the same workaround as with block arrival class, instead of tracking the source of each block (`local/live/bootsrap`) next to the block and then choosing broadcasting strategy based on that, we try to synchronise state across multiple components.

I can see the reasoning for that and I think we'd be able to do something like that with the block_pipeline::context when merged. https://github.com/nanocurrency/nano-node/blob/6299819eaeaafcd302ca42170e6532de9272df6f/nano/node/block_pipeline/context.hpp

pwojcikdev:
> > Not sure about the way block origin is removed. You are trying to split data into multiple containers that naturally wants to be wrapped in a structure and move through the pipeline. It is more error prone and unnecessary, and in few places that it happens (election related code) it's a pain to work with. It's more or less the same workaround as with block arrival class, instead of tracking the source of each block (`local/live/bootsrap`) next to the block and then choosing broadcasting strategy based on that, we try to synchronise state across multiple components.
> 
> I can see the reasoning for that and I think we'd be able to do something like that with the block_pipeline::context when merged. https://github.com/nanocurrency/nano-node/blob/6299819eaeaafcd302ca42170e6532de9272df6f/nano/node/block_pipeline/context.hpp

That looks like a perfect fit. The current solution is OK as a temporary measure until we get the block pipeline merged.

pwojcikdev (dismissed):

PR #4181, opened 2023-03-10
Add the process_live_dipatcher class

thsfs:
Move the `process_live` function to the `process_live_dispatcher` class. This is an observer for the confirmed processed blocks, that also calls `process_live`.

Remove the `block_post_events` class.

### DIFF
--- a/nano/node/blockprocessor.hpp
+++ b/nano/node/blockprocessor.hpp
@@ -68,7 +68,6 @@ private:
 	nano::process_return process_one (nano::write_transaction const &, block_post_events &, std::shared_ptr<nano::block> block, bool const = false);
 	void queue_unchecked (nano::write_transaction const &, nano::hash_or_account const &);
 	std::deque<processed_t> process_batch (nano::unique_lock<nano::mutex> &);
-	void process_live (nano::transaction const &, std::shared_ptr<nano::block> const &);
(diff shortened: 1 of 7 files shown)


nano/node/blockprocessor.hpp
```diff
@@ -56,7 +56,7 @@ class block_processor final
 
 public: // Events
 	using processed_t = std::pair<nano::process_return, std::shared_ptr<nano::block>>;
-	nano::observer_set<nano::process_return const &, std::shared_ptr<nano::block>> processed;
+	nano::observer_set<nano::process_return const &, std::shared_ptr<nano::block>, nano::transaction const &> processed;
```
clemahieu: I'd rather we didn't put transactions through this callback. Rather I think this should use the batch_processed observer and within that observer, create the required transaction so the observers themselves don't need to be modified.

nano/node/process_live_dispatcher.cpp
```diff
@@ -0,0 +1,59 @@
...
+	inactive_vote_cache{ inactive_vote_cache },
+	websocket{ websocket }
+{
+}
+
+void nano::process_live_dispatcher::connect (nano::block_processor & block_processor)
+{
+	block_processor.batch_processed.add ([this] (auto const & batch) {
+		auto const transaction = ledger.store.tx_begin_read ();
+		for (auto const & [result, block] : batch)
```
thsfs, line 23: One thing I didn't like is that each use of this observer is multiplier to the iteration list. Not every data is used, the size of the list could be a smaller "target" set in a future improvement.

@clemahieu

clemahieu (approved):

Issue #4057, opened 2023-01-18
Convert boost::filesystem references to std::filesystem


clemahieu:
std::filesystem is supported as of C++17, replace usages of boost::filesystem with std::filesystem.

meyediri:
I would like to take this one

RickiNano:
@qwahzi 
This issue can be closed now
There is also a very old PR #4183 that can also be closed

qwahzi:
Awesome, thank you Ricki!

PR #4183, opened 2023-03-15
Convert boost::filesystem references to std::filesystem

meyediri:
associated issue #4057

### DIFF
--- a/nano/core_test/common.hpp
+++ b/nano/core_test/common.hpp
@@ -1,0 +1,41 @@
+#pragma once
+
+#include <nano/core_test/testutil.hpp>
+#include <nano/node/testing.hpp>
+
+#include <gtest/gtest.h>
+
+#include <numeric>
+
+using namespace std::chrono_literals;
+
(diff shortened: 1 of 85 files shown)


nano/nano_rpc/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ target_link_libraries(
   nano_rpc
   rpc
   secure
-  Boost::filesystem
+  std::filesystem
```
clemahieu, line 7: This line should be removed.

nano/core_test/common.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
```
clemahieu, line 1: It looks like some other files got mixed in.

clemahieu (changes requested):
Also needs a formatting run ci/clang-format-do.sh

qwahzi:
Closed by #4057 and #4315

PR #4195, opened 2023-03-21
Fix or suppress Windows build warnings

thsfs:
Fixes Windows build warnings from nano source-code. This doesn't fix for the warnings for the submodules.

### DIFF
--- a/nano/node/bootstrap/bootstrap_ascending.cpp
+++ b/nano/node/bootstrap/bootstrap_ascending.cpp
@@ -445,8 +445,7 @@ void nano::bootstrap_ascending::send (std::shared_ptr<nano::transport::channel> 
 	nano::asc_pull_req::blocks_payload request_payload;
 	request_payload.start = tag.start;
 	request_payload.count = pull_count;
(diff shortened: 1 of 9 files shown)


nano/node/bootstrap/bootstrap_ascending.cpp
```diff
@@ -434,6 +434,10 @@ nano::bootstrap_ascending::id_t nano::bootstrap_ascending::generate_id ()
 	return id;
 }
 
+#ifdef _WIN32
+#pragma warning(push)
+#pragma warning(disable : 4267) // warning C4267: '=': conversion from 'size_t' to 'uint8_t', possible loss of data -- wrongly infered by the compiler
```
clemahieu: Is this not able to be solved without disabling the warning?

clemahieu (dismissed):

nano/node/bootstrap/bootstrap_ascending.cpp
thsfs: It seems only using curly braces is enough. Removed the `#pragma warning`.

PR #4200, opened 2023-03-27
Changes the semantics of election::confirmed and clean up several unit tests

clemahieu:
Changes the semantics of election::confirmed to return if the block is durably confirmed on disk, rather than simply being confirmed in memory. This was causing subtle race conditions or unnecessary extra checking to ensure the confirmation has been committed to disk.

Rewriting test to use a block that is not confirmed on disk.
Cleaning up confirmation_height.conflict_rollback_cemented and reduce complexity.
Cleaning up election.quorum_minimum_confirm_success and fixing race condition where confirmation happens asynchronously.
Fixing race condition in election.quorum_minimum_update_weight_before_quorum_checks when checking asynchronous election confirmation.
Cleaning up and simplifying node.rollback_gap_source Removing some test calls to block_processor::flush which is being phased out.

### DIFF
--- a/nano/node/election.hpp
+++ b/nano/node/election.hpp
@@ -116,6 +116,12 @@ public: // State transitions
 	void transition_active ();
 
 public: // Status
+	// Returns true when the election is confirmed in memory
+	// Elections will first confirm in memory once sufficient votes have been received
+	bool status_confirmed () const;
+	// Returns true when the winning block is durably confirmed in the ledger.
+	// Later once the confirmation height processor has updated the confirmation height it will be confirmed on disk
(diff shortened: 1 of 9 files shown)


nano/node/election.hpp
```diff
@@ -116,7 +116,10 @@ class election final : public std::enable_shared_from_this<nano::election>
 	void transition_active ();
 
 public: // Status
+	// Returns true when the winning block is durably confirmed in the ledger.
```
dsiganos, line 122: From the comments in these 4 lines, I do not fully understand the difference between confirmed and status_confirmed. The comments need to be improved.

PR #4205, opened 2023-04-04
Add throttling functionality to the ascending bootstrapper

clemahieu:
This adds the ability for the ascending bootstrapper to slow itself down when no progress is being made. This inability to slow down had a side effect of impacting our unit testing and required us to turn the request rate very low compared to synthetic and beta tests.

With the exception of the aggressive_flooding test, unit tests will now pass when the rate is up to 1024 from 128 having tested with 4096 previously.

This disables the aggressive_flooding test, which will be moved to load tests, since its resource requirement is high.

### DIFF
--- a/nano/node/bootstrap/bootstrap_ascending.cpp
+++ b/nano/node/bootstrap/bootstrap_ascending.cpp
@@ -9,6 +9,35 @@
 
 using namespace std::chrono_literals;
 
+nano::bootstrap_ascending::throttle::throttle (size_t count) :
+	successes{ count },
+	samples{ count, true }
+{
+}
+
+bool nano::bootstrap_ascending::throttle::throttled () const
+{
+	return successes == 0;
+}
+
+void nano::bootstrap_ascending::throttle::add (bool sample)
+{
+	if (samples.front ())
+	{
+		--successes;
+	}
+	samples.push_back (sample);
+	if (sample)
+	{
+		++successes;
+	}
+}
+
+size_t nano::bootstrap_ascending::throttle::success_count () const
+{
+	return successes;
+}
+
 /*
  * database_iterator
  */
@@ -89,6 +118,11 @@ nano::account nano::bootstrap_ascending::buffered_iterator::next ()
 	return *(*this);
 }
 
+bool nano::bootstrap_ascending::buffered_iterator::warmup () const
+{
+	return warmup_m;
+}
+
 void nano::bootstrap_ascending::buffered_iterator::fill ()
 {
 	debug_assert (buffer.empty ());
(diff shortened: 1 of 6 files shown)


nano/node/bootstrap/bootstrap_ascending.hpp
```diff
@@ -35,6 +35,24 @@ class bootstrap_ascending
 {
 	using id_t = uint64_t;
 
+	// Class used to throttle the ascending bootstrapper once it reaches a steady state
+	// Tracks verify_result samples and signals throttling if no tracked samples have gotten results
+	class throttle
+	{
+	public:
+		// Initialized with all true samples
+		throttle (size_t size);
```
thsfs, line 44: good place for using explicit.

nano/node/bootstrap/bootstrap_ascending.cpp
```diff
@@ -9,6 +9,35 @@
...
+		--successes;
+	}
+	samples.push_back (sample);
+	if (sample)
+	{
+		++successes;
+	}
+}
+
+size_t nano::bootstrap_ascending::throttle::success_count () const
```
thsfs, line 36: this has no use yet.

nano/node/bootstrap/bootstrap_ascending.cpp
```diff
@@ -112,6 +146,10 @@ void nano::bootstrap_ascending::buffered_iterator::fill ()
 		{
 			buffer.push_back (*pending_iterator);
 		}
+		else
+		{
+			warmup_m = false;
```
thsfs, line 151: this can be true only for the first time the fill() function is called. Is that right?

nano/node/bootstrap/bootstrap_ascending.cpp
```diff
@@ -678,12 +717,23 @@ bool nano::bootstrap_ascending::run_one ()
 	return success;
 }
 
+void nano::bootstrap_ascending::throttle_if_needed ()
+{
+	if (!iterator.warmup () && throttle.throttled ())
```
pwojcikdev, line 722: Doesn't 'throttle' object access need a lock?

clemahieu: As it's written it assumes synchronisation is handled externally.
This seemed appropriate since it's a 1-to-1 association between bootstrap_ascending instance and throttle instance.

nano/node/bootstrap/bootstrap_ascending.cpp
clemahieu: This should be true until after the iterator has gone through both the account and pending tables once.

nano/node/bootstrap/bootstrap_ascending.cpp
pwojcikdev: Why then are you locking accesses of this class in other places, I believe your explanation is not correct and it is in fact being used from multiple threads.

pwojcikdev: In other words, just move the lock acquired on line 725 above this if statement.

PR #4207, opened 2023-04-05
Updating throttle class

clemahieu:
Given feedback from https://github.com/nanocurrency/nano-node/pull/4205

Marking throttle constructor explicit.
Removing unused function.


thsfs (commented):
LGTM.

PR #4208, opened 2023-04-05
Stats bug fix

Fiono11:
Small bug in the unchecked_map class.


clemahieu:
Looks like a change to the flat buffers submodule got mixed in.

Fiono11:
Fixed it.

PR #4209, opened 2023-04-05
Fixing threading errors from review and TSAN

clemahieu:
The rationale for not making throttle thread-safe itself is the fine-grained versus coarse-grained lock tradeoff.
Fine grained:
- Increased lock aquisition/release overhead
- Decreased lock aquisition latency
- Increased chance of deadlock
- More self-contained

Coarse grained:
- Decreased lock acquistion overhead
- Increased lock acquisition latency
- Decrease chance of deadlock
- Less self contained

I don't think there is a chance for deadlock here since throttle doesn't do any callbacks which may induce extra locking
I think the biggest concern is this is already a hot codepath and even the stat counters are starting to show up in profiles because of stat mutex aquisition overhead
I believe the reusability chance of the throttle class is low and it's highly associated with the ascending bootstrapper that already has a mutex that can be used.

The rationale on not using an atomic<bool> for stopped is because it can effectively silence data races when mixed with condition variables.
For instance the schedule:
T1: read stopped (false)
T2: acquire mutex
T2: write stopped (true)
T2: condition notify
T1: condition wait <- Waiting on notification that has already ocurred

### DIFF
--- a/nano/node/bootstrap/bootstrap_ascending.cpp
+++ b/nano/node/bootstrap/bootstrap_ascending.cpp
@@ -590,9 +590,10 @@ void nano::bootstrap_ascending::wait_blockprocessor ()
 
 void nano::bootstrap_ascending::wait_available_request ()
 {
+	nano::unique_lock<nano::mutex> lock{ mutex };
 	while (!stopped && !limiter.should_pass (1))
 	{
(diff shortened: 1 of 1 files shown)


pwojcikdev (approved):

pwojcikdev:
No need to make the throttle class thread safe, not sure why the confusion is there, but its usages absolutely needed to be protected by mutex.

PR #4211, opened 2023-04-05
Convert node-wide request rate limiter to a per-channel outstanding request limiter.

clemahieu:
This patch adds channel scoring functionality to the ascending bootstrapper. It prioritises using the most responsive channels first, e.g. the channels with the lowest outstanding requests.

Since there is no guarantee requests will be responded to, the scores periodically "decay" when timeouts are calculated.

This scoring opens up future possibilities for additional metrics; an example would be tracking telemetry information and not requesting nodes that report low block counts.

Performance wise this looks to have ~2x improvement over randomly selecting channels as was done previously.

### DIFF
--- a/nano/node/bootstrap_ascending/peer_scoring.cpp
+++ b/nano/node/bootstrap_ascending/peer_scoring.cpp
@@ -1,0 +1,135 @@
+#include <nano/node/bootstrap/bootstrap_config.hpp>
+#include <nano/node/bootstrap_ascending/peer_scoring.hpp>
+#include <nano/node/transport/channel.hpp>
+
+/*
+ * peer_scoring
+ */
+
+nano::bootstrap_ascending::peer_scoring::peer_scoring (nano::bootstrap_ascending_config & config, nano::network_constants const & network_constants) :
+	network_constants{ network_constants },
+	config{ config }
+{
+}
+
+bool nano::bootstrap_ascending::peer_scoring::try_send_message (std::shared_ptr<nano::transport::channel> channel)
+{
+	auto & index = scoring.get<tag_channel> ();
+	auto existing = index.find (channel.get ());
+	if (existing == index.end ())
+	{
+		index.emplace (channel, 1, 1, 0);
+	}
+	else
+	{
+		if (existing->outstanding < config.requests_limit)
+		{
+			[[maybe_unused]] auto success = index.modify (existing, [] (auto & score) {
+				++score.outstanding;
+				++score.request_count_total;
+			});
+			debug_assert (success);
+		}
+		else
+		{
+			return true;
+		}
+	}
+	return false;
+}
+
+void nano::bootstrap_ascending::peer_scoring::received_message (std::shared_ptr<nano::transport::channel> channel)
+{
+	auto & index = scoring.get<tag_channel> ();
+	auto existing = index.find (channel.get ());
+	if (existing != index.end ())
+	{
+		if (existing->outstanding > 1)
+		{
+			[[maybe_unused]] auto success = index.modify (existing, [] (auto & score) {
+				--score.outstanding;
+				++score.response_count_total;
+			});
+			debug_assert (success);
+		}
+	}
+}
+
+std::shared_ptr<nano::transport::channel> nano::bootstrap_ascending::peer_scoring::channel ()
+{
+	auto & index = scoring.get<tag_outstanding> ();
+	for (auto const & score : index)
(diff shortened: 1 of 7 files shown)


nano/node/bootstrap_ascending/peer_scoring.cpp
```diff
@@ -0,0 +1,127 @@
...
+	return scoring.size ();
+}
+
+void nano::bootstrap_ascending::peer_scoring::timeout ()
+{
+	auto & index = scoring.get<tag_channel> ();
+	for (auto i = index.begin (), n = index.end (); i != n;)
+	{
+		auto & score = *i;
+		//std::cerr << boost::str (boost::format ("%1%: %2% %3% %4%\n") % score.endpoint % score.outstanding % score.request_count_total % score.response_count_total);
```
thsfs: These "std::cerr" output comments can be removed.

nano/node/bootstrap_ascending/peer_scoring.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+std::size_t nano::bootstrap_ascending::peer_scoring::size () const
+{
+	return scoring.size ();
+}
+
+void nano::bootstrap_ascending::peer_scoring::timeout ()
+{
+	auto & index = scoring.get<tag_channel> ();
+	for (auto i = index.begin (), n = index.end (); i != n;)
```
thsfs: suggest naming the "i" variable as "peer_score" for better readability.

nano/node/bootstrap_ascending/peer_scoring.cpp
```diff
@@ -0,0 +1,127 @@
...
+	{
+		auto & score = *i;
```
thsfs: Not used.

nano/node/bootstrap_ascending/peer_scoring.cpp
```diff
@@ -0,0 +1,127 @@
...
+		scoring.modify (i, [] (auto & score) {
+			score.decay ();
+		});
+	}
+}
+
+void nano::bootstrap_ascending::peer_scoring::sync (std::deque<std::shared_ptr<nano::transport::channel>> const & list)
+{
+	auto & index = scoring.get<tag_channel> ();
+	for (auto channel : list)
```
thsfs: channel can be a const reference.

nano/node/bootstrap_ascending/peer_scoring.cpp
```diff
@@ -109,9 +113,24 @@ void nano::bootstrap_ascending::peer_scoring::sync (std::deque<std::shared_ptr<n
 			{
 				if (!channel->max (nano::transport::traffic_type::bootstrap))
 				{
-					index.insert ({ channel, channel.get (), channel->get_endpoint (), 1, 1, 0 });
+					index.emplace (channel, 1, 1, 0);
```
thsfs, line 116: The counters could go also as default values in the constructor, not sure which way is better.

nano/node/bootstrap_ascending/peer_scoring.cpp
```diff
@@ -0,0 +1,136 @@
...
+		}
+		else
+		{
+			return true;
+		}
+	}
+	return false;
+}
+
+void nano::bootstrap_ascending::peer_scoring::received_message (std::shared_ptr<nano::transport::channel> channel)
```
thsfs, line 41: This function has no use yet.

nano/node/bootstrap_ascending/peer_scoring.hpp
```diff
@@ -0,0 +1,88 @@
...
+	private:
+		class peer_score
+		{
+		public:
+			explicit peer_score (std::shared_ptr<nano::transport::channel> const &, uint64_t, uint64_t, uint64_t);
+			std::weak_ptr<nano::transport::channel> channel;
+			// std::weak_ptr does not provide ordering so the naked pointer is also tracked and used for ordering channels
+			// This pointer may be invalid if the channel has been destroyed
+			nano::transport::channel * channel_ptr;
+			nano::endpoint endpoint;
```
thsfs: The endpoint has no use yet and it can be retrieved from the channel.

nano/node/bootstrap_ascending/peer_scoring.cpp
clemahieu: This was missed when rebasing.

PR #4219, opened 2023-04-24
Remove live updating bandwidth limits 

clemahieu:
Live updating of bandwidth parameters was added in https://github.com/nanocurrency/nano-node/pull/3257 though a thread safety issue with this was identified by TSAN.

Since config reloading was never fully implemented, removing this ability until thread safety can be assured.

### DIFF
--- a/nano/core_test/network.cpp
+++ b/nano/core_test/network.cpp
@@ -897,7 +897,7 @@ TEST (network, duplicate_revert_publish)
 }
 
 // The test must be completed in less than 1 second
-TEST (network, bandwidth_limiter)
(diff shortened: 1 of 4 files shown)


nano/core_test/network.cpp
```diff
@@ -897,7 +897,7 @@ TEST (network, duplicate_revert_publish)
 }
 
 // The test must be completed in less than 1 second
-TEST (network, bandwidth_limiter)
+TEST (network, bandwidth_limiter_4)
```
thsfs: Suggest to use better suffixes than _4 and _2

thsfs (approved):

PR #4235, opened 2023-05-17
Moving responsibility of trimming active_transaction container

clemahieu:
Moving responsibility of trimming active_transaction container in to the container itself rather than in the election scheduler.


pwojcikdev (approved):
I had a similar refactoring in mind, good to see it done.

PR #4237, opened 2023-05-19
Create namespace 'scheduler' and class scheduler::component

clemahieu:
This component encloses the buckets/hinted/optimistic schedulers. There are an increasing number of classes that are associated with election scheduling so there is a desire to put these related concepts in their own namespace and file folder.

Move optimistic_scheduler under scheduler/optimistic scheduler::optimistic node.scheduler.optimistic
Move hinted_scheduler under scheduler/hinted scheduler::hinted node.scheduler.hinted
Move election_scheduler under scheduler/component scheduler::component node.scheduler.buckets
Created scheduler::component which contains each of the separate schedulers as one component

Node and components use PIML pattern with convenience references.

Fixed circular reference between active_transactions and scheduler::buckets by using the nano::node & already inside active_transactions.

### DIFF
--- a/nano/node/scheduler/buckets.hpp
+++ b/nano/node/scheduler/buckets.hpp
@@ -15,12 +15,15 @@ namespace nano
 {
 class block;
 class node;
+}
 
-class election_scheduler final
+namespace nano::scheduler
+{
+class buckets final
 {
 public:
-	election_scheduler (nano::node &, nano::stats &);
-	~election_scheduler ();
+	buckets (nano::node &, nano::stats &);
+	~buckets ();
 
 	void start ();
 	void stop ();
--- a/nano/node/scheduler/component.hpp
+++ b/nano/node/scheduler/component.hpp
@@ -1,0 +1,25 @@
+#pragma once
+
+#include <memory>
+
+namespace nano::scheduler
+{
+class buckets;
+class hinted;
+class node;
+class optimistic;
+
+class component
+{
+	std::unique_ptr<nano::scheduler::optimistic> optimistic_impl;
+	std::unique_ptr<nano::scheduler::buckets> buckets_impl;
+	std::unique_ptr<nano::scheduler::hinted> hinted_impl;
+
+public:
+	explicit component (nano::node & node);
+
+	nano::scheduler::buckets & buckets;
+	nano::scheduler::hinted & hinted;
+	nano::scheduler::optimistic & optimistic;
+};
+}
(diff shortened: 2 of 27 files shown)


nano/node/scheduler/buckets.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include <condition_variable>
+#include <deque>
+#include <memory>
+#include <thread>
+
+namespace nano
+{
+class block;
+class node;
+namespace scheduler
```
pwojcikdev: Maybe consider specifying the `nano::scheduler` namespace in a single statement (with forward declarations in separate block). Nesting namespaces introduces additional indentation, we do both in our codebase which is inconsistent. Using non-nested version makes it consistent with the most common case.

nano/node/scheduler/component.hpp
```diff
@@ -0,0 +1,28 @@
...
+	class optimistic;
+
+	class component
+	{
+		std::unique_ptr<nano::scheduler::optimistic> optimistic_impl;
+		std::unique_ptr<nano::scheduler::buckets> buckets_impl;
+		std::unique_ptr<nano::scheduler::hinted> hinted_impl;
+
+	public:
+		component (nano::node & node);
```
pwojcikdev: Code analyzer is going to complain about missing explicit

nano/node/scheduler/hinted.hpp
```diff
@@ -0,0 +1,68 @@
...
+class node;
+class node_config;
+class active_transactions;
+class vote_cache;
+class online_reps;
+
+namespace scheduler
+{
+	/*
+ * Monitors inactive vote cache and schedules elections with the highest observed vote tally.
```
pwojcikdev, line 21: Nesting namespaces introduces problems with inconsistent indentation exactly like visible in this file.

pwojcikdev (dismissed):
It's OK. However, I don't think "Fixed circular reference between active_transactions and scheduler::buckets by using the nano::node & already inside active_transactions." is entirely accurate, since circular dependency is still there, it's just hidden by node which acts as service locator. One could argue that this pattern is even worse. The dependency is only inside `nano::active_transactions::block_cemented_callback` which definitely does too many things and can be split further to properly remove the ugly interdependencies (similar refactoring to what was done for block processor recently).

clemahieu:
Incorporated non-nested namespaces and converted free-function to constructor for hinted::config

pwojcikdev (approved):

PR #4267, opened 2023-08-22
Remove separate signature checking thread classes

clemahieu:
This removes the use of separate signature-checking threads/classes from the node.
These classes add complexity to critical code. Even where performance optimizations could be made through multi-threading, it would be better implemented using standard C++ instead of a custom class.
It's unclear if these classes are helping performance at all so we're opting to remove them until/if a performance improvement is needed.

This PR contains 4 commits:
- Adds explicit negative tests for state and epoch blocks to verify incorrect signatures are rejected
- All blocks passed to block_processor::add_impl are enqueued in the block processor, rather than splitting them based on the type, and passed to separate signature checking threads. This remove use of signature checking threads from block_processor.
- Remove use of signature checking threads from vote_processor and check each vote directly in the vote_processor thread
- Remove unused signature checking classes.


pwojcikdev (dismissed):
Good PR that removes unnecessary complexity, all the checks were performed inside the `ledger_processor` anyway.

pwojcikdev (dismissed):

pwojcikdev (dismissed):

PR #4276, opened 2023-09-04
Remove priority::flush and rename node::block_confirm to node::start_election

clemahieu:
This PR contains 3 commits that:
- Removes blocking nano::test::confirm functions which have functionality superceded by nano::test::start_elections which polls a nano::system rather than using a thread wait
- Renames nano::block_confirm to node::start_election. This adresses a TODO item to better name this function and make it not block
- Removes priority::flush function as flush functions are being removed from the codebase.

### DIFF
--- a/nano/test_common/testutil.hpp
+++ b/nano/test_common/testutil.hpp
@@ -333,18 +333,6 @@ namespace test
 	 */
 	bool process_live (nano::node & node, std::vector<std::shared_ptr<nano::block>> blocks);
 	/*
-	 * Convenience function to confirm a list of blocks
-	 * The actual confirmation will happen asynchronously, check for that with `nano::test::confirmed (..)` function
-	 * @return true if successfully scheduled blocks to be confirmed
-	 */
-	bool confirm (nano::node & node, std::vector<std::shared_ptr<nano::block>> blocks);
-	/*
-	 * Convenience function to confirm a list of hashes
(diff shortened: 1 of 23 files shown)


nano/test_common/testutil.hpp
```diff
@@ -332,22 +332,6 @@ namespace test
...
-	 * Convenience function to confirm a list of hashes
-	 * The actual confirmation will happen asynchronously, check for that with `nano::test::confirmed (..)` function
-	 * @return true if successfully scheduled blocks to be confirmed
-	 */
-	bool confirm (nano::node & node, std::vector<nano::block_hash> hashes);
-	/*
-	 * Convenience function to check whether a list of blocks is confirmed.
-	 * @return true if all blocks are confirmed, false otherwise
-	 */
 	bool confirmed (nano::node & node, std::vector<std::shared_ptr<nano::block>> blocks);
```
pwojcikdev: I think the docs for `confirmed` got eaten when removing preceding code

pwojcikdev (dismissed):
The change doesn't look bad, but I don't think the usage and implementation of `nano::test::start_elections` is correct, since it uses a `release_assert(..)` internally. That will just abort the whole testsuite run on the first error. The correct way would be to have `nano::test::start_elections` return a boolean and call it with `ASSERT_TRUE(nano::test::start_elections(..))` similar to the way it was done so far.

PR #4277, opened 2023-09-05
Return error code from nano::test::start_election

clemahieu:
This addresses points brought up in review of https://github.com/nanocurrency/nano-node/pull/4276

### DIFF
--- a/nano/test_common/testutil.cpp
+++ b/nano/test_common/testutil.cpp
@@ -222,20 +222,24 @@ std::shared_ptr<nano::election> nano::test::start_election (nano::test::system &
 	return election;
 }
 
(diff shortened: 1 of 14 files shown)


nano/test_common/testutil.cpp
```diff
@@ -254,20 +254,25 @@ std::shared_ptr<nano::election> nano::test::start_election (nano::test::system &
...
+bool nano::test::start_elections (nano::test::system & system_a, nano::node & node_a, std::vector<nano::block_hash> const & hashes_a, bool const forced_a)
 {
+	auto result = true;
 	for (auto const & hash_l : hashes_a)
 	{
 		auto election = nano::test::start_election (system_a, node_a, hash_l);
-		release_assert (election);
+		if (!election)
+		{
+			result = false;
```
pwojcikdev: I think we should just return here, otherwise we'll be accessing null pointer in the next branch

pwojcikdev (dismissed):

Issue #3911, opened 2022-08-25
Assert in nano::rocksdb::store::clear()


dsiganos:
I hit this assert whilst running a unit test.
```
[ RUN      ] active_transactions.fork_replacement_tally
Assertion (status.ok ()) failed
int nano::rocksdb::store::clear(rocksdb::ColumnFamilyHandle*)
/home/ds/CLionProjects/work/nano/node/rocksdb/rocksdb.cpp:602
```

dsiganos:
This is the function that contains the assert statement:
```
int nano::rocksdb::store::clear (::rocksdb::ColumnFamilyHandle * column_family)
{
	// Dropping completely removes the column
	auto name = column_family->GetName ();
	auto status = db->DropColumnFamily (column_family);
	release_assert (status.ok ());

	// Need to add it back as we just want to clear the contents
	auto handle_it = std::find_if (handles.begin (), handles.end (), [column_family] (auto & handle) {
		return handle.get () == column_family;
	});
	debug_assert (handle_it != handles.cend ());
	status = db->CreateColumnFamily (get_cf_options (name), name, &column_family);
	release_assert (status.ok ());
	handle_it->reset (column_family);
	return status.code ();
}
```

It appears to be deleting something and then trying to add it back again.
That sounds racy... it possibly needs thread protection.

brandon-bb:
Currently, working on this issue - had the idea of rewriting the function to utilise the atomic update feature in RocksDB. Will check to see if we could apply a lock to the column being cleared to prevent access by other threads.

RickiNano:
This issue has been fixed by brandon-bb and can be closed @qwahzi

qwahzi:
> This issue has been fixed by brandon-bb and can be closed @qwahzi

Thank you Ricki and Brandon!

PR #4279, opened 2023-09-06
fixed bug in nano::rocksdb::store::clear()

brandon-bb:
Addresses #3911 - issue before was that the function was deleting the column entirely to clear the contents then adding it back. This solution makes use of the atomic update feature in RocksDB by creating a batch of delete operations for all keys in the column and applying them with a write operation. This approach ensures that either all operations are applied or none are, thus guaranteeing database consistency.


clemahieu (approved):

clemahieu:
If you can rebase this I will merge it.

RickiNano:
I took the liberty of cherrypicking the commits to a new PR to resolve the merge conflict. 
I hope this is ok. All credits go to @brandon-bb 
New pr: https://github.com/nanocurrency/nano-node/pull/4304

PR #4281, opened 2023-09-12
Separate manual scheduler in to its own class

clemahieu:
- This creates the nano::scheduler::manual class which uses functionality moved from nano::scheduler::priority.
- Previously, the manual and priority queues were being run by the same thread within the same class.
- Starting and stopping the scheduler component threads is now the responsibility of scheduler::component, rather than being handled directly within nano::node.
- Rename scheduler component threads reported thread names to be more uniform.


clemahieu:
ct.bnano.info tests are green.

PR #4288, opened 2023-09-21
Force use of windows.h instead of default winnt.h

RickiNano:
This fixes the "No target Architecture" error when building on windows.
Clang format will report an error but the windows.h include must be before processthreadapi.h to work properly.


clemahieu:
Looks like clang format doesn't like the non-alphabetical order of headers.

Is processthreadsapi.h needed at all with windows.h?

RickiNano:
You are right. It still builds without the processthreadsapi.h include.
I've updated the PR

PR #4306, opened 2023-10-02
Refactor block_cemented_callback for readability

gr0vity-dev:
- Modularized the block confirmation logic into separate functions.
- Separated active and inactive block confirmation processes.
- Simplified the main block_cemented_callback() function for easier understanding.
- Reduced nested conditionals by early returning and using helper methods.
- Improved code readability and maintainability.

Marked as draft PR to discuss if such a breakup into smaller methods is desirable

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -67,25 +67,27 @@ void nano::active_transactions::stop ()
 
 void nano::active_transactions::block_cemented_callback (std::shared_ptr<nano::block> const & block_a)
 {
-	nano::store::read_transaction transaction = node.store.tx_begin_read ();
-	auto status_type = determine_election_status (transaction, block_a);
+	auto transaction = node.store.tx_begin_read ();
+	auto status_type = election_status (transaction, block_a);
 
 	if (!status_type)
 		return;
 
-	if (*status_type == nano::election_status_type::inactive_confirmation_height)
+	switch (*status_type)
 	{
-		process_inactive_confirmation (transaction, block_a);
-	}
-	else
-	{
-		process_active_confirmation (transaction, block_a, *status_type);
+		case nano::election_status_type::inactive_confirmation_height:
+			process_inactive_confirmation (transaction, block_a);
+			break;
+
+		default:
+			process_active_confirmation (transaction, block_a, *status_type);
+			break;
 	}
 
 	handle_final_votes_confirmation (block_a, transaction, *status_type);
 }
 
-boost::optional<nano::election_status_type> nano::active_transactions::determine_election_status (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block)
+boost::optional<nano::election_status_type> nano::active_transactions::election_status (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block)
 {
 	boost::optional<nano::election_status_type> status_type;
 
@@ -103,8 +105,8 @@ boost::optional<nano::election_status_type> nano::active_transactions::determine
 
 void nano::active_transactions::process_inactive_confirmation (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block)
 {
-	nano::account account{};
-	nano::uint128_t amount (0);
+	nano::account account;
+	nano::uint128_t amount{ 0 };
 	bool is_state_send = false;
 	bool is_state_epoch = false;
 	nano::account pending_account{};
@@ -124,12 +126,12 @@ void nano::active_transactions::process_active_confirmation (nano::store::read_t
 		election_winners_lk.unlock ();
 		if (election->confirmed () && election->winner ()->hash () == hash)
 		{
-			execute_confirmed (transaction, block, election, status_type);
+			handle_confirmation (transaction, block, election, status_type);
 		}
 	}
 }
 
-void nano::active_transactions::execute_confirmed (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block, std::shared_ptr<nano::election> election, nano::election_status_type status_type)
+void nano::active_transactions::handle_confirmation (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block, std::shared_ptr<nano::election> election, nano::election_status_type status_type)
 {
 	nano::block_hash hash = block->hash ();
 	update_recently_cemented (election);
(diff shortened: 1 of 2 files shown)


nano/node/active_transactions.cpp
```diff
@@ -67,99 +67,152 @@ void nano::active_transactions::stop ()
 
 void nano::active_transactions::block_cemented_callback (std::shared_ptr<nano::block> const & block_a)
 {
-	auto transaction = node.store.tx_begin_read ();
+	nano::store::read_transaction transaction = node.store.tx_begin_read ();
```
clemahieu: Is using an explicit type intentional here? I think in most places we create transactions we use 'auto'

nano/node/active_transactions.cpp
```diff
@@ -67,99 +67,152 @@ void nano::active_transactions::stop ()
...
-		// This block was explicitly added to the confirmation height_processor
-		election_status_type = nano::election_status_type::active_confirmed_quorum;
+		process_active_confirmation (transaction, block_a, *status_type);
 	}
 
-	if (election_status_type.is_initialized ())
+	handle_final_votes_confirmation (block_a, transaction, *status_type);
+}
+
+boost::optional<nano::election_status_type> nano::active_transactions::determine_election_status (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block)
```
clemahieu: I'd just call this "election_status"

nano/node/active_transactions.cpp
```diff
@@ -67,99 +67,152 @@ void nano::active_transactions::stop ()
...
-	auto transaction = node.store.tx_begin_read ();
+	nano::store::read_transaction transaction = node.store.tx_begin_read ();
+	auto status_type = determine_election_status (transaction, block_a);
 
-	boost::optional<nano::election_status_type> election_status_type;
-	if (!confirmation_height_processor.is_processing_added_block (block_a->hash ()))
+	if (!status_type)
+		return;
+
+	if (*status_type == nano::election_status_type::inactive_confirmation_height)
```
clemahieu: I think a switch statement would be better here, generally better for any enum value. Then the compiler can warn if there are missing cases which is important when we add new values to enums.

nano/node/active_transactions.cpp
```diff
@@ -67,99 +67,152 @@ void nano::active_transactions::stop ()
...
+		status_type = nano::election_status_type::active_confirmed_quorum;
+	}
+
+	return status_type;
+}
+
+void nano::active_transactions::process_inactive_confirmation (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block)
+{
+	nano::account account{};
+	nano::uint128_t amount (0);
```
clemahieu: This should use initialization syntax:
nano::uint128_t amount{ 0 };

The way it's currently written is many other places in the code but it's an old style that's been replaced with braced initialization.

nano/node/active_transactions.cpp
```diff
@@ -67,99 +67,152 @@ void nano::active_transactions::stop ()
...
+	{
+		status_type = nano::election_status_type::active_confirmed_quorum;
+	}
+
+	return status_type;
+}
+
+void nano::active_transactions::process_inactive_confirmation (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block)
+{
+	nano::account account{};
```
clemahieu: The {} is mostly redundant. For some types this forces the default constructor to be run, which could give it an initialized value versus an uninitialized one.

nano/node/active_transactions.cpp
```diff
@@ -67,99 +67,152 @@ void nano::active_transactions::stop ()
...
+	bool cemented_bootstrap_count_reached = node.ledger.cache.cemented_count >= node.ledger.bootstrap_weight_max_blocks;
+	bool was_active = status == nano::election_status_type::active_confirmed_quorum || status == nano::election_status_type::active_confirmation_height;
+
+	if (cemented_bootstrap_count_reached && was_active)
+	{
+		activate_scheduler_for_account_and_destination (account, block, transaction);
+	}
+}
+
+void nano::active_transactions::activate_scheduler_for_account_and_destination (const nano::account & account, std::shared_ptr<nano::block> const & block, nano::store::read_transaction const & transaction)
```
clemahieu: This name is a bit long. Maybe "activate_dependents".

nano/node/active_transactions.cpp
```diff
@@ -67,99 +67,152 @@ void nano::active_transactions::stop ()
...
-						}
-					}
-				}
-			}
+			execute_confirmed (transaction, block, election, status_type);
 		}
+	}
+}
+
+void nano::active_transactions::execute_confirmed (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block, std::shared_ptr<nano::election> election, nano::election_status_type status_type)
```
clemahieu: Maybe "handle_confirmation" to maintain the "handle" nomenclature.

clemahieu (commented):
I added a couple style comments but since this is just moving code let's ignore those for now and do the direct code copying.

nano/node/active_transactions.cpp
gr0vity-dev: I agree that it's too long. Since it activates the next blocks, what about "activate_successors" ?

gr0vity-dev:
Thanks for the feedback. I applied all your suggestions.
One thing I noticed is that we mostly use {variable}_a in the method definition, to indicate it's an argument ?
Should I keep that convention when splitting into smaller methods ?

clemahieu:
> Thanks for the feedback. I applied all your suggestions. One thing I noticed is that we mostly use {variable}_a in the method definition, to indicate it's an argument ? Should I keep that convention when splitting into smaller methods ?

I've been moving away from that convention. Since argument names are looked at by the users I think it's better to keep those clean.

I still use _m (class member) and _l (local variable) when there are name shadowing issues because those aren't often looked at by end users.

nano/node/active_transactions.cpp
```diff
@@ -68,98 +68,155 @@ void nano::active_transactions::stop ()
...
+	if (!status_type)
+		return;
+
+	switch (*status_type)
+	{
+		case nano::election_status_type::inactive_confirmation_height:
+			process_inactive_confirmation (transaction, block_a);
+			break;
+
+		default:
```
clemahieu, line 82: I don't think we should use a default clause here but that's the existing functionality so I guess we can keep it.

clemahieu (approved):

Issue #4102, opened 2023-02-03
Review and remove enumerations from stats_enums.hpp that are not used


thsfs:
As the code evolves and the old portions of it gets removed, some enumerations might not be necessary anymore.

This task is for reviewing the enumerations that don´t have any use in the code and remove them.

PR #4307, opened 2023-10-03
Remove enumerations from stats_enums.hpp that are not used

RickiNano:
Clean up enumerations that are no longer in use. This pr targets #4102

### DIFF
--- a/nano/lib/stats_enums.hpp
+++ b/nano/lib/stats_enums.hpp
@@ -19,7 +19,6 @@ enum class type : uint8_t
 	vote,
 	election,
 	http_callback,
-	peering,
 	ipc,
 	tcp,
 	confirmation_height,
@@ -30,7 +29,6 @@ enum class type : uint8_t
 	filter,
 	telemetry,
 	vote_generator,
-	vote_cache,
 	hinting,
 	blockprocessor,
 	bootstrap_server,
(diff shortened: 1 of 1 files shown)


gr0vity-dev:
I'm wondering if it makes sense to have enums that are only used in the testcases ?
Do they make sense in a testcase when the codebase doesn't use them ?

Looking at the `vote_processed` enum, it got removed with the Optimistic elections PR in these commits :
[e1c893b7dcbcfd9cf53eb1b0239e9febc62bb508](https://github.com/nanocurrency/nano-node/commit/e1c893b7dcbcfd9cf53eb1b0239e9febc62bb508)
[62811054380185c376df600501d22101b389d975](https://github.com/nanocurrency/nano-node/commit/62811054380185c376df600501d22101b389d975)

clemahieu:
> I'm wondering if it makes sense to have enums that are only used in the testcases ? Do they make sense in a testcase when the codebase doesn't use them ?
> 
> Looking at the `vote_processed` enum, it got removed with the Optimistic elections PR in these commits : [e1c893b7dcbcfd9cf53eb1b0239e9febc62bb508](https://github.com/nanocurrency/nano-node/commit/e1c893b7dcbcfd9cf53eb1b0239e9febc62bb508) [62811054380185c376df600501d22101b389d975](https://github.com/nanocurrency/nano-node/commit/62811054380185c376df600501d22101b389d975)

That is a good point and really I agree. It's occurred to me that we shouldn't be checking counters in tests at all. The stat counters are for users, and just as we shouldn't be depending on the output of logs, we shouldn't be depending on the counters.

RickiNano:
Do you want me to remove the enums again, and also remove the unit tests they are used in?

clemahieu:
No, let's do that separately.  This change is fine on its own to remove enums that are completely unused.

PR #4308, opened 2023-10-05
Replace boost filesystem with std filesystem

RickiNano:
This pull request replaces boost filesystem with the c++17 std filesystem

It builds and runs on my pc but it does have at least two issues that I hope someone more knowledgeable can either fix or tell me how to fix:
1.  In daemon.ccp line 186 is commented to avoid a build error. This breaks child process RPC, while in-process RPC still works
~~2. In secure/utility.cpp we used boost::filesystem::unique_path () but this is not available in std filesystem. I have changed the unique_path function to always return "temp" as a workaround for now~~

EDIT: The second issue has been resolved with a custom function but I need help with the first one
This PR targets: https://github.com/nanocurrency/nano-node/issues/4057

### DIFF
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -674,14 +674,14 @@ TEST (mdb_block_store, bad_path)
 		GTEST_SKIP ();
 	}
 	nano::logger_mt logger;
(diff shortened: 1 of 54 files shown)


RickiNano:
I got both issues fixed now. I'll close this draft PR and I will reopen a "real" pr when everything is ready

PR #4309, opened 2023-10-08
Use c++ std filesystem

RickiNano:
This replaces all uses of boost filesystem with standard c++ filesystem.

A couple of notes:
boost::filesystem::unique_path is not supported in std::filesystem. I have re-written the nano::unique_path function found in /secure/utility.cpp so that it mimics the same behavior.
The "bad_path" unit test fails in LMDB mode when the path input is set to "///" because of the way std::filesystem handles invalid paths. I have removed this test.

Target issue: https://github.com/nanocurrency/nano-node/issues/4057

### DIFF
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -674,13 +674,14 @@ TEST (mdb_block_store, bad_path)
 		GTEST_SKIP ();
 	}
 	nano::logger_mt logger;
+
 	try
 	{
 		auto path = nano::unique_path ();
 		{
 			std::ofstream stream (path.c_str ());
 		}
-		boost::filesystem::permissions (path, boost::filesystem::perms::no_perms);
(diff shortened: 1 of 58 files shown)


clemahieu (changes requested):
I changed how file opening errors are handled in https://github.com/nanocurrency/nano-node/pull/4314 which means we won't need to remove the associated test here.

I tried to push a commit on top of this branch but it looks like GitHub branch protection rules prevent this.

Please rebase and this should be good to go.

RickiNano:
Github runners are reporting some strange linker errors after rebasing this PR on current develop branch. 
I've cherry picked the commits into a new branch and created a new PR here:
https://github.com/nanocurrency/nano-node/pull/4315
Feel free to merge or close either of them.

clemahieu:
The linker errors were related to GitHub caching. I cleared the cache and re-ran and they passed so I committed the other PR.

Thanks!

Issue #4169, opened 2023-03-04
Improve AEC alignment


RickiNano:
### Summary

Nodes receives blocks asynchronously, resulting in each node having its own perceptions on when exactly a block was received and thus affecting the associated LRU (Least Recent Use). When a node selects a block for the AEC, it sorts the bucket by LRU and choose the block from the account with the least recent use. This works well under normal circumstances. 

However, during a spam attack, the LRU timings become extremely tight and the diverging values from one node to another leads to nodes picking their own set of blocks for the AEC and reducing the chances of consensus. AEC alignment could be reached if blocks were sorted by block hash instead, since these values are the same for all nodes. However, sorting by hash only would remove the prioritization of accounts with low usage

Proposed solution:
1. Round the LRU value to nearest minute
2. AEC candidates are still prioritized by LRU first, but then sorted by block hash

Since local LRU values are not precise to begin with, I suggest reducing the precision of the LRU data type to one minute. This will group all blocks into one-minute "segments" that are much more likely to be the same across the network. 
With many blocks now sharing the same LRU value (or segment), the node should sort these candidates by block hash, as the hash value will always be the same across the network.

This approach can be improved further by sorting the last two LRU segments instead just one. This is because node A may receive a block at 11:26:59 while node B gets the block at 11:27:00 and thus putting it into another segment. By always considering the last two LRU segments we have the required overlap to always have perfect AEC alignment

### What problem would be solved by this feature?

Faster consensus
Reduced voting traffic
Higher CPS

### Are there any previous requests for this feature?

This has been discussed multiple times on Discord, reddit and the Nano forum. This is an attempt to make a concrete solution description

### Do you have a suggested solution?

Described in the summary

### If this feature is approved, would you be willing to submit a pull request with the solution?

I am willing to collaborate

### Possible solution

Described in the summary

RickiNano:
Gr0vity compiled a version of the node that rounds the LRU time to the nearest minute and simulated on a local desynced network
![image](https://user-images.githubusercontent.com/81099017/225373606-61237b68-c3a1-4bc1-bc93-330e45a3a2d0.png)
![image](https://user-images.githubusercontent.com/81099017/225373638-9c7f3226-3ea1-47bc-a3b7-13c313d0b782.png)
Discussion can be found here> https://discord.com/channels/370266023905198083/769209197333053511/1082049233886134393

ATXMJ:
Out of curiosity, why was "nearest minute" chosen for this concept?

[...]

PR #4317, opened 2023-10-25
Round LRU time to nearest 32 seconds

RickiNano:
By rounding LRU time we are more likely to get AEC aligned across the network because blocks that are published very fast are grouped in 32 seconds "slots" that are sorted by hash.
This pr uses a 5 bit mask for rounding for best performance.
60 seconds rounding has been suggested but I chose 32 seconds to err on the side of caution and because Piotr mentioned that down to 15 seconds might be enough

### DIFF
--- a/nano/node/scheduler/bucket.cpp
+++ b/nano/node/scheduler/bucket.cpp
@@ -35,7 +35,10 @@ void nano::scheduler::bucket::pop ()
 
 void nano::scheduler::bucket::push (uint64_t time, std::shared_ptr<nano::block> block)
 {
-	queue.insert ({ time, block });
+	const uint64_t bitmask = ~0b11111;
+	uint64_t rounded_time = time & bitmask;
+
+	queue.insert ({ rounded_time, block });
 	if (queue.size () > maximum)
 	{
 		debug_assert (!queue.empty ());


qwahzi:
Related issue with more details, for reference: https://github.com/nanocurrency/nano-node/issues/4169

nano/node/scheduler/bucket.cpp
```diff
@@ -35,7 +35,10 @@ void nano::scheduler::bucket::pop ()
 
 void nano::scheduler::bucket::push (uint64_t time, std::shared_ptr<nano::block> block)
 {
-	queue.insert ({ time, block });
+	const uint64_t bitmask = ~0b11111;
+	uint64_t rounded_time = time & bitmask;
```
pwojcikdev, line 39: For readability I would strongly suggest to keep a normal division. Our bottleneck is IO which is many orders of magnitude slower anyway, this division is noop when it comes to performance

RickiNano:
This PR has been tested on the beta network.
The saturation point before introducing LRU rounding was 650 bps. With a 32 seconds LRU rounding, saturation increased to 900 bps.
We never tested 900 bps with the previous version but here is a comparison of 850 bps (no rounding) vs 900 bps (with 32 seconds rounding)
image

RickiNano:
The performance improvement that was shown on the beta network turned out to be caused by a problem with one of the nodes that was fixed just after the LRU rounding version was installed.
From further testing it seems that the performance gained is minimal so I will close this PR for now

PR #4318, opened 2023-10-26
Remove atomic enum election::state_m and instead require locking election::mutex.

clemahieu:


dsiganos:
@clemahieu, could you explain why this change is needed, what problem does it solve.

clemahieu:
This change is removing the atomic election::state_m variable in favor of using the election mutex for synchronization. The issue is there are two synchronization mechanisms in use, locking the mutex or doing an atomic compare exchange, which has race conditions.

From what I recall the election can be doing an operation with a locked mutex, and a separate action can change the election state because it doesn't respect the mutex being locked.

dsiganos (approved):

clemahieu:
LGTM

PR #4322, opened 2023-11-01
Move mutex acquisition outside write transaction inside vote generator.

clemahieu:


nano/node/voting.cpp
```diff
@@ -184,7 +184,7 @@ nano::vote_generator::~vote_generator ()
 	stop ();
 }
 
-void nano::vote_generator::process (store::write_transaction const & transaction, nano::root const & root_a, nano::block_hash const & hash_a)
+bool nano::vote_generator::process (store::write_transaction const & transaction, nano::root const & root_a, nano::block_hash const & hash_a)
```
pwojcikdev: After those modifications a name better suiting this method seems to be `should_vote`, probably can be made const too.

pwojcikdev (approved):

PR #4324, opened 2023-11-02
Make hinted and optimistic AEC allocation configurable

gr0vity-dev:
During the beta tests I noticed that modifying `active_elections_hinted_limit_percentage` in the config-node.toml isn't applied correctly.

This PR adds
- active_elections_hinted_limit_percentage
- active_elections_optimistic_limit_percentage
to nodeconfig.cpp


pwojcikdev:
I think we should make a separate config subcategory at `[node.active_elections]` with a dedicated `active_elections_config` class. Instead of prefixing new entries with `active_elections_...` put them in this config file for better long term organisation.

gr0vity-dev:
Closed in favor of #4604

PR #4336, opened 2023-11-08
Revert timed block_uniquer cleanup and replace.

clemahieu:
Rather than cleaning up on an interval, continue cleaning up items progressively but rather than selecting random hashes to attempt to clean up, iterate the next items following the item just inserted for attempted cleanup.


pwojcikdev:
This approach is strictly inferior to a properly implemented linear cleanup, in the ideal case you can make a cleanup that runs in constant amortized time.

PR #4342, opened 2023-11-12
Set maximum unchecked blocks in config

RickiNano:
This PR makes the maximum number of unchecked blocks to store in memory configurable by using a new "max_unchecked_blocks" setting in config-node.toml. It defaults to 65536 blocks (the current hardcoded number)

### DIFF
--- a/nano/node/nodeconfig.cpp
+++ b/nano/node/nodeconfig.cpp
@@ -465,9 +465,9 @@ nano::error nano::node_config::deserialize_toml (nano::tomlconfig & toml)
 		{
 			toml.get_error ().set ("active_elections_size must be greater than 250");
 		}
-		if (max_unchecked_blocks < 1)
+		if (max_unchecked_blocks < 0)
 		{
-			toml.get_error ().set ("max_unchecked_blocks must be greater than 0");
+			toml.get_error ().set ("max_unchecked_blocks must be 0 or higher");
 		}
 		if (bandwidth_limit > std::numeric_limits<std::size_t>::max ())
 		{
--- a/nano/node/nodeconfig.cpp
+++ b/nano/node/nodeconfig.cpp
@@ -128,6 +128,7 @@ nano::error nano::node_config::serialize_toml (nano::tomlconfig & toml) const
 	toml.put ("max_work_generate_multiplier", max_work_generate_multiplier, "Maximum allowed difficulty multiplier for work generation.\ntype:double,[1..]");
 	toml.put ("frontiers_confirmation", serialize_frontiers_confirmation (frontiers_confirmation), "Mode controlling frontier confirmation rate.\ntype:string,{auto,always,disabled}");
 	toml.put ("max_queued_requests", max_queued_requests, "Limit for number of queued confirmation requests for one channel, after which new requests are dropped until the queue drops below this value.\ntype:uint32");
+	toml.put ("max_unchecked_blocks", max_unchecked_blocks, "Maximum number of unchecked blocks to store in memory. Defaults to 65536. \ntype:uint64,[1..11]");
 	toml.put ("rep_crawler_weight_minimum", rep_crawler_weight_minimum.to_string_dec (), "Rep crawler minimum weight, if this is less than minimum principal weight then this is taken as the minimum weight a rep must have to be tracked. If you want to track all reps set this to 0. If you do not want this to influence anything then set it to max value. This is only useful for debugging or for people who really know what they are doing.\ntype:string,amount,raw");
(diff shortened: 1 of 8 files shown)


nano/node/nodeconfig.cpp
```diff
@@ -473,6 +476,10 @@ nano::error nano::node_config::deserialize_toml (nano::tomlconfig & toml)
 		{
 			toml.get_error ().set ("active_elections_size must be greater than 250");
 		}
+		if (max_unchecked_blocks < 1)
```
pwojcikdev: There is practically no difference between 1 and 0, so if someone wants to disable unchecked blocks let's allow it. It might be useful for testing.

RickiNano: Good point. Updated

nano/node/nodeconfig.cpp
```diff
@@ -128,6 +128,7 @@ nano::error nano::node_config::serialize_toml (nano::tomlconfig & toml) const
 	toml.put ("max_work_generate_multiplier", max_work_generate_multiplier, "Maximum allowed difficulty multiplier for work generation.\ntype:double,[1..]");
 	toml.put ("frontiers_confirmation", serialize_frontiers_confirmation (frontiers_confirmation), "Mode controlling frontier confirmation rate.\ntype:string,{auto,always,disabled}");
 	toml.put ("max_queued_requests", max_queued_requests, "Limit for number of queued confirmation requests for one channel, after which new requests are dropped until the queue drops below this value.\ntype:uint32");
+	toml.put ("max_unchecked_blocks", max_unchecked_blocks, "Maximum number of unchecked blocks to store in memory. Defaults to 65536. \ntype:uint64,[1..11]");
```
dsiganos: What does 1..11 mean here?

nano/node/nodeconfig.cpp
```diff
@@ -462,6 +465,10 @@ nano::error nano::node_config::deserialize_toml (nano::tomlconfig & toml)
 		{
 			toml.get_error ().set ("active_elections_size must be greater than 250");
 		}
+		if (max_unchecked_blocks < 0)
```
dsiganos: There is not much point in this check since max_unchecked_block is unsigned.
But it s also harmless unless it starts throwing compiler warnings.

nano/node/unchecked_map.cpp
```diff
@@ -5,7 +5,8 @@
 #include <nano/lib/timer.hpp>
 #include <nano/node/unchecked_map.hpp>
 
-nano::unchecked_map::unchecked_map (nano::stats & stats, bool const & disable_delete) :
+nano::unchecked_map::unchecked_map (unsigned & max_unchecked_blocks, nano::stats & stats, bool const & disable_delete) :
```
dsiganos: If you are going to use a reference then you should declare it as const since this function shouldn't be changing it. But I wouldn't use a reference for an int, just pass it by value.

nano/node/unchecked_map.hpp
```diff
@@ -22,7 +22,7 @@ class stats;
 class unchecked_map
 {
 public:
-	unchecked_map (nano::stats &, bool const & do_delete);
+	unchecked_map (unsigned & max_unchecked_blocks, nano::stats &, bool const & do_delete);
```
dsiganos: No need for reference and use const

nano/node/unchecked_map.hpp
```diff
@@ -62,11 +62,10 @@ class unchecked_map
 	nano::condition_variable condition;
 	nano::mutex mutex;
 	std::thread thread;
+	unsigned max_unchecked_blocks;
```
dsiganos: You could also use const here but only if we will not allow changing this dynamically.

dsiganos (changes requested):

dsiganos:
You should add some code to check this new config in the test cases:
* `toml.daemon_config_deserialize_no_defaults`
* `toml.daemon_config_deserialize_defaults`

nano/node/nodeconfig.cpp
RickiNano: That is the value range. However, I copied from another line and forgot to change the values. That is updated now

nano/node/nodeconfig.cpp
RickiNano: You are right. A negative value in the config is not able to trigger this warning anyway. I have removed the check

nano/node/unchecked_map.cpp
RickiNano: Fixed

nano/node/unchecked_map.hpp
RickiNano: Fixed

nano/node/unchecked_map.hpp
RickiNano: Fixed

dsiganos (approved):

PR #4344, opened 2023-11-14
Reduce mutex locked time in block scheduler

RickiNano:
Mutex is unlocked early by destroying the locked object it was created in and storing the block in a variable. The subsequent block processing is done outside the lock, reducing the total lock time.
Unit tests are successful.
Disclaimer: I'm no c++ expert. Chatgpt has helped me with parts of this. Review carefully, and feel free to dismiss it.

### DIFF
--- a/nano/node/scheduler/manual.cpp
+++ b/nano/node/scheduler/manual.cpp
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
 
 void nano::scheduler::manual::run ()
 {
-	nano::unique_lock<nano::mutex> lock{ mutex };
-	while (!stopped)
+	while (true)
 	{
-		condition.wait (lock, [this] () {
-			return stopped || predicate ();
-		});
-		debug_assert ((std::this_thread::yield (), true)); // Introduce some random delay in debug builds
-		if (!stopped)
-		{
-			node.stats.inc (nano::stat::type::election_scheduler, nano::stat::detail::loop);
+		std::tuple<std::shared_ptr<nano::block>, boost::optional<nano::uint128_t>, nano::election_behavior> task;
 
-			if (predicate ())
+		{
+			nano::unique_lock<nano::mutex> lock{ mutex };
+			condition.wait (lock, [this] () {
+				return stopped || predicate ();
+			});
+			debug_assert ((std::this_thread::yield (), true)); // Introduce some random delay in debug builds
+			if (stopped)
 			{
-				auto const [block, previous_balance, election_behavior] = queue.front ();
-				queue.pop_front ();
-				lock.unlock ();
-				node.stats.inc (nano::stat::type::election_scheduler, nano::stat::detail::insert_manual);
-				auto result = node.active.insert (block, election_behavior);
-				if (result.election != nullptr)
-				{
-					result.election->transition_active ();
-				}
+				break;
 			}
-			else
-			{
-				lock.unlock ();
-			}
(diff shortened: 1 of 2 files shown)


nano/node/scheduler/manual.cpp
```diff
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
 
 void nano::scheduler::manual::run ()
 {
-	nano::unique_lock<nano::mutex> lock{ mutex };
-	while (!stopped)
```
clemahieu, line 55: Needs lock

nano/node/scheduler/manual.cpp
```diff
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
...
+	while (true)
 	{
-		condition.wait (lock, [this] () {
```
clemahieu, line 57: Needs lock

nano/node/scheduler/manual.cpp
```diff
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
...
-			return stopped || predicate ();
-		});
-		debug_assert ((std::this_thread::yield (), true)); // Introduce some random delay in debug builds
-		if (!stopped)
-		{
-			node.stats.inc (nano::stat::type::election_scheduler, nano::stat::detail::loop);
```
clemahieu, line 63: Doesn't need lock but the stats class needs its own improvement

nano/node/scheduler/manual.cpp
```diff
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
...
+		std::tuple<std::shared_ptr<nano::block>, boost::optional<nano::uint128_t>, nano::election_behavior> task;
 
-			if (predicate ())
```
clemahieu, line 65: Needs lock

nano/node/scheduler/manual.cpp
```diff
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
...
-			{
-				auto const [block, previous_balance, election_behavior] = queue.front ();
```
clemahieu, line 67: Needs lock

nano/node/scheduler/manual.cpp
```diff
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
...
-				queue.pop_front ();
-				lock.unlock ();
```
clemahieu, line 69: Lock released

nano/node/scheduler/manual.cpp
```diff
@@ -51,36 +51,37 @@ bool nano::scheduler::manual::predicate () const
...
-				node.stats.inc (nano::stat::type::election_scheduler, nano::stat::detail::insert_manual);
-				auto result = node.active.insert (block, election_behavior);
```
clemahieu, line 71: Could be slower but the lock is already released for this.

clemahieu (commented):
I don't think this change improves much. In the existing code the locks are already held for the minimum amount needed.

I put some inline comments just to show the parts that need the lock held and the parts that already avoid lock holding.

For performance things like this we'd first look at flamegraph profiles to see if there's an actual performance issue, otherwise it's premature optimization.

RickiNano:
I think you are right. I will look into flamegraph and see if I can find something that needs optimization. 
Gr0vity has already provided me with a flamegraph chart

PR #4354, opened 2024-01-09
Merge server_socket in to tcp_listener

clemahieu:
Merges server_socket into tcp_listener. Functionality from nano::socket wasn't being used within nano::server_socket except for using the inherited boost strand. tcp_listener is changed to derive from shared_from_this since it's captured in network operation callbacks as sever_socket was previously.
Removes delays when dropping connections due to IP limits.
Improves unit test assertions.


dsiganos:
The removing of the delay in accepting sockets should really be in a separate PR. It is a feature change whereas the server socket removal is a refactoring.

nano/node/transport/tcp_server.cpp
```diff
@@ -10,59 +10,57 @@
 
 #include <memory>
 
+namespace
```
dsiganos, line 13: Is there any reason for the forcing of the namespace to be global?
It seems unnecessary. Probably a historical artifact.

clemahieu:
Removed commit which modified socket timeouts. Added an additional commit which removed client_socket which was just a passthrough for socket. Marked socket as final and removed virtual functions.

pwojcikdev (approved):
> tcp_listener is changed to derive from shared_from_this since it's captured in network operation callbacks as sever_socket was previously.


This should not be the case in a properly designed code.

nano/node/transport/tcp_server.cpp
```diff
@@ -10,59 +10,57 @@
...
-		node.logger.always_log (boost::str (boost::format ("Network: Error while binding for incoming TCP/bootstrap on port %1%: %2%") % listening_socket->listening_port () % ec.message ()));
-		throw std::runtime_error (ec.message ());
+		acceptor.listen (boost::asio::socket_base::max_listen_connections, ec);
 	}
-
-	// the user can either specify a port value in the config or it can leave the choice up to the OS:
-	// (1): port specified
-	// (2): port not specified
-	//
-	const auto listening_port = listening_socket->listening_port ();
```
dsiganos, line 40: This block of code was written by Theo and deals with setting the port number in a semi automatic way. I remember he spent a lot of time on this. What are we losing by removing this? We are probably losing the ability to automatically let the operating system give a node a listening address. I do not know what depends on that ability though and whether it is important.

nano/node/transport/tcp_server.cpp
clemahieu: This is putting it into an anonymous namespace, the c++ equivalent of a C static function.

nano/node/transport/tcp_server.cpp
clemahieu: The ability to have an OS selected port remains. This is done by setting the peering_port=0 or on the command line --config node.peering_port=0.

this->local is set in the constructor to the port passed in to the constructor and just above you can see the acceptor binds to the specified port.

Tests make use of the OS-selected port so tests would be failing but I also verified by running a node and overriding the peering_port to 0 and checking the port it's bound to:

nano_node 69415      clemahieu   18u  IPv6 0x5bc40e581518c7f9      0t0    TCP *:54688 (LISTEN)

The previous functionality was copying the port selected in to this->port and using it only in tcp_listener::endpoint (). endpoint() was changed to directly query the listening port from the acceptor rather than caching a copy locally.

clemahieu:
> > tcp_listener is changed to derive from shared_from_this since it's captured in network operation callbacks as sever_socket was previously.
> 
> This should not be the case in a properly designed code.

This was copied from the socket class but that's a good point, I'll see if that can be removed.

nano/node/transport/tcp_server.cpp
dsiganos: I understand and I do not see the point of it. But it is a very minor point.

dsiganos (approved):

PR #4361, opened 2024-01-15
Fix conflicts.add_two unit test

dsiganos:


clemahieu (dismissed):

clemahieu:
Useful utility functions.

nano/core_test/conflicts.cpp
```diff
@@ -1,3 +1,5 @@
+#include "nano/test_common/chains.hpp"
```
pwojcikdev: In our codebase there seems to be a convention of only using < > includes

nano/test_common/chains.hpp
```diff
@@ -32,6 +32,18 @@ std::vector<std::pair<nano::account, nano::block_list_t>> setup_chains (nano::te
...
+ * \brief Create a pair of send/receive blocks to implement the transfer of "amount" raw from "source" to the unopened account "dest".
+ * \param system
+ * \param node
+ * \param amount the amount of raw to transfer
+ * \param source the source account
+ * \param dest the destination account
+ * \param dest_rep the rep that the dest account should have
+ * \param force_confirm force confirm the blocks
+ */
+std::pair<std::shared_ptr<nano::block>, std::shared_ptr<nano::block>> setup_new_account (nano::test::system & system, nano::node & node, nano::uint128_t const amount, nano::keypair source, nano::keypair dest, nano::account dest_rep, bool force_confirm);
```
pwojcikdev, line 45: Also, setup_account name should be fine, there is no setup_old_account variant I suppose. To make it less error prone it should generate and return newly created account keypair to prevent accidental reuse, similar to setup rep.

dsiganos: If do not pass in the dest account, I lose the ability to select the representative of the newly created account in a straightforward way. and the number of returned items would becomes too many.

pwojcikdev: Okay that makes sense

pwojcikdev (approved):

Issue #3946, opened 2022-09-12
Improve polymorphism in `rep_crawler::cleanup_reps()`


simpago:
### Broken Polymorphism
The method `rep_crawler::cleanup_reps()` is responsible for removing representatives that have inactive channels.
The problem is that a `channel` cannot be asked whether it is closed, so rep_crawler checks for the concrete type of the channel and asks for example `tcp_channels` whether that channel still exists. Checking the concrete type of the channel breaks polymorphism!

Here is the relevant code snippet:
```
void nano::rep_crawler::cleanup_reps ()
{
	std::vector<std::shared_ptr<nano::transport::channel>> channels;
	{
		// Check known rep channels
		nano::lock_guard<nano::mutex> lock (probable_reps_mutex);
		auto iterator (probable_reps.get<tag_last_request> ().begin ());
		while (iterator != probable_reps.get<tag_last_request> ().end ())
		{
			if (iterator->channel->get_tcp_endpoint ().address () != boost::asio::ip::address_v6::any ())
			{
				channels.push_back (iterator->channel);
				++iterator;
			}
			else
			{
				// Remove reps with closed channels
				iterator = probable_reps.get<tag_last_request> ().erase (iterator);
			}
		}
	}
	// Remove reps with inactive channels
	for (auto const & i : channels)
	{
		bool equal (false);
		if (i->get_type () == nano::transport::transport_type::tcp)
		{
			auto find_channel (node.network.tcp_channels.find_channel (i->get_tcp_endpoint ()));
			if (find_channel != nullptr && *find_channel == *static_cast<nano::transport::channel_tcp *> (i.get ()))
			{
				equal = true;
			}
		}
		else if (i->get_type () == nano::transport::transport_type::udp)
		{
			auto find_channel (node.network.udp_channels.channel (i->get_endpoint ()));
			if (find_channel != nullptr && *find_channel == *static_cast<nano::transport::channel_udp *> (i.get ()))
			{
				equal = true;
			}
		}
		else if (i->get_type () == nano::transport::transport_type::fake)
		{
			equal = true;
		}
		if (!equal)
		{
			nano::lock_guard<nano::mutex> lock (probable_reps_mutex);
			probable_reps.get<tag_channel_ref> ().erase (*i);
		}
	}
}
```

### Possible Solution
A possible fix for this design issue could be the following:
`tcp_channels` currently removes inactive channels in its `purge (...)` method. It however does _not_ change the state of the channel.

[...]

PR #4362, opened 2024-01-16
Improved polymorphism in rep_crawler::cleanup_reps()

brandon-bb:
Addresses #3946 

This solution is an implementation of the proposed one by @simpago -  channel state is represented using a boolean variable. `tcp_channels::purge()` toggles this variable to close the channel. As this was crafted to introduce minimal changes, there could be a better way to represent channel state.


gr0vity-dev:
I had a quick look at the test results.
It doesn't seem to pass the unit tests yet.
Also when spinning up a local nano network with this version, most of the blocks don't confirm.

Taken from https://ct.bnano.info :
Screenshot 2024-01-16 at 23 13 55

dsiganos:
Not relevant anymore after the rep_crawler overhaul and recent networking improvements.

PR #4365, opened 2024-01-17
Search for receivables as a wallet action, rather than a background node task.

clemahieu:
While working on having a boost::asio::io_context for each nano::node instance rather than in nano::system instances, it uncovered a lifetime issue with background tasks queued by the wallet. nano::wallet::enter_password queues a background task on the node to search for receivable blocks when a wallet is unlocked.

Typically this is not an issue as wallet instances have the same lifetime as node objects. However, in unit tests, a standalone nano::wallets instance is created and destroyed before the node or io_context are destroyed. The body of nano::wallet::search_receivable could still be executing when nano::wallets and nano::wallet instances were destroyed yet the background task survived. This could cause wallet unit tests to fail intermittently if these race conditions ocurred. 

The tests could be rewritten to not create these standalone instances which matches how they're actually used within nano::node instances, however, the wallet already has a way to queue background tasks which is uses for work generation and generating blocks.

Queueing this search as a wallet action also feels like better encapsulation of wallet functionality and doesn't execute this potentially very slow task on an io_context which could affect io performance.

### DIFF
--- a/nano/node/wallet.cpp
+++ b/nano/node/wallet.cpp
@@ -699,8 +699,9 @@ bool nano::wallet::enter_password (store::transaction const & transaction_a, std
 	auto result (store.attempt_password (transaction_a, password_a));
 	if (!result)
 	{
-		auto this_l (shared_from_this ());
-		wallets.node.background ([this_l] () {
+		auto this_l = shared_from_this ();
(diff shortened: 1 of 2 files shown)


pwojcikdev (dismissed):

dsiganos:
The qt unit test wallet.enter_password consistently fails.

dsiganos (approved):

PR #4367, opened 2024-01-18
Store pointer to boost::asio::signal_set and cancel when ipc_server::stop is called.

clemahieu:
Previously the signals weren't cancelled on ipc_server and would remain running until the io_context was destroyed.

### DIFF
--- a/nano/node/ipc/ipc_server.cpp
+++ b/nano/node/ipc/ipc_server.cpp
@@ -14,7 +14,6 @@
 #include <nano/node/json_handler.hpp>
 #include <nano/node/node.hpp>
 
-#include <boost/asio/signal_set.hpp>
 #include <boost/endian/conversion.hpp>
 #include <boost/property_tree/json_parser.hpp>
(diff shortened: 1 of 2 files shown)


nano/node/ipc/ipc_server.cpp
```diff
@@ -650,6 +649,7 @@ nano::ipc::ipc_server::ipc_server (nano::node & node_a, nano::node_rpc_config co
 
 nano::ipc::ipc_server::~ipc_server ()
 {
+	stop ();
```
dsiganos: I believe we said that we would not call stop() from destructors.

dsiganos:
I think we should try to follow the pattern we started of calling stop() to an object and not stopping from the destructor.
In order to do that, we need to rethink the start and stop of all the pic and roc classes.
I started to do that but I found the code very unreadable. So I started with a some refactoring first:
https://github.com/nanocurrency/nano-node/pull/4370

PR #4368, opened 2024-01-18
Convert unit test make_shared<nano::node> to system.add_node.

clemahieu:
When changing the lifetime of io_context it ocurred that a lot of unit tests constructed nodes manually via make_shored.

This change converts many usages to system.add_node. Some usages like the networking tests are not upgraded as they're testing specific operations on how nodes connect which system::add_node assumes.

### DIFF
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -820,12 +820,10 @@ TEST (bootstrap_processor, lazy_hash_pruning)
 	node_flags.disable_bootstrap_bulk_push_client = true;
 	node_flags.enable_pruning = true;
 	auto node0 = system.add_node (config, node_flags);
-	nano::keypair key1;
-	nano::keypair key2;
-	// Generating test chain
 
 	nano::state_block_builder builder;
 
+	// send Gxrb_ratio raw from genesis to genesis
 	auto send1 = builder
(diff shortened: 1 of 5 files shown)


dsiganos (dismissed):

dsiganos:
I think some of these tests manually create a node so that there are no automatic network connections between the nodes.

dsiganos:
For example, the test case bootstrap_processor.lazy_hash_pruning intentionally creates the nano_node instance without using add_node so that there are no connections between the 2 nodes and then manually does a tcp_establish to connect the two nodes when the test wishes to connect them.

dsiganos:
@clemahieu I made some fixes to the tests and committed, please check that you agree with what I am doing, before I carry on and do the rest

clemahieu:
Yes these fixes look good. I thought I had avoided conversions in tests that needed a starting disconnected node, for instance the network.* tests, but I missed these bootstrap ones.

dsiganos:
The principle idea behind this PR is invalid. The idea is to modify the bootstrap and rep_remove tests to system.add_node() instead of manually making their own node. But they manually make their own node because they want a disconnected node. I have made a lot of improvements to the bootstrap unit tests, including comments to stop this mistake from happening again (it has happened before too), but I will commit them as a new PR.

Issue #4364, opened 2024-01-17
Improve assert statements of unit tests


dsiganos:
We have many assert statements in unit test that are checking for equality like this:

ASSERT_TRUE(a == b)
or
ASSERT_TIMELY(5s, a == b)

They should be converted to:
ASSERT_EQ(a, b)
and
ASSERT_TIMELY_EQ(5s, a, b)

RickiNano:
I'd like to take this one

dsiganos:
Great, it's yours!

RickiNano:
@dsiganos 
Do you want a line like this:
`ASSERT_TRUE (node.weight (key1.pub) == amount && node.weight (key2.pub) == amount && node.weight (key3.pub) == amount);`
To be changed into 3 ASSERT_EQ (one for each of the ANDed parts), or should I just leave them?

dsiganos:
Splitting that one makes sense.

RickiNano:
Hey @dsiganos 
I've created a PR for this. I made a small program that automated the process.
If you have more similar changes that you want me to do, I can easily modify the program

dsiganos:
You could also convert (they are less valuable but still good to do):
* ASSERT_TRUE (a > b) -> ASSERT_GT(a, b)
* ASSERT_TRUE (a >= b) -> ASSERT_GE(a, b)
* ASSERT_TRUE (a < b) -> ASSERT_LT(a, b)
* ASSERT_TRUE (a <= b) -> ASSERT_LE(a, b)
Create 4 different commits for each type so that the review is easier to do.

PR #4372, opened 2024-01-18
Use assert_eq and assert_timely_eq

RickiNano:
This PR targets issue #4364 

The way we do assert equals have often been done this way:
ASSERT_TRUE(a == b)
but is now replaced with
ASSERT_EQ(a, b)

Also this format
ASSERT_TIMELY(5s, a == b)
is replaced with
ASSERT_TIMELY_EQ(5s, a, b)

The replacement was automated by a script to avoid human error.
Two asserts had multiple conditions and have manually been split into the individual asserts

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -219,7 +219,7 @@ TEST (active_transactions, keep_local)
 	node.process_active (receive3);
 
 	/// bound elections, should drop after one loop
-	ASSERT_TIMELY (5s, node.active.size () == node_config.active_elections_size);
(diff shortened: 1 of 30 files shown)


dsiganos:
Wow, that's a lot of lines to check!

clemahieu (dismissed):
Took a pass through and it looks good.
A few of the compound expressions were converted to multiple checks which is fine.

RickiNano:
I think I just found a small problem in wallet.cpp:
ASSERT_TIMELY_EQ (5s, node.wallets.mutex.try_lock (), 1);
The try_lock function returns a bool, so I'm not sure why it worked with ASSERT_TIMELY before.
I'll commit a fix

dsiganos (approved):

PR #4376, opened 2024-01-19
Assert conversions part 2

RickiNano:


nano/core_test/uint256_union.cpp
```diff
@@ -549,7 +549,7 @@ void test_union_operator_less_than ()
 template <typename Num>
 void check_operator_less_than (Num lhs, Num rhs)
 {
-	ASSERT_TRUE (lhs < rhs);
```
dsiganos: We should leave the check_operator_ functions as they are since they are specifically checking the operator. ASSERT_LT likely uses the operator internally but in theory it doesn't. It could check for less than like this for example: !(a >= b), which wouldn't be testing the less than operator.

nano/core_test/uint256_union.cpp
```diff
@@ -578,7 +578,7 @@ void test_union_operator_greater_than ()
 template <typename Num>
 void check_operator_greater_than (Num lhs, Num rhs)
 {
-	ASSERT_TRUE (lhs > rhs);
```
dsiganos: Leave it as it was.

dsiganos (changes requested):

nano/core_test/uint256_union.cpp
RickiNano: Ok, I've reverted this

dsiganos (approved):

nano/core_test/active_transactions.cpp
```diff
@@ -983,7 +983,7 @@ TEST (active_transactions, fork_replacement_tally)
 	ASSERT_TIMELY_EQ (5s, count_rep_votes_in_election (), 8);
 
 	auto votes2 (election->votes ());
-	ASSERT_TRUE (votes2.find (nano::dev::genesis_key.pub) != votes2.end ());
```
dsiganos: Leave it this as to was, iterators and equality asserts could cause problems. Leave as it was .

nano/core_test/bootstrap_server.cpp
```diff
@@ -315,7 +315,7 @@ TEST (bootstrap_server, serve_multiple)
 		{
 			// Find matching response
 			auto response_it = std::find_if (all_responses.begin (), all_responses.end (), [&] (auto ack) { return ack.id == next_id; });
-			ASSERT_TRUE (response_it != all_responses.end ());
```
dsiganos: Leave it this as to was, iterators and equality asserts could cause problems. Leave as it was .

nano/core_test/unchecked_map.cpp
```diff
@@ -249,9 +249,9 @@ TEST (unchecked, multiple_get)
 		unchecked1.push_back (i.block->hash ());
 	}
 	// Asserts the payloads where correclty saved
-	ASSERT_TRUE (std::find (unchecked1.begin (), unchecked1.end (), block1->hash ()) != unchecked1.end ());
```
dsiganos: Leave it this as to was, iterators and equality asserts could cause problems. Leave as it was .

nano/core_test/unchecked_map.cpp
```diff
@@ -261,8 +261,8 @@ TEST (unchecked, multiple_get)
 		unchecked2.push_back (i.block->hash ());
 	}
 	// Asserts the payloads where correctly saved
-	ASSERT_TRUE (std::find (unchecked2.begin (), unchecked2.end (), block1->hash ()) != unchecked2.end ());
```
dsiganos: Leave it this as to was, iterators and equality asserts could cause problems. Leave as it was .

dsiganos (changes requested):

nano/core_test/active_transactions.cpp
RickiNano: Done

dsiganos (approved):

dsiganos:
resolves #4364

PR #4380, opened 2024-01-21
During release process, create pull request to increment major version on instead of a direct commit

gr0vity-dev:
Instead of directly trying to commit to the protected develop branch, we can create a pull request with the version increment like this : https://github.com/gr0vity-dev/nano-node/pull/9 which can then be merged after the workflow has ended.

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -65,7 +65,3 @@ jobs:
           git add CMakeLists.txt
           git commit -m "Update CPACK_PACKAGE_VERSION_PRE_RELEASE to 0"
           git push origin ${{ steps.check-release-branch.outputs.release-branch-name }}
-        shell: bash
-        env:
-          GITHUB_ACTOR: ${{ github.actor }}
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -1,4 +1,4 @@
-name:  Prepare Release
+name: Prepare Release
 on:
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/prepare_release.yml
```diff
@@ -37,25 +37,28 @@ jobs:
...
           sed -i.bak "s/set(CPACK_PACKAGE_VERSION_MAJOR \"[0-9]*\")/set(CPACK_PACKAGE_VERSION_MAJOR \"$new_version_major\")/g" CMakeLists.txt
           rm CMakeLists.txt.bak
           git add CMakeLists.txt
           git commit -m "Update CPACK_PACKAGE_VERSION_MAJOR to $new_version_major"
-          git push origin ${{ steps.get-default-branch.outputs.default-branch }}
-          # Reset CPACK_PACKAGE_VERSION_MAJOR to its original value for the release branch
-          git reset --hard HEAD~1 
+          git push origin $NEW_BRANCH
+          # Create pull request
+          export GH_TOKEN=${{ secrets.GITHUB_TOKEN }}
```
dsiganos, line 60: Is it safe to put the GitHub token in an enviroment variable? I don't know if it is or isn't just asking the question.

dsiganos (dismissed):

dsiganos:
LGTM but @clemahieu  should really review it too since he is more familiar now with the release process than I am.

.github/workflows/prepare_release.yml
gr0vity-dev: In this case it should be safe as it never leaves the workflow. 
Its validity is limited to the duration and the scope of the current workflow. (It's not a PAT)
Also its not passed to an external script but required for `gh pr create`

gr0vity-dev:
Just noticed that I dropped `git reset --hard HEAD~1 ` by mistake during the rework.
The MAJOR_VERSION only needs to be incremented for the pull request and not for the releases/v branch

PR #4382, opened 2024-01-21
Fix for node.fork_publish_inactive

dsiganos:
By taking a copy of blocks and saving it into a variable we are not giving a chance to the ASSET_TIMELY to work if the result is not immediately correct.


pwojcikdev (dismissed):
`ASSERT_NE (election->blocks ().end (), election->blocks ().find (send1->hash ()));`
I don't think `ASSERT_NE` is a good choice for comparing iterators

clemahieu (dismissed):

PR #4383, opened 2024-01-21
Fix bootstrap unit tests

### LINKED ISSUE #3517: Unit test `bootstrap_processor.push_diamond_pruning` intermittently failing
Disabled in https://github.com/nanocurrency/nano-node/pull/3512/commits/e1d0a51253d19fa725fb30fc84880c89ddc50527

### LINKED ISSUE #3559: Unit test `bootstrap_processor.push_diamond` intermittently failing
Disabled in https://github.com/nanocurrency/nano-node/pull/3558

CI run in which it failed: [here](https://github.com/nanocurrency/nano-node/runs/4280675502?check_suite_focus=true#step:6:398)

### LINKED ISSUE #3613: Unit test `bulk.genesis_pruning` intermittently failing
Disabled in https://github.com/nanocurrency/nano-node/pull/3611

CI run in which it failed: [here](https://github.com/nanocurrency/nano-node/runs/4521441878?check_suite_focus=true#step:6:425)

### LINKED ISSUE #3640: Unit test `bootstrap_processor.lazy_unclear_state_link` intermittently failing
Disabled in https://github.com/nanocurrency/nano-node/pull/3629

CI run in which it failed: [here](https://github.com/nanocurrency/nano-node/runs/4507281068?check_suite_focus=true#step:5:410)

dsiganos:
This fixes a number of issues with the bootstrap tests.
This work was triggered by https://github.com/nanocurrency/nano-node/pull/4368, which was flawed.

A common mistake that I have seen before is to convert manually created instances of nano::node (without using `system.add_node`). Often people assume that this is just old code and that the manual construction of the object can be tidied up by the add_node call. But in many cases these manual node constructions are there so that the node is not connected to anything, add_node runs code to connect all nodes in a system.

Many tests depended on or simply called block_processor.flush(). I removed all instances of block processor flush as it shouldn't be needed and is known to be buggy.

Some tests were supposedly disabling the legacy bootstrap and testing the lazy bootstrap. But since we now have ascending bootstrap, these tests had to modified to disable the ascending bootstrap too.

Many tests were setting the network port semi-automatically but there is no need to do so, it is more flexible to let the OS pick a port number.

I re-enabled some tests that were disabled due to failing intermittedly.

resolves #3559
resolves #3517
resolves #3640
resolves #3613

### DIFF
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -1458,9 +1458,10 @@ TEST (bootstrap_processor, wallet_lazy_frontier)
 	node0->block_processor.add (receive1);
 	node0->block_processor.add (send2);
 	node0->block_processor.add (receive2);
-	node0->block_processor.flush ();
+	ASSERT_TIMELY (5s, nano::test::exists (*node0, { send1, receive1, send2, receive2 }));
+
 	// Start wallet lazy bootstrap
-	auto node1 (std::make_shared<nano::node> (system.io_ctx, system.get_available_port (), nano::unique_path (), system.work));
+	auto node1 = system.make_disconnected_node ();
 	nano::test::establish_tcp (system, *node1, node0->network.endpoint ());
 	auto wallet (node1->wallets.create (nano::random_wallet_id ()));
 	ASSERT_NE (nullptr, wallet);
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -280,17 +280,19 @@ TEST (bulk_pull, count_limit)
 	ASSERT_EQ (nullptr, block);
 }
 
-TEST (bootstrap_processor, DISABLED_process_none)
+TEST (bootstrap_processor, process_none)
 {
 	nano::test::system system (1);
-	auto node1 (std::make_shared<nano::node> (system.io_ctx, system.get_available_port (), nano::unique_path (), system.work));
-	ASSERT_FALSE (node1->init_error ());
-	auto done (false);
+	auto node0 = system.nodes[0];
+	auto node1 = system.make_disconnected_node ();
+
+	bool done = false;
+	node0->observers.socket_accepted.add ([&] (nano::transport::socket & socket) {
+		done = true;
+	});
+
 	node1->bootstrap_initiator.bootstrap (system.nodes[0]->network.endpoint (), false);
-	while (!done)
-	{
-		system.io_ctx.run_one ();
-	}
+	ASSERT_TIMELY (5s, done);
 	node1->stop ();
 }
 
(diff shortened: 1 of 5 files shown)


nano/test_common/system.cpp
```diff
@@ -563,7 +576,7 @@ void nano::test::system::stop ()
 
 nano::node_config nano::test::system::default_config ()
 {
-	nano::node_config config{ get_available_port (), logging };
+	nano::node_config config{ std::nullopt, logging };
```
pwojcikdev: Historically, `get_available_port ()` was used to guarantee deterministic port allocation when running tests. Later it was modified to use ephemeral ports by default, but still make it possible to have them deterministic if needed. My understanding is that this won't allow that anymore

dsiganos: Yes, you are probably right about that. I had forgotten about that option. 
Should we instead do:
```
nano::node_config config{ get_available_port (true), logging };
```
Would that work both ways?

pwojcikdev: It's defined as `uint16_t get_available_port (bool can_be_zero = true)`, so `nano::node_config config{ get_available_port (), logging };`

dsiganos: I reinstated the previous behaviour.

nano/core_test/bootstrap.cpp
```diff
@@ -400,26 +394,33 @@ TEST (bootstrap_processor, process_new)
...
+
+	// send amount raw from genesis to key2, the wallet will autoreceive
+	auto amount = node1->config.receive_minimum.number ();
+	auto send = system.wallet (0)->send_action (nano::dev::genesis_key.pub, key2.pub, amount);
 	ASSERT_NE (nullptr, send);
-	ASSERT_TIMELY (10s, !node1->balance (key2.pub).is_zero ());
-	auto receive (node2->block (node2->latest (key2.pub)));
+	ASSERT_TIMELY (5s, !node1->balance (key2.pub).is_zero ());
+	auto receive = node2->block (node2->latest (key2.pub));
 	ASSERT_NE (nullptr, receive);
```
clemahieu: There might have been an existing issue here. I don't see how there would be a guarantee that node2 has received the block that was generated on node1 immediately.

nano/core_test/bootstrap.cpp
```diff
@@ -734,9 +757,10 @@ TEST (bootstrap_processor, lazy_hash)
 	node0->block_processor.add (receive1);
 	node0->block_processor.add (send2);
 	node0->block_processor.add (receive2);
-	node0->block_processor.flush ();
```
clemahieu, line 737: Nice to see these disappear.

clemahieu (dismissed):

PR #4384, opened 2024-01-21
Remove cleanup functions related to unchecked blocks.

clemahieu:
Since the addition of the ascending bootstrapped, all unchecked blocks are held in memory and there is already a size limit on how many are kept in the container. Previously these blocks were stored on disk and the number of items was unbounded which necessitated periodic cleanup.

Cleaning the unchecked blocks could actually negatively affect the ascending bootstrapper and clobber critical blocks it needs causing the need to download them again.

### DIFF
--- a/nano/core_test/node.cpp
+++ b/nano/core_test/node.cpp
@@ -3254,45 +3254,6 @@ TEST (node, peer_cache_restart)
 	}
 }
 
-TEST (node, unchecked_cleanup)
-{
-	nano::test::system system{};
-	nano::node_flags node_flags{};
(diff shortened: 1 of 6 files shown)


dsiganos:
We do not need a periodic cleanup as long as new entries replace old ones.
And ideally entries that have entered the unchecked table should be guaranteed some minimum time before deletion (some hysterisis) to avoid attacks where the unchecked tables is constantly filled with rubbish to flush out good entries.

dsiganos (dismissed):

pwojcikdev (dismissed):

dsiganos (approved):

Issue #4379, opened 2024-01-21
Introduce nano::test::block_or_pruned_exists ()


dsiganos:
We would like to have a function called `nano::test::block_or_pruned_exists()`, which would be defined as follows:

```
	/*
	 * Convenience function to check whether a list of hashes exists in node ledger or in the pruned table.
	 * @return true if all blocks are fully processed and inserted in the ledger, false otherwise
	 */
	bool block_or_pruned_exists (nano::node & node, std::vector<nano::block_hash> hashes);

	/*
	 * Convenience function to check whether a list of blocks exists in node ledger or their hash exists in the pruned table.
	 * @return true if all blocks are fully processed and inserted in the ledger, false otherwise
	 */
	bool block_or_pruned_exists (nano::node & node, std::vector<std::shared_ptr<nano::block>> blocks);
```

It would follow a similar pattern to `nano::test::exists pair of functions`.

Then convert multiple lines of unit test code like this:
```
	ASSERT_TRUE (node1.ledger.block_or_pruned_exists (publish1.block->hash ()));
	ASSERT_TRUE (node2.ledger.block_or_pruned_exists (publish2.block->hash ()));
	ASSERT_TRUE (node2.ledger.block_or_pruned_exists (publish3.block->hash ()));
```
into this:
```
ASSERT_TRUE (nano::test::ledger.block_or_pruned_exists (*node1, {publish1, publish2, publish3}));
```

RickiNano:
I'm going to give this a stab

PR #4393, opened 2024-01-24
Convenience function for block_or_pruned_exists

RickiNano:
This PR targets issue #4379

### DIFF
--- a/nano/core_test/bootstrap.cpp
+++ b/nano/core_test/bootstrap.cpp
@@ -1126,7 +1126,7 @@ TEST (bootstrap_processor, DISABLED_lazy_unclear_state_link)
 	node2->bootstrap_initiator.bootstrap_lazy (receive->hash ());
 	// Check processed blocks
 	ASSERT_TIMELY (10s, !node2->bootstrap_initiator.in_progress ());
-	ASSERT_TIMELY (5s, nano::test::block_or_pruned_exists (*node2, { send1, send2, open, receive }));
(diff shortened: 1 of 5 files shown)


pwojcikdev (dismissed):

nano/core_test/bootstrap.cpp
```diff
@@ -1345,10 +1334,7 @@ TEST (bootstrap_processor, lazy_pruning_missing_block)
 	// Some blocks cannot be retrieved from pruned node
 	node2->block_processor.flush ();
 	ASSERT_EQ (1, node2->ledger.cache.block_count);
-	ASSERT_FALSE (node2->ledger.block_or_pruned_exists (send1->hash ()));
-	ASSERT_FALSE (node2->ledger.block_or_pruned_exists (send2->hash ()));
-	ASSERT_FALSE (node2->ledger.block_or_pruned_exists (open->hash ()));
-	ASSERT_FALSE (node2->ledger.block_or_pruned_exists (state_open->hash ()));
+	ASSERT_FALSE (nano::test::block_or_pruned_exists (*node2, { send1, send2, open, state_open }));
```
clemahieu: The ASSERT_FALSE variants of this aren't equivalent to what they replace. This old test is that none of the blocks exist but the new test is only in any of them don't exist.

It would need a different function similar to std::all_of and std::none_of.  https://en.cppreference.com/w/cpp/algorithm/all_any_none_of

PR #4394, opened 2024-01-25
Fix failing TEST (active_transactions, fifo) 

gr0vity-dev:
This testcase manually pushes two blocks to the election scheduler.
However the backlog population sometimes also pushes the block to the election scheduler resulting in it being dropped twice.
--> The solution is to disable the backlog population for this testcase

Screenshot 2024-01-25 at 15 46 22

After this  change I ran the testcase 100k times and saw it failing only once. (Compared to ~1 in 1000 before)

The `stats` indicate that an election can still get activated by priority scheduler:
 `type: election_scheduler detail: insert_priority dir: in value: 1`
 
The only other place I found that runs a priority scheduler is [here](https://github.com/nanocurrency/nano-node/blob/9b6165519d7dd8b029debec20f0d6418170e83a2/nano/node/active_transactions.cpp#L198) in ` nano::active_transactions::activate_successors`


dsiganos:
Nice one! This test case been bugging me! Detailed tracing looks extremely useful!

dsiganos:
`nano::active_transactions::activate_successors` is called from 
`nano::active_transactions::handle_final_votes_confirmation` which is called from
`nano::active_transactions::block_cemented_callback` which is called for every cemented block.

So it appears that every newly cemented block triggers an election for its successor, if we are past the `bootstrap_weight_max_blocks` point.

dsiganos:
Try setting `node.ledger.bootstrap_weight_max_blocks` to a large number so that that activate_successors is not triggered (e.g. 100).

dsiganos (approved):

gr0vity-dev:
> Nice one! This test case been bugging me! Detailed tracing looks extremely useful!

This is simply printing the `stats`to cout. Here is a .patch file to apply those for debugging purposes.
[stats_to_cout.patch](https://github.com/nanocurrency/nano-node/files/14057160/stats_to_cout.patch)

PR #4404, opened 2024-01-26
Remove block_processor::flush function

clemahieu:
The ::flush style functions have been disfavored as they're error and deadlock-prone.  Depending on them in unit tests is problematic as their effect can change over time as components are added or removed from the node.

Removing the block_processor::flush function is done in 5 commits with increasing complexity and finally removing it from the legacy bootstrap processor as well its implementation in the final commit.

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -255,7 +255,6 @@ TEST (active_transactions, inactive_votes_cache)
 	node.vote_processor.vote (vote, std::make_shared<nano::transport::inproc::channel> (node, node));
 	ASSERT_TIMELY_EQ (5s, node.vote_cache.size (), 1);
 	node.process_active (send);
-	node.block_processor.flush ();
 	ASSERT_TIMELY (5s, node.ledger.block_confirmed (node.store.tx_begin_read (), send->hash ()));
 	ASSERT_EQ (1, node.stats.count (nano::stat::type::election, nano::stat::detail::vote_cached));
 }
@@ -356,7 +355,6 @@ TEST (active_transactions, inactive_votes_cache_existing_vote)
 				.build_shared ();
 	node.process_active (send);
 	node.block_processor.add (open);
-	node.block_processor.flush ();
 	ASSERT_TIMELY_EQ (5s, node.active.size (), 1);
 	auto election (node.active.election (send->qualified_root ()));
 	ASSERT_NE (nullptr, election);
@@ -725,7 +723,7 @@ TEST (active_transactions, republish_winner)
 				 .build_shared ();
 
 	node1.process_active (send1);
-	node1.block_processor.flush ();
+	ASSERT_TIMELY (5s, nano::test::exists (node1, { send1 }));
 	ASSERT_TIMELY_EQ (3s, node2.stats.count (nano::stat::type::message, nano::stat::detail::publish, nano::stat::dir::in), 1);
 
 	// Several forks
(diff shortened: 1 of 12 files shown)


nano/core_test/active_transactions.cpp
```diff
@@ -714,7 +712,6 @@ TEST (active_transactions, republish_winner)
 				 .build_shared ();
 
 	node1.process_active (send1);
-	node1.block_processor.flush ();
```
dsiganos, line 728: I think here we need:
```
nano::test::exists(node1, send1)
```

nano/core_test/active_transactions.cpp
```diff
@@ -731,7 +728,6 @@ TEST (active_transactions, republish_winner)
 					.build_shared ();
 		node1.process_active (fork);
 	}
-	node1.block_processor.flush ();
```
dsiganos, line 745: Maybe here we need the use nano::test:exists()...

nano/core_test/confirmation_height.cpp
```diff
@@ -61,7 +61,6 @@ TEST (confirmation_height, single)
 		ASSERT_EQ (nano::dev::genesis->hash (), confirmation_height_info.frontier);
 
 		node->process_active (send1);
-		node->block_processor.flush ();
```
dsiganos, line 64: nano::test::exists(send1)?

nano/node/bootstrap/bootstrap_legacy.cpp
```diff
@@ -223,10 +223,10 @@ void nano::bootstrap_attempt_legacy::run ()
 			condition.wait (lock, [&stopped = stopped, &pulling = pulling] { return stopped || pulling == 0; });
 		}
 
-		// Flushing may resolve forks which can add more pulls
-		lock.unlock ();
-		node->block_processor.flush ();
-		lock.lock ();
+		while (!stopped && !node->active.empty ())
```
dsiganos: If we are in operating in saturation mode, which we have to assume we will be, then this will mean that the legacy bootstrap will never try to do a bootstrap? I think it can't be a binary yes or no, if the AEC is busy/ful the bootstrap should be discouraged but never stopped completely.

nano/core_test/active_transactions.cpp
clemahieu: Added.

nano/core_test/confirmation_height.cpp
clemahieu: Added.

nano/core_test/active_transactions.cpp
clemahieu: Added.

nano/node/bootstrap/bootstrap_legacy.cpp
clemahieu: Looking at this, I made an incorrect chang by looking at the AEC rather than at the emptiness of the block_processor. I changed the implementation to check the block_processor is empty rather than the AEC is empty.

dsiganos: But even then, my comment still holds, under saturation, will then block processor ever be empty?
And if it is never empty, will the bootstrap never kick in?

clemahieu: I updated this wait to have a timeout and added a TODO comment to improve it as the existing behavior didn't have this timeout. The legacy bootstrap is planned to be removed and there are improvements for tagging blocks through the block processor that @pwojcikdev has in the pipeline.

dsiganos (approved):

PR #4405, opened 2024-01-26
Fix #4390: Wanted: to_json() function for nano::block_sideband

### LINKED ISSUE #4390: Wanted: to_json() function for nano::block_sideband
We currently have no way to convert a block sideband object to json representation for debugging purposes.
It would be nice to have that ability.

The task is to implement the function, which returns a string representing the state of the object in json format:
```
std::string nano::block_sideband::to_json() const;
```

Napetrou:
to_json() function converts a block sideband object to json representation.


nano/lib/blocks.cpp
```diff
@@ -1866,6 +1866,11 @@ bool nano::block_sideband::deserialize (nano::stream & stream_a, nano::block_typ
 	return result;
 }
 
+std::string nano::block_sideband::to_json () const
```
dsiganos, line 1869: Thank you for the effort, but I would expect to see something much more elegant and much easier to read.

Issue #4402, opened 2024-01-26
Use nano::test::make_final_vote() instead of manually constructing final votes in unit tests


dsiganos:
Replace final votes construction of the type:
```
auto vote (std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, nano::vote::timestamp_max, nano::vote::duration_max, std::vector<nano::block_hash> (1, send->hash ())));
```

with:

```
nano::test:make_final_vote(nano::dev::genesis_key, { send });
```

Be careful, we have many types of votes, only replace the final vote generation in this task to keep the change easily reviewable.
There should roughly 10 instances of it.

RickiNano:
I'm working on this

PR #4407, opened 2024-01-26
Use make_final_vote function

RickiNano:
This PR target issue #4402 
Uses the make_final_vote function in unit tests to improve readability

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -882,7 +882,7 @@ TEST (active_transactions, fork_replacement_tally)
 					.build_shared ();
 		node1.process_active (open);
 		// Confirmation
-		auto vote = nano::test::make_final_vote (nano::dev::genesis_key, { open });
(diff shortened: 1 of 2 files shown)


nano/core_test/active_transactions.cpp
```diff
@@ -882,7 +882,7 @@ TEST (active_transactions, fork_replacement_tally)
 					.build_shared ();
 		node1.process_active (open);
 		// Confirmation
-		auto vote (std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, nano::vote::timestamp_max, nano::vote::duration_max, std::vector<nano::block_hash>{ send->hash (), open->hash () }));
```
dsiganos, line 885: This is wrong, you missed the block "send."

RickiNano: You're right. I've added it now

dsiganos (approved):

PR #4410, opened 2024-01-27
Use the make_vote function in unit tests

### LINKED ISSUE #4403: Use nano::test::make_vote() instead of manually constructing votes in unit tests
NOTE: Issue https://github.com/nanocurrency/nano-node/issues/4402 should be implemented before doing this.

Replace votes construction of the type:
```
auto vote (std::make_shared<nano::vote> (nano::dev::genesis_key.pub, nano::dev::genesis_key.prv, X, Y, std::vector<nano::block_hash> (1, send->hash ())));
```

where X is not nano::vote::timestamp_max and Y is not nano::vote::duration_max

with:

```
nano::test:make_vote(nano::dev::genesis_key, { send }, X, Y);
```

Only replace code in unit test files. there should be roughly 30-40 instances to change.

RickiNano:
Uses test_util function to create votes in unit tests

resolves #4403

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -362,7 +362,7 @@ TEST (active_transactions, inactive_votes_cache_existing_vote)
 	ASSERT_NE (nullptr, election);
 	ASSERT_GT (node.weight (key.pub), node.minimum_principal_weight ());
 	// Insert vote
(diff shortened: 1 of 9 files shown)


dsiganos (approved):

dsiganos:
The nano::test::make_vote with zero timestamp and zero duration can take advantage of the arguments having defaults arguments set to 0 and not set the last 2 arguments (0, 0). But it is fine as it is, it isn't really a problem as it is.

PR #4413, opened 2024-01-30
Ignore selected peers

RickiNano:
This PR introduces an optional list of ip addresses that a node operator wants to ignore. The peers are inserted in the config file as "blocked_peers"
Both ip v4 and v6 address formats are supported and can be mixed. Example:
blocked_peers = ["::ffff:10.0.0.1", "192.168.0.1"]
This could be useful for blocking malicious nodes.

### DIFF
--- a/nano/core_test/toml.cpp
+++ b/nano/core_test/toml.cpp
@@ -180,6 +180,7 @@ TEST (toml, daemon_config_deserialize_defaults)
 	ASSERT_EQ (conf.node.peering_port, defaults.node.peering_port);
 	ASSERT_EQ (conf.node.pow_sleep_interval, defaults.node.pow_sleep_interval);
 	ASSERT_EQ (conf.node.preconfigured_peers, defaults.node.preconfigured_peers);
+	ASSERT_EQ (conf.node.blocked_peers, defaults.node.blocked_peers);
 	ASSERT_EQ (conf.node.preconfigured_representatives, defaults.node.preconfigured_representatives);
 	ASSERT_EQ (conf.node.receive_minimum, defaults.node.receive_minimum);
 	ASSERT_EQ (conf.node.signature_checker_threads, defaults.node.signature_checker_threads);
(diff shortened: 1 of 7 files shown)


nano/core_test/toml.cpp
```diff
@@ -407,6 +408,7 @@ TEST (toml, daemon_config_deserialize_no_defaults)
 	peering_port = 999
 	pow_sleep_interval= 999
 	preconfigured_peers = ["dev.org"]
+	blocked_peers = ["192.168.0.1"]
```
clemahieu: This should also test ipv6 decoding.

nano/node/network.cpp
```diff
@@ -347,6 +349,39 @@ void nano::network::broadcast_confirm_req_many (std::deque<std::pair<std::shared
 	}
 }
 
+void nano::network::Configure_blocked_peers ()
```
clemahieu: This shouldn't have a capital in the name.

nano/node/network.cpp
```diff
@@ -468,6 +503,12 @@ class network_message_visitor : public nano::message_visitor
 
 void nano::network::process_message (nano::message const & message, std::shared_ptr<nano::transport::channel> const & channel)
 {
+	if (is_ip_blocked (channel->get_tcp_endpoint ().address ()))
```
clemahieu: We shouldn't be tracking a channel for a blocked ip at all.

I think a better place to handle this is in tcp_listener::on_connection within the asinc_accept callback. The first check can be if it's a blocked IP and if so, it can issue a new call to on_connection which will wait for the next connection.

clemahieu:
We should also have an affirmative test that shows the ip blocking works.

RickiNano:
I've moved the logic to the tcp_listener::on_connection function. I will try to add some unit tests next

nano/node/transport/tcp_server.cpp
```diff
@@ -136,6 +137,13 @@ void nano::transport::tcp_listener::on_connection (std::function<bool (std::shar
 		[this_l, new_connection, cbk = std::move (callback)] (boost::system::error_code const & ec_a) mutable {
 			this_l->evict_dead_connections ();
 
+			if (this_l->is_ip_blocked (new_connection->remote.address ()))
+			{
+				this_l->node.logger.info (nano::log::type::tcp_listener, "Connection refused from blocked IP: {}", new_connection->remote_endpoint ().address ().to_string ());
```
clemahieu, line 142: I think this should remain a debug log as it could be high-volume.

PR #4420, opened 2024-02-05
Prevent elections from being started after being processed by `confirmation_heigh_processor`

gr0vity-dev:
Currently it is possible that an election gets scheduled for a block that is being processed or has been processed by the `confirmation_heigh_processor`.

This introduces a deque of recently confirmed hashes by the `confirmation_height_unbounded` processor and keeps track of the last 65536 confirmed hashes, similar to the `active_transactions`class.

It addresses the intermittent failures of the following test cases :
- `election_scheduler.no_vacancy`
- `active_transactions.inactive_votes_cache_election_start`

### DIFF
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -428,7 +428,7 @@ nano::election_insertion_result nano::active_transactions::insert (std::shared_p
 	auto const existing = roots.get<tag_root> ().find (root);
 	if (existing == roots.get<tag_root> ().end ())
 	{
-		if (!recently_confirmed.exists (root) && !node.confirmation_height_processor.is_processing_or_confirmed (hash))
+		if (!recently_confirmed.exists (root) && !confirmation_height_processor.is_processing_or_confirmed (hash))
 		{
 			result.inserted = true;
 			auto observe_rep_cb = [&node = node] (auto const & rep_a) {
--- a/nano/node/active_transactions.cpp
+++ b/nano/node/active_transactions.cpp
@@ -428,7 +428,7 @@ nano::election_insertion_result nano::active_transactions::insert (std::shared_p
 	auto const existing = roots.get<tag_root> ().find (root);
 	if (existing == roots.get<tag_root> ().end ())
 	{
-		if (!recently_confirmed.exists (root))
+		if (!recently_confirmed.exists (root) && !node.confirmation_height_processor.is_processing_or_confirmed (hash))
 		{
 			result.inserted = true;
 			auto observe_rep_cb = [&node = node] (auto const & rep_a) {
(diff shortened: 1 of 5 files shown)


nano/node/confirmation_height_unbounded.cpp
```diff
@@ -469,6 +481,15 @@ bool nano::confirmation_height_unbounded::has_iterated_over_block (nano::block_h
 	return block_cache.count (hash_a) == 1;
 }
 
+bool nano::confirmation_height_unbounded::has_iterated_or_confirmed (nano::block_hash const & hash_a) const
```
dsiganos, line 484: Can you add a comment to explain what this function does?
It is not immediately obvious what iterated means in this context.
Also, it would be nice to make it clear, if confirmed means cemented.

dsiganos:
If you understand what confirmation_height_unbounded does then could you add some comments in the header file to explain?

nano/node/active_transactions.cpp
```diff
@@ -428,7 +428,7 @@ nano::election_insertion_result nano::active_transactions::insert (std::shared_p
 	auto const existing = roots.get<tag_root> ().find (root);
 	if (existing == roots.get<tag_root> ().end ())
 	{
-		if (!recently_confirmed.exists (root))
+		if (!recently_confirmed.exists (root) && !node.confirmation_height_processor.is_processing_or_confirmed (hash))
```
dsiganos: There is no need to refer to `node`. nano::active_transactions has `confirmation_height_processor`.
```
!confirmation_height_processor.is_processing_or_confirmed (hash)
```

nano/node/confirmation_height_unbounded.cpp
gr0vity-dev: There is an existing function called `has_iterated_over_block` .
Here `has_iterated` means the same.
A straightforward implementation would involve calling the existing `has_iterated_over_block` and a new `is_recently_confirmed` each acquiring and releasing the mutex lock independently.
Something like : `return has_iterated_over_block() || is_recently_confirmed()`

 I opted to implement `has_iterated_or_confirmed` by duplicating the logic from both methods under a single mutex lock acquisition.

gr0vity-dev:
My understanding of the `confirmation_height_unbounded` is that it is responsible to recursively confirm blocks if a successor has already been confirmed.
However in the context of this pull_request I didn't look at the class closely to understand what it does in detail. 
I only saw that `recently_confirmed.exists`inside `active_transactions` class doesn't cover recursively confirmed blocks by the `confirmation_height_processor`.

gr0vity-dev:
Closing this while waiting for bounded backlog to simplify the confirmation height processor

PR #4425, opened 2024-02-10
Remove frontiers table from ledger

clemahieu:
The frontiers table is redundant and only used for legacy block processing. It tracks a mapping of head block hash to account number. This information can already be obtained by the previous block's sideband information, which contains the account number, and the account_info which contains the head block.

This rewrites the ledger checks in terms of the already existing information and removes updates to the frontiers table and associated tests. This also does a ledger upgrade which drops the frontiers table.


nano/core_test/block_store.cpp
```diff
@@ -1400,6 +1395,42 @@ TEST (mdb_block_store, upgrade_v21_v22)
 	// Testing the upgrade code worked
 	check_correct_state ();
 }
+
+TEST (mdb_block_store, upgrade_v22_v23)
+{
+	if (nano::rocksdb_config::using_rocksdb_in_tests ())
+	{
+		// Don't test this in rocksdb mode
```
dsiganos: This comment is not very useful. The fact that we are skipping this test in RocksDB mode is obvious from the code. What I would really like to know is why we are skipping this test in RocksDB mode.

nano/secure/ledger.cpp
```diff
@@ -466,12 +446,12 @@ void ledger_processor::change_block (nano::change_block & block_a)
 			result.code = block_a.valid_predecessor (*previous) ? nano::process_result::progress : nano::process_result::block_position;
 			if (result.code == nano::process_result::progress)
 			{
-				auto account (ledger.store.frontier.get (transaction, block_a.hashables.previous));
-				result.code = account.is_zero () ? nano::process_result::fork : nano::process_result::progress;
+				auto account = ledger.account (*previous);
```
dsiganos: Can we trust the sideband to have the account information?
I noticed recently that side band dos not always hold the account of a block (copied from slack chat):
```
The field account of class nano::block_sideband is always set to all zeros for state blocks and for open blocks read from the ledger. Is that a bug? It seems very strange to have a field that is mostly incorrect
```

clemahieu: This is using ledger::account which will get it either from the sideband when appropriate or directly from the block

https://github.com/nanocurrency/nano-node/blob/a39259bdfdca747ff0f3514d1c5dfd6389398744/nano/secure/ledger.cpp#L1137

I do have some concern with this implementation as it's not obviously how it works, it returns the non-zero value from either sideband or the block, or zero if they're both zero. The way I have this implemented in the block checker context is more direct:

https://github.com/nanocurrency/nano-node/blob/d6e172eb5551e5a952d57dca32de7e6bec09bbc0/nano/node/block_check_context.cpp#L79

dsiganos (dismissed):

nano/secure/ledger.cpp
```diff
@@ -466,12 +446,12 @@ void ledger_processor::change_block (nano::change_block & block_a)
...
+				auto info = ledger.account_info (transaction, account);
+				debug_assert (info);
```
pwojcikdev: From std::optional
> The behavior is undefined if *this does not contain a value.

Let's use `release_asserts` in such a critical piece of code (almost everywhere in ledger.cpp). There's no recovery codepath, so this should completely crash.

nano/secure/ledger.cpp
```diff
@@ -563,8 +539,10 @@ void ledger_processor::receive_block (nano::receive_block & block_a)
 			result.code = block_a.valid_predecessor (*previous) ? nano::process_result::progress : nano::process_result::block_position;
 			if (result.code == nano::process_result::progress)
 			{
-				auto account (ledger.store.frontier.get (transaction, block_a.hashables.previous));
-				result.code = account.is_zero () ? nano::process_result::gap_previous : nano::process_result::progress; // Have we seen the previous block? No entries for account at all (Harmless)
+				auto account = ledger.account (*previous);
+				auto info = ledger.account_info (transaction, account);
+				debug_assert (info);
+				result.code = info->head != block_a.hashables.previous ? nano::process_result::gap_previous : nano::process_result::progress; // Have we seen the previous block? No entries for account at all (Harmless)
```
pwojcikdev: This does not seem equivalent to the previous check:
```
result.code = account.is_zero () ? nano::process_result::gap_previous : nano::process_result::progress; // Have we seen the previous block? No entries for account at all (Harmless)
```
If account info head is not empty and does not point to the previous hash, isn't it a fork?

pwojcikdev: The check for `gap_previous` is done a few lines above:
```
result.code = previous != nullptr ? nano::process_result::progress : nano::process_result::gap_previous;
```

nano/secure/ledger.cpp
clemahieu: Agreed, I will raise a separate PR to replace the debug_asserts in the ledger_processor with release_asserts.

nano/secure/ledger.cpp
clemahieu: Fixed this, the fork checking is now done in a similar fashion to other blocks.

clemahieu:
Split this commit. This is the second part which includes database changes which we can merge later in the release cycle since it upgrades the database. The first part is in this commit https://github.com/nanocurrency/nano-node/pull/4460

PR #4437, opened 2024-02-17
work_pool: move stopping from destructor to stop function

dsiganos:
The work_pool starts a number of threads and joins them in the destructor.
This is pattern we want to break. We want to have a separate stop stage before destruction.
This change makes work_pool follow the new pattern.

### DIFF
--- a/nano/core_test/block_store.cpp
+++ b/nano/core_test/block_store.cpp
@@ -1012,14 +1012,14 @@ TEST (mdb_block_store, sideband_height)
 	nano::block_builder builder;
 	auto transaction (store.tx_begin_write ());
 	store.initialize (transaction, ledger.cache, nano::dev::constants);
-	nano::work_pool pool{ nano::dev::network_params.network, std::numeric_limits<unsigned>::max () };
+	nano::test::start_stop_container<nano::work_pool> pool{ nano::dev::network_params.network, std::numeric_limits<unsigned>::max () };
 	auto send = builder
 				.send ()
 				.previous (nano::dev::genesis->hash ())
 				.destination (nano::dev::genesis_key.pub)
 				.balance (nano::dev::constants.genesis_amount - nano::Gxrb_ratio)
 				.sign (nano::dev::genesis_key.prv, nano::dev::genesis_key.pub)
-				.work (*pool.generate (nano::dev::genesis->hash ()))
+				.work (*pool.obj.generate (nano::dev::genesis->hash ()))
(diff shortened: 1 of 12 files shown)


nano/core_test/block_store.cpp
```diff
@@ -1012,30 +1012,30 @@ TEST (mdb_block_store, sideband_height)
...
-	nano::work_pool pool{ nano::dev::network_params.network, std::numeric_limits<unsigned>::max () };
+	nano::test::start_stop_container<nano::work_pool> pool{ nano::dev::network_params.network, std::numeric_limits<unsigned>::max () };
 	auto send = builder
 				.send ()
 				.previous (nano::dev::genesis->hash ())
 				.destination (nano::dev::genesis_key.pub)
 				.balance (nano::dev::constants.genesis_amount - nano::Gxrb_ratio)
 				.sign (nano::dev::genesis_key.prv, nano::dev::genesis_key.pub)
-				.work (*pool.generate (nano::dev::genesis->hash ()))
+				.work (*pool.obj.generate (nano::dev::genesis->hash ()))
```
pwojcikdev: Why do we need this strange pattern of `pool.obj...`? Keep it as just `pool`. Simpler is better.

dsiganos: Can you suggest how? this is the simplest I could think of.
I thought of overloading the -> operator but too much overloading is not simple and I decided against it.
Do you mean by using some kind of overloaded operator?

nano/test_common/testutil.hpp
```diff
@@ -254,6 +266,28 @@ namespace test
...
 
+	/**
+	 * A container that constructs and stores an object and calls `start` from constructor and `stop` from destructor (RAII)
+	 */
+	template <typename T>
+		requires start_stoppable<T>
+	struct start_stop_container
+	{
+		template <typename... Args>
+		explicit start_stop_container (Args &&... args) :
```
pwojcikdev: I'm having trouble understanding why this class is implement this way, this doesn't seem to be in-line with the goal of having a variadic start stop guard.

dsiganos: I am trying to create a class that will hold an object and call start and stop on it.
Is there a bug or are you against the idea?

pwojcikdev (approved):

dsiganos:
I missed a whole bunch of work_pool cases that are exercised when start node in non-daemon modes.
I will fix them.

PR #4442, opened 2024-02-19
Clean up ledger::process return value.

clemahieu:
Remove process_return class which only contained process_result.
Rename process_result to block_status to better indicate its purpose.

### DIFF
--- a/nano/core_test/active_transactions.cpp
+++ b/nano/core_test/active_transactions.cpp
@@ -149,7 +149,7 @@ TEST (active_transactions, confirm_frontier)
 		node2.rep_crawler.probable_reps.emplace (nano::dev::genesis_key.pub, *peers.begin ());
 	}
 
-	ASSERT_EQ (nano::process_result::progress, node2.process (*send).code);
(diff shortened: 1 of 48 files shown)


RickiNano:
I'm new to the codebase so it may just be me who is confused, but I find the name ledger_code to be confusing and not descriptive of what it is. The word "code" is very generic and can be so many things.
How about ledger_result? Or ledger_status?
Or if code needs to be in the name, then ledger_result_code or ledger_status_code

clemahieu:
I think a longer name here is overly descriptive. The main c++ type to report errors is std::error_code so I think nano::ledger_code matches this.

I would have made this a nested enum of ledger nano::ledger::code, though nested types cannot be forward declared so I named it nano::ledger_code.

This is some external discussion on overly long names: https://wiki.c2.com/?VeryLongDescriptiveNamesThatProgrammingPairsThinkProvideGoodDescriptions

RickiNano:
I was not advocating for longer naming per se, but just replacing "code" with "status" or "result" that has more meaning (at least to me :) )

clemahieu:
Changed to nano::block_status

PR #4443, opened 2024-02-20
Convert ledger::process to take a shared_ptr argument instead of a block reference.

clemahieu:


nano/core_test/active_transactions.cpp
```diff
@@ -1041,7 +1041,10 @@ TEST (active_transactions, confirmation_consistency)
 }
 }
 
-TEST (active_transactions, confirm_new)
+// Test disabled because it's failing intermittently.
+// PR in which it got disabled: https://github.com/nanocurrency/nano-node/pull/3629
+// Issue for investigating it: https://github.com/nanocurrency/nano-node/issues/3634
+TEST (active_transactions, DISABLED_confirm_new)
```
dsiganos: This is removing functionality that should not be removed.

dsiganos: I pushed a commit that solves this.

dsiganos (approved):

PR #4447, opened 2024-02-23
Node shutdown reliability improvements

clemahieu:
Cleans up connections waiting to connect when tcp_channels::stop is called which would otherwise leave waiting connection attempts in the io_context queue.
Wait for vote_generator thread to start before returning from vote_generator::start which may run in to an issue where an immediate call to vote_generator::stop would find a non-joinable thread.


pwojcikdev:
> Wait for vote_generator thread to start before returning from vote_generator::start which may run in to an issue where an immediate call to vote_generator::stop would find a non-joinable thread.

How is it possible? If initialization, starting and stopping is done by the main thread, this should never happen.

PR #4460, opened 2024-03-05
Remove usages of frontiers table from ledger_processor block checks

clemahieu:
This is the first part of two commits to remove the frontier table from the database. This removes usages of the values in the frontier table from ledger_processor.

The second part https://github.com/nanocurrency/nano-node/pull/4425 will do the database upgrade which can be merged later in the release cycle.

### DIFF
--- a/nano/secure/ledger.cpp
+++ b/nano/secure/ledger.cpp
@@ -463,12 +463,12 @@ void ledger_processor::change_block (nano::change_block & block_a)
 			result = block_a.valid_predecessor (*previous) ? nano::block_status::progress : nano::block_status::block_position;
 			if (result == nano::block_status::progress)
 			{
-				auto account (ledger.store.frontier.get (transaction, block_a.hashables.previous));
-				result = account.is_zero () ? nano::block_status::fork : nano::block_status::progress;
+				auto account = previous->account ();
+				auto info = ledger.account_info (transaction, account);
+				debug_assert (info);
+				result = info->head != block_a.hashables.previous ? nano::block_status::fork : nano::block_status::progress;
 				if (result == nano::block_status::progress)
 				{
-					auto info = ledger.account_info (transaction, account);
-					debug_assert (info);
 					debug_assert (info->head == block_a.hashables.previous);
 					result = validate_message (account, hash, block_a.signature) ? nano::block_status::bad_signature : nano::block_status::progress; // Is this block signed correctly (Malformed)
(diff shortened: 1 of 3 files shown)


nano/secure/ledger.cpp
```diff
@@ -563,8 +563,10 @@ void ledger_processor::receive_block (nano::receive_block & block_a)
 			result = block_a.valid_predecessor (*previous) ? nano::block_status::progress : nano::block_status::block_position;
 			if (result == nano::block_status::progress)
 			{
-				auto account (ledger.store.frontier.get (transaction, block_a.hashables.previous));
-				result = account.is_zero () ? nano::block_status::gap_previous : nano::block_status::progress; // Have we seen the previous block? No entries for account at all (Harmless)
+				auto account = ledger.account (*previous);
+				auto info = ledger.account_info (transaction, account);
+				debug_assert (info);
+				result = info->head != block_a.hashables.previous ? nano::block_status::fork : nano::block_status::progress; // Have we seen the previous block? No entries for account at all (Harmless)
```
pwojcikdev: I see the handling of forks is fixed, but I believe the comment on this line should be updated to `// If we have the block but it's not the latest we have a signed fork `

pwojcikdev (dismissed):

nano/secure/ledger.cpp
```diff
@@ -574,8 +576,6 @@ void ledger_processor::receive_block (nano::receive_block & block_a)
 						result = ledger.block_or_pruned_exists (transaction, block_a.hashables.source) ? nano::block_status::progress : nano::block_status::gap_source; // Have we seen the source block already? (Harmless)
 						if (result == nano::block_status::progress)
 						{
-							auto info = ledger.account_info (transaction, account);
-							debug_assert (info);
 							result = info->head == block_a.hashables.previous ? nano::block_status::progress : nano::block_status::gap_previous; // Block doesn't immediately follow latest block (Harmless)
```
dsiganos, line 576: Why do we need this check and why is it returning gap_previous?

dsiganos: The setting of the variable `info` moved upwards but this check, which is now redundant, was left behind.

dsiganos (dismissed):

nano/secure/ledger.cpp
dsiganos: I created a separate issue for this: https://github.com/nanocurrency/nano-node/issues/4473

PR #4461, opened 2024-03-05
Ledger query simplification

clemahieu:
This PR merges the ledger::account/balance/amount variants in to a single function for each, returning an optional associated value.

The major advantage is a simplification in the nano::ledger interface and increased type-checking in values returned.  Previously there were separate functions for assumed-correct function calls protected with debug_assert, and variants permitted to fail that returned error codes or sentinel values.

These changes replace each account/amount/balance with functions that return std::optional values for each associated type. Variants that used assumed-correct functions simply call std::optional::value, which does runtime value checking which increases correctness safety. Variants that used _safe variants or checked sentinel values are replaced with std::optional bool conversion checks and operate the same way.

### DIFF
--- a/nano/nano_node/entry.cpp
+++ b/nano/nano_node/entry.cpp
@@ -1494,7 +1494,7 @@ int main (int argc, char * const * argv)
 							bool error_or_pruned (false);
 							if (!state_block.hashables.previous.is_zero ())
 							{
-								prev_balance = node->ledger.balance_safe (transaction, state_block.hashables.previous, error_or_pruned);
+								prev_balance = node->ledger.balance (transaction, state_block.hashables.previous).value_or (0);
 							}
 							if (node->ledger.is_epoch_link (state_block.hashables.link))
 							{
@@ -1518,11 +1518,10 @@ int main (int argc, char * const * argv)
 					}
 					else
 					{
-						bool error_or_pruned (false);
-						auto prev_balance (node->ledger.balance_safe (transaction, block->previous (), error_or_pruned));
-						if (!node->ledger.pruning || !error_or_pruned)
+						auto prev_balance = node->ledger.balance (transaction, block->previous ());
+						if (!node->ledger.pruning || prev_balance)
 						{
-							if (block->balance () < prev_balance)
+							if (block->balance () < prev_balance.value ())
 							{
 								// State send
 								block_details_error = !sideband.details.is_send || sideband.details.is_receive || sideband.details.is_epoch;
@@ -1534,7 +1533,7 @@ int main (int argc, char * const * argv)
 									// State change
 									block_details_error = sideband.details.is_send || sideband.details.is_receive || sideband.details.is_epoch;
 								}
-								else if (block->balance () == prev_balance && node->ledger.is_epoch_link (block->link ()))
+								else if (block->balance () == prev_balance.value () && node->ledger.is_epoch_link (block->link ()))
 								{
 									// State epoch
 									block_details_error = !sideband.details.is_epoch || sideband.details.is_send || sideband.details.is_receive;
--- a/nano/node/bootstrap/bootstrap_server.cpp
+++ b/nano/node/bootstrap/bootstrap_server.cpp
@@ -286,7 +286,7 @@ nano::asc_pull_ack nano::bootstrap_server::process (const store::transaction & t
 		case asc_pull_req::hash_type::block:
 		{
 			// Try to lookup account assuming target is block hash
-			target = ledger.account_safe (transaction, request.target.as_block_hash ());
+			target = ledger.account (transaction, request.target.as_block_hash ()).value_or (0);
 		}
 		break;
 	}
--- a/nano/node/bootstrap_ascending/service.cpp
+++ b/nano/node/bootstrap_ascending/service.cpp
@@ -166,7 +166,7 @@ void nano::bootstrap_ascending::service::inspect (store::transaction const & tx,
 		break;
 		case nano::block_status::gap_source:
 		{
-			const auto account = block.previous ().is_zero () ? block.account () : ledger.account (tx, block.previous ());
+			const auto account = block.previous ().is_zero () ? block.account () : ledger.account (tx, block.previous ()).value ();
 			const auto source = block.source ().is_zero () ? block.link ().as_block_hash () : block.source ();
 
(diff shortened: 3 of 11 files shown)


nano/nano_node/entry.cpp
```diff
@@ -1494,7 +1494,10 @@ int main (int argc, char * const * argv)
 							bool error_or_pruned (false);
 							if (!state_block.hashables.previous.is_zero ())
 							{
-								prev_balance = node->ledger.balance_safe (transaction, state_block.hashables.previous, error_or_pruned);
+								if (auto balance = node->ledger.balance (transaction, state_block.hashables.previous))
+								{
+									prev_balance = balance.value ();
+								}
```
pwojcikdev: Those `if (opt = ...) { value = opt.value () }` blocks could in many cases be replaced simply by one-liner: `opt.value_or({})` but it's a cosmetic improvement - new style is better either way.

nano/node/bootstrap_ascending/service.cpp
```diff
@@ -132,7 +132,7 @@ void nano::bootstrap_ascending::service::inspect (store::transaction const & tx,
 	{
 		case nano::block_status::progress:
 		{
-			const auto account = ledger.account (tx, hash);
+			const auto account = ledger.account (block);
```
pwojcikdev, line 135: Not reusing existing db transaction is a big red flag, given that this is called very frequently, what's the performance impact here?

nano/node/json_handler.cpp
```diff
@@ -1150,13 +1150,12 @@ void nano::json_handler::block_info ()
...
-			auto amount (node.ledger.amount_safe (transaction, hash, error_or_pruned));
-			if (!error_or_pruned)
+			auto amount = node.ledger.amount (transaction, hash);
+			if (amount)
 			{
-				response_l.put ("amount", amount.convert_to<std::string> ());
+				response_l.put ("amount", amount.value ().convert_to<std::string> ());
 			}
-			auto balance (node.ledger.balance (transaction, hash));
+			auto balance = node.ledger.balance (*block);
```
pwojcikdev, line 1158: Not reusing existing db transaction.

nano/node/json_handler.cpp
```diff
@@ -1214,20 +1213,19 @@ void nano::json_handler::block_confirm ()
 				nano::election_status status{ block_l, 0, 0, std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now ().time_since_epoch ()), std::chrono::duration_values<std::chrono::milliseconds>::zero (), 0, 1, 0, nano::election_status_type::active_confirmation_height };
 				node.active.recently_cemented.put (status);
 				// Trigger callback for confirmed block
-				auto account (node.ledger.account (transaction, hash));
-				bool error_or_pruned (false);
-				auto amount (node.ledger.amount_safe (transaction, hash, error_or_pruned));
+				auto account = node.ledger.account (*block_l);
```
pwojcikdev, line 1216: Not reusing existing db transaction - there are many call sites like this. Is there any reason for those changes? Reusing db transactions seems to me like a simple and positive thing, it's both faster and guarantees consistent view - something that might be important down the road with database sharding.

clemahieu: These conversions from (tx, hash) to (block &) are misleading, they don't open new transactions on the database, these are utility functions that are hung on the ledger class that pull the values from the block data or sideband.

I'm going to make a followup PR that moves these on to the block itself.

nano/nano_node/entry.cpp
clemahieu: That's a good point. I'll convert these as I like that style better.

nano/node/json_handler.cpp
pwojcikdev: Ahh, I see, it now makes sense. I should have started reviewing this PR by looking at the ledger class itself.

nano/secure/ledger.cpp
```diff
@@ -834,30 +831,18 @@ nano::uint128_t nano::ledger::balance (nano::block const & block)
 }
 
 // Balance for account containing hash
-nano::uint128_t nano::ledger::balance (store::transaction const & transaction, nano::block_hash const & hash) const
+std::optional<nano::uint128_t> nano::ledger::balance (store::transaction const & transaction, nano::block_hash const & hash) const
 {
 	if (hash.is_zero ())
 	{
 		return 0;
```
pwojcikdev: Being consistent with the new style of using optionals to indicate failed ops, should this perhaps return `std::nullopt` ?

nano/secure/ledger.cpp
```diff
@@ -1156,68 +1141,32 @@ nano::account nano::ledger::account (nano::block const & block) const
...
-	auto block_balance (balance (transaction_a, hash_a));
-	auto previous_balance (balance (transaction_a, block_l->previous ()));
-	return block_balance > previous_balance ? block_balance - previous_balance : previous_balance - block_balance;
-}
-
-nano::uint128_t nano::ledger::amount_safe (store::transaction const & transaction_a, nano::block_hash const & hash_a, bool & error_a) const
+std::optional<nano::uint128_t> nano::ledger::amount (store::transaction const & transaction_a, nano::block_hash const & hash_a)
 {
 	auto block_l = block (transaction_a, hash_a);
 	debug_assert (block_l);
```
pwojcikdev: Shouldn't we check if block is empty and return nullopt instead of keeping this debug assert?

nano/secure/ledger.cpp
clemahieu: Made this change. It was a bit more tricky because ledger::amount was depending on this returning 0 if it couldn't find the hash.  I added an extra commit to change ledger::amount so it no longer depends on this behaviour.

nano/node/bootstrap/bootstrap_server.cpp
```diff
@@ -286,7 +286,10 @@ nano::asc_pull_ack nano::bootstrap_server::process (const store::transaction & t
 		case asc_pull_req::hash_type::block:
 		{
 			// Try to lookup account assuming target is block hash
-			target = ledger.account_safe (transaction, request.target.as_block_hash ());
+			if (auto account = ledger.account (transaction, request.target.as_block_hash ()))
```
clemahieu: Missed one value_or

pwojcikdev (approved):

PR #4465, opened 2024-03-06
Improve block field functions by accessing sideband information internally

clemahieu:
This change moves the responsibility of determining if a field is in the block or in the block's sideband on to the block class.

Previously call sites would need to either inspect the block type or check for a sentinel value to determine where the block field information was stored. This complicates call sites, duplicates code, and could be error prone.

Direct block field access e.g. account/link/destination/balance are suffixed with _field and return an std::optional depending if the block type contains the requested field.

The replacement field functions check block type and determine if the value should be retrieved from the field or the block sideband.

The conversion for each field is done with two commits in order to ensure all call sites are inspected and updated to the new semantics.
- The first commit changes the return type of of the block field which allowed easy locating / updating of all the call sites through compiler type checking.
- The second commit renames the direct field access function and adds a helper function to pull the field value or the sideband value depending on block type.


nano/node/bootstrap_ascending/service.cpp
```diff
@@ -166,7 +166,7 @@ void nano::bootstrap_ascending::service::inspect (store::transaction const & tx,
 		break;
 		case nano::block_status::gap_source:
 		{
-			const auto account = block.previous ().is_zero () ? block.account ().value () : ledger.account (tx, block.previous ()).value ();
+			const auto account = block.previous ().is_zero () ? block.account_field ().value () : ledger.account (tx, block.previous ()).value ();
```
pwojcikdev, line 153: This line looks a bit suspicious, it seems to rely on the fact that if both source and previous block is missing, then ledger processing code will return block_status::gap_previous. Otherwise it would crash. No need to do anything, just something that might bite later down the road, with subsequent refactorings.

clemahieu: Yea it relies on the fact that if previous.is_zero (), then it must be an open or state block and therefore the account field is present.  Since the block failed to be inserted it won't have sideband at this point.

pwojcikdev (dismissed):

nano/lib/blocks.cpp
```diff
@@ -161,45 +208,118 @@ bool nano::block::has_sideband () const
...
+	release_assert (has_sideband ());
+	switch (type ())
+	{
+		case nano::block_type::send:
+			return destination_field ().value ();
+		case nano::block_type::state:
+			release_assert (sideband ().details.is_send);
+			return link_field ().value ().as_account ();
+		default:
+			release_assert (false);
```
pwojcikdev, line 276: There is some inconsistency in the way unexpected behavior is handled, `.value ()` will throw an exception while `release_assert` will terminate the program. Maybe those accessors should be marked with `noexcept` to terminate instantly if they find the program in undefined state.

clemahieu: Yea I can make that change. The intent with the value () calls is to terminate, at least in these inner functions.

nano/lib/blocks.hpp
```diff
@@ -63,6 +52,35 @@ class block
...
+	nano::amount balance () const;
+	// Balance field for open/send/state blocks
+	virtual std::optional<nano::amount> balance_field () const;
+	// Returns the destination account for send/state blocks that are sends
+	nano::account destination () const;
+	// Destination account for send blocks
+	virtual std::optional<nano::account> destination_field () const;
+	// Link field for state blocks
+	virtual std::optional<nano::link> link_field () const;
+	// Previous block if field exists and is non-zero
```
clemahieu: Comment needs to be corrected.

nano/node/active_transactions.cpp
```diff
@@ -165,8 +165,10 @@ void nano::active_transactions::handle_confirmation (nano::store::read_transacti
 
 void nano::active_transactions::handle_block_confirmation (nano::store::read_transaction const & transaction, std::shared_ptr<nano::block> const & block, nano::block_hash const & hash, nano::account & account, nano::uint128_t & amount, bool & is_state_send, bool & is_state_epoch, nano::account & pending_account)
 {
-	auto destination = block->link ().is_zero () ? block->destination () : block->link ().as_account ();
-	node.receive_confirmed (transaction, hash, destination);
+	if (block->sideband ().details.is_send)
```
pwojcikdev: Could be using `block->is_send()` helper

nano/node/json_handler.cpp
```diff
@@ -3679,7 +3679,7 @@ void nano::json_handler::republish ()
 							while (block_d != nullptr && hash != source)
 							{
 								hashes.push_back (previous);
-								source = node.ledger.block_source (transaction, *block_d);
+								source = block_d->source_field ().value_or (block_d->is_send () ? 0 : block_d->link_field ().value_or (0).as_block_hash ());
```
pwojcikdev, line 3682: A short `node.ledger.block_source ( .. )` call changed into this relatively long conditional statement, is this intentional?

pwojcikdev (dismissed):

nano/node/json_handler.cpp
```diff
@@ -3641,7 +3641,7 @@ void nano::json_handler::republish ()
 				block = node.ledger.block (transaction, hash);
 				if (sources != 0) // Republish source chain
 				{
-					nano::block_hash source (node.ledger.block_source (transaction, *block));
+					nano::block_hash source = block->source_field ().value_or (block->link_field ().value_or (0).as_block_hash ());
```
pwojcikdev, line 3644: Another line that got more complicated after the refactor than before.

PR #4469, opened 2024-03-07
Multithreaded request aggregator

RickiNano:
Made by Colin. I just made it configurable in the config file


pwojcikdev:
> Defaults to half the number of CPU threads, and at least 2.

This seems a bit excessive, especially since this is an auxiliary mechanism. We've seen good results even with 2-4 threads allocated.

RickiNano:
I have changed it to use max 4 threads as default

dsiganos:
You also need to fix the test case "toml.daemon_config_deserialize_defaults" and "daemon_config_deserialize_no_defaults"

dsiganos:
Ideally, we want these tests to fail when we forget them, as they are, they not very good.

RickiNano:
> You also need to fix the test case "toml.daemon_config_deserialize_defaults" and "daemon_config_deserialize_no_defaults"

It's included now

nano/node/nodeconfig.cpp
```diff
@@ -129,6 +129,7 @@ nano::error nano::node_config::serialize_toml (nano::tomlconfig & toml) const
 	toml.put ("max_work_generate_multiplier", max_work_generate_multiplier, "Maximum allowed difficulty multiplier for work generation.\ntype:double,[1..]");
 	toml.put ("frontiers_confirmation", serialize_frontiers_confirmation (frontiers_confirmation), "Mode controlling frontier confirmation rate.\ntype:string,{auto,always,disabled}");
 	toml.put ("max_queued_requests", max_queued_requests, "Limit for number of queued confirmation requests for one channel, after which new requests are dropped until the queue drops below this value.\ntype:uint32");
+	toml.put ("request_aggregator_threads", request_aggregator_threads, "Number of threads dedicated to the request aggregator. Defaults to use 4 threads if available, otherwise uses all cpu threads available");
```
pwojcikdev: > Defaults to use 4 threads if available, otherwise uses all cpu threads available

This is a bit confusing, I'm not sure if I understand the description. But it's a detail.

pwojcikdev (dismissed):

nano/node/nodeconfig.cpp
dsiganos: Yes, it is confusing. How about this?

Number of threads to dedicate to request aggregator. The default value is the minimum of 4 or the number returned by nano::hardware_concurency(), which is the number of hardware threads or the value of the environment variable NANO_HARDWARE_CONCURRENCY.

dsiganos (dismissed):

nano/node/nodeconfig.cpp
pwojcikdev: From a perspective of non-dev user, this is even worse. A simple "Defaults to using at most 4 threads" would suffice, I see no reason to make it that complicated.

RickiNano: I agree with @pwojcikdev 
This is much worse.
How about "Defaults to 4 threads if hardware allows, or else all available threads"

pwojcikdev: Sounds good for me @RickiNano

dsiganos: I merged it now, so you need to create a new PR to change it

dsiganos: I do not agree. If we say it defaults to 4, that is technically wrong and I do not agree with having wrong explanations. In any case, I have been outvoted, so I leave out to you guys to create a new PR and change it, if you prefer.

pwojcikdev: @dsiganos This is user facing config, it should be clear if somewhat simplified. What you did is describe how code works, that anyone with dev skills can lookup in 30 seconds.

PR #4475, opened 2024-03-10
Reduce recompile time of ledger.hpp.

clemahieu:
Reduce ledger.hpp dependencies and remove ledger.hpp inclusion from other header files.

Previous recompile time: 131 files, 3:00 min
New recompile time: 74 files, 1:30 min


nano/secure/generate_cache.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+namespace nano
+{
+/* Holds flags for various cacheable data. For most CLI operations caching is unnecessary
+ * (e.g getting the cemented block count) so it can be disabled for performance reasons. */
+class generate_cache
```
pwojcikdev: This is an unfortunate name, `generate_cache_flags` would be more fitting and it's a bit of an overkill to have it in separate file, should probably be merged with config/flags related classes.

pwojcikdev (dismissed):

nano/secure/generate_cache.hpp
clemahieu: Agree, it's a bad name and will rename it as such.

It can't be merged with node_flags because it's in the "secure" library and if it wasn't in a separate file, nodeconfig.hpp would need to include ledger.hpp which would have a chain-effect.

PR #4477, opened 2024-03-11
Changed description of request_aggregator_threads

RickiNano:


nano/node/nodeconfig.cpp
```diff
@@ -130,7 +130,7 @@ nano::error nano::node_config::serialize_toml (nano::tomlconfig & toml) const
 	toml.put ("max_work_generate_multiplier", max_work_generate_multiplier, "Maximum allowed difficulty multiplier for work generation.\ntype:double,[1..]");
 	toml.put ("frontiers_confirmation", serialize_frontiers_confirmation (frontiers_confirmation), "Mode controlling frontier confirmation rate.\ntype:string,{auto,always,disabled}");
 	toml.put ("max_queued_requests", max_queued_requests, "Limit for number of queued confirmation requests for one channel, after which new requests are dropped until the queue drops below this value.\ntype:uint32");
-	toml.put ("request_aggregator_threads", request_aggregator_threads, "Number of threads to dedicate to request aggregator. The default value is the minimum of 4 or the number returned by nano::hardware_concurency(), which is the number of hardware threads or the value of the environment variable NANO_HARDWARE_CONCURRENCY.");
+	toml.put ("request_aggregator_threads", request_aggregator_threads, "Number of threads to dedicate to request aggregator. Defaults to 4 threads if hardware allows, or else all available threads.");
```
clemahieu: Should this be "Defaults to using a thread per core, up to a maximum of 4"

RickiNano: Most cpu's have more threads than cores so I think it's better to just refer to what hardware allows, or maybe what cpu allows.

RickiNano: Maybe "Defaults to using all cpu threads, up to a maximum of 4" ?

dsiganos (approved):

dsiganos (approved):

PR #4480, opened 2024-03-11
Remove pragma from pending_info.cpp

clemahieu:


dsiganos:
I saw that too earlier... how did that get in?

dsiganos (approved):

PR #4482, opened 2024-03-12
rep_weights part 1 - Add data store for representative weights

simpago:
This is the first part of multiple stacked and self-contained pull requests. The next part is #4483.

### Overall Goal

### What this PR does
This PR introduces the store class `rep_weight` for storing the rep weights in the database. The new store is not used yet. This is the basis for the next PR.

### DIFF
--- a/nano/node/make_store.cpp
+++ b/nano/node/make_store.cpp
@@ -1,3 +1,4 @@
+#include <nano/lib/logging.hpp>
 #include <nano/node/make_store.hpp>
 #include <nano/store/lmdb/lmdb.hpp>
 #include <nano/store/rocksdb/rocksdb.hpp>
--- a/nano/store/lmdb/lmdb.cpp
+++ b/nano/store/lmdb/lmdb.cpp
@@ -24,7 +24,8 @@ nano::store::lmdb::component::component (nano::logger & logger_a, std::filesyste
 		peer_store,
 		confirmation_height_store,
 		final_vote_store,
-		version_store
+		version_store,
+		rep_weight_store
 	},
 	// clang-format on
 	block_store{ *this },
@@ -37,6 +38,7 @@ nano::store::lmdb::component::component (nano::logger & logger_a, std::filesyste
 	confirmation_height_store{ *this },
 	final_vote_store{ *this },
 	version_store{ *this },
+	rep_weight_store{ *this },
 	logger{ logger_a },
 	env (error, path_a, nano::store::lmdb::env::options::make ().set_config (lmdb_config_a).set_use_no_mem_init (true)),
 	mdb_txn_tracker (logger_a, txn_tracking_config_a, block_processor_batch_max_time_a),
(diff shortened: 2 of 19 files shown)


nano/node/make_store.cpp
```diff
@@ -1,7 +1,15 @@
+#include "nano/lib/logging.hpp"
```
pwojcikdev: The convention so far was to always use `<>` for includes.

nano/node/make_store.cpp
```diff
@@ -1,7 +1,15 @@
...
+#include "nano/secure/utility.hpp"
+
 #include <nano/node/make_store.hpp>
 #include <nano/store/lmdb/lmdb.hpp>
 #include <nano/store/rocksdb/rocksdb.hpp>
 
+std::unique_ptr<nano::store::component> nano::make_store ()
+{
+	return nano::make_store (nano::default_logger (), nano::unique_path (), nano::dev::constants);
```
pwojcikdev: This should be in `nano::test` namespace (test_common folder iirc) as this should never be used by non-test code.

nano/store/lmdb/lmdb.cpp
```diff
@@ -204,6 +206,7 @@ void nano::store::lmdb::component::open_databases (bool & error_a, store::transa
 	pending_store.pending_handle = pending_store.pending_v0_handle;
 	error_a |= mdb_dbi_open (env.tx (transaction_a), "final_votes", flags, &final_vote_store.final_votes_handle) != 0;
 	error_a |= mdb_dbi_open (env.tx (transaction_a), "blocks", MDB_CREATE, &block_store.blocks_handle) != 0;
+	error_a |= mdb_dbi_open (env.tx (transaction_a), "rep_weights", flags, &rep_weight_store.rep_weights_handle) != 0;
```
pwojcikdev: Old version of node used the same "rep_weights" table. I'm not sure how reliable was the version upgrade code, but it might be worth considering using unique name to avoid potential conflicts if a node happens to run with a very old ledger.

nano/node/make_store.cpp
simpago: Fixed the include style

nano/node/make_store.cpp
simpago: Moved it to nano::test_common::make_store.cpp

nano/store/lmdb/lmdb.cpp
simpago: Good point. I've just checked the old code and luckily the table name was different back then. It was named "representation":
https://github.com/nanocurrency/nano-node/blob/3209fbec1897772e5615108f7691d34d03359a6e/nano/node/lmdb/lmdb.cpp#L136

pwojcikdev (dismissed):

simpago:
rebased on top of develop

dsiganos (dismissed):

nano/store/lmdb/lmdb.cpp
dsiganos: This line fails when trying to start anode with an existing database.
Maybe ignore the error until we add the db upgrade code?

simpago: I didn't think about this when I split the branch into multiple pull requests. Ok, I'm going to ignore the error and add the error check back in PR part 2.

simpago:
Removed error check when opening rep_weights table

dsiganos (approved):

PR #4483, opened 2024-03-12
rep_weights part 2 - Store representative weights in database

simpago:
This is the second part of multiple stacked and self-contained pull requests. The previous part is #4482. The next part is #4485.

### Overall Goal

### What this PR does
This PR inserts all representative weights into the database and also adjusts those weights whenever the rep_weights cache is updated. Still all representatives are in memory, but this will change in the next PR.


nano/store/lmdb/lmdb.cpp
```diff
@@ -242,6 +251,35 @@ void nano::store::lmdb::component::upgrade_v21_to_v22 (store::write_transaction
 	logger.info (nano::log::type::lmdb, "Upgrading database from v21 to v22 completed");
 }
 
+// Fill rep_weights table with all existing representatives and their vote weight
+void nano::store::lmdb::component::upgrade_v22_to_v23 (store::write_transaction const & transaction_a)
```
simpago, line 254: I was wondering if it is a problem if the write_transaction is long running?
Also should we indicate progress here? Like printing a dot after every x accounts?

pwojcikdev: Even a basic progress indicator would be great, from a user perspective it might not be clear if the process hanged or is progressing.

nano/store/lmdb/lmdb.cpp
```diff
@@ -242,6 +251,35 @@ void nano::store::lmdb::component::upgrade_v21_to_v22 (store::write_transaction
...
+		}
+		nano::uint128_t total{ 0 };
+		nano::store::lmdb::db_val value;
+		auto status = get (transaction_a, tables::rep_weights, i->second.representative, value);
+		if (success (status))
+		{
+			total = nano::amount{ value }.number ();
+		}
+		total += i->second.balance.number ();
+		status = put (transaction_a, tables::rep_weights, i->second.representative, nano::amount{ total });
```
simpago: cryptocode suggested to use MDB_APPEND here, which speeds inserts up significantly

nano/store/lmdb/lmdb.cpp
simpago: I've added a progress log statement for every 250,000 accounts

nano/secure/account_info.hpp
```diff
@@ -29,4 +29,24 @@ class account_info final
 	uint64_t block_count{ 0 };
 	nano::epoch epoch_m{ nano::epoch::epoch_0 };
 };
+
+/**
+ * Account info as of DB version 22.
+ * This class protects the DB upgrades from future changes of the account_info class.
```
clemahieu: It wasn't immediately clear to me the purpose of this class or how it protected DB upgrades.

Maybe "This is a snapshot of the account_info table at v22 which needs to be read for the v22 to v23 upgrade.

nano/core_test/ledger.cpp
```diff
@@ -657,15 +658,36 @@ TEST (ledger, open_fork)
...
+	auto store{ nano::test::make_store () };
+	nano::rep_weights rep_weights{ store->rep_weight };
+	auto txn{ store->tx_begin_write () };
+	rep_weights.representation_add (txn, 1, 100);
+	rep_weights.representation_add_dual (txn, 2, 100, 3, 100);
+	ASSERT_EQ (3, rep_weights.size ());
+	ASSERT_EQ (3, store->rep_weight.count (txn));
+
+	// set rep weights to 0
+	rep_weights.representation_add (txn, 1, std::numeric_limits<nano::uint128_t>::max () - 99);
```
clemahieu: I'd change these to 0 - 100 instead of ::max() - 99 so the 100s match up.

nano/secure/rep_weights.cpp
```diff
@@ -0,0 +1,139 @@
...
+	put_store (txn_a, rep_a, previous_weight, new_weight);
+	nano::lock_guard<nano::mutex> guard (mutex);
+	put_cache (rep_a, new_weight);
+}
+
+void nano::rep_weights::representation_add_dual (store::write_transaction const & txn_a, nano::account const & rep_1, nano::uint128_t const & amount_1, nano::account const & rep_2, nano::uint128_t const & amount_2)
+{
+	if (rep_1 != rep_2)
+	{
+		auto previous_weight_1{ rep_weight_store.get (txn_a, rep_1) };
```
clemahieu, line 23: Can this just be written as two calls to representation_add?

nano/secure/rep_weights.cpp
```diff
@@ -0,0 +1,139 @@
...
+/** Makes a copy */
+std::unordered_map<nano::account, nano::uint128_t> nano::rep_weights::get_rep_amounts () const
+{
+	nano::lock_guard<nano::mutex> guard (mutex);
+	return rep_amounts;
+}
+
+void nano::rep_weights::copy_from (nano::rep_weights & other_a)
+{
+	nano::lock_guard<nano::mutex> guard_this (mutex);
```
clemahieu, line 60: Is this really needed? I see the existing implementation loads the weights in to a local variable and then copies them; I don't see why it can't load them in-place without the copy.

nano/secure/rep_weights.cpp
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+std::size_t nano::rep_weights::size () const
+{
+	nano::lock_guard<nano::mutex> guard (mutex);
+	return rep_amounts.size ();
+}
+
+std::unique_ptr<nano::container_info_component> nano::collect_container_info (nano::rep_weights const & rep_weights, std::string const & name)
```
clemahieu: We've been changing these friend + free functions that collect weight to instead be member functions.

nano/secure/rep_weights.cpp
simpago: The issue here is with the lock. Two calls to representation add would cause two separate locks. We could introduce a helper function `representation_add_locked`, but then the LMDB table write would be inside the lock, which I think is not optimal, because the lock would be held longer.

nano/secure/rep_weights.cpp
simpago: This `copy_from` is called during `ledger::initialize()` from multiple threads. When a thread is finished with its rep_weights workload, it copies the result into the final rep_weights instance. This happens in parallel with the other threads. If it was loaded in-place, then we would need to lock the rep_weights instance for every insert.

nano/secure/account_info.hpp
simpago: Changed the comment

nano/core_test/ledger.cpp
simpago: Changed it

nano/secure/rep_weights.cpp
simpago: Converted it into a member function

PR #4484, opened 2024-03-12
Simplify the nano::ledger::successor function to make its implementation more obvious.

clemahieu:
These changes simplifies the way block successors are accessed.

Usages of store::block::successor are changed to ledger::successor.
ledger::successor returns a std::optional<nano::block_hash> instead of a std::shared_ptr<nano::block> which allows users to determine if they want to load the block.
A convenience overload of ledger::successor is added which takes a block_hash and expands it to a qualified_root.

### DIFF
--- a/nano/secure/ledger.hpp
+++ b/nano/secure/ledger.hpp
@@ -52,6 +52,7 @@ public:
 	nano::uint128_t account_receivable (store::transaction const &, nano::account const &, bool = false);
 	nano::uint128_t weight (nano::account const &);
 	std::optional<nano::block_hash> successor (store::transaction const &, nano::qualified_root const &) const noexcept;
+	std::optional<nano::block_hash> successor (store::transaction const & transaction, nano::block_hash const & hash) const noexcept;
 	std::shared_ptr<nano::block> forked_block (store::transaction const &, nano::block const &);
 	std::shared_ptr<nano::block> head_block (store::transaction const &, nano::account const &);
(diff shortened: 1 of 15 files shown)


nano/secure/ledger.hpp
```diff
@@ -51,7 +51,8 @@ class ledger final
 	nano::uint128_t account_balance (store::transaction const &, nano::account const &, bool = false);
 	nano::uint128_t account_receivable (store::transaction const &, nano::account const &, bool = false);
 	nano::uint128_t weight (nano::account const &);
-	std::shared_ptr<nano::block> successor (store::transaction const &, nano::qualified_root const &);
+	std::optional<nano::block_hash> successor (store::transaction const &, nano::qualified_root const &) const noexcept;
```
pwojcikdev, line 54: Depending on whether qualified root represents a block ( root and previous fields are non zero) or an account ( root is set account and previous is zero) this returns either the next block in chain or account. Is this accurate? How exactly the situation looks like with epoch or epoch-open blocks?

pwojcikdev (dismissed):

nano/secure/ledger.hpp
clemahieu: There's an ambiguity there that appeared when epoch blocks were created. Roots are only used for block PoW and vote_req hinting. The successor function gives precedence to the block hash rather than the account number. The net result is that vote_req root hints don't work for epoch-open blocks.

PR #4485, opened 2024-03-12
rep_weights part 3 - Add mininum weight to representative cache

simpago:
This is the last part of multiple stacked and self-contained pull requests. The previous part is #4483.

### Overall Goal

### What this PR does
This PR introduces a minimum weight for the representatives cache (`rep_weights`). All votes from representatives below that minimum will be ignored. The minimum weight can be configured via the new option `representative_vote_weight_minimum` in the file `config-node.toml`


nano/node/nodeconfig.hpp
```diff
@@ -64,6 +64,11 @@ class node_config
 	std::chrono::milliseconds vote_generator_delay{ std::chrono::milliseconds (100) };
 	unsigned vote_generator_threshold{ 3 };
 	nano::amount online_weight_minimum{ 60000 * nano::Gxrb_ratio };
+	/*
+	 * The minimum vote weight that a representative must have for its vote to be counted.
+	 * All representatives above this weight will be kept in memory!
+	 */
+	nano::amount representative_vote_weight_minimum{ 10 * nano::Gxrb_ratio };
```
simpago: I picked 10 nanos as the default mininum vote weight. Is this a good value? It results in less than 200 reps with the current ledger.

gr0vity-dev: `10 * nano::Gxrb_ratio` is 10000 Nano. Was that intentional ?

Here is the [definition](https://github.com/nanocurrency/nano-node/blob/ef985e3cf1dcd026ba8a6df1e6f3f685934e1be9/nano/lib/numbers.hpp#L15) of that value

I tried setting it to 10 Nano instead which results in about 1000 reps on the live network.

simpago: Good catch, that was a mistake. I've just pushed a fix.

nano/core_test/ledger.cpp
```diff
@@ -688,6 +690,34 @@ TEST (ledger, delete_rep_weight_of_zero)
...
+	ASSERT_EQ (1, rep_weights.size ());
+	ASSERT_EQ (1, store->rep_weight.count (txn));
+
+	// above min weight
+	rep_weights.representation_add (txn, 1, 1);
+	ASSERT_EQ (1, rep_weights.size ());
+	ASSERT_EQ (1, store->rep_weight.count (txn));
+
+	// fall blow min weight
+	rep_weights.representation_add (txn, 1, std::numeric_limits<nano::uint128_t>::max () - 5);
```
clemahieu: I think the "0 - x" style is more clear since it's exactly like subtracting x.

nano/secure/ledger.hpp
```diff
@@ -53,9 +53,16 @@ class ledger final
...
+	/**
+	 * Returns the cached vote weight for the given representative.
+	 * If the weight is below the cache limit it returns 0.
+	 * During bootstrap it returns the preconfigured bootstrap weights.
+	 */
 	nano::uint128_t weight (nano::account const &);
 	std::optional<nano::block_hash> successor (store::transaction const &, nano::qualified_root const &) const noexcept;
 	std::optional<nano::block_hash> successor (store::transaction const & transaction, nano::block_hash const & hash) const noexcept;
+	/* Returns the exact vote weight for the given representative by doing a database lookup */
+	nano::uint128_t weight_exact (store::transaction const &, nano::account const &);
```
clemahieu, line 65: On the naming of "weight_exact" it seems like the other weight function isn't exact though they should always be in sync shouldn't they?

simpago: Let's say a representative has a vote weight of Ӿ0.1. `weight_exact` would return Ӿ0.1, but `weight` would return 0, because it wasn't in the cache. I wanted to rename `weight` to `weight_cached`, but this would pollute this pull request with the renames. Or should I rename it now? Do you have better names in mind?
Maybe let's rename `weight`=>`weight_cached` and `weight_exact` => `weight`?

nano/core_test/ledger.cpp
simpago: Fixed it

nano/secure/ledger.hpp
pwojcikdev: Renaming is going to impact work in progress code, the current "non-exact" weight function is what node components should use 99% of the time.

PR #4486, opened 2024-03-12
Introduce ledger confirmed_set/any_set classes to separate confirmed-only and confirmed/unconfirmed operations on the ledger

clemahieu:
The motivation for this PR is in preparation for bounded backlog changes. Currently, both confirmed and unconfirmed transactions are together on disk and the ACID properties of the ledger are encapsulated with a database transaction.

With the bounded backlog, blocks will be stored in memory so there will need to be memory synchronization in addition to database synchronization. Rather than filling the ledger class with unconfirmed/confirmed variants for each operation.

The ledger is divided in to several sets and subsets:
ledger::confirmed - The set of confirmed transactions
ledger::any - The superset of confirmed and unconfirmed transactions

When unconfirmed blocks are held in memory the set ledger::unconfirmed will be added which holds only unconfirmed transactions.

In the future the confirming_set can be moved on to ledger which is the subset of unconfirmed blocks that are queued to be moved in to the confirmed set.


clemahieu:
Changed base file class name to ledger_view_base and renamed the user-facing class to ledger_view.

nano/node/bootstrap_ascending/service.cpp
```diff
@@ -150,7 +151,7 @@ void nano::bootstrap_ascending::service::inspect (store::transaction const & tx,
 		break;
 		case nano::block_status::gap_source:
 		{
-			const auto account = block.previous ().is_zero () ? block.account_field ().value () : ledger.account (tx, block.previous ()).value ();
+			const auto account = block.previous ().is_zero () ? block.account_field ().value () : view.account (block.previous ()).value ();
```
pwojcikdev: Ledger view should be passed instead of database transaction, otherwise data races are possible. Assuming bounded backlog is implemented, with current code it's possible that unconfirmed blocks will be dropped in between calls to inspect (). Another aspect is that creating a new view for each call is very suboptimal.

pwojcikdev: To clarify, this is not an issue right now thanks to guarantees provided by database transactions, but will be once unconfirmed set is no longer stored on disk.

clemahieu:
Rebased on top of https://github.com/nanocurrency/nano-node/pull/4484

clemahieu:
I reworked this from the previous ledger_view design for several reasons:

- The confirmed/unconfirmed views were almost never interchangeable so using inheritance wasn't really appropriate.
- Capturing the transaction in the view was never really sufficient so both transaction/view had to be passed around in cases
- Creating the views littered the code with view variables right next to transaction variables
- Creating the view had an allocation overhead and many of its uses were short-lived
- The overwhelming majority of accesses are looking for an unconfirmed view so terseness / syntax sugar is desirable.
- Fewer operations are needed on the confirmed view than the unconfirmed view which tended to break the inheritence abstraction.

nano/node/node.cpp
```diff
@@ -1182,12 +1189,12 @@ void nano::node::start_election (std::shared_ptr<nano::block> const & block)
...
 {
 	auto transaction (store.tx_begin_read ());
-	return ledger.block_confirmed (transaction, hash_a);
+	return ledger.confirmed.exists_or_pruned (transaction, hash_a);
 }
 
 bool nano::node::block_confirmed_or_being_confirmed (nano::store::transaction const & transaction, nano::block_hash const & hash_a)
 {
-	return confirming_set.exists (hash_a) || ledger.block_confirmed (transaction, hash_a);
+	return confirming_set.exists (hash_a) || ledger.confirmed.exists_or_pruned (transaction, hash_a);
```
pwojcikdev: With the way it's implemented now, a big warning should be put there. In LMDB write transactions do not block readers, nor do readers block writers. So it's possible that a read transaction is started with block being unconfirmed, then confirming set starts confirming and successfully processes and flushes it (`confirming_set.exists (...)` returns false) and a subsequent check `ledger.confirmed.exists_or_pruned ()` still returns false because read transaction was started before the block was confirmed.

pwojcikdev: Perhaps this could accept a write transaction (LMDB guarantees that only a single writer is active), however this could introduce a significant performance degradation, depending on how it's used.

nano/secure/receivable_iterator_impl.hpp
```diff
@@ -0,0 +1,56 @@
...
+template <typename T>
+bool nano::receivable_iterator<T>::operator== (receivable_iterator const & other) const
+{
+	debug_assert (view == nullptr || other.view == nullptr || view == other.view);
+	debug_assert (account.is_zero () || other.account.is_zero () || account == other.account);
+	return item == other.item;
+}
+
+template <typename T>
+bool nano::receivable_iterator<T>::operator!= (receivable_iterator const & other) const
```
pwojcikdev: This should be auto generated since C++ 20.

nano/secure/ledger_set_any.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+	ledger_set_any (nano::ledger const & ledger);
+
+	std::optional<nano::account> account (store::transaction const & transaction, nano::block_hash const & hash) const;
+	account_iterator account_begin (store::transaction const & transaction) const;
+	account_iterator account_end () const;
+	// Returns the next receivable entry equal or greater than 'key'
+	account_iterator account_lower_bound (store::transaction const & transaction, nano::account const & account) const;
+	// Returns the next receivable entry for an account greater than 'account'
+	account_iterator account_upper_bound (store::transaction const & transaction, nano::account const & account) const;
```
pwojcikdev: There is a typo in all comments next to account_* functions, those are not receivable entries

nano/secure/ledger_set_confirmed.cpp
```diff
@@ -0,0 +1,181 @@
...
+		return std::nullopt;
+	}
+	return *result;
+}
+
+std::optional<nano::block_hash> nano::ledger_set_confirmed::successor (store::transaction const & transaction, nano::qualified_root const & root) const
+{
+	if (root.previous ().is_zero ())
+	{
+		return ledger.store.block.successor (transaction, root.previous ());
```
pwojcikdev: Is this supposed to be like this? `root.previous ()` is zero here. Seems like code duplication is leading to subtle bugs.

nano/secure/ledger_set_confirmed.cpp
```diff
@@ -0,0 +1,181 @@
...
+	auto block = get (transaction, head (transaction, account_a));
+	if (!block)
+	{
+		return std::nullopt;
+	}
+	return block->balance ().number ();
+}
+
+// Balance for account containing hash
+std::optional<nano::uint128_t> nano::ledger_set_confirmed::balance (store::transaction const & transaction, nano::block_hash const & hash) const
```
pwojcikdev: This seems to return balance for a specific block, not the latest value for account. Is this intentional?

nano/secure/receivable_iterator_impl.hpp
clemahieu: Good to know. Removed and allowed defaults.

nano/secure/ledger_set_confirmed.cpp
clemahieu: This function isn't actually used anywhere in the code; it looks left over from when this PR used polymorphism where this implementation was created but never used or tested.  Since it's not used I deleted it rather than fixing / testing.

nano/secure/ledger_set_confirmed.cpp
clemahieu: I removed the comment since the intent is to return the balance of the associated block by hash.

With the new block query functions we could potentially remove these convenience functions head/balance/height for block_hash and call sites can pull the value directly from the block.

nano/node/node.cpp
clemahieu: This is a good point, I didn't see that race condition before.  Since this isn't new behavior I opened a separate PR for adding the comment. https://github.com/nanocurrency/nano-node/pull/4540

nano/secure/ledger_set_confirmed.cpp
dsiganos: I think it is worth renaming it to `block_balance` to emphasise the fact that it returns the balance at that block and behaves differently to the function definition (one line above) with the same name.

nano/secure/account_iterator.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
```
dsiganos, line 1: This file could do with some comments.
How is the end of a sequence detected?
What does the second constructor do?
What is a view expected to be?
A quick example would be nice

nano/secure/ledger_set_any.hpp
```diff
@@ -0,0 +1,66 @@
...
+	ledger_set_any (nano::ledger const & ledger);
+
+	std::optional<nano::account> account (store::transaction const & transaction, nano::block_hash const & hash) const;
+	account_iterator account_begin (store::transaction const & transaction) const;
+	account_iterator account_end () const;
+	// Returns the next account entry equal or greater than 'account'
+	account_iterator account_lower_bound (store::transaction const & transaction, nano::account const & account) const;
+	// Returns the next account entry greater than 'account'
+	account_iterator account_upper_bound (store::transaction const & transaction, nano::account const & account) const;
+	std::optional<nano::uint128_t> amount (store::transaction const & transaction, nano::block_hash const & hash) const;
```
dsiganos: Should these functions return `nano::amount` type?

dsiganos: It is not clear what guarantees this function gives.
For example, in what scenario will a previous block not have a balance and hence an amount will not be returned?
image

nano/secure/ledger_set_confirmed.cpp
```diff
@@ -0,0 +1,161 @@
...
+	return block->sideband ().height <= info.value ().height;
+}
+
+bool nano::ledger_set_confirmed::exists_or_pruned (store::transaction const & transaction, nano::block_hash const & hash) const
+{
+	if (ledger.store.pruned.exists (transaction, hash))
+	{
+		return true;
+	}
+	auto block = ledger.store.block.get (transaction, hash);
```
dsiganos: This could call `nano::ledger_set_confirmed::exists()` and avoid code repetition.

nano/secure/ledger_set_confirmed.cpp
```diff
@@ -0,0 +1,161 @@
...
+}
+
+std::optional<nano::account_info> nano::ledger_set_confirmed::get (store::transaction const & transaction, nano::account const & account) const
+{
+	auto info = ledger.store.confirmation_height.get (transaction, account);
+	if (!info)
+	{
+		return std::nullopt;
+	}
+	debug_assert (false);
```
dsiganos: This doesn't look right. And the fact that no test fails with this assert means that we are missing an important test case or we have a bug (or it is one of the test we have disabled and still have not re-enabled).

nano/secure/ledger_set_confirmed.cpp
```diff
@@ -0,0 +1,161 @@
...
+	}
+	auto block = get (transaction, hash);
+	if (!block)
+	{
+		return std::nullopt;
+	}
+	return block->balance ().number ();
+}
+
+bool nano::ledger_set_confirmed::exists (store::transaction const & transaction, nano::block_hash const & hash) const
```
dsiganos: Code repetition. This can be implemented by calling `nano::ledger_set_confirmed::get`.

nano/secure/ledger_set_confirmed.cpp
```diff
@@ -0,0 +1,161 @@
...
+	{
+		return nullptr;
+	}
+	return block->sideband ().height <= info.value ().height ? block : nullptr;
+}
+
+std::optional<nano::pending_info> nano::ledger_set_confirmed::get (store::transaction const & transaction, nano::pending_key const & key) const
+{
+	auto result = ledger.store.pending.get (transaction, key);
+	if (!result && !exists_or_pruned (transaction, key.hash))
```
dsiganos: Is this expression right? it looks wrong to me. I find it very difficult to make sense of it. Splitting into 2 if statements would improve readability. It seems to always return pending entry if it exists.

nano/secure/ledger_set_confirmed.cpp
```diff
@@ -0,0 +1,161 @@
...
+	{
+		return std::nullopt;
+	}
+	return result;
+}
+
+nano::block_hash nano::ledger_set_confirmed::head (store::transaction const & transaction, nano::account const & account) const
```
dsiganos: Should this return std::optional for consistency?

nano/secure/ledger_set_confirmed.hpp
```diff
@@ -0,0 +1,49 @@
...
+	std::optional<nano::uint128_t> balance (store::transaction const & transaction, nano::block_hash const & hash) const;
+	bool exists (store::transaction const & transaction, nano::block_hash const & hash) const;
+	bool exists_or_pruned (store::transaction const & transaction, nano::block_hash const & hash) const;
+	std::optional<nano::account_info> get (store::transaction const & transaction, nano::account const & account) const;
+	std::shared_ptr<nano::block> get (store::transaction const & transaction, nano::block_hash const & hash) const;
+	std::optional<nano::pending_info> get (store::transaction const & transaction, nano::pending_key const & key) const;
+	nano::block_hash head (store::transaction const & transaction, nano::account const & account) const;
+	uint64_t height (store::transaction const & transaction, nano::account const & account) const;
+	receivable_iterator receivable_end () const;
+	// Returns the next receivable entry equal or greater than 'key'
```
dsiganos, line 42: This comment seems incorrect or incomplete.

nano/secure/ledger_set_confirmed.hpp
```diff
@@ -0,0 +1,49 @@
...
+	std::optional<std::pair<nano::pending_key, nano::pending_info>> receivable_lower_bound (store::transaction const & transaction, nano::account const & account, nano::block_hash const & hash) const;
+	// Returns the next receivable entry for an account greater than 'account'
+	receivable_iterator receivable_upper_bound (store::transaction const & transaction, nano::account const & account) const;
```
dsiganos: The receivable upper/lower bound functions are very confusing. More verbose comments are needed and possibly examples too.

nano/secure/ledger_set_any.hpp
```diff
@@ -0,0 +1,66 @@
...
+	bool exists (store::transaction const & transaction, nano::block_hash const & hash) const;
+	bool exists_or_pruned (store::transaction const & transaction, nano::block_hash const & hash) const;
+	std::optional<nano::account_info> get (store::transaction const & transaction, nano::account const & account) const;
+	std::shared_ptr<nano::block> get (store::transaction const & transaction, nano::block_hash const & hash) const;
+	std::optional<nano::pending_info> get (store::transaction const & transaction, nano::pending_key const & key) const;
+	nano::block_hash head (store::transaction const & transaction, nano::account const & account) const;
+	uint64_t height (store::transaction const & transaction, nano::account const & account) const;
+	uint64_t height (store::transaction const & transaction, nano::block_hash const & hash) const;
+	// Returns whether there are any receivable entries for 'account'
+	bool receivable_any (store::transaction const & transaction, nano::account const & account) const;
```
dsiganos: The naming of this is not ideal given that the struct is also called 'any'.
Maybe `receivables_exist`?

I wonder why there is no equivalent function in the confirmed set.

nano/secure/ledger_set_confirmed.hpp
dsiganos: I think this explanation should be in the comments.
I initially assumed that the comment was wrong and I ended up down a rabbit hole and wasted half an hour.
image

dsiganos:
General feedback:
* I personally would prefer the 3 get functions of the ledger sets to be more explicit about what they return to make code reading easier (get_account_info, get_block, etc).
* The passing of store::transaction objects around seems a little wrong in the sense that it should be an implementation detail hidden to the user. I think it would be better to introduce a ledger context object which holds the store::transaction inside it.
* There is a lack of comments. I think most member functions should be commented even if the behaviour appears obvious. It is always good to document the intention in English and what is obvious to the writer of the code is maybe not obvious to someone else.
* There is some obvious code repetition that can be eliminated.

nano/secure/ledger_set_confirmed.cpp
dsiganos: Fixed in:
https://github.com/nanocurrency/nano-node/pull/4486/commits/4c02ce53d17b0dca1a46b7fa3d945bfdf4018b59

nano/node/bootstrap_ascending/service.cpp
clemahieu: To address the allocations concern, these changes were redesigned.

nano/secure/ledger_set_confirmed.cpp
clemahieu: I experimented with it but it required some more extensive rewriting at call sites that numerically depend on the 0 sentinel value.

We can raise an issue.

nano/secure/ledger_set_confirmed.cpp
clemahieu: Fixed

nano/secure/ledger_set_any.hpp
clemahieu: Renamed

nano/secure/ledger_set_confirmed.hpp
clemahieu: Added

nano/secure/account_iterator.hpp
clemahieu: Added

nano/secure/ledger_set_confirmed.cpp
clemahieu: I reorganised and renamed the functions, I agree this is much clearer now.

nano/secure/ledger_set_confirmed.cpp
clemahieu: Fixed

nano/secure/ledger_set_confirmed.cpp
clemahieu: It wasn't but the code also wasn't used so I removed it.

nano/secure/ledger_set_any.hpp
clemahieu: Fixed

clemahieu: I agree these are confusing, they're to deal with block pruning.  I tried to simplify this but the only way I could fix the issue would be to store the amount as sideband which might be a good idea but requires a block table upgrade.

This matches existing behavior.

nano/secure/ledger_set_any.cpp
```diff
@@ -0,0 +1,213 @@
...
+	}
+	auto previous_balance = this->block_balance (transaction, block_l->previous ());
+	if (!previous_balance)
+	{
+		return std::nullopt;
+	}
+	return block_balance > previous_balance.value () ? block_balance.number () - previous_balance.value ().number () : previous_balance.value ().number () - block_balance.number ();
+}
+
+// Balance for account containing hash
```
dsiganos, line 105: This comment seems wrong, it can be removed since the name of the function is now obvious.

nano/secure/ledger_set_any.cpp
```diff
@@ -0,0 +1,213 @@
...
+#include <nano/store/pruned.hpp>
+
+nano::ledger_set_any::ledger_set_any (nano::ledger const & ledger) :
+	ledger{ ledger }
+{
+}
+
+std::optional<nano::amount> nano::ledger_set_any::account_balance (secure::transaction const & transaction, nano::account const & account_a) const
+{
+	auto block = block_get (transaction, ledger.any.account_head (transaction, account_a));
```
dsiganos, line 15: `ledger.any` seems redundant

nano/secure/ledger_set_any.hpp
```diff
@@ -0,0 +1,79 @@
...
+	bool block_exists (secure::transaction const & transaction, nano::block_hash const & hash) const;
+	bool block_exists_or_pruned (secure::transaction const & transaction, nano::block_hash const & hash) const;
+	std::shared_ptr<nano::block> block_get (secure::transaction const & transaction, nano::block_hash const & hash) const;
+	uint64_t block_height (secure::transaction const & transaction, nano::block_hash const & hash) const;
+	std::optional<nano::block_hash> block_successor (secure::transaction const & transaction, nano::block_hash const & hash) const;
+	std::optional<nano::block_hash> block_successor (secure::transaction const & transaction, nano::qualified_root const & root) const;
+
+public: // Operations on pending entries
+	std::optional<nano::pending_info> pending_get (secure::transaction const & transaction, nano::pending_key const & key) const;
+	bool receivable_any (secure::transaction const & transaction, nano::account const & account) const;
```
dsiganos, line 61: This should be deleted

PR #4492, opened 2024-03-15
Update flatbuffers to v24.3.7

clemahieu:


ci/build.sh
```diff
@@ -93,11 +93,11 @@ parallel_build_flag() {
             echo "-- -m"
             ;;
         *)
-            echo "--parallel $(number_of_processors)"
```
dsiganos, line 96: What is the reason for removing parallel building?

clemahieu:
This was a test to see if updating flatc or disabling parallelism would fix the GH builders.

Closing for now.

PR #4496, opened 2024-03-16
Adding abstraction of iteration over receivable entries

clemahieu:
This series of changes is aimed at decoupling queriers of receivable entries from the implementation of how they're stored. Queries of pending entries fall into three major categories

**Ledger processing**
This is an operation that checks the correctness of source blocks for blocks that receive a balance. If each block that sends a balance inserts an entry into an O(1) unordered_set (destination, block_hash), every receive block can be checked in O(1) time by using the (account, source) fields of blocks that receive. The existence of that value indicates the block is receivable.

**Iterating blocks that can be received by an account.**
This is done by a lot of wallet operations and RPCs related to wallet operations that want to check particular items available to be received for a particular account. A wallet would do this operation when looking for items it could process if wasn't signaled to that event through another means, e.g. websockets.

These operations are currently implemented with sorted queries over a single account searching for values. It might be possible to implement this without a sorted query.

**Iterating accounts that have items to receive**
The "unopened" rpc operates this way, iterating up to a certain number of accounts and summing the amount that can be received.

**Iterating all items**
Epoch upgrades, diagnostic functions

# There are 3 functions that iterate over receivable entries:
ledger.receivable_upper_bound (tx, account) - This will search for the first receivable item for an account strictly greater than "account".
ledger.receivable_upper_bound (tx, account, hash) - This will search for the next receivable item for a particular account greater than "hash". This will return nullopt if iteration goes past items in this account.

Upper_bound is used for the convenience of being able to feed in the current value of an iteration without modification and still iterate entries.

ledger.receivable_lower_bound (tx, account, hash) - This search finds the first item equal to or greater than the items passed in. This is the base function used by upper_bound functions.

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -1041,42 +1041,41 @@ void nano::json_handler::accounts_receivable ()
 	bool const sorting = request.get<bool> ("sorting", false);
 	auto simple (threshold.is_zero () && !source && !sorting); // if simple, response is a list of hashes for each account
 	boost::property_tree::ptree pending;
-	auto transaction (node.store.tx_begin_read ());
+	auto transaction = node.store.tx_begin_read ();
 	for (auto & accounts : request.get_child ("accounts"))
 	{
 		auto account (account_impl (accounts.second.data ()));
 		if (!ec)
 		{
 			boost::property_tree::ptree peers_l;
-			for (auto i (node.store.pending.begin (transaction, nano::pending_key (account, 0))), n (node.store.pending.end ()); i != n && nano::pending_key (i->first).account == account && peers_l.size () < count; ++i)
+			for (auto current = node.ledger.receivable_upper_bound (transaction, account, 0), end = node.ledger.receivable_end (); current != end; ++current)
 			{
-				nano::pending_key const & key (i->first);
-				if (block_confirmed (node, transaction, key.hash, include_active, include_only_confirmed))
+				auto const & [key, info] = *current;
+				if (include_only_confirmed && !node.ledger.block_confirmed (transaction, key.hash))
 				{
-					if (simple)
-					{
-						boost::property_tree::ptree entry;
-						entry.put ("", key.hash.to_string ());
-						peers_l.push_back (std::make_pair ("", entry));
-					}
-					else
-					{
-						nano::pending_info const & info (i->second);
-						if (info.amount.number () >= threshold.number ())
-						{
-							if (source)
-							{
-								boost::property_tree::ptree pending_tree;
-								pending_tree.put ("amount", info.amount.number ().convert_to<std::string> ());
-								pending_tree.put ("source", info.source.to_account ());
-								peers_l.add_child (key.hash.to_string (), pending_tree);
-							}
-							else
-							{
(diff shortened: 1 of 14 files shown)


nano/test_common/system.cpp
```diff
@@ -416,11 +416,10 @@ void nano::test::system::generate_receive (nano::node & node_a)
 		auto transaction (node_a.store.tx_begin_read ());
 		nano::account random_account;
 		random_pool::generate_block (random_account.bytes.data (), sizeof (random_account.bytes));
-		auto i (node_a.store.pending.begin (transaction, nano::pending_key (random_account, 0)));
-		if (i != node_a.store.pending.end ())
+		auto item = node_a.ledger.receivable_upper_bound (transaction, random_account);
+		if (item)
```
pwojcikdev: > Returns the next receivable entry for an account greater than 'account'

Is this condition checking the same thing as before?

nano/nano_node/entry.cpp
```diff
@@ -1735,8 +1735,9 @@ int main (int argc, char * const * argv)
 
 			start_threads (check_pending, pending);
 
+			auto & ledger = node->ledger;
 			size_t const pending_deque_overflow (64 * 1024);
-			for (auto i (node->store.pending.begin (transaction)), n (node->store.pending.end ()); i != n; ++i)
+			for (auto i = ledger.receivable_upper_bound (transaction, 0, 0); i; i = ledger.receivable_upper_bound (transaction, i.value ().first.account, i.value ().first.hash))
```
pwojcikdev: I see `ledger.receivable_upper_bound (transaction, i.value ().first.account, i.value ().first.hash))` pattern is used in quite a few places, it would be more readable and ergonomic to provide an override for upper_bound (..., <account, hash> pair)

nano/node/json_handler.cpp
```diff
@@ -1041,42 +1041,41 @@ void nano::json_handler::accounts_receivable ()
...
-								pending_tree.put ("amount", info.amount.number ().convert_to<std::string> ());
-								pending_tree.put ("source", info.source.to_account ());
-								peers_l.add_child (key.hash.to_string (), pending_tree);
-							}
-							else
-							{
-								peers_l.put (key.hash.to_string (), info.amount.number ().convert_to<std::string> ());
-							}
-						}
-					}
```
pwojcikdev, line 1079: An if/else decision tree is often easier to read than sequential code with continue statements.

pwojcikdev (dismissed):

nano/test_common/system.cpp
clemahieu: It should be. Since random_accout is a random number i.e. it's not an actual account that has a receivable entry, this is finding the next-biggest account that has a receivable entry.

nano/nano_node/entry.cpp
clemahieu: Agreed. With the rewritten iterator classes there are now done with a normal ++i.

clemahieu:
Rewrote these functions to return an iterator object to match standard c++ iteration.

pwojcikdev (approved):

PR #4502, opened 2024-03-19
Remove final vote canary code as final votes are now always enabled.

clemahieu:


dsiganos:
Will this interfere with our test setups? I think they issue canary blocks.
@gr0vity-dev

dsiganos:
The following variables can also be removed:
* live_canary_public_key_data
* beta_canary_public_key_data
* test_canary_public_key_data

pwojcikdev (dismissed):

gr0vity-dev:
I wondered the same, however I like seeing old code removed from the codebase.
The local test networks use pre made ledger mostly. In case we need to create a new ledger we can always initialise the network with an older version and upgrade the db.

What about resetting the beta network ? Will that still work, since we used to issue a canary block in that event.

clemahieu:
It should have no effect, the blocks should simply be treated like normal.

Though it would be good to test this.

gr0vity-dev:
I tested the changes in a local network :
- **Without a canary** block in the ledger
  - With **this PR** an election is only confirmed **after 2 rounds of voting**
  - With  **V26.1** an election is confirmed **after 1 round of voting**
- **With a canary** in the ledger (for the sake of completeness)
  - **both versions** require **2 rounds of voting**

Screenshot 2024-03-19 at 12 24 38

PR #4508, opened 2024-03-20
Block cemented callback simplify

clemahieu:
Simplifies and merges logic that was spread across multiple functions and coupled with nano::election.

Behaviour changed with respect to callbacks that were *not* called for confirmed blocks in certain circumstances. This was a purely synthetic case where an election was explicitly confirmed in code but didn't have an associated election. In other cases where there was no election yet the block was confirmed indirectly, the callbacks were still called. This behaviour change calls the callback for all blocks that are confirmed.

Behaviour changed with respect to entries in the recently_cemented list. Previously blocks implicitly confirmed were not placed in this list yet were reported through callbacks. This behaviour was arbitrary and could be confusing. Now all blocks that are reported through callbacks are placed in the recently_cemented list.

### DIFF
--- a/nano/secure/common.hpp
+++ b/nano/secure/common.hpp
@@ -360,14 +360,14 @@ class election_status final
 {
 public:
 	std::shared_ptr<nano::block> winner;
-	nano::amount tally;
-	nano::amount final_tally;
-	std::chrono::milliseconds election_end;
-	std::chrono::milliseconds election_duration;
-	unsigned confirmation_request_count;
-	unsigned block_count;
-	unsigned voter_count;
-	election_status_type type;
+	nano::amount tally{ 0 };
+	nano::amount final_tally{ 0 };
+	std::chrono::milliseconds election_end{ std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now ().time_since_epoch ()) };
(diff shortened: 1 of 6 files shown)


nano/secure/common.hpp
```diff
@@ -360,14 +360,14 @@ class election_status final
...
-	std::chrono::milliseconds election_end;
-	std::chrono::milliseconds election_duration;
-	unsigned confirmation_request_count;
-	unsigned block_count;
-	unsigned voter_count;
-	election_status_type type;
+	nano::amount tally{ 0 };
+	nano::amount final_tally{ 0 };
+	std::chrono::milliseconds election_end{ std::chrono::duration_cast<std::chrono::milliseconds> (std::chrono::system_clock::now ().time_since_epoch ()) };
+	std::chrono::milliseconds election_duration{ std::chrono::duration_values<std::chrono::milliseconds>::zero () };
```
pwojcikdev, line 366: Can't this be just `{ 0 }` or `{}`?

pwojcikdev (approved):
Good refactoring, new code makes much more sense.

nano/secure/common.hpp
clemahieu: I didn't check, I can't quite remember the rules for default initialization versus being uninitialized, this just seems more explicit.

PR #4517, opened 2024-03-22
Remove debug_assert for a tautology which was causing a warning. uint8_t can't be larger than 256

clemahieu:


pwojcikdev (approved):
I added this check just in case someone in the future changes the uint8 parameter type to something else, but if it causes warnings then it's ok to remove.

PR #4519, opened 2024-03-23
Reimplement and simplify confirmation_height_processor

clemahieu:
This PR replaces the confirmation_height_processor class with a simpler implementation named confirming_set. confirming_set represents the set of blocks that have been confirmed through elections but not yet confirmed on disk. The old confirmation_height_processor(bounded/unbounded) are removed.

The existing confirmation_height_processor was designed before the ascending bootstrapper when unchecked blocks were unbounded on disk. Under the original conditions, hundreds of thousands of blocks dependent blocks could be implicitly confirmed when confirming a block which greatly complicated the design out of the need for memory efficiency.

With the ascending bootstrapper and with the bounded backlog the number of dependent blocks that could be confirmed will never exceed the size of the bounded backlog so the upper end of memory requirements is greatly reduced.

This change adds the function ledger::confirm which recursively confirms a block and all its dependents.

Items to be confirmed on disk are aggregated in the confirming_set class and confirmed in batches within a single database transaction.

The confirmation_height tests are split between those that check the correctness of the ledger::confirm function and testing the confirming_set behavior within the node.


nano/secure/ledger.cpp
```diff
@@ -865,6 +865,50 @@ std::optional<nano::pending_info> nano::ledger::pending_info (store::transaction
...
+		{
+			// unconfirmed dependencies were added
+		}
+	}
+	return result;
+}
+
+void nano::ledger::confirm (nano::store::write_transaction const & transaction, nano::block const & block)
+{
+	debug_assert (!store.confirmation_height.get (transaction, block.account ()) || store.confirmation_height.get (transaction, block.account ()).value ().height < block.sideband ().height);
```
pwojcikdev: Can this check be tightened to assert that current confirmation height is exactly one less than block height? From what I understand this function is supposed to be used to incrementally confirm blocks in account chain, one by one. Is my understanding correct?

clemahieu: Yep that's true. Changed.

nano/core_test/ledger_confirm.cpp
```diff
@@ -0,0 +1,886 @@
...
+
+	auto transaction = node1->store.tx_begin_write ();
+	ASSERT_EQ (nano::block_status::progress, node1->ledger.process (transaction, send1));
+	node1->ledger.confirm (transaction, send1->hash ());
+	ASSERT_TRUE (node1->ledger.block_confirmed (transaction, send1->hash ()));
+	ASSERT_EQ (1, node1->stats.count (nano::stat::type::confirmation_height, nano::stat::detail::blocks_confirmed, nano::stat::dir::in));
+	ASSERT_EQ (2, node1->ledger.cache.cemented_count);
+}
+
+TEST (ledger_confirm, election_winner_details_clearing_node_process_confirmed)
```
pwojcikdev, line 755: The same test case exists in previous testsuite for `confirming_set`.

nano/core_test/confirmation_height.cpp
```diff
@@ -1,2187 +0,0 @@
...
-
-		ASSERT_EQ (0, node->active.election_winner_details_size ());
-	};
-
-	test_mode (nano::confirmation_height_mode::bounded);
-	test_mode (nano::confirmation_height_mode::unbounded);
-}
-
-// This test checks that a receive block with uncemented blocks below cements them too.
-TEST (confirmation_height, cemented_gap_below_receive)
```
pwojcikdev, line 1625: This and `cemented_gap_below_no_cache` appears to be removed from the new set of tests. Is this intentional?

nano/core_test/ledger_confirm.cpp
```diff
@@ -0,0 +1,886 @@
...
+	ASSERT_EQ (nano::block_status::progress, ledger.process (transaction, open2));
+	ASSERT_EQ (2, ledger.pruning_action (transaction, send2->hash (), 2));
+	ASSERT_FALSE (ledger.block_exists (transaction, send2->hash ()));
+	ASSERT_FALSE (ledger.block_confirmed (transaction, open2->hash ()));
+	auto confirmed = ledger.confirm (transaction, open2->hash ());
+	ASSERT_TRUE (ledger.block_confirmed (transaction, open2->hash ()));
+}
+
+// Test that if a block is marked to be confirmed that doesn't exist in the ledger the program aborts
+TEST (ledger_confirm, rollback_added_block_death)
```
pwojcikdev: This should either follow *DeathTest naming pattern (as to why is explained here https://github.com/google/googletest/blob/main/docs/advanced.md#death-tests-and-threads) or be modified to use exceptions which would avoid a lot of problems mentioned above.

nano/node/confirming_set.cpp
```diff
@@ -0,0 +1,120 @@
...
+		{
+			std::deque<std::shared_ptr<nano::block>> cemented;
+			std::deque<nano::block_hash> already;
+			// Move items in to back buffer and release lock so more items can be added to the front buffer
+			processing = std::move (this->set);
+			// Process all items in the back buffer
+			for (auto i = processing.begin (), n = processing.end (); !stopped && i != n;)
+			{
+				lock.unlock (); // Waiting for db write is potentially slow
+				write_queue.wait (nano::writer::confirmation_height);
```
pwojcikdev: This call returns a `write_guard` object that in previous version of conf height processor was kept in scope through the processing. I'm not that familiar with exact mechanics and purpose of it, but it appears to prevent acquiring other writers in parallel.

nano/node/node.hpp
```diff
@@ -168,7 +168,8 @@ class node final : public std::enable_shared_from_this<node>
 	nano::node_observers observers;
 	nano::port_mapping port_mapping;
 	nano::block_processor block_processor;
-	nano::confirmation_height_processor confirmation_height_processor;
+	std::shared_ptr<nano::confirming_set> confirming_set_impl;
```
pwojcikdev: Can this be unique_ptr? Understanding shared_ptr lifetime is generally harder with the assumption that it can outlive the parent node, is this the case here?

nano/core_test/ledger_confirm.cpp
clemahieu: Fixed

nano/node/confirming_set.cpp
clemahieu: Good catch, that wasn't clear from the description. I opened https://github.com/nanocurrency/nano-node/pull/4522 to add [[nodiscard]] to help avoid this in the future.

nano/node/node.hpp
clemahieu: This was unintentional.  Fixed.

nano/core_test/confirmation_height.cpp
clemahieu: I put these test deletions into a single commit to single out their removal. https://github.com/nanocurrency/nano-node/pull/4519/commits/abc127ad5999530e738da2bcbe72a201360ec055

These tests seemed to be testing behavior unrelated to confirming: gaps in source/previous or bootstrapping/live block sources. I'm not fully sure the original intent whether it was pedantic or it might have to do with multiple db transactions being used per traversal.

The new implementation is strongly intolerant to missing blocks if they are requested to be confirmed and will release_assert.

clemahieu:
Changed the ledger::confirm algorithm to use a stack rather than a deque since we're not using it as a queue.

nano/node/confirming_set.cpp
pwojcikdev: It might be worth to consider whether this write queue is needed at all, from what I understand a single database write transaction can be active at a time anyway.

nano/core_test/confirmation_height.cpp
pwojcikdev: Okay, this makes total sense. I was confused by those tests and didn't spend too much time analyzing them, thanks for the explanation. I agree that crashing loudly with release asserts is the way to go in such cases, if chains with gaps are being cemented then something must have gone terribly wrong.

pwojcikdev (approved):

PR #4521, opened 2024-03-24
CLI option to merge new node config changes into current config

RickiNano:
The node config options are constantly changing with each new version of the node. New ones are added and old ones are removed. As a node operator it's not easy to keep your config file up to date because it currently needs to be done manually.
This PR adds a new CLI option called "update_config" that will read the current configuration file and find all non-default values. It will then insert them into a fresh default node configuration, and output it to the console.


RickiNano:
Rebased on top of current develop to fix merge issue

RickiNano:
I've discovered potential issue with this PR. I'll close it and create a new one when the problem is solved

PR #4522, opened 2024-03-24
Improve write_database_queue::wait description and add [[nodiscard]]

clemahieu:


pwojcikdev (dismissed):
> Blocks until we are at the head of the queue and blocks other waiters until write_guard goes out of scope

I'm having a bit of trouble understanding why this `write_database_queue` is needed in the first place, this description sounds like just acquiring a mutex.

clemahieu:
Added a better description. The write queue gives fair access to the exclusive database write transaction which is guarded by a mutex with no fairness guarantee.

PR #4526, opened 2024-03-28
Show `receivable` and `weight` for unopened accounts in `account_info`rpc

gr0vity-dev:
Currently when an account is not opened,  the rpc command `account_info` returns an error `Account not found`.
Even if we provide the `receivable` or `weight` flags, the receivable blocks or the delegated weight is not shown.

```json
{
  "action": "account_info",
  "account": "nano_1dckaox4wondc4dtsn9fnw1wzgehb8q9agx4e6rch43kaadrobciepmbjcd3",
  "receivable": "true",
  "weight" : "true"
}
```
returns
```json
{
    "error": "Account not found"
}
```

Even though it has receivable blocks.
This pull request allows ordinary response keys to coexist with an error key
The responsefor the example above would look like this:
```json
{
    "error": "Account not found",
    "pending": "1100000000000000000000000000000",
    "receivable": "1100000000000000000000000000000",
    "weight": "0"
}
```

All the rpc unit tests are passing. There is no test that explicitly prohibits this behavior.
This change in behavior is probably not be wanted, but it greatly simplified the rpc requests required for nanobrowse explorer.

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -624,6 +624,24 @@ void nano::json_handler::account_info ()
 		auto info (account_info_impl (transaction, account));
 		nano::confirmation_height_info confirmation_height_info;
 		node.store.confirmation_height.get (transaction, account, confirmation_height_info);
+		if (weight)
+		{
+			auto account_weight (node.ledger.weight (account));
+			response_l.put ("weight", account_weight.convert_to<std::string> ());
+		}
+		if (receivable)
+		{
+			auto account_receivable = node.ledger.account_receivable (transaction, account);
+			response_l.put ("pending", account_receivable.convert_to<std::string> ());
+			response_l.put ("receivable", account_receivable.convert_to<std::string> ());
+
+			if (include_confirmed)
+			{
+				auto account_receivable = node.ledger.account_receivable (transaction, account, true);
+				response_l.put ("confirmed_pending", account_receivable.convert_to<std::string> ());
+				response_l.put ("confirmed_receivable", account_receivable.convert_to<std::string> ());
+			}
+		}
 		if (!ec)
 		{
 			response_l.put ("frontier", info.head.to_string ());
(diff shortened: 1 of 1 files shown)


dsiganos:
If this is a feature that we add then it should have its own test case.

dsiganos:
The concept of returning partial results together with an error is a little messy.
Have we considered if there are better ways to achieve a similar result?
Can you explain what you have to do to get this info without having this feature?

gr0vity-dev:
I agree that it is messy and I don't like the solution very much. It's more of a hack.
The main issue is that an `Account not found` is returned even though the account has operations on it. (delegators or receivable blocks)

In order to make the same work for unopened accounts, I have to do 2 more rpc calls
https://docs.nano.org/commands/rpc-protocol/#account_weight
https://docs.nano.org/commands/rpc-protocol/#accounts_receivable
and merge the data.

pwojcikdev:
RPC commands should be designed in such a way to make it as easy as possible to create integrations. Instead of returning "Account not found", maybe "Account not opened" should be returned or perhaps do not return error at all? If I was writing a python parser I would probably check for "error" field and throw if it's found. Returning error even when account is found obviously breaks this design.

nano/node/json_handler.cpp
```diff
@@ -169,18 +169,12 @@ void nano::json_handler::response_errors ()
...
-		std::stringstream ostream;
-		boost::property_tree::write_json (ostream, response_error);
-		response (ostream.str ());
-	}
-	else
-	{
-		std::stringstream ostream;
-		boost::property_tree::write_json (ostream, response_l);
-		response (ostream.str ());
+		response_l.put ("error", ec.message ());
```
dsiganos, line 172: This is changing the behaviour of `nano::json_handler::response_errors ()` to allow for returning an error mixed with partial results. Previously, error was always returned alone, even if partial results existed. Since response_errors () is called in many places, and it is difficult to check all of them, I suggest we do not change its behaviour and instead create a special case function for returning error with partial results (if that is what we decide to do). For example, `nano::json_handler::response_errors_with_data ()`.

gr0vity-dev:
> RPC commands should be designed in such a way to make it as easy as possible to create integrations. 

I agree with this, as I'm using a custom python library to interact with rpc endpoint which doesn't map the output but keeps it as json. So the library simply forwards what the rpc endpoint returns so the user can decide whether to throw or not.
I decided to add a special flag to ignore matching errors. In this case "Account not found" and not throw.

I can see that partial results can be an issue for integrations...
There are probably many ways we could take, like a return a warning "Account not opened".

Maybe I'll just close the pull request and we'll discuss when we focus on making rpc usage easier.

nano/node/json_handler.cpp
gr0vity-dev: I put it in a dedicated commit to make this change in behavior clear.
I think that the error "Account not found" is incorrect in this case and it should be "Account not opened"

We could have an additional flag `partial_data` in `account_info` rpc call.
This way the end-user would make it clear he wants to be shown partial data for this call and it would be easy on integrations. Also it would limit the impact on this rpc call alone.

gr0vity-dev:
Closing for now because the approach of mixing partial responses with the error key is a bit messy.

PR #4533, opened 2024-04-03
RPC endpoint for election statistics

RickiNano:
This pull request adds a new election_statistics RPC endpoint to help identify bottlenecks within the AEC.
The endpoint will respond with the number of active elections for each electiontype. It will also return the percentage of AEC slots in use.
If this is merged, I intend to expand the dataset with the average and max election age.

Request:
```
{
    "action": "election_statistics" 
}
```

Response example:
```
{
    "normal": "1595",
    "hinted": "70",
    "optimistic": "222",
    "total": "1887",
    "aec_utilization_percentage": "37.74"
}
```

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -6866,3 +6866,18 @@ TEST (rpc, confirmation_info)
 		ASSERT_EQ (0, response.get<unsigned> ("total_tally"));
 	}
 }
+
+TEST (rpc, election_statistics)
+{
+	nano::test::system system;
+	auto node1 = add_ipc_enabled_node (system);
+	auto const rpc_ctx = add_rpc (system, node1);
+	boost::property_tree::ptree request1;
+	request1.put ("action", "election_statistics");
+	auto response1 (wait_response (system, rpc_ctx, request1));
+	ASSERT_EQ ("0", response1.get<std::string> ("normal"));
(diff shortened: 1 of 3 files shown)


clemahieu:
This seems useful. The PR should include a test that at least runs the code.

nano/rpc_test/rpc.cpp
```diff
@@ -6866,3 +6866,18 @@ TEST (rpc, confirmation_info)
...
+	auto node1 = add_ipc_enabled_node (system);
+	auto const rpc_ctx = add_rpc (system, node1);
+	boost::property_tree::ptree request1;
+	request1.put ("action", "election_statistics");
+	auto response1 (wait_response (system, rpc_ctx, request1));
+	ASSERT_EQ ("0", response1.get<std::string> ("normal"));
+	ASSERT_EQ ("0", response1.get<std::string> ("hinted"));
+	ASSERT_EQ ("0", response1.get<std::string> ("optimistic"));
+	ASSERT_EQ ("0", response1.get<std::string> ("total"));
+	ASSERT_EQ ("0.00", response1.get<std::string> ("aec_utilization_percentage"));
```
pwojcikdev, line 6882: This could at least start a single election which would ensure reported values work as expected

RickiNano: @pwojcikdev I'm not sure how to do this. If I start a normal election it might get confirmed before the RPC request is executed.

PR #4535, opened 2024-04-04
Move write database queue

clemahieu:
This renames write_database_queue to write_queue and moves the class into the nano::store library. It also moves the object from nano::node to nano::store::component.

This is a more natural place for this class/object to be placed and also preps for confirming_set to be moved on to the ledger class which would otherwise have a cyclic dependency secure -> store -> node -> secure


nano/node/blockprocessor.hpp
```diff
@@ -86,7 +86,7 @@ class block_processor final
 	};
 
 public:
-	block_processor (nano::node &, nano::write_database_queue &);
+	block_processor (nano::node &, nano::store::write_queue &);
```
pwojcikdev: Given that write_queue is now part of store object, is it necessary to separately pass it in constructors?

nano/store/write_queue.hpp
```diff
@@ -38,27 +38,27 @@ class write_guard final
  * Allocates database write access in a fair maner rather than directly waiting for mutex aquisition
  * Users should wait() for access to database write transaction and hold the write_guard until complete
  */
-class write_database_queue final
+class write_queue final
 {
 public:
-	write_database_queue (bool use_noops_a);
+	write_queue (bool use_noops_a);
```
pwojcikdev: Since the class is getting a small refresh, adding explicit here will silence warnings about implicit conversions.

pwojcikdev (dismissed):

nano/node/blockprocessor.hpp
clemahieu: Good point, not sure why they were being passed in. Fixed.

PR #4537, opened 2024-04-04
Election age statistics

RickiNano:
This PR adds information to the `election_statistics` RPC call about the maximum and average age of elections in AEC .
This will help identify long running elections.
I have also updated the unit test to create an election

Request:
```
{
  "action": "election_statistics"
}
```
Response example:
```
{
    "normal": "200",
    "hinted": "22",
    "optimistic": "425",
    "total": "647",
    "aec_utilization_percentage": "8.09",
    "max_election_age": "6",
    "average_election_age": "2.46"
}
```

### DIFF
--- a/nano/node/election.hpp
+++ b/nano/node/election.hpp
@@ -118,6 +118,10 @@ public: // Interface
 	nano::vote_info get_last_vote (nano::account const & account);
 	void set_last_vote (nano::account const & account, nano::vote_info vote_info);
 	nano::election_status get_status () const;
+	std::chrono::steady_clock::time_point get_election_start () const
+	{
+		return election_start;
+	}
 
 private: // Dependencies
 	nano::node & node;
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -2009,10 +2009,21 @@ void nano::json_handler::election_statistics ()
 	unsigned hinted_count = 0;
 	unsigned optimistic_count = 0;
 	unsigned total_count = 0;
+	std::chrono::steady_clock::duration total_age{};
+	auto now = std::chrono::steady_clock::now ();
+	std::chrono::steady_clock::time_point oldest_election_start = now;
 
 	for (auto const & election : active_elections)
 	{
 		total_count++;
+		auto election_start = election->get_election_start ();
+		auto age = now - election_start;
+		total_age += age;
+		if (election_start < oldest_election_start)
+		{
+			oldest_election_start = election_start;
+		}
+
 		switch (election->behavior ())
 		{
 			case election_behavior::normal:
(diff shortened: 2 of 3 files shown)


nano/node/election.hpp
```diff
@@ -118,6 +118,10 @@ class election final : public std::enable_shared_from_this<election>
 	nano::vote_info get_last_vote (nano::account const & account);
 	void set_last_vote (nano::account const & account, nano::vote_info vote_info);
 	nano::election_status get_status () const;
+	std::chrono::steady_clock::time_point get_election_start () const
```
clemahieu, line 121: Since the start time is public and const I think it's fine to just pull the value directly and this accessor isn't needed.

nano/node/json_handler.cpp
```diff
@@ -2009,10 +2009,21 @@ void nano::json_handler::election_statistics ()
...
+	auto now = std::chrono::steady_clock::now ();
+	std::chrono::steady_clock::time_point oldest_election_start = now;
 
 	for (auto const & election : active_elections)
 	{
 		total_count++;
+		auto election_start = election->get_election_start ();
+		auto age = now - election_start;
+		total_age += age;
+		if (election_start < oldest_election_start)
```
clemahieu, line 2022: Can oldest_election_start = std::min(oldest_election_start, election->election_start) be used instead of branching?

nano/node/election.hpp
RickiNano: election_start is private so I can't access it directly

nano/node/json_handler.cpp
RickiNano: I don't think I can do that if using the getter function. That would probably require election_start to be public which I think is a bad idea. It shouldn't be modified outside the election class

clemahieu: This should be compatible with the getter:

oldest_election_start = std::min(oldest_election_start, election->get_election_start())

nano/rpc_test/rpc.cpp
```diff
@@ -6870,14 +6870,38 @@ TEST (rpc, confirmation_info)
 TEST (rpc, election_statistics)
 {
 	nano::test::system system;
-	auto node1 = add_ipc_enabled_node (system);
+	nano::node_config node_config;
+	node_config.ipc_config.transport_tcp.enabled = true;
+	node_config.ipc_config.transport_tcp.port = system.get_available_port ();
+	nano::node_flags node_flags;
+	node_flags.disable_request_loop = true;
```
dsiganos, line 6877: It would be nice to explain why you are disabling the request loop.

nano/rpc_test/rpc.cpp
```diff
@@ -6870,14 +6870,38 @@ TEST (rpc, confirmation_info)
...
+	auto send1 = builder
+				 .send ()
+				 .previous (nano::dev::genesis->hash ())
+				 .destination (nano::public_key ())
+				 .balance (nano::dev::constants.genesis_amount - 100)
+				 .sign (nano::dev::genesis_key.prv, nano::dev::genesis_key.pub)
+				 .work (*system.work.generate (nano::dev::genesis->hash ()))
+				 .build ();
+	node1->process_active (send1);
+	ASSERT_TIMELY (5s, !node1->active.empty ());
```
dsiganos, line 6891: This line looks strange to me. Maybe there is a race condition between process_active() and start_elections(). The usual way to use start_elections() is:
```
	ASSERT_TRUE (nano::test::process (*node1, {send1}));
	ASSERT_TRUE (nano::test::start_elections (system, *node1, {send1}));
```

nano/rpc_test/rpc.cpp
```diff
@@ -6870,14 +6870,38 @@ TEST (rpc, confirmation_info)
...
+	boost::property_tree::ptree request;
+	request.put ("action", "election_statistics");
+	{
+		auto response (wait_response (system, rpc_ctx, request));
+		ASSERT_EQ ("1", response.get<std::string> ("normal"));
+		ASSERT_EQ ("0", response.get<std::string> ("hinted"));
+		ASSERT_EQ ("0", response.get<std::string> ("optimistic"));
+		ASSERT_EQ ("1", response.get<std::string> ("total"));
+		ASSERT_NE ("0.00", response.get<std::string> ("aec_utilization_percentage"));
+		ASSERT_NO_THROW (response.get<std::string> ("max_election_age"));
```
dsiganos, line 6904: These 2 checks could check that the values are within an expected range.

PR #4547, opened 2024-04-09
Fixes TSAN issues.

clemahieu:
TSAN identified issues in peer_container.tcp_channel_cleanup_works and node.fork_keep

### DIFF
--- a/nano/test_common/network.cpp
+++ b/nano/test_common/network.cpp
@@ -24,6 +24,14 @@ std::shared_ptr<nano::transport::channel_tcp> nano::test::establish_tcp (nano::t
 	return result;
 }
 
+std::shared_ptr<nano::node> nano::test::add_outer_node (nano::test::system & system_a, nano::node_config const & config_a, nano::node_flags flags_a)
+{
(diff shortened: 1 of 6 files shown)


nano/test_common/network.cpp
```diff
@@ -24,6 +24,14 @@ std::shared_ptr<nano::transport::channel_tcp> nano::test::establish_tcp (nano::t
 	return result;
 }
 
+std::shared_ptr<nano::node> nano::test::add_outer_node (nano::test::system & system_a, nano::node_config const & config_a, nano::node_flags flags_a)
+{
+	auto outer_node = std::make_shared<nano::node> (system_a.io_ctx, nano::unique_path (), config_a, system_a.work, flags_a);
+	outer_node->start ();
+	system_a.nodes.push_back (outer_node);
```
dsiganos, line 31: By adding it to nodes, any subsequent nodes will connect to this node.
This should be added to the disconnected nodes set.

dsiganos: I fixed it here:
https://github.com/nanocurrency/nano-node/pull/4548/files

PR #4549, opened 2024-04-10
Max election age precision

RickiNano:
`max_election_age` is currently being casted as `std::chrono::seconds` which causes the value to be rounded to nearest second. This pr changes it to use `double` and returns the value with 2 decimals.
It's unfortunate that I did not notice this issue in the original PR.
The unit test is also updated to test that both election age values are greater than zero. On my machine they are 0.09 seconds during unit test.
I also implemented the syntax Colin suggested for `oldest_election_start` to avoid the if statement

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -6898,8 +6898,8 @@ TEST (rpc, election_statistics)
 	ASSERT_EQ ("0", response.get<std::string> ("optimistic"));
 	ASSERT_EQ ("1", response.get<std::string> ("total"));
 	ASSERT_NE ("0.00", response.get<std::string> ("aec_utilization_percentage"));
-	ASSERT_GT (response.get<double> ("max_election_age"), 0);
-	ASSERT_GT (response.get<double> ("average_election_age"), 0);
-	ASSERT_LT (response.get<double> ("max_election_age"), 5);
-	ASSERT_LT (response.get<double> ("average_election_age"), 5);
+	ASSERT_GT (response.get<int> ("max_election_age"), 0);
+	ASSERT_GT (response.get<int> ("average_election_age"), 0);
+	ASSERT_LT (response.get<int> ("max_election_age"), 5000);
+	ASSERT_LT (response.get<int> ("average_election_age"), 5000);
 }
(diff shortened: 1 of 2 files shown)


nano/rpc_test/rpc.cpp
```diff
@@ -6901,7 +6901,7 @@ TEST (rpc, election_statistics)
 		ASSERT_EQ ("0", response.get<std::string> ("optimistic"));
 		ASSERT_EQ ("1", response.get<std::string> ("total"));
 		ASSERT_NE ("0.00", response.get<std::string> ("aec_utilization_percentage"));
-		ASSERT_NO_THROW (response.get<std::string> ("max_election_age"));
```
dsiganos, line 6904: You could also check that the value is less than 5 seconds.

dsiganos:
The unit test is a little messy and more complicated than it needs to be.
I will tidy it up so that you can see what are the unnecessary elements.

dsiganos:
There is no need to disable_request_loop because there are no voters on the network and the election will not be voted on and it will sit and wait.

dsiganos (approved):

dsiganos (dismissed):

clemahieu (dismissed):

simpago:
Wouldn't it be simpler to return integer milliseconds instead of floating point seconds?

RickiNano:
Good point @simpago 
What do you think @dsiganos and @clemahieu ?

dsiganos:
In general, I always favour integers and fixed point to floating point. Floating point comes with too many traps. I agree that returning milliseconds is preferable.

RickiNano:
I have updated it to use milliseconds

dsiganos:
But now the other floating point sticks out like a sore thumb...
We can avoid the other floating point by diving it into smaller units too.

RickiNano:
I don't see anything wrong with having the percentage value as a floating point value. Without any decimals it lacks the required precision.

dsiganos (approved):

dsiganos:
Floats are inaccurate and cause problems and it is good to avoid them. But  in RPC code, I don't mind having floats, I would be worried if the floats were in the protocol code.

PR #4552, opened 2024-04-12
Fix and re-enable unit test rpc.wallet_history

### LINKED ISSUE #3514: Unit test `rpc.wallet_history` intermittently failing
Disabled in https://github.com/nanocurrency/nano-node/pull/3512/commits/e1d0a51253d19fa725fb30fc84880c89ddc50527

dsiganos:
It appears that the problme was that a timestamp was checked for exactness instead of chacking for a time range.

fixes #3514

### DIFF
--- a/nano/rpc_test/rpc.cpp
+++ b/nano/rpc_test/rpc.cpp
@@ -5622,21 +5622,21 @@ TEST (rpc, wallet_history)
 	ASSERT_EQ (send2->hash ().to_string (), std::get<3> (history_l[0]));
 	ASSERT_EQ (nano::dev::genesis_key.pub.to_account (), std::get<4> (history_l[0]));
 	ASSERT_LE (timestamp, std::stoull (std::get<5> (history_l[0])));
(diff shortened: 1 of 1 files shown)


clemahieu (dismissed):

nano/rpc_test/rpc.cpp
```diff
@@ -5613,6 +5608,7 @@ TEST (rpc, DISABLED_wallet_history)
 	request.put ("action", "wallet_history");
 	request.put ("wallet", node->wallets.items.begin ()->first.to_string ());
 	auto response (wait_response (system, rpc_ctx, request));
+	boost::property_tree::write_json (std::cout, response);
```
pwojcikdev: Having logs from tests is great, but please only write to stdout if logging is enabled.

dsiganos: No, was not meant to  be there. I will remove it.

dsiganos:
Looks like I messed up the last tw check, I will fix them.

PR #4553, opened 2024-04-12
Documentation for pending key and info classes and some other improvements whilst investigating

dsiganos:
The motivation for this PR was a desire to document pending_key and pending_info classes.
However, before I could document them, I needed to understand them.
I created a test case for pending table to help me understand it.
I created a function to print the pending table for learning and debugging purposes.
And whilst creating that test case, I noticed that there were refactoring improvements that could be made.

### DIFF
--- a/nano/secure/pending_info.hpp
+++ b/nano/secure/pending_info.hpp
@@ -18,6 +18,7 @@ namespace nano
 {
 /**
  * Information on an uncollected send
+ * This class captures the data stored in a pending table entry
  */
 class pending_info final
 {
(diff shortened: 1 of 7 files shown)


clemahieu (approved):
We've used the "east-const" style where the type can always be read from right to left.

Though not worth redoing these commits for that.

https://isocpp.org/wiki/faq/const-correctness#const-ref-alt

pwojcikdev (approved):

nano/secure/pending_info.hpp
```diff
@@ -27,10 +28,20 @@ class pending_info final
...
 	bool deserialize (nano::stream &);
 	bool operator== (nano::pending_info const &) const;
-	nano::account source{};
-	nano::amount amount{ 0 };
-	nano::epoch epoch{ nano::epoch::epoch_0 };
+	nano::account source{}; // the account sending the funds
+	nano::amount amount{ 0 }; // amount receivable in this transaction
+	nano::epoch epoch{ nano::epoch::epoch_0 }; // epoch of sending block, this info is stored here to make it possible to prune the send block
+
+	friend std::ostream & operator<< (std::ostream & os, const nano::pending_info & info)
```
pwojcikdev, line 35: This obviously won't work well with tracing or serialization, not sure how useful it is aside from primitive test logging.

dsiganos: It is mainly for use for primitive logging and for printing the value of the struct from inside the debugger.
I am happy to make it compatible with tracing library but I would like to do that as a separate step so that it doesn't become a big task.

PR #4554, opened 2024-04-13
disable rpm builds

gr0vity-dev:
Currently our linux rpm build fails because a different set of scripts is used. 
Here is a [recent example](https://github.com/nanocurrency/nano-node/actions/runs/8669382623) of a failed rpm build.
Currently there is no maintainer for these scripts.

This PR simply removes the rpm build from the github workflow to make the builds green again.
TODO: Remove all the unused scripts related to the rpm builds.

### DIFF
--- a/.github/workflows/build_deploy.yml
+++ b/.github/workflows/build_deploy.yml
@@ -102,51 +102,6 @@ jobs:
           AWS_DEFAULT_REGION: us-east-2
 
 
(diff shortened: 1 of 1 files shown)


clemahieu:
For reference, this error looks related to a false-positive that was fixed in GCC 13. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651

PR #4556, opened 2024-04-13
Vote generator write queue

clemahieu:
Changes the vote_generators to respect the database write queue.

Add consistency check to ensure the writer token is not already queued since they're waited for by value.
Change wait-loop to use the predicated overload of wait.

### DIFF
--- a/nano/node/vote_generator.cpp
+++ b/nano/node/vote_generator.cpp
@@ -94,6 +94,7 @@ void nano::vote_generator::process_batch (std::deque<queue_entry_t> & batch)
 {
 	std::deque<candidate_t> candidates_new;
 	{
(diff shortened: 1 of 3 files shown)


pwojcikdev (approved):
Was there any performance or stability problem with previous code? Or it's just for consistency?

clemahieu:
> Was there any performance or stability problem with previous code? Or it's just for consistency?

Yea there was a weird behavior where one of the schedulers (I think hinted) would 95% of the time execute something instantly but sometimes it would take multiple seconds.

This was on the transaction locks branch so it's hard to tell since that's still being debugged. It's worth a new performance analysis with this.

PR #4557, opened 2024-04-13
Move node::receive_confirmed to wallets::receive_confirmed

clemahieu:
This is inherently a wallet operation so it's more appropriate to be on the wallets class.

This also breaks a direct dependency/cycle from block processing to the wallet operations.

This also breaks a lock order inversion where a ledger transaction mutex is held while acquiring the wallet mutexes while most operations do it in the reverse order.

### DIFF
--- a/nano/node/wallet.cpp
+++ b/nano/node/wallet.cpp
@@ -1712,6 +1712,40 @@ void nano::wallets::ongoing_compute_reps ()
 	});
 }
 
+void nano::wallets::receive_confirmed (nano::block_hash const & hash_a, nano::account const & destination_a)
+{
+	nano::unique_lock<nano::mutex> lk{ mutex };
+	auto wallets_l = get_wallets ();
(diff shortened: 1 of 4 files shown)


nano/node/wallet.cpp
```diff
@@ -1712,6 +1712,40 @@ void nano::wallets::ongoing_compute_reps ()
...
+			else
+			{
+				if (!node.ledger.block_or_pruned_exists (node.ledger.store.tx_begin_read (), hash_a))
+				{
+					node.logger.warn (nano::log::type::node, "Confirmed block is missing: {}", hash_a.to_string ());
+					debug_assert (false, "Confirmed block is missing");
+				}
+				else
+				{
+					node.logger.warn (nano::log::type::node, "Block has already been received: {}", hash_a.to_string ());
```
pwojcikdev: `nano::log::type` should be changed to `wallet`

pwojcikdev (dismissed):

PR #4558, opened 2024-04-13
Generate indented toml config files for improved legibility

RickiNano:
The current config layout is quite hard to navigate.
This PR adds indentation to clearly show what is a child object of a node. Indentations are allowed for the TOML format according to the official documentation.
`--generate_config` cli command has been tested with both node, rpc and log configs. It also supports the `--use_defaults` command with correct output.
Here is a small sample of the new config file for illustration:
```
[node.httpcallback]

        # Callback address.
        # type:string,ip
        # address = ""

        # Callback port number.
        # type:uint16
        # port = 0

        # Callback target path.
        # type:string,uri
        # target = ""

[node.ipc.flatbuffers]

        # Allow client to send unknown fields in json messages. These will be ignored.
        # type:bool
        # skip_unexpected_fields_in_json = true

        # Verify that the buffer is valid before parsing. This is recommended when receiving data from untrusted sources.
        # type:bool
        # verify_buffers = true

[node.ipc.local]

        # If enabled, certain unsafe RPCs can be used. Not recommended for production systems.
        # type:bool
        # allow_unsafe = false

        # Enable or disable IPC via local domain socket.
        # type:bool
        # enable = false

        # Timeout for requests.
        # type:seconds
        # io_timeout = 15

        # Path to the local domain socket.
        # type:string
        # path = "/tmp/nano"

[node.ipc.tcp]

        # Enable or disable IPC via TCP server.
        # type:bool
        # enable = false

        # Timeout for requests.
        # type:seconds
        # io_timeout = 15

        # Server listening port.
        # type:uint16
        # port = 7077
```

### DIFF
--- a/nano/lib/tomlconfig.cpp
+++ b/nano/lib/tomlconfig.cpp
@@ -186,15 +186,7 @@ void nano::tomlconfig::erase_default_values (tomlconfig & defaults_a)
 	erase_defaults (defaults_l.get_tree (), self.get_tree (), get_tree ());
 }
 
-std::string nano::tomlconfig::to_string ()
-{
-	std::stringstream ss;
-	cpptoml::toml_writer writer{ ss, "" };
-	tree->accept (writer);
-	return ss.str ();
-}
-
-std::string nano::tomlconfig::to_string_commented_entries ()
+std::string nano::tomlconfig::to_string (bool comment_values)
 {
 	std::stringstream ss, ss_processed;
 	cpptoml::toml_writer writer{ ss, "" };
@@ -202,9 +194,16 @@ std::string nano::tomlconfig::to_string_commented_entries ()
 	std::string line;
 	while (std::getline (ss, line, '\n'))
 	{
-		if (!line.empty () && line[0] != '#' && line[0] != '[')
+		if (!line.empty () && line[0] != '[')
 		{
-			line = "#" + line;
+			if (line[0] == '#') // Already commented
+			{
+				line = "\t" + line;
+			}
+			else
+			{
+				line = comment_values ? "\t# " + line : "\t" + line;
+			}
 		}
 		ss_processed << line << std::endl;
 	}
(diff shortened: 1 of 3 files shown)


pwojcikdev:
Do we really want to do it? Some time ago I researched a similar approach and was discouraged after reading through https://github.com/toml-lang/toml/issues/608 especially since most tools should offer pretty good toml syntax highlighting.

RickiNano:
The TOML language is not that widespread. Even notepad++ that supports 50+ languages doesn't come with a toml parser.
I prefer it to be indented. It makes it super easy to read no matter what text editor you are using.

pwojcikdev:
Okay, without a syntax highligher indentation definitely helps.

nano/lib/tomlconfig.cpp
```diff
@@ -186,25 +186,24 @@ void nano::tomlconfig::erase_default_values (tomlconfig & defaults_a)
...
-	cpptoml::toml_writer writer{ ss, "" };
-	tree->accept (writer);
-	return ss.str ();
-}
-
-std::string nano::tomlconfig::to_string_commented_entries ()
+std::string nano::tomlconfig::to_string (bool comment_values)
 {
 	std::stringstream ss, ss_processed;
 	cpptoml::toml_writer writer{ ss, "" };
```
clemahieu, line 192: Looking at the toml_writer constructor it looks like there is already support for indenting that we're overriding by passing in an empty string for the second argument.


`    toml_writer(std::ostream& s, const std::string& indent_space = "\t")`
    
 I believe we should be using that rather than implementing our own.

clemahieu (changes requested):

nano/lib/tomlconfig.cpp
RickiNano: Unfortunately the build in indent function only indents the non-commented lines. Looks like this:
```
[node.rep_crawler]
                query_timeout = 60000

[node.rocksdb]

# Whether to use the RocksDB backend for the ledger database.
# type:bool
                enable = false

# Number of threads to use with the background compaction and flushing. Number of hardware threads is recommended.
# type:uint32
                io_threads = 20

# This will modify how much memory is used represented by 1 (low), 2 (medium), 3 (high). Default is 2.
# type:uint8
                memory_multiplier = 2

[node.statistics.log]

# Log file name for counters.
# type:string
                        filename_counters = "counters.stat"
```

clemahieu (approved):

PR #4564, opened 2024-04-14
Fix race conditions in accessing election confirmation counters 

dsiganos:


clemahieu (dismissed):

nano/core_test/confirming_set.cpp
```diff
@@ -165,7 +165,7 @@ TEST (confirmation_callback, confirmed_history)
 		ASSERT_TIMELY (10s, node->store.write_queue.contains (nano::store::writer::confirmation_height));
 
 		// Confirm that no inactive callbacks have been called when the confirmation height processor has already iterated over it, waiting to write
-		ASSERT_EQ (0, node->stats.count (nano::stat::type::confirmation_observer, nano::stat::detail::inactive_conf_height, nano::stat::dir::out));
+		ASSERT_ALWAYS_EQ (1ms, 0, node->stats.count (nano::stat::type::confirmation_observer, nano::stat::detail::inactive_conf_height, nano::stat::dir::out));
```
pwojcikdev: Is 1ms long enough for this assert?

pwojcikdev (dismissed):

nano/core_test/confirming_set.cpp
dsiganos: It should be, we only need to catch it most of the time, or even just sometimes, not all the time. It is not worth delaying the tests more.

pwojcikdev: Are you 100% sure about that? 1ms is probably shorter than a single context switch on a congested GH runner, it's likely that this loop won't ever run twice, especially after system io context refactorings. I don't really see a problem with this particular check as it's not important, but this is obviously setting a bad precedent.

dsiganos: Ok, suggest a value. I am happy to change it.

pwojcikdev: Maybe 50ms? Test runtimes won't be noticeably affected.

dsiganos: Ok

PR #4566, opened 2024-04-15
Shorten transaction lifetimes by using temporary expressions in some trivial cases.

clemahieu:
Eliminate holding ledger transactions while making or servicing callbacks. This is to facilitate using a shared_mutex to synchronize memory objects with the ledger https://github.com/nanocurrency/nano-node/pull/4567

### DIFF
--- a/nano/node/scheduler/optimistic.cpp
+++ b/nano/node/scheduler/optimistic.cpp
@@ -127,8 +127,6 @@ void nano::scheduler::optimistic::run ()
 
 		if (predicate ())
 		{
-			auto transaction = ledger.store.tx_begin_read ();
-
 			while (predicate ())
 			{
 				debug_assert (!candidates.empty ());
@@ -137,7 +135,7 @@ void nano::scheduler::optimistic::run ()
 
 				lock.unlock ();
 
-				run_one (transaction, candidate);
+				run_one (ledger.store.tx_begin_read (), candidate);
 
 				lock.lock ();
 			}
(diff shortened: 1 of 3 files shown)


nano/node/scheduler/optimistic.cpp
```diff
@@ -137,7 +135,7 @@ void nano::scheduler::optimistic::run ()
 
 				lock.unlock ();
 
-				run_one (transaction, candidate);
+				run_one (ledger.store.tx_begin_read (), candidate);
```
pwojcikdev, line 138: Was this tested to actually improve performance? I know that for write transactions it's better to have longer running instances and the difference is huge.

clemahieu: The main purpose is to eliminate holding transactions while making callbacks or notifying observers. I added a PR comment linking to the PR that adds a shared_lock to ledger transactions to guard memory objects.

pwojcikdev: Okay, the intention is good but I'm not sure it's the way to fix it, especially since this would imply many other components need to be changed too.

pwojcikdev: A call to `transaction.refresh_is_needed` might be helpful here, but requiring starting a fresh transaction for each iteration should never be necessary in a properly designed system.

clemahieu: There are options to batch these items if there's a performance problem, which we don't know if there is.

This loop could: copy the candidates, unlock, create db transaction, and run operation on each copied item, lock.

pwojcikdev: So the problem is with keeping long lived transaction while holding its own mutex, is this correct? I might have misunderstood the problem.

pwojcikdev (approved):

PR #4571, opened 2024-04-17
Fix windows tests

RickiNano:
The timeout command behaves differently in windows than linux and mac. In windows it will pause for the specified value before continuing the script.
This PR enables windows system tests to be launched from a native Windows command prompt instead of the bash and use the build in timeout function (in milliseconds)

### DIFF
--- a/ci/tests/run-system-tests.sh
+++ b/ci/tests/run-system-tests.sh
@@ -27,8 +27,9 @@ for script in ${NANO_SYSTEST_DIR}/*.sh; do
     # Redirecting output to a file to prevent it from being mixed with the output of the action
     # Using timeout command to enforce time limits
     if [[ "$OSTYPE" == "msys" ]]; then
-        # Windows minimal system (msys) detected. Launch a command prompt for better compatibility and timeout function
-        cmd.exe /C "timeout /T $((TEST_TIMEOUT * 1000)) /NOBREAK & ./$script > ${name}.log 2>&1"
+        # Windows minimal system (msys) detected. Launch a command prompt for better compatibility
+		# Todo: Add timeout logic to limit execution time. This will probably require a powershell script
+        cmd.exe /C "./$script > ${name}.log 2>&1"
     else
         # Other systems like Mac or Linux
         timeout $TEST_TIMEOUT ./$script > "${name}.log" 2>&1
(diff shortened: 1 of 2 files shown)


ci/tests/run-system-tests.sh
```diff
@@ -26,7 +26,14 @@ for script in ${NANO_SYSTEST_DIR}/*.sh; do
 
     # Redirecting output to a file to prevent it from being mixed with the output of the action
     # Using timeout command to enforce time limits
-    timeout $TEST_TIMEOUT ./$script > "${name}.log" 2>&1
+    if [[ "$OSTYPE" == "msys" ]]; then
+        # Windows minimal system (msys) detected. Launch a command prompt for better compatibility and timeout function
+        cmd.exe /C "timeout /T $((TEST_TIMEOUT * 1000)) /NOBREAK & ./$script > ${name}.log 2>&1"
```
dsiganos: Is the /NOBREAK switch necessary?

dsiganos: I don't think this does what we want.
This timeout behaves like a `sleep` command on linux.

dsiganos: What we want is a program to be executed with a ceiling on how long it can remain running.
For example, we start a program and if it does not complete in 5 minutes, then kill it.
That is what timeout does on linux.

dsiganos:
I suggest you get it working on windows without a timeout command first.
Maybe that's good enough since the timeout should not really be needed.
Then we can see if we can introduce a timeout command on windows.

RickiNano:
@dsiganos Yeah, I think you are right. I have removed the timeout part for now. 
I'll research if could use a more advanced powershell script to limit execution time on windows but that will be a separate PR.

pwojcikdev:
What's wrong with current timeout? From what I see it correctly timeouts the test script. It's delivery/handling of signals that's somehow not working on Windows.

dsiganos:
@pwojcikdev I don't know if timeout exists and works the same way on windows.
However, the timeout command is not absolutely necessary.
So I suggested we split the task into 2 parts:
* getting it working without a timeout restriction
* then adding a timeout restriction

pwojcikdev:
Okay, I don't think it's necessary, but if you want to go this route please work in a separate branch for now instead of pushing untested code to this PR. Running workflows without a timeout command will block runners for excessive amounts of time, and in turn this blocks running workflows for other PRs. 

For reference, I believe originally system tests were running on Linux only. I enabled running them on Windows some time ago and they're using bash environment. Therefore timeout command *should* be equivalent to other unix oses.

RickiNano:
I'll close this PR for now and see if I can come up with another approach

PR #4573, opened 2024-04-17
A system test to exercise the rpc stop command

dsiganos:


nano/rpc/rpc.cpp
```diff
@@ -50,6 +50,7 @@ void nano::rpc::start ()
 		logger.critical (nano::log::type::rpc, "Error while binding for RPC on port: {} ({})", endpoint.port (), ec.message ());
 		throw std::runtime_error (ec.message ());
 	}
+	logger.info (nano::log::type::rpc, "Nano node RPC listening address: {}", fmt::streamed (acceptor.local_endpoint ()));
```
pwojcikdev: In general it doesn't have to be "Nano node" because we have this separate `nano_rpc` which could also be running this listener, but well it's just a log.

pwojcikdev (dismissed):

nano/rpc/rpc.cpp
dsiganos: It's a good point, I'll remove the "Nano node " since it is a trivial fix.

pwojcikdev (approved):

clemahieu (approved):

PR #4575, opened 2024-04-18
Fix unit test bootstrap_processor.lazy_pruning_missing_block

dsiganos:


nano/core_test/bootstrap.cpp
```diff
@@ -1306,6 +1306,12 @@ TEST (bootstrap_processor, lazy_pruning_missing_block)
 	config.frontiers_confirmation = nano::frontiers_confirmation_mode::disabled;
 	config.enable_voting = false; // Remove after allowing pruned voting
 	nano::node_flags node_flags;
+	// It looks like force confirmed elections remain in the AEC and
+	// they keep publishing blocks, disabling the request loop stops this.
+	// A better fix would be to create a way to add and cement the blocks
+	// without involving an election at all or have a way to delete an
+	// election from the AEC.
+	node_flags.disable_request_loop = true;
```
pwojcikdev: This comment suggests that inserting and confirming blocks directly in the ledger `node->ledger.process` and `node->ledger.confirm` would be a better solution.

dsiganos: If we have a way to insert and confirm without creating an election then that might be a cleaner solution, if I understood what is happening well.
I was not aware of `nano::ledger::confirm`.
I will try it later.

clemahieu: For reference it was recently added here https://github.com/nanocurrency/nano-node/pull/4519

nano/secure/ledger.hpp
```diff
@@ -116,6 +116,8 @@ class ledger final
 	// Returns the next receivable entry equal or greater than 'key'
 	std::optional<std::pair<nano::pending_key, nano::pending_info>> receivable_lower_bound (secure::transaction const & tx, nano::account const & account, nano::block_hash const & hash) const;
 	void initialize (nano::generate_cache_flags const &);
+
+public: // for unit tests only
```
pwojcikdev: This is mainly used by confirming set, so the comment is inaccurate.

dsiganos: I mean making the member function public for use in unit tests only.
I will make it more explicit.

pwojcikdev: This is still less than ideal. A safe pattern similar to the way it's done in other classes should be used that provides a public `force_confirm` method with a proper release_assert that guards it.

pwojcikdev: Let's follow good patterns that are already established and work. A single assert is more descriptive than any comment.

pwojcikdev (approved):

PR #4579, opened 2024-04-20
Node configuration auto updater

RickiNano:
This pull request introduces a new command line feature `--update_config` , which updates the nodes current configuration by merging custom settings into the latest configuration template. Custom settings will be uncommented in the new configuration, while default values will be commented.
This feature simplifies the process for node operators to maintain an up-to-date configuration without manually merging the current configuration with the new one.

Features:
* Updates config file with new entries, tables and documentation
* Removes entries that are no longer referenced in code
* Preserves custom values from current configuration
* Formatted output

Limitations:
* Currently only works with node configuration file but could easily be extended to rpc an log configs

### DIFF
--- a/nano/core_test/toml.cpp
+++ b/nano/core_test/toml.cpp
@@ -1038,3 +1038,39 @@ TEST (toml, log_config_no_required)
 
 	ASSERT_FALSE (toml.get_error ()) << toml.get_error ().get_message ();
 }
+
+TEST (toml, merge_config_files)
+{
+	nano::network_params network_params{ nano::network_constants::active_network };
+	nano::tomlconfig default_toml;
+	nano::tomlconfig current_toml;
+	nano::tomlconfig merged_toml;
+	nano::daemon_config default_config{ ".", network_params };
+	nano::daemon_config current_config{ ".", network_params };
+
+	std::stringstream ss;
+
+	ss << R"toml(
+	[node]
+	 active_elections_size = 999
+	 # backlog_scan_batch_size = 7777
+	[node.bootstrap_ascending]
+	 block_wait_count = 33333
+	 old_entry = 34
+	)toml";
+
+	current_toml.read (ss);
+	current_config.deserialize_toml (current_toml);
+
+	current_config.serialize_toml (current_toml);
+	default_config.serialize_toml (default_toml);
+
+	auto merged_config = current_toml.merge_defaults (current_toml, default_toml);
+
+	ASSERT_TRUE (merged_config.find ("active_elections_size = 999") != std::string::npos);
+	ASSERT_FALSE (merged_config.find ("active_elections_size = 5000") != std::string::npos);
+	ASSERT_TRUE (merged_config.find ("# backlog_scan_batch_size = 10000") != std::string::npos);
+	ASSERT_FALSE (merged_config.find ("backlog_scan_batch_size = 7777") != std::string::npos);
+	ASSERT_TRUE (merged_config.find ("block_wait_count = 33333") != std::string::npos);
+	ASSERT_FALSE (merged_config.find ("old_entry") != std::string::npos);
+}
--- a/nano/core_test/toml.cpp
+++ b/nano/core_test/toml.cpp
@@ -1079,6 +1079,6 @@ TEST (toml, merge_config_files)
 	ASSERT_NE (merged_config.node.active_elections_size, default_config.node.active_elections_size);
 	ASSERT_EQ (merged_config.node.active_elections_size, 999);
 	ASSERT_NE (merged_config.node.backlog_scan_batch_size, 7777);
-	ASSERT_EQ (merged_config.node.active_elections_size, 999);
+	ASSERT_EQ (merged_config.node.bootstrap_ascending.block_wait_count, 33333);
 	ASSERT_TRUE (merged_config_string.find ("old_entry") == std::string::npos);
 }
--- a/nano/core_test/toml.cpp
+++ b/nano/core_test/toml.cpp
@@ -1047,6 +1047,7 @@ TEST (toml, merge_config_files)
 	nano::tomlconfig merged_toml;
 	nano::daemon_config default_config{ ".", network_params };
 	nano::daemon_config current_config{ ".", network_params };
+	nano::daemon_config merged_config{ ".", network_params };
 
 	std::stringstream ss;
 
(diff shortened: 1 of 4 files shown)


nano/node/cli.cpp
```diff
@@ -62,6 +62,7 @@ void nano::add_node_options (boost::program_options::options_description & descr
 	("migrate_database_lmdb_to_rocksdb", "Migrates LMDB database to RocksDB")
 	("diagnostics", "Run internal diagnostics")
 	("generate_config", boost::program_options::value<std::string> (), "Write configuration to stdout, populated with defaults suitable for this system. Pass the configuration type node, rpc or log. See also use_defaults.")
+	("update_config", boost::program_options::value<std::string> (), "Reads the current node configuration and updates it with missing keys and values and delete keys that are no longer used. Updated configuration is written to stdout.")
```
dsiganos: The command takes an argument but ignores it.

nano/node/cli.cpp
```diff
@@ -711,6 +712,31 @@ std::error_code nano::handle_node_options (boost::program_options::variables_map
...
+		nano::tomlconfig current_toml;
+		nano::daemon_config default_config{ data_path, network_params };
+		nano::daemon_config current_config{ data_path, network_params };
+
+		std::vector<std::string> config_overrides;
+		auto error = nano::read_node_config_toml (data_path, current_config, config_overrides);
+		if (error)
+		{
+			std::cerr << "Could not read existing config file\n";
+			ec = nano::error_cli::invalid_arguments;
```
dsiganos: I think the error `nano::error_cli::reading_config` is more appropriate here.

nano/lib/tomlconfig.cpp
```diff
@@ -186,6 +186,41 @@ void nano::tomlconfig::erase_default_values (tomlconfig & defaults_a)
 	erase_defaults (defaults_l.get_tree (), self.get_tree (), get_tree ());
 }
 
+// Merges two TOML configurations and commenting values that are identical
+std::string nano::tomlconfig::merge_defaults (nano::tomlconfig & current_config, nano::tomlconfig & default_config)
+{
+	auto default_tree = default_config.get_tree ();
```
dsiganos: Why are default_tree and current_tree needed?

nano/lib/tomlconfig.cpp
```diff
@@ -186,6 +186,41 @@ void nano::tomlconfig::erase_default_values (tomlconfig & defaults_a)
...
+	// Serialize both configs to commented strings
+	std::string defaults_str = default_config.to_string (true);
+	std::string current_str = current_config.to_string (true);
+
+	// Read both configs line by line
+	std::istringstream stream_defaults (defaults_str);
+	std::istringstream stream_current (current_str);
+	std::string line_defaults, line_current, result;
+
+	while (std::getline (stream_defaults, line_defaults) && std::getline (stream_current, line_current))
```
dsiganos, line 201: This loop will only work correctly if the defaults and current lists are sorted and have the same number of elements. Is this guranteed to happen? If yes, how do you know?

dsiganos:
The algorithm converts the structured data in plain text and then does the comparison.
That will work if the output string is always printed in a normalised way.
But in general, I would expect the comparison to be done while the data is held in structured format.

dsiganos:
A test case would be nice.

nano/node/cli.cpp
RickiNano: The intention was for future support for updating rpc and log configs too. However, they are quite small configs and they don't change as often the node config.
I'll remove it for now

nano/node/cli.cpp
RickiNano: Agreed. I'll change it

nano/lib/tomlconfig.cpp
RickiNano: You are right. It's not needed anymore. I'll change the assert to test the toml configs instead, and delete these lines

nano/lib/tomlconfig.cpp
RickiNano: The current list and default list will always follow the exact same layout. This is because we use the serialize_toml function to create the configs. That function will create all the tree nodes, config items, and docs based on the input

RickiNano:
> The algorithm converts the structured data in plain text and then does the comparison. That will work if the output string is always printed in a normalised way. But in general, I would expect the comparison to be done while the data is held in structured format.

Unfortunately the TOML library will only write all key/value lines as uncommented line entries. So even if we merge the TOML trees we still have to compare the final string output to the default string output to find what lines should be commented (default values)
It's simpler to directly convert both of them to string and compare them line by line, since we know that they always have the same structure

nano/lib/tomlconfig.cpp
dsiganos: You can either remove it or check that it is "node".

nano/node/cli.cpp
dsiganos: You can either remove it or make sure it is set to "node".

RickiNano:
@dsiganos I have added a unit test

nano/core_test/toml.cpp
```diff
@@ -1038,3 +1038,47 @@ TEST (toml, log_config_no_required)
 
 	ASSERT_FALSE (toml.get_error ()) << toml.get_error ().get_message ();
 }
+
```
dsiganos, line 1041: I think in this case, a systest would had been a better option.
But you have done it now...

nano/lib/tomlconfig.cpp
```diff
@@ -186,6 +186,36 @@ void nano::tomlconfig::erase_default_values (tomlconfig & defaults_a)
...
+	{
+		if (line_defaults == line_current)
+		{
+			// Use default value
+			result += line_defaults + "\n";
+		}
+		else
+		{
+			// Non default value. Removing the # to uncomment
```
dsiganos, line 210: This code assumes that the line is commented.
I do not understand why it is guaranteed that the line will be commented.

RickiNano: It is always commented because the tostring() function takes a boolean value that determines if all parameters should be commented or not:
```
	// Serialize both configs to commented strings
	std::string defaults_str = default_config.to_string (true);
	std::string current_str = current_config.to_string (true);
```

dsiganos: It is also assuming that there are characters after the '#' character.
How about you put an assert to ensure that pos will never be `npos` and an assert to check that the string length is larger than pos.
Let's make the assumptions explicit.

dsiganos (approved):

PR #4582, opened 2024-04-23
Fix a TSAN issue in active_transactions.confirm_new

dsiganos:


clemahieu (dismissed):

pwojcikdev:
Is this caused by a race condition when lazily calculating block hash? I'm not sure how many test have similar pattern, but seems like it could be quite a few.

dsiganos:
@pwojcikdev Yes, I assume that the caching of the hash is causing this TSAN warning.
Many tests will likely have this problem but I have not thought about we would solve all of them at once yet.
Using this as a test run to see if the problem in this case goes away.

PR #4588, opened 2024-04-26
Fix TSAN issue with test receivable_processor.send_with_receive

dsiganos:


pwojcikdev:
I'm able to easily identify at least 2 other call sites where this problem happens. It's not clear to me how this one-by-one approach tackles the root of the problem and makes future work easier.

PR #4593, opened 2024-04-30
Redesign the priority scheduler to maintain a 1-to-1 link with running elections

clemahieu:
This change appropriately prioritises elections by bucket and then by LRU account. Overfilled buckets have the lowest priority block election cancelled and rolled back.


qwahzi:
Removing from V27 milestone due to: https://github.com/nanocurrency/nano-node/pull/4626

PR #4604, opened 2024-05-06
Introduce `active_transactions_config`

gr0vity-dev:
Introduce `active_transactions_config` :

| Old config                                | New config                                   |
|-------------------------------------------|----------------------------------------------|
| active_elections_size                     | active_transactions.size                     |
| active_elections_hinted_limit_percentage | active_transactions.hinted_limit_percentage |
| active_elections_optimistic_limit_percentage | active_transactions.optimistic_limit_percentage |
| confirmation_history_size                 | active_transactions.confirmation_history_size|
|  /                 | active_transactions.confirmation_cache|

Newly introduced `confirmation_cache` is used in `recently_confirmed{ config.confirmation_cache },`  instead of the hardcoded value `65536`

### DIFF
--- a/nano/core_test/CMakeLists.txt
+++ b/nano/core_test/CMakeLists.txt
@@ -3,7 +3,7 @@ add_executable(
   entry.cpp
   fakes/websocket_client.hpp
   fakes/work_peer.hpp
-  active_transactions.cpp
+  active_elections.cpp
   async.cpp
   backlog.cpp
   block.cpp
--- a/nano/core_test/active_elections.cpp
+++ b/nano/core_test/active_elections.cpp
@@ -1,6 +1,6 @@
 #include <nano/lib/blocks.hpp>
 #include <nano/lib/jsonconfig.hpp>
-#include <nano/node/active_transactions.hpp>
+#include <nano/node/active_elections.hpp>
 #include <nano/node/confirming_set.hpp>
 #include <nano/node/election.hpp>
 #include <nano/node/scheduler/component.hpp>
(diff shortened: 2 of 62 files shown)


pwojcikdev (dismissed):
Nice cleanup. The question is whether we want to call this config `[nano.active_transactions]` to keep consistency with class naming (transaction name is a bit unfortunate in my opinion) or `[nano.active_elections]` to keep consistency with previous config naming.

gr0vity-dev:
I asked myself the same question. 
Since the config also includes something unrelated like `confirmation_history_size` I opted to keep keep consistency with class naming rather than previous naming. (It's a breaking change either way)
Maybe we should name the entire class `active_elections` because it's basically a container for all ongoing elections ?

pwojcikdev:
Renaming the class to `active_elections` sounds better and should be quite easy.

gr0vity-dev:
I renamed `active_transactions` to `active_elections`.  
So the update config looks like this :

| Old config                                | New config                                   |
|-------------------------------------------|----------------------------------------------|
| active_elections_size                     | active_elections.size                     |
| active_elections_hinted_limit_percentage | active_elections.hinted_limit_percentage |
| active_elections_optimistic_limit_percentage | active_elections.optimistic_limit_percentage |
| confirmation_history_size                 | active_elections.confirmation_history_size|
|  /                 | active_elections.confirmation_cache|

And it can be used in the `config-node.toml` like this:
```toml
[node.active_elections]
    size = 5000
    hinted_limit_percentage = 20
    optimistic_limit_percentage = 10
    confirmation_history_size = 2048
    confirmation_cache = 65536
```

clemahieu (dismissed):
LGTM. Nice cleanup and agree on the class rename.

pwojcikdev (dismissed):

PR #4607, opened 2024-05-07
Separate vote_router from active_transactions

clemahieu:
This change is to simplify the active_transactions class which has many responsibilities around election management. This change separates the responsibility of routing votes to their respective elections from other responsibilities inside active_transactions.

The mapping hash -> election in nano::active_transactions::blocks is moved into a new class nano::vote_router. Vote processing logic is moved from nano::active_transactions::vote to nano::vote_router::vote.

nano::vote_router changes the behavior from active_transactions::blocks in that it does not own the lifetime of elections. Only a weak_ptr to election is held and a background thread cleans up orphaned routes periodically.

nano::vote_code and nano::vote_source are moved into the vote_router.hpp file and forward declared when possible.

### DIFF
--- a/nano/node/vote_router.cpp
+++ b/nano/node/vote_router.cpp
@@ -1,0 +1,187 @@
+#include <nano/lib/enum_util.hpp>
+#include <nano/lib/thread_roles.hpp>
+#include <nano/lib/utility.hpp>
+#include <nano/node/active_elections.hpp>
+#include <nano/node/election.hpp>
+#include <nano/node/vote_cache.hpp>
+#include <nano/node/vote_router.hpp>
+
+#include <chrono>
+
+using namespace std::chrono_literals;
+
+nano::stat::detail nano::to_stat_detail (nano::vote_code code)
+{
+	return nano::enum_util::cast<nano::stat::detail> (code);
+}
+
+nano::stat::detail nano::to_stat_detail (nano::vote_source source)
+{
+	return nano::enum_util::cast<nano::stat::detail> (source);
+}
+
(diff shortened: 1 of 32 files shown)


nano/node/vote_router.cpp
```diff
@@ -0,0 +1,190 @@
...
+	std::unique_lock lock{ mutex };
+	elections.insert_or_assign (hash, election);
+}
+
+void nano::vote_router::erase (nano::election const & election)
+{
+	std::unique_lock lock{ mutex };
+	for (auto const & [hash, _] : election.blocks ())
+	{
+		elections.erase (hash);
```
pwojcikdev, line 49: Election can only hold a limited number of blocks and can drop blocks during its lifetime, so this erase function is not robust.

pwojcikdev (commented):
This is really cool. I can already see how this can be improved further with read/write locks to allow for faster multithreaded vote processing (no need for an exclusive mutex when we're not modifying the mapping). I struggled to get it done in the past when election routing was part of the AEC, but this makes it much easier.

nano/node/vote_router.cpp
pwojcikdev, line 46: Election object can hold a limited number of blocks and some can be dropped during its lifetime. While the cleanup thread should still clean up the remaining hashes, at first sight this might be confusing.

pwojcikdev (dismissed):

nano/node/vote_router.cpp
clemahieu: Not sure if this directly helps but I renamed these to connect/disconnect to represent the looser ownership nature.

PR #4611, opened 2024-05-10
Remove active_elections::trim

clemahieu:
This change removes the active_elections::trim function which is problematic with more advanced scheduler logic.

The code is not hit during normal operation because each scheduler already checks its own internal limits before starting elections.


nano/node/json_handler.cpp
```diff
@@ -2046,7 +2050,7 @@ void nano::json_handler::election_statistics ()
 	std::stringstream stream_utilization;
 	stream_utilization << std::fixed << std::setprecision (2) << utilization_percentage;
 
-	response_l.put ("normal", normal_count);
+	response_l.put ("priority", priority_count);
```
pwojcikdev, line 2054: Missing manual count entry

pwojcikdev (dismissed):

PR #4612, opened 2024-05-10
Add ability for channels to pause/resume processing of incoming messages

clemahieu:
This is a follow on to PR https://github.com/nanocurrency/nano-node/pull/4609 and responds to the fair queue for a channel being full by pausing incoming message processing for the channel.

Pausing incoming message handling will cause TCP flow control to back up to the sender and limit incoming data.


pwojcikdev:
I don't think it works right now, I added a new core test to ensure proper behavior. Throttling is a bit tricky with the current networking architecture. Depending on how critical it is to get it working right now, this will probably be much easier to implement with networking code implemented using coroutines, using primitives like asynchronous conditions.

PR #4615, opened 2024-05-12
Fix Test `request_aggregator.one_update`

gr0vity-dev:
I noticed that  `request_aggregator.one_update` is failing intermittently (1/100 cases) because `node.stats.count(nano::stat::type::requests, nano::stat::detail::requests_generated_votes))` is 2 instead of expected 1. However it will eventually always be 2.

I changed the test-case accordingly, but maybe there is un underlying issue if `requests_generated_votes` should be 1 rather than 2.

I also noticed that `node.aggregator.empty ()`is not guaranteed to be true immediately in about 1/2000 cases for me.

### DIFF
--- a/nano/core_test/request_aggregator.cpp
+++ b/nano/core_test/request_aggregator.cpp
@@ -111,7 +111,7 @@ TEST (request_aggregator, one_update)
 	node.aggregator.request (request, dummy_channel);
 	// In the ledger but no vote generated yet
 	ASSERT_TIMELY (3s, 0 < node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_votes))
-	ASSERT_TRUE (node.aggregator.empty ());
+	ASSERT_TIMELY (3s, node.aggregator.empty ());
 	ASSERT_TIMELY_EQ (3s, 2, node.stats.count (nano::stat::type::aggregator, nano::stat::detail::aggregator_accepted));
 	ASSERT_TIMELY_EQ (3s, 0, node.stats.count (nano::stat::type::aggregator, nano::stat::detail::aggregator_dropped));
 	ASSERT_TIMELY_EQ (3s, 0, node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_unknown));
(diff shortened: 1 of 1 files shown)


nano/core_test/request_aggregator.cpp
```diff
@@ -111,17 +111,17 @@ TEST (request_aggregator, one_update)
...
 	ASSERT_TIMELY (3s, 0 < node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_votes))
-	ASSERT_TRUE (node.aggregator.empty ());
+	ASSERT_TIMELY (3s, node.aggregator.empty ());
 	ASSERT_TIMELY_EQ (3s, 2, node.stats.count (nano::stat::type::aggregator, nano::stat::detail::aggregator_accepted));
 	ASSERT_TIMELY_EQ (3s, 0, node.stats.count (nano::stat::type::aggregator, nano::stat::detail::aggregator_dropped));
 	ASSERT_TIMELY_EQ (3s, 0, node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_unknown));
 	ASSERT_TIMELY_EQ (3s, 2, node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_hashes));
 	size_t count = 0;
-	ASSERT_TIMELY_EQ (3s, 1, (count = node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_votes)));
+	ASSERT_TIMELY_EQ (3s, 2, (count = node.stats.count (nano::stat::type::requests, nano::stat::detail::requests_generated_votes)));
```
pwojcikdev, line 120: This should probably check for greater or equal to one. Not sure how exactly it's implemented in vote generator, but depending on timing, two hashes could be bundled into one vote.

gr0vity-dev: In this testcase, it always returns 2 but I can change it to >= 1.
Should we have another testcase that explicitly tests this bundling ?

pwojcikdev: I think it's tested somewhere. It might be always 2 just because the load is light, not sure.

gr0vity-dev: I've seen you've fixed this test already in a recent commit. I'll close this one

pwojcikdev: I think I only fixed one check, so this PR might still be relevant.

gr0vity-dev: created a new one without conflicts that addresses the remaining issue: https://github.com/nanocurrency/nano-node/pull/4622

PR #4623, opened 2024-05-14
Add per-bucket election limiter

clemahieu:
This change is a simpler version of the priority scheduler per-bucket limiter that tracks the number of elections it's started while retaining the ability to queue blocks for activation once elections already started for the bucket complete.

This removes the priority scheduler thread. Elections are activated at time of call to ::activate or when a different election is being cleaned up.


nano/node/scheduler/buckets.cpp
```diff
@@ -62,8 +62,10 @@ nano::scheduler::buckets::~buckets ()
 
 auto nano::scheduler::buckets::bucket (nano::uint128_t const & balance) const -> scheduler::bucket &
 {
-	auto index = std::upper_bound (minimums.begin (), minimums.end (), balance) - minimums.begin () - 1;
-	return *buckets_m[index];
+	auto iter = buckets_m.upper_bound (balance);
+	--iter; // Iterator points to bucket after the target priority
```
clemahieu: This should be on the previous line.

PR #4627, opened 2024-05-17
Address performance regression in vote_cache::top

clemahieu:
Revert https://github.com/nanocurrency/nano-node/pull/4518/commits/c734caf4d1ad840df83983e40b86e5d2c1f56a69 which removed loop short-circuit checking for top voted-for elections.
Eliminate duplicate expression evaluation.
Use deque rather than vector for results to reduce backing reallocation.

### DIFF
--- a/nano/node/vote_cache.cpp
+++ b/nano/node/vote_cache.cpp
@@ -231,11 +231,11 @@ void nano::vote_cache::clear ()
 	cache.clear ();
 }
 
-std::vector<nano::vote_cache::top_entry> nano::vote_cache::top (const nano::uint128_t & min_tally)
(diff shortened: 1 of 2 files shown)


pwojcikdev (approved):
I can't remember why I removed the by tally index, seems to make no sense.

EDIT:
Okay, so `tally ()` is not a constant time function:
```
return std::accumulate (voters.begin (), voters.end (), nano::uint128_t{ 0 }, [] (auto sum, auto const & item) {
   return sum + item.weight;
});
```
Did you measure that the additional overhead on each insertion is worth the speedup during lookup? In any case, it's possible to implement some kind of smart tally caching to alleviate this issue and have both fast insertion and top lookup.

PR #4638, opened 2024-05-23
Add flamegraph generation in CI build

clemahieu:
Flamegraphs are attached to the workflow build as artifacts.

New tests can be added by creating a new gtest in slow_test.
Tests like TEST (flamegraph, testname_x) will be executed if testname_x is added to the flamegraph.yml file test matrix.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -56,7 +56,6 @@ pushd $BUILD_DIR
 
 cmake \
 -DCMAKE_BUILD_TYPE=${BUILD_TYPE:-"Debug"} \
--DCMAKE_CXX_FLAGS="-g -Og" \
 -DPORTABLE=ON \
 -DACTIVE_NETWORK=nano_${NANO_NETWORK:-"live"}_network \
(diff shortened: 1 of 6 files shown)


ci/build.sh
```diff
@@ -56,6 +56,7 @@ pushd $BUILD_DIR
 
 cmake \
 -DCMAKE_BUILD_TYPE=${BUILD_TYPE:-"Debug"} \
+-DCMAKE_CXX_FLAGS="-g -Og" \
```
pwojcikdev: Does this need to be enabled globally for all builds? In my experience with flamegraphs, it was enough to set build type to `RelWithDebInfo`. Do github runners need a workaround?

clemahieu: While getting it to work I was having issues with no symbols, though that was also before I set RelWithDebInfo in the flamegraph tests themselves.

Looks like these aren't needed.

pwojcikdev (approved):
Looks good.

PR #4645, opened 2024-05-31
Fix rocksdb migration

RickiNano:
Database migrations are currently failing on RocksDb because the rep_weights table does not exist. This PR creates the table if it does not exist.
It also fixes the wrong log types in the same migration


gr0vity-dev:
This makes my local testsuite run fine again when using rocksdb!

PR #4647, opened 2024-06-13
migrate_database_lmdb_to_rocksdb improvements

RickiNano:
The `migrate_database_lmdb_to_rocksdb` option is running for a very long time with the current ledger size (almost 200 million blocks). On my local machine it took 65 minutes to complete. Nothing is written on screen during this process and users may think the process has stalled.
This PR adds some progress feedback. One update for each of the 7 tables that are migrated.
It also adds a simple disk space check to warn users if they might not have enough space to complete the migration.
~~The current converted RocksDb database is 73 GB, and the warning is given if the system has less than 75GB available.~~
The warning is given based on the size of the LMDB database that is being migrated. The final RocksDb size is approximately 65% of the LMDB space.

### DIFF
--- a/nano/secure/ledger.cpp
+++ b/nano/secure/ledger.cpp
@@ -1265,7 +1265,7 @@ bool nano::ledger::migrate_lmdb_to_rocksdb (std::filesystem::path const & data_p
 	// Open rocksdb database
 	nano::rocksdb_config rocksdb_config;
 	rocksdb_config.enable = true;
-	rocksdb_config.memory_multiplier = std::numeric_limits<uint8_t>::max ();
+	//rocksdb_config.memory_multiplier = 4;
 	auto rocksdb_store = nano::make_store (logger, data_path_a, nano::dev::constants, false, true, rocksdb_config);
 
 	if (!rocksdb_store->init_error ())
@@ -1273,7 +1273,7 @@ bool nano::ledger::migrate_lmdb_to_rocksdb (std::filesystem::path const & data_p
 		auto table_size = store.count (store.tx_begin_read (), tables::blocks);
 		logger.info (nano::log::type::ledger, "Step 1 of 7: Converting {} million entries from blocks table", table_size / 1000000);
 		std::atomic<std::size_t> count = 0;
-		auto refresh_interval = 20ms;
+		auto refresh_interval = 100ms;
 		store.block.for_each_par (
 		[&] (store::read_transaction const & /*unused*/, auto i, auto n) {
 			auto rocksdb_transaction (rocksdb_store->tx_begin_write ({}, { nano::tables::blocks }));
--- a/nano/secure/ledger.cpp
+++ b/nano/secure/ledger.cpp
@@ -1241,13 +1241,18 @@ uint64_t nano::ledger::pruning_action (secure::write_transaction & transaction_a
 // A precondition is that the store is an LMDB store
 bool nano::ledger::migrate_lmdb_to_rocksdb (std::filesystem::path const & data_path_a) const
 {
+	nano::logger logger;
+	nano::logger::initialize (nano::log_config::daemon_default (), data_path_a);
+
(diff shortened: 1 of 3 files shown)


pwojcikdev:
This should be using node logger not cout, otherwise logs aren't saved to disk. There are some places in the existing code that are using cout too, but those are leftovers. Also, printing progress every x converted entries would be nice.

RickiNano:
I have been unable to get node logger implemented. I don't think it's important since the migration is a one time process.
I've updated the code with more granular updates to indicate that progress is being made. 
Here is the output of a full production lmdb to rocksDb migration:

image

pwojcikdev:
> I have been unable to get node logger implemented. I don't think it's important since the migration is a one time process.


And that's exactly the reason why it should be logged well, in line with production ready code. In case something goes wrong there should be a persistent record.

RickiNano:
@pwojcikdev 
I got the Nano logger implemented now. 
Instead of outputting dots for each step it is now giving actual numbers to the log.
image

nano/nano_node/entry.cpp
```diff
@@ -55,13 +55,23 @@ class address_library_pair
...
 }
 
 int main (int argc, char * const * argv)
 {
 	nano::set_umask (); // Make sure the process umask is set before any files are created
 	nano::logger::initialize (nano::log_config::cli_default ());
 
+	nano::set_file_descriptor_limit (OPEN_FILE_DESCRIPTORS_LIMIT);
+	auto const file_descriptor_limit = nano::get_file_descriptor_limit ();
+	nano::default_logger ().info (nano::log::type::daemon, "File descriptors limit: {}", file_descriptor_limit);
```
pwojcikdev: This shouldn't be moved, this won't log anything.

nano/secure/ledger.cpp
```diff
@@ -1242,18 +1242,20 @@ bool nano::ledger::migrate_lmdb_to_rocksdb (std::filesystem::path const & data_p
...
 	nano::rocksdb_config rocksdb_config;
 	rocksdb_config.enable = true;
+	rocksdb_config.memory_multiplier = std::numeric_limits<uint8_t>::max ();
 	auto rocksdb_store = nano::make_store (logger, data_path_a, nano::dev::constants, false, true, rocksdb_config);
 
 	if (!rocksdb_store->init_error ())
 	{
 		logger.info (nano::log::type::ledger, "Step 1 of 7: Converting blocks table");
 		std::atomic<std::size_t> count = 0;
+		auto refresh_interval = 20ms;
```
pwojcikdev: Why such a short refresh interval?

nano/nano_node/entry.cpp
clemahieu: We could split setting the file descriptor limit with logging the warning about it being too low, though setting the file descriptor limit was moved here since it was only set when called with "--daemon" before and it needed to be set at least for "--migrate_database_lmdb_to_rocksdb" if not other commands when used with rocksdb when it has many files.

pwojcikdev: Setting the limit here is not a problem, it's a good place for it. The problem is trying to log when logging is still in CLI mode. CLI commands usually have a very specific output format and to avoid polluting it with node details, only critical errors are logged. The warning probably needs to be logged via cerr, and the `"File descriptors limit: {}"...` line moved back into the daemon code, so it leaves a trace in log files if we ever need to debug it. 
 
```
nano::log_config nano::log_config::cli_default ()
{
	log_config config{};
	config.default_level = nano::log::level::critical;
	config.console.colors = false; // to avoid printing warning about cerr and colors
	config.console.to_cerr = true; // Use cerr to avoid interference with CLI output that goes to stdout
	config.file.enable = false;
	return config;
}
```

pwojcikdev: I pushed a commit which should fix the logging. Also, as setting the limit was moved away from daemon class, targets other than cli wallet would no longer set it.

PR #4675, opened 2024-07-08
Add database information to telemetry data

RickiNano:
Adds telemetry information about the type of database backend the node is using.
The telemetry RPC function reports both individual database backend information or consolidated (average). The reported database backend for each node will be either 'RocksDb' , 'LMDB' or "Unknown". 
'Unknown' is for nodes running v26.1 or earlier that does not send any database information in telemetry data.

### DIFF
--- a/nano/core_test/message.cpp
+++ b/nano/core_test/message.cpp
@@ -389,7 +389,7 @@ TEST (message, confirm_req_hash_batch_serialization_v2)
 /**
  * Test that a confirm_ack can encode an empty hash set
  */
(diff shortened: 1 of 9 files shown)


pwojcikdev:
I added a testcase that ensures proper telemetry serialization, which currently fails. It's a bit concerning that existing tests don't ensure that telemetry is correctly transmitted between nodes.

I also noticed that we're missing testcases for serialization of `telemetry_ack` & `telemetry_req` messages, which might need some additional modifications to make this work correctly, not sure.

PR #4683, opened 2024-07-17
Enable large votes

clemahieu:
This bumps the protocol version and also adds the "--disable_large_votes" command line option to disable large votes.


nano/node/network.hpp
```diff
@@ -171,7 +171,7 @@ class network final
 	static std::size_t const buffer_size = 512;
 
 	static std::size_t const confirm_req_hashes_max = 7;
```
pwojcikdev: This also needs to be set to 255

pwojcikdev (approved):

Issue #4702, opened 2024-08-13
Make bootstrap weights file plaintext


clemahieu:
Currently this file in binary but there's no need for it to be binary and a text-based file would make it easier for manual inspection and diffing.

PR #4703, opened 2024-08-14
Bootstrap weights in plain text

RickiNano:
Replaces the binary file with embedded bootstrap weight in plain text.
The weight values and max block counts in this PR are from 14-08-2024


pwojcikdev:
We also need the script you used to generate `bootstrap_weights_live.hpp` and `bootstrap_weights_beta.hpp`. It would be better to place the globals in a `namespace nano::weights` and a separate `weights` library.

RickiNano:
@pwojcikdev 
I have added the script for generating the bootstrap weights data.
I tried updating the namespace but I get some weird build errors. Probably just me lacking the knowledge to do this correctly. Could you change it for me?

RickiNano:
I think I got the namespace change working now

RickiNano:
Rebased on top of develop 25-10-2024

pwojcikdev:
The biggest issue I can see with this is that it introduces `create_weights_file` node option, which requires node shutdown and restart to generate weights. This might not be that big deal, but why do this at all when there is already a nice `record_rep_weights.py` script that can pull weights from any node via rpc, so it doesn't require full ssh access or node restarts.

RickiNano:
@pwojcikdev 
The `create_weights_file` option is meant to be used by us developers just once for each node release to update the embedded weights before the node version is shipped. This can be done on any node that is up to date.
It addresses issue #4702 that is part of the V28 milestones
I wouldn't mind taking the responsibility of updating the weights for both live and beta before releases.

pwojcikdev:
Okay, but why change this at all when working and maintaining the python file was strictly easier and more ergonomic?

RickiNano:
I can look into converting the existing python script so that it creates the weights as text files. I'm not a Python guy though.

RickiNano:
I have removed the `create_weights_file` option and modified the python script. Turned out that it was super easy to do even though I had no experience with Python.
The python script now takes a network name as input and creates the complete `bootstrap_weights_live.hpp` or `bootstrap_weights_beta.hpp` file that can overwrite the existing file.
I also updated the current `bootstrap_weights_live.hpp` and `bootstrap_weights_beta.hpp` with the python created ones.

nano/node/bootstrap_weights_live.hpp
```diff
@@ -0,0 +1,148 @@
+#pragma once
+
+#include <string>
+#include <vector>
+namespace nano::weights
```
pwojcikdev, line 5: Just my ocd, but a newline between these would be nice

pwojcikdev (dismissed):
LGTM, thanks for modifying the script, this will make maintenance easier in the future

PR #4719, opened 2024-09-03
Fix correctness, sanity check, and resolve shutdown issue.

clemahieu:
There was a missing mutex lock entries_mutex in unchecked_map::collect_container_info.

Sanity check the equality of the iterator qualified root to the item searched for.

Consider stopped flag when wait-looping. Use condition_variable::wait rather than this_thread::sleep_for

### DIFF
--- a/nano/node/confirming_set.cpp
+++ b/nano/node/confirming_set.cpp
@@ -132,11 +132,15 @@ void nano::confirming_set::run_batch (std::unique_lock<std::mutex> & lock)
 		notification.cemented.swap (cemented);
 		notification.already_cemented.swap (already);
 
-		// Wait for the worker thread if too many notifications are queued
+		std::unique_lock lock{ mutex };
 		while (notification_workers.num_queued_tasks () >= config.max_queued_notifications)
 		{
 			stats.inc (nano::stat::type::confirming_set, nano::stat::detail::cooldown);
-			std::this_thread::sleep_for (100ms);
+			condition.wait_for (lock, 100ms, [this] { return stopped; });
+			if (stopped)
+			{
+				return;
+			}
 		}
 
 		notification_workers.push_task ([this, notification = std::move (notification)] () {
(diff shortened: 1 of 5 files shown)


nano/node/confirming_set.cpp
```diff
@@ -133,10 +133,14 @@ void nano::confirming_set::run_batch (std::unique_lock<std::mutex> & lock)
 		notification.already_cemented.swap (already);
 
 		// Wait for the worker thread if too many notifications are queued
-		while (notification_workers.num_queued_tasks () >= config.max_queued_notifications)
+		std::unique_lock lock{ mutex };
+		bool stopped_l;
+		condition.wait (lock, [this, &stopped_l] {
+			return (stopped_l = stopped) || notification_workers.num_queued_tasks () < config.max_queued_notifications;
```
pwojcikdev: If this is using condition variables now, then condition needs to be notified after each processed entry

clemahieu: What do you think of using the wait_for variant? That would emulate previous behavior if we're worried about excessive notifications.

pwojcikdev: I think the impact of additional notifications will be zero, this is bound by IO anyway.

clemahieu: thread_pool doesn't have a notification when the number of queue tasks changes so I'll leave this to poll using wait_for instead of sleep.

pwojcikdev: Is this the same? Won't `condition.wait_for (lock, 100ms,` wait for 100ms and then continue regardless of whether the predicate is fulfilled?

pwojcikdev: In your code the `stats.inc (nano::stat::type::confirming_set, nano::stat::detail::cooldown);` is lost, which is necessary to properly understand live node bottlenecks. Code below should be most similar to the previous logic:

```
while (notification_workers.num_queued_tasks () >= config.max_queued_notifications)
 {
 	stats.inc (nano::stat::type::confirming_set, nano::stat::detail::cooldown);
 	condition.wait_for(..., 100ms, [...] { return stopped; }
 	if (stopped) { return; }
 }
 ```

pwojcikdev (approved):

PR #4741, opened 2024-10-01
Make max votes static again

RickiNano:
Make `confirm_req_hashes_max` and `confirm_ack_hashes_max` static and set to 255. 
Removed `disable_large_votes` flag


pwojcikdev:
This should not be static in network class but part of protocol constants.

RickiNano:
In what file are the protocol constants stored?

PR #4744, opened 2024-10-06
QT wallet: proper display of state open block in GUI history

SergiySW:
State open block should be marked as "Receive".
For open blocks `previous` == 0 and function `nano::ledger_set_any::block_balance ()` will return `std::nullopt` resulting in displaying "Unknown" block type.


nano/qt/qt.cpp
```diff
@@ -585,7 +585,7 @@ class short_text_visitor : public nano::block_visitor
 	{
 		auto balance (block_a.hashables.balance.number ());
 		auto previous_balance = ledger.any.block_balance (transaction, block_a.hashables.previous);
-		if (!previous_balance)
+		if (!block_a.hashables.previous.is_zero () && !previous_balance)
```
clemahieu: This solution doesn't look right as the next else-if dereferences .value().

This shows a gap in our testing for a pruned send that would hit this branch.

We should add a test for this missed case and also the main thing we're fixing in addition to a correct solution.

SergiySW: As state block sideband contains is_send, is_receive, is_epoch data, it can be done. But without amount unfortunantely. Actually unknown subtype should just never exists then.
To add amount here it would require breaking change for pruned table to include pruned block balance. Pruned usage is probably quite low so breaking change is a good option.
Same would probably be required for RPC history.

pwojcikdev:
Thanks for submitting this. Merged and will be included in V28.

PR #4754, opened 2024-10-15
Automatic database backend

RickiNano:
Switching database backend between LMDB and RocksDb is currently done by setting `RocksDb.enabled` to true or false in the config.

This PR removes that setting and adds a new `Database_backend` config setting.
`Database_backend` can be set to one of the following:
* lmdb
* rocksdb
* auto

The first two will force the use of the specific backend.
When using auto mode the node will look for an existing ledger and use this. If no ledger is found it will initialize a new ~~RocksDb~~ LMDB ledger. If multiple ledgers are found it will display a warning message and use the RocksDb ledger.
The main purpose is to make new node operators use RocksDb but without forcing existing LMDB users to make changes.
In future node versions (V29?) we can show messages to LMDB users that encourage them to update to RocksDb and instructions on how to do this.
Even later versions (V30?) could halt the node initialization if LMDB is used and display instructions on updating. If they insist on using LMDB they can set `Database_backend` to lmdb

### DIFF
--- a/nano/core_test/toml.cpp
+++ b/nano/core_test/toml.cpp
@@ -578,7 +578,6 @@ TEST (toml, daemon_config_deserialize_no_defaults)
 	vacancy_threshold = 99
 
 	[node.rocksdb]
-	enable = true
 	io_threads = 99
 	read_cache = 99
 	write_cache = 99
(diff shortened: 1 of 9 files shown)


RickiNano:
Closing this PR as it has been superseded by #4828

PR #4759, opened 2024-10-20
Convert store iterators to variants rather than using polymorphism.

clemahieu:
This converts the store iterators from version that use polymorphism to using std::variant.

The primary drivers for this change is to simplify and allow more complex iterators to be created where this was previously fairly difficult to do.

The iterators are separated in the 3 parts:
- The base database iterators LMDB/RocksDB that use the library's native data type, MDB_val/Slice.
- A variant database iterator that can hold any of the raw database iterators and adapts the native database type to std::span<uint8_t>
- A strong-typed database iterator that mirrors existing iterator behavior with strong typing for key/value pairs.

The iterators are bi-directional with a sentinel value to signify end of iteration. The iterators are also circular so after reaching the end-sentinel value, further incrementing will wrap around to the beginning, or decrementing will position the iterator at the last key/value.

Using std::variant also eliminates heap allocations when constructing iterators which can improve performance in some circumstances.


nano/store/iterator.cpp
```diff
@@ -1 +1,104 @@
...
+	std::visit ([&] (auto && arg) {
+		using T = std::remove_cvref_t<decltype (arg)>;
+		if constexpr (std::is_same_v<T, lmdb::iterator>)
+		{
+			if (!arg.is_end ())
+			{
+				auto & current = *arg;
+				std::span<uint8_t const> key{ reinterpret_cast<uint8_t const *> (current.first.mv_data), current.first.mv_size };
+				std::span<uint8_t const> value{ reinterpret_cast<uint8_t const *> (current.second.mv_data), current.second.mv_size };
+				this->current = std::make_pair (key, value);
```
pwojcikdev: It seems wrong to include this code here, similar to all the other methods this should be a simple:
```
std::visit ([] (auto && arg) {
 	this->current = arg.current ();
 },
 ```
 With the database specific code hidden in a particular backend implementation.

Comment (clemahieu) [reply nano/store/iterator.cpp]: I added a span() function on the concrete iterators. Internally these want to hold their native pointers for iteration e.g. MDB_val and rocksdb::Slice but this way the database-specific part is in the database-specific iterator.

**Comment by clemahieu** [comment]
I checked on using std::reverse_iterator rather than implementing our own though it looks like it only supports copyable iterators. https://en.cppreference.com/w/cpp/iterator/reverse_iterator/reverse_iterator

File: nano/store/rocksdb/iterator.cpp [hunk 2]
```diff
@@ -0,0 +1,146 @@
+#include "rocksdb/iterator.h"
```
pwojcikdev: #include  <...> instead of "..." for consistency

nano/store/rocksdb/utility.cpp
```diff
@@ -0,0 +1,11 @@
+#include <nano/store/rocksdb/transaction_impl.hpp>
+#include <nano/store/rocksdb/utility.hpp>
+
+auto nano::store::rocksdb::tx (store::transaction const & transaction_a) -> std::variant<::rocksdb::Transaction *, ::rocksdb::ReadOptions *>
+{
+	if (dynamic_cast<nano::store::read_transaction const *> (&transaction_a) != nullptr)
+	{
+		return static_cast<::rocksdb::ReadOptions *> (transaction_a.get_handle ());
```
pwojcikdev, line 8: Is it possible for the handle to store the pointer as `std::any` and use type safe cast `std::any_cast`? Or is that not necessary here

nano/store/typed_iterator_templ.hpp
```diff
@@ -0,0 +1,88 @@
...
+namespace nano::store
+{
+template <typename Key, typename Value>
+void typed_iterator<Key, Value>::update ()
+{
+	if (!iter.is_end ())
+	{
+		// FIXME Don't convert via lmdb::db_val, this is just a placeholder
+		auto const & data = *iter;
+		lmdb::db_val key_val{ MDB_val{ data.first.size (), const_cast<void *> (reinterpret_cast<void const *> (data.first.data ())) } };
```
pwojcikdev, line 15: Well, if it works it works 😆

pwojcikdev (dismissed):
Really, really good. Already makes working with database backends simpler.

nano/store/rocksdb/utility.cpp
clemahieu: Yes, I think applying a similar change to transactions would be another good improvement. I'll leave that to another PR though.

PR #4766, opened 2024-10-24
Optimize vote distribution between PRs and non-PRs

gr0vity-dev:
Currently, PRs receive votes both from flood_vote and flood_vote_pr, leading to duplicate voting, while fewer non-PR receive votes. This is particularly noticeable in small test networks.

This change ensures votes are distributed more efficiently:
- PRs receive votes only through flood_vote_pr
- Non-PRs receive votes through flood_vote_non_pr
- Overall outgoing vote load remains unchanged

Here is the difference seen in a local test network:

Screenshot 2024-10-24 at 20 38 12

I included a small optimization to `nano::network::list_non_pr` since it's called often.

```
Test Configuration & Benchmark Results:
PR Count: 100
Non-PR Count: 200
Iterations: 10000

--------------------------------------------------
Current method   - Duration: -- 521 ms
Partition method - Duration: -- 354 ms
--------------------------------------------------

Target count per iteration: 29 [ceil( 2* sqrt(200))]
Actual Current method:      29
Actual Partition method:    29
```

### DIFF
--- a/nano/node/network.cpp
+++ b/nano/node/network.cpp
@@ -386,11 +386,13 @@ std::deque<std::shared_ptr<nano::transport::channel>> nano::network::list_non_pr
 {
 	std::deque<std::shared_ptr<nano::transport::channel>> result;
 	tcp_channels.list (result);
+
+	auto partition_point = std::partition (result.begin (), result.end (),
+	[this] (std::shared_ptr<nano::transport::channel> const & channel) {
+		return !node.rep_crawler.is_pr (channel);
+	});
+	result.resize (std::distance (result.begin (), partition_point));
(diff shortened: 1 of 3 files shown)


pwojcikdev (approved):

pwojcikdev:
Looks good. Do I understand correctly that before votes were being sent twice to PRs?

gr0vity-dev:
> Looks good. Do I understand correctly that before votes were being sent twice to PRs?

Not systematically but they were sent:
- to all PRs in `flood_vote_pr`
- to 2 * sqrt of all peers (including PRs) in `flood_vote`

This means that some PRs might get the same votes twice

PR #4771, opened 2024-10-27
Link boost stacktrace appropriately for line information.

clemahieu:
The previous code was overcomplicating a relatively straightforward requirement to link against librt and libdynamic to get symbolic stacktraces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -122,18 +122,6 @@ set(NANO_ROCKSDB_TOOLS
     OFF
     CACHE BOOL "")
 
-option(NANO_STACKTRACE_BACKTRACE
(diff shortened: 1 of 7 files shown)


pwojcikdev:
Don't we need `add_definitions(-DBOOST_STACKTRACE_USE_BACKTRACE)` otherwise a default mode will be used?

pwojcikdev:
I was confused why github runners were having problems , but I noticed that in failed run annotations there is this:
```
Unhandled exception. System.IO.IOException: No space left on device : '/home/runner/runners/2.321.0/_diag/Worker_20250128-155555-utc.log'...
```
Perhaps the the additional stacktrace processing code pushed our build directory over github limits...

PR #4776, opened 2024-10-29
Compile time autumn2024

clemahieu:
This is a series of header cleanups aimed at improving compile time.

Templates for hashing numbers was moved to numbers_templ.hpp

A forward declaration header for nano::stream was added to stream_fwd.hpp

Total compile time including submodules was reduced from 10m to 9m, about a 10% improvement overall and likely more significant for incremental recompiles.


nano/node/confirming_set.hpp
```diff
@@ -20,6 +21,14 @@
 
 namespace mi = boost::multi_index;
 
+namespace nano
+{
+class election;
+class ledger;
+class logger;
+class stats;
```
pwojcikdev: This could be in fwd header to avoid constantly repeating the same class names, it's just noise

pwojcikdev (dismissed):

nano/node/confirming_set.hpp
clemahieu: Good point. Moved all of these forward declarations plus some other opportunistic ones in to the fwd.hpp files you started.

PR #4778, opened 2024-11-02
Exit migration if existing rocksdb ledger found

RickiNano:
Halt migration if rocksdb folder already exists instead of deleting it.

### DIFF
--- a/nano/secure/ledger.cpp
+++ b/nano/secure/ledger.cpp
@@ -1280,7 +1280,12 @@ bool nano::ledger::migrate_lmdb_to_rocksdb (std::filesystem::path const & data_p
 	boost::system::error_code error_chmod;
 	nano::set_secure_perm_directory (data_path_a, error_chmod);
 	auto rockdb_data_path = data_path_a / "rocksdb";
-	std::filesystem::remove_all (rockdb_data_path);
+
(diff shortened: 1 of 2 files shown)


nano/secure/ledger.cpp
```diff
@@ -1280,7 +1280,12 @@ bool nano::ledger::migrate_lmdb_to_rocksdb (std::filesystem::path const & data_p
 	boost::system::error_code error_chmod;
 	nano::set_secure_perm_directory (data_path_a, error_chmod);
 	auto rockdb_data_path = data_path_a / "rocksdb";
-	std::filesystem::remove_all (rockdb_data_path);
+
+	if (std::filesystem::exists (rockdb_data_path))
+	{
+		logger.error (nano::log::type::ledger, "Existing Rocksdb folder found in '{}'. Please remove it and try again.", rockdb_data_path.string ());
```
pwojcikdev: They stylize the name as "RocksDB"

pwojcikdev (dismissed):

nano/secure/ledger.cpp
RickiNano: I've force pushed a commit that fixes the naming

PR #4798, opened 2024-11-30
Optimize Vote Request Behavior During Bootstrapping

gr0vity-dev:
With bounded backlog there's a need to minimize the gap between checked and cemented blocks especially during bootstrapping phase. Currently, elections stay passive for 5 seconds by default, during which we don't request votes. However, during bootstrapping, we need to request votes to confirm blocks.

This PR implements a heuristic to identify if a block is actively being voted on. If there are no recent votes in the vote cache for a block, it likely indicates we're in bootstrapping phase, and we should request votes immediately rather than waiting for the default 5-second passive period.

The election will transition to active state immediately if the block's hash is not found in the vote cache.

### DIFF
--- a/nano/node/active_elections.cpp
+++ b/nano/node/active_elections.cpp
@@ -416,6 +416,14 @@ nano::election_insertion_result nano::active_elections::insert (std::shared_ptr<
 			debug_assert (count_by_behavior[result.election->behavior ()] >= 0);
 			count_by_behavior[result.election->behavior ()]++;
 
+			// Skip passive phase for blocks without cached votes to avoid bootstrap delays
+			bool active_immediately = false;
+			if (node.vote_cache.contains (hash))
+			{
+				result.election->transition_active ();
+				active_immediately = true;
+			}
+
 			node.stats.inc (nano::stat::type::active_elections, nano::stat::detail::started);
 			node.stats.inc (nano::stat::type::active_elections_started, to_stat_detail (election_behavior_a));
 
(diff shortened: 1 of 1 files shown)


nano/node/active_elections.cpp
```diff
@@ -396,6 +396,7 @@ nano::election_insertion_result nano::active_elections::insert (std::shared_ptr<
 		return result;
 	}
 
+	auto active_state = false;
```
pwojcikdev: `bool active_state = ...` this can be moved to inner scope

nano/node/active_elections.cpp
```diff
@@ -416,16 +417,24 @@ nano::election_insertion_result nano::active_elections::insert (std::shared_ptr<
...
+
 			node.stats.inc (nano::stat::type::active_elections, nano::stat::detail::started);
 			node.stats.inc (nano::stat::type::active_elections_started, to_stat_detail (election_behavior_a));
 
 			node.logger.trace (nano::log::type::active_elections, nano::log::detail::active_started,
 			nano::log::arg{ "behavior", election_behavior_a },
 			nano::log::arg{ "election", result.election });
 
-			node.logger.debug (nano::log::type::active_elections, "Started new election for block: {} (behavior: {})",
+			node.logger.debug (nano::log::type::active_elections, "Started new election for block: {} (behavior: {}, active: {})",
```
pwojcikdev: For clarity this should say something like `active immediately: {}`

nano/node/active_elections.cpp
```diff
@@ -416,16 +416,25 @@ nano::election_insertion_result nano::active_elections::insert (std::shared_ptr<
 			debug_assert (count_by_behavior[result.election->behavior ()] >= 0);
 			count_by_behavior[result.election->behavior ()]++;
 
+			// Skip passive phase for blocks without cached votes to avoid bootstrap delays
+			bool active_immediately = false;
+			if (node.vote_cache.find (hash).empty ())
```
pwojcikdev: This should use vote_cache.contains(hash) to avoid an unnecessary allocation

gr0vity-dev: Good catch!

pwojcikdev (approved):

PR #4799, opened 2024-11-30
Make election behavior mutable to support dynamic behaviour transition

gr0vity-dev:
Problem:
- Elections are currently immutable and can't be transitioned to priority status
- Causing delays for optimistic elections rescheduled by priority scheduler as the next Vote broadcast is fixed to 15-second intervals. (by `network.vote_broadcast_interval`)

Changes:
- Modifies election behavior to be mutable, allowing elections (optimistic/hinted/...) to be transitioned to priority status
- Added ability to trigger immediate vote broadcasts on transition without waiting for the next 15-second round

Benefits:
- It would allow for more aggressive use of optimistic elections (to narrow the gap between checked and cemented during bootstrapping) 
  - currently configured to accounts with gap_threshold of 32 could be reduced to 4
- Maintain quick response times for live traffic by allowing election behavior changes.

### DIFF
--- a/nano/node/active_elections.cpp
+++ b/nano/node/active_elections.cpp
@@ -435,6 +435,23 @@ nano::election_insertion_result nano::active_elections::insert (std::shared_ptr<
 	else
 	{
 		result.election = existing->election;
+
+		// Upgrade to priority election to enable immediate vote broadcasting.
+		auto previous_behavior = result.election->behavior ();
+		if (election_behavior_a == nano::election_behavior::priority && previous_behavior != nano::election_behavior::priority)
+		{
+			bool transitioned = result.election->transition_priority ();
+			if (transitioned)
+			{
+				count_by_behavior[previous_behavior]--;
+				count_by_behavior[election_behavior_a]++;
+				node.stats.inc (nano::stat::type::active_elections, nano::stat::detail::transition_priority);
+			}
+			else
+			{
+				node.stats.inc (nano::stat::type::active_elections, nano::stat::detail::transition_priority_failed);
+			}
+		}
 	}
 
 	lock.unlock ();
--- a/nano/node/election.cpp
+++ b/nano/node/election.cpp
@@ -183,6 +183,25 @@ void nano::election::transition_active ()
 	state_change (nano::election_state::passive, nano::election_state::active);
 }
 
+bool nano::election::transition_priority ()
+{
+	nano::lock_guard<nano::mutex> guard{ mutex };
+
+	if (behavior_m == nano::election_behavior::priority || behavior_m == nano::election_behavior::manual)
+	{
+		return false;
+	}
+
+	behavior_m = nano::election_behavior::priority;
+	last_vote = std::chrono::steady_clock::time_point{}; // allow new outgoing votes immediately
+
+	node.logger.debug (nano::log::type::election, "Transitioned election behavior to priority from {} for root: {}",
+	to_string (behavior_m),
+	qualified_root.to_string ());
+
+	return true;
+}
+
 void nano::election::cancel ()
 {
 	nano::lock_guard<nano::mutex> guard{ mutex };
--- a/nano/node/election.cpp
+++ b/nano/node/election.cpp
@@ -27,7 +27,7 @@ nano::election::election (nano::node & node_a, std::shared_ptr<nano::block> cons
 	live_vote_action (live_vote_action_a),
 	node (node_a),
 	behavior_m (election_behavior_a),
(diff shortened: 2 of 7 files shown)


pwojcikdev:
Can you include a test or two that ensures this works? So I imagine and election would be manually inserted via `active.insert ( ..., election_behavior::optimistic )` and then upgraded to priority.

gr0vity-dev:
I added a simple testcase to validate the behavior transitioning works.

Ideally I'd like to also make sure the `vote_broadcast_interval` is reset and a rep votes again immediately after transitioning.
My problem is to create a network with one rep that doesn't have enough weight to confirm an election.
It seems :
- we need to start from genesis and transfer some weight over to one rep
- by creating a  send + open block and confirm these.  

If that approach is right I can create another testcase that makes sure voting happens as intended.

nano/node/active_elections.cpp
```diff
@@ -287,8 +287,9 @@ void nano::active_elections::cleanup_election (nano::unique_lock<nano::mutex> &
 	debug_assert (!election->confirmed () || recently_confirmed.exists (election->qualified_root));
 
 	// Keep track of election count by election type
-	debug_assert (count_by_behavior[election->behavior ()] > 0);
-	count_by_behavior[election->behavior ()]--;
+	auto election_behavior = election->behavior_locked (lock_a);
```
pwojcikdev: This change seems wrong, I'm surprised it compiles. 1) it's passing `active_elections`s lock into the election, 2) behavioud_locked should be private function of the election, not available outside

nano/node/election.cpp
```diff
@@ -776,8 +776,15 @@ std::vector<nano::vote_with_weight_info> nano::election::votes_with_weight () co
 	return result;
 }
 
+nano::election_behavior nano::election::behavior_locked (nano::unique_lock<nano::mutex> & lock_a) const
+{
+	debug_assert (lock_a.owns_lock ());
```
pwojcikdev: Missing assert for` !mutex.try_lock ()`, it doesn't check that the lock passed specifically is locking this election mutex

pwojcikdev: It doesn't use the lock anyway, so it's unnecessary, only the assert for `!mutex.try_lock ()` should be here.

nano/node/active_elections.cpp
```diff
@@ -435,6 +436,19 @@ nano::election_insertion_result nano::active_elections::insert (std::shared_ptr<
...
 	{
 		result.election = existing->election;
+
+		// Upgrade to priority election to enable immediate vote broadcasting.
+		auto previous_behavior = result.election->behavior ();
+		if (election_behavior_a == nano::election_behavior::priority && previous_behavior != nano::election_behavior::priority)
+		{
+			count_by_behavior[previous_behavior]--;
+			count_by_behavior[election_behavior_a]++;
+			result.election->transition_priority ();
```
pwojcikdev: Is this guaranteed to always succeed? This probably works but is unnecessary leaving space for subtle bugs. Should be something like:
```
bool transitioned = result.election->transition_priority ();
if (transitioned)
{
// success, update counts, stats, log
}
else
{
// failed for some reason, inc failed stats
}
```

pwojcikdev:
Regarding testing the voting behaviour, I don't think having two nodes is necessary. Each election should keep counters for number of vote broadcasts and confirmation requests and those can be queried directly. There is still a need to setup a rep key for node with weight higher than primary rep threshold but not enough to confirm blocks on its own, but that's rather simple.

nano/node/active_elections.cpp
gr0vity-dev: You are right.  I'll revert the change, and look more closely at why TSAN was complaining.

gr0vity-dev:
> Each election should keep counters for number of vote broadcasts and confirmation requests
My issue is
-  by default the testcase node is not voting and doesn't have a rep wallet in it. 
- So it doesn't broadcast a vote.
https://github.com/nanocurrency/nano-node/blob/a4614215d030cfcb3b39799fd4ded2f72bbb3f63/nano/node/election.cpp#L429
I wanted to add a rep wallet, but
-  if I use the genesis key, blocks are confirmed by its single vote 
- So the optimistic election is not activated anymore...

pwojcikdev:
>if I use the genesis key, blocks are confirmed by its single vote
So the optimistic election is not activated anymore...


You should be able to setup a representative with eg. 111K nano weight which will be generating votes but not be able to confirm elections. Am I missing something in this setup?

gr0vity-dev:
Yes exactly. My initial question was how to best do it. I didn't see see a helper method to set this up.
The best I found was :
 - Start from genesis and transfer some weight over to one rep by creating a send + open block and force confirm these blocks.

This requires quite a bit of code. If that's the right way, I'll go for it. Just wanted to make sure this is the right approach.

pwojcikdev:
There is `nano::test::setup_rep ( ... )` if that's helpful.

pwojcikdev (approved):

PR #4801, opened 2024-12-03
Do not add processed live traffic into priorities

gr0vity-dev:
I noticed that processed live traffic is inserted into priorities.
My understanding is that processed blocks from live traffic don't need further bootstrapping and they should not be a bootstrap priority.

### DIFF
--- a/nano/node/bootstrap/bootstrap_service.cpp
+++ b/nano/node/bootstrap/bootstrap_service.cpp
@@ -254,17 +254,21 @@ void nano::bootstrap_service::inspect (secure::transaction const & tx, nano::blo
 	{
 		case nano::block_status::progress:
 		{
-			const auto account = block.account ();
-
-			// If we've inserted any block in to an account, unmark it as blocked
-			accounts.unblock (account);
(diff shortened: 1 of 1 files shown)


clemahieu:
I think a full bootstrap profile would give an answer.

gr0vity-dev:
When the live traffic block is processed (with result == progress). Doesn't this mean that the block was inserted into the ledger and the account chain is fully in sync with the network ?

I'm running a live bootstrap with a modified version, having among others this  What should I be looking at ?

Screenshot 2024-12-03 at 14 42 07

pwojcikdev:
I don't think profiling the from scratch bootstrap is a proper way to judge an impact of this. Rather this should affect how node handles live spam, as this should reduce bootstrap traffic in this case. If spam tests look OK we should merge this.

pwojcikdev:
We looked at the impact of this on live spam and it looks like it improves things slightly and avoids inserting unnecessary accounts into the bootstrapping set.

pwojcikdev (approved):

PR #4802, opened 2024-12-04
Fix timing issue in active_elections.fork_replacement_tally

gr0vity-dev:
This testcase fails in ~10% of cases on my machine with
```
  max_blocks
    Which is: 10
  election->blocks ().size ()
    Which is: 1
```
By waiting for election insertion for the first block, before processing forks, this timing issue is eliminated.

### DIFF
--- a/nano/core_test/active_elections.cpp
+++ b/nano/core_test/active_elections.cpp
@@ -901,6 +901,7 @@ TEST (active_elections, fork_replacement_tally)
 					 .build ();
 
 	// Forks without votes
+	std::shared_ptr<nano::election> election;
 	for (auto i (0); i < reps_count; i++)
 	{
 		auto fork = builder.make_block ()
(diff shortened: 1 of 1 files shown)


nano/core_test/active_elections.cpp
```diff
@@ -913,10 +914,15 @@ TEST (active_elections, fork_replacement_tally)
 					.work (*system.work.generate (latest))
 					.build ();
 		node1.process_active (fork);
+
+		// Assert election exists on first iteration
+		if (i == 0)
+		{
+			ASSERT_TIMELY (5s, election = node1.active.election (fork->qualified_root ()));
```
pwojcikdev: Shouldn't this be true for every iteration, not just the first?

gr0vity-dev: Yes, we could make the assert more general for every iteration.

pwojcikdev (approved):

PR #4805, opened 2024-12-06
Fix ledger_priority.sequential_blocks testcase

gr0vity-dev:
I saw this testcase failing on the windows github runner.

Different account chains are not guaranteed to have the same priority timestamp. 
The opening could happen much later than the sending. 
If we add a 1 second delay between processing `send1` and `open1`, the testcase previously always failed.
The opening timestamp of "key1 account" can be equal or greater than the matching send timestamp of "genesis account".

### DIFF
--- a/nano/core_test/ledger_priority.cpp
+++ b/nano/core_test/ledger_priority.cpp
@@ -191,9 +191,10 @@ TEST (ledger_priority, sequential_blocks)
 	ASSERT_EQ (100, priority_balance2);
 	ASSERT_EQ (100, priority_balance3); // Max of current (50) and previous (100)
 
-	ASSERT_EQ (nano::dev::genesis->sideband ().timestamp, timestamp1);
-	ASSERT_EQ (send1->sideband ().timestamp, timestamp2);
-	ASSERT_EQ (open->sideband ().timestamp, timestamp3);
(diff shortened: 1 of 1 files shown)


nano/core_test/ledger_priority.cpp
```diff
@@ -191,9 +191,10 @@ TEST (ledger_priority, sequential_blocks)
 	ASSERT_EQ (100, priority_balance2);
 	ASSERT_EQ (100, priority_balance3); // Max of current (50) and previous (100)
 
-	ASSERT_EQ (nano::dev::genesis->sideband ().timestamp, timestamp1);
-	ASSERT_EQ (send1->sideband ().timestamp, timestamp2);
-	ASSERT_EQ (open->sideband ().timestamp, timestamp3);
+	ASSERT_EQ (nano::dev::genesis->sideband ().timestamp, timestamp1); // genesis account
+	// Opening account must have equal or greater timestamp than sending counterpart.
+	ASSERT_TRUE (timestamp2 >= send1->sideband ().timestamp); // key1 account
```
pwojcikdev: This can use ASSERT_GE (...)

pwojcikdev:
Good fix. I added this test recently, it was generated by Claude. I didn't notice it was comparing timestamps in this way.

pwojcikdev (approved):

PR #4807, opened 2024-12-07
Enable debug logging for core tests

gr0vity-dev:
Failures on GitHub runners are sometimes hard to reproduce locally. Adding detailed logging helps analyze test failures directly from CI logs.

- Enables NANO_LOG_STATS and debug level logging for core tests across all CI platforms
- Helps diagnose intermittent test failures by providing more detailed execution traces

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -35,6 +35,9 @@ jobs:
         if: steps.build.outcome == 'success' && (success() || failure())
         run: ../ci/tests/run-core-tests.sh
         working-directory: build
+        env:
+          NANO_LOG_STATS: 1
+          NANO_LOG: debug
 
       - name: RPC Tests
         if: steps.build.outcome == 'success' && (success() || failure())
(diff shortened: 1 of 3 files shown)


pwojcikdev:
I wouldn't enable this, the amount of logs generated is just enormous to the point where it's not possible to easily see which tests failed anymore because github truncates viewable log size. An approach similar to gtest-parallel might be better where only log output from failed tests is printed.

gr0vity-dev:
Oh right. when just running `core_test` without `gtest-parallel`, the output is always shown and will be huge. That's not what we want indeed!
Should we switch to `gtest-parallel` instead.
It will hide the output for all passing tests and just show failing tests. 
We could set the `--worker=1` so tests don't actually run in parallel

pwojcikdev:
Gtest-parallel was added as a submodule in `submodules/gtest-parallel` so perhaps the script can invoke it from there

pwojcikdev:
For some reason it no longer is able to show the sanitizer report https://github.com/nanocurrency/nano-node/actions/runs/12215820961/job/34078250929?pr=4807 It detects the problem, but the report is empty. Not sure what's going on here.

PR #4828, opened 2025-01-19
Database backend configuration

RickiNano:
In the current implementation, the database backend is determined by the setting of `RocksDb.Enable` config setting which is not very intuitive.
This PR replaces that setting with a new `database_backend` setting. It can be set to either `lmdb` or `rocksdb`. If not set, it defaults to `lmdb`.
To make it easier for node operators that are currently using RocksDb, the node will notify the operator with an error message if the config file has the old `RocksDb.Enable` set to true and `database_backend` set to lmdb (for example when using an old config file that does not have the new database_backend property). The error message can be removed in future versions when operators have migrated to use the new `database_backend` type
This PR also makes it easier to add more backends in the future.

### DIFF
--- a/nano/node/nodeconfig.cpp
+++ b/nano/node/nodeconfig.cpp
@@ -701,18 +701,18 @@ nano::database_backend nano::node_config::get_config_backend (nano::tomlconfig &
 
 nano::database_backend nano::node_config::get_default_backend ()
 {
-	auto backend_str = nano::env::get<std::string> ("BACKEND");
+	auto backend_str = nano::env::get<std::string> ("NANO_BACKEND");
 
 	if (backend_str.has_value ())
 	{
 		auto backend = deserialize_database_backend (backend_str.value ());
 		if (backend.has_value ())
 		{
-			std::cerr << "Database backend overridden by BACKEND environment variable: " << *backend_str << std::endl;
+			std::cerr << "Database backend overridden by NANO_BACKEND environment variable: " << *backend_str << std::endl;
 			return backend.value ();
 		}
 
-		std::cerr << "Unknown database backend in BACKEND environment variable: " << *backend_str << std::endl;
+		std::cerr << "Unknown database backend in NANO_BACKEND environment variable: " << *backend_str << std::endl;
 	}
 	return database_backend::lmdb;
 }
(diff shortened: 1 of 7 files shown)


pwojcikdev:
I agree with doing the backend configuration this way, but it should not throw an error if legacy `rocksdb.enable` is set. There should be a warning, but the node should check for this case and use rocksdb, there is no reason not to do that. It's not a good user experience otherwise.

The only case where we should throw an error is if there is a mismatch between `rocksdb.enable` and `database_backend`, ie. one is explicitly set to rocksdb and the other to lmdb.

RickiNano:
The way we currently handle configuration errors are by logging an error an exit the application, so this is not really any different, imo.
If we don't let the users know that they have to update the config file, they will keep using the deprecated `RocksDb.Enable` without even knowing it. We have to get rid of this option in the config file somehow and use the new setting instead.

I only see two solutions to this:
1. Notify the user and let them update the config themselves (as in this PR)
2. Automatically update the config file on launch

Option 2 is definitely a possibility. We already have the `--update_config` option that migrates the current configuration into the latest one. I could change that function so that it also migrates the `RocksDb.Enable` into the appropriate `database_backend` setting. I feel that it is a bit controversial to change the users config file without permission, though.

nano/node/nodeconfig.cpp
```diff
@@ -659,6 +661,62 @@ void nano::node_config::deserialize_address (std::string const & entry_a, std::v
...
+		toml.get_error ().set ("Unknown database_backend type: " + backend_str);
+		return nano::database_backend::lmdb;
+	}
+	// Default to LMDB
+	return nano::database_backend::lmdb;
+}
+
+nano::database_backend nano::node_config::get_default_backend ()
+{
+	auto backend_str = nano::env::get<std::string> ("BACKEND");
```
Exxenoz: Why don't you name it `DATABASE_BACKEND`?

RickiNano: > Why don't you name it `DATABASE_BACKEND`?

It's an existing environment key. I didn't name it

pwojcikdev: NANO_BACKEND would fit better with our convention of prefixing all nano specific env variables with NANO_

RickiNano: I have renamed the environment variable to NANO_BACKEND now

PR #4835, opened 2025-01-27
Change fanout to scale by std::log rather than std::sqrt

clemahieu:


pwojcikdev:
I rebased it on top of develop after merging https://github.com/nanocurrency/nano-node/pull/4837 which seems to have fixed it. Looks good.

nano/node/network.cpp
```diff
@@ -430,7 +430,8 @@ std::deque<std::shared_ptr<nano::transport::channel>> nano::network::list_non_pr
 // Simulating with sqrt_broadcast_simulate shows we only need to broadcast to sqrt(total_peers) random peers in order to successfully publish to everyone with high probability
```
not-Ian, line 430: Is this comment out of date after this change?

`sqrt(total_peers)` should now be `log(total_peers)`, no?

PR #4840, opened 2025-02-05
Fix vote_spacing logic 

gr0vity-dev:
The `votable()` method had incorrect logic that allowed voting on recently
voted blocks when it should have prevented them. Fix inverts the logic to:
- Prevent repeated votes on the same root/hash pairs within spacing window
- Allow immediate voting on different blocks with same root (competing forks)

Additional changes:
- Remove vote spacing check from vote_generator reply()
- Update vote_spacing to track latest fork votes per root
- Fix tests to reflect correct spacing behavior

### DIFF
--- a/nano/node/vote_generator.cpp
+++ b/nano/node/vote_generator.cpp
@@ -230,6 +230,7 @@ void nano::vote_generator::reply (nano::unique_lock<nano::mutex> & lock_a, reque
 			auto const & [root, hash] = *i;
 			if (std::find (roots.begin (), roots.end (), root) == roots.end ())
 			{
+				// Vote spacing skipped for replies - allows all peers to request same vote
 				roots.push_back (root);
 				hashes.push_back (hash);
 			}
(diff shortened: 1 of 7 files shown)


nano/node/vote_generator.cpp
```diff
@@ -230,15 +230,8 @@ void nano::vote_generator::reply (nano::unique_lock<nano::mutex> & lock_a, reque
...
-				if (spacing.votable (root, hash))
-				{
-					roots.push_back (root);
-					hashes.push_back (hash);
-				}
-				else
-				{
-					stats.inc (nano::stat::type::vote_generator, nano::stat::detail::generator_spacing);
-				}
+				roots.push_back (root);
```
pwojcikdev, line 234: There should be a comment here explaining that we don't want to use spacing for on-demand vote replies. I think it's a good modification but evidently it wasn't that obvious since we are only changing it now.

pwojcikdev:
I modified the vote spacing class to track votes by qualified root, which simplified the code and it eliminated the need to handle ambiguity.

PR #4841, opened 2025-02-09
RPC: Include optional `linked_account` field in account_history, block_info, and blocks_info RPCs

Exxenoz:
This PR adds an optional parameter `include_linked_account` to the account_history, block_info, and blocks_info RPC endpoints.

When enabled, the response will include a new field `linked_account`, which returns the linked account associated with the block (or "0" if no linked account exists/was found e.g. due to pruning). The related tests have been updated accordingly.

**Issues resolved**
This PR is an extension to PR #4833

PR #4842, opened 2025-02-09
Ledger: Use representative for change blocks and epoch signer for epoch blocks in linked_account

Exxenoz:
- For change blocks, `ledger::linked_account` now returns the block's representative
- For epoch blocks, it returns the corresponding epoch signer

**Issues resolved**
This PR is an extension to PR #4841

This PR simplifies the detection of related accounts for block explorer applications. As a result, the following logic can be completely removed, reducing complexity:

```
switch (historyEntry.type) {
  case "change":
	relatedAccount = historyEntry.representative;
	break;
  case "open":
  case "receive":
	relatedAccount = historyEntry.account;
	break;
  case "send":
	relatedAccount = historyEntry.destination;
	break;
  case "state":
	switch (historyEntry.subtype) {
	  case "change":
		relatedAccount = historyEntry.representative;
		break;
	  case "epoch": // ToDo: Find epoch signer
		relatedAccount = historyEntry.account;
		break;
	  case "receive":
	  case "send":
		relatedAccount = historyEntry.account;
		break;
	}
	break;
}
```

### DIFF
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -463,6 +463,177 @@ TEST (websocket, confirmation_options_votes)
 	}
 }
 
+TEST (websocket, confirmation_options_linked_account)
+{
+	nano::test::system system;
+	nano::node_config config = system.default_config ();
+	config.websocket_config.enabled = true;
+	config.websocket_config.port = system.get_available_port ();
+	auto node1 (system.add_node (config));
+
+	std::atomic<bool> ack_ready{ false };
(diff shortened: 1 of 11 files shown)


Exxenoz:
It's about this commit: https://github.com/nanocurrency/nano-node/pull/4842/commits/6de112d1a15d3f83edc731b71e33ad874d02d05a

Can someone tell me if this makes sense? If so, I'll do a rebase and fix the web socket tests; otherwise the PR can be closed.

PR #4843, opened 2025-02-11
Updated bootstrap weights

RickiNano:
Updated bootstrap weights from 11-02-2025 for live and beta network.
Also modified the python script to insert the date that the new weight file is created


RickiNano:
Weights updated again 25-02-2025

PR #4855, opened 2025-03-04
Fix websocket confirmation options handling

gr0vity-dev:
This PR fixes an issue with the websocket confirmation message handling. Previously, the system would cache messages based solely on the `include_block` option, causing other subscriber options (like `include_linked_account` and `include_sideband_info`) to be ignored if another subscriber had already created a cached message.

Changes made:
- Removed the message caching mechanism that was causing options to be ignored
- Each subscriber now receives a message built specifically with their configuration options
- Simplified parameter naming for better readability
- Made options handling more consistent across the codebase
- Added a failing testcase that is passing by the removal of the cache

These changes ensure that all subscribers receive confirmation messages that correctly respect their individual option settings.

### DIFF
--- a/nano/core_test/websocket.cpp
+++ b/nano/core_test/websocket.cpp
@@ -1146,3 +1146,107 @@ TEST (websocket, new_unconfirmed_block)
 	ASSERT_EQ ("state", message_contents.get<std::string> ("type"));
 	ASSERT_EQ ("send", message_contents.get<std::string> ("subtype"));
 }
+
+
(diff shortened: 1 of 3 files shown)


pwojcikdev (approved):

gr0vity-dev:
This was previous code + testcase only:

Screenshot 2025-03-04 at 20 33 25

[Here is the run](https://github.com/nanocurrency/nano-node/actions/runs/13660701620/job/38190921681)

PR #4863, opened 2025-03-17
Remove unnecessary com file for Win

RickiNano:
The build process is creating both nano_wallet.exe and nano_wallet.com for the Windows build. The dot com file is not needed. 
Running the gui wallet from the console doesn't make much sense. If you want to use the console you should be using nano_node.exe
It is also possible to add arguments to nano_wallet.exe if you launch it from a console.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -844,18 +844,6 @@ if(NANO_GUI OR RAIBLOCKS_GUI)
 
   target_link_libraries(nano_wallet Boost::process rpc node qt)
 
-  if(WIN32)
-    target_link_libraries(nano_wallet Qt5::WinExtras)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -844,18 +844,6 @@ if(NANO_GUI OR RAIBLOCKS_GUI)
 
   target_link_libraries(nano_wallet Boost::process rpc node qt)
 
-  if(WIN32)
-    target_link_libraries(nano_wallet Qt5::WinExtras)
```
pwojcikdev, line 848: We most likely don't want to remove `Qt5::WinExtras` from the normal `nano_wallet` target.

PR #4928, opened 2025-07-12
Update RocksDb to V10.3.4

RickiNano:
Updates from V9.7.2 to V10.3.4
I have been testing this version on live network for a week without any issues.


pwojcikdev:
Can we upgrade straight to v10.4.2? I'm not familiar with rocksdb release practices, but 10.3.4 is not tagged anywhere (https://github.com/facebook/rocksdb/releases)

RickiNano:
I did try with 10.4.2, but it triggers some kind of build error on Windows Github runners. I can't reproduce that error locally. It builds fine on my windows pc and it uses same msvc version.
Here is the failing unit test with 10.4.2:
https://github.com/RickiNano/nano-node/actions/runs/16235666353

It may have something to do with the API change in 10.4:
`Change NewExternalTableFactory to return a unique_ptr instead of shared_ptr.`
https://github.com/facebook/rocksdb/blob/main/HISTORY.md#public-api-changes-1

RickiNano:
@pwojcikdev 
I found the cause of the failing windows build on V10.4.2. It seems that they have unreachable code in auto_skip_compressor.cc.
I made a second PR for 10.4.2 that suppresses this warning
Not sure if we should go with 10.4.2 or just 10.3.4

PR #4961, opened 2025-11-12
Disable LMDB read ahead on Windows

RickiNano:
LMDB's read-ahead feature works well on Linux and Mac, but causes excessive memory usage on Windows, consuming all 
available RAM and making the node slow to a crawl after a few hours.
This PR disables read ahead for windows but maintains it on Mac and Linux

### DIFF
--- a/nano/store/lmdb/lmdb_env.cpp
+++ b/nano/store/lmdb/lmdb_env.cpp
@@ -40,9 +40,12 @@ void nano::store::lmdb::env::init (std::filesystem::path const & path_a, nano::s
 
 			// It seems if there's ever more threads than mdb_env_set_maxreaders has read slots available, we get failures on transaction creation unless MDB_NOTLS is specified
 			// This can happen if something like 256 io_threads are specified in the node config
-			// MDB_NORDAHEAD will allow platforms that support it to load the DB in memory as needed.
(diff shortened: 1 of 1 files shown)


qwahzi:
@RickiNano - Were you able to do any performance tests for read-ahead on Mac/Linux? I'm wondering if there's any benefit to keeping read-ahead enabled on Mac/Linux, since I'm guessing Nano is pretty heavy on random reads (except maybe when bootstrapping)?

> The excessive reads were caused by OS read-ahead. By default, when we read a page, the OS will prefetch the pages following that page into memory. However, this is usually undesirable in a DB, since we are usually doing random read. So we need to disable read-ahead to get optimal performance for random read.

https://stackoverflow.com/a/49354913

> [MDB_NORDAHEAD](http://www.lmdb.tech/doc/group__mdb__env.html#ga0fcdaf42f0c45749a2dff11ee322303f) Turn off readahead. Most operating systems perform readahead on read requests by default. This option turns it off if the OS supports it. Turning it off may help random read performance when the DB is larger than RAM and system RAM is full. The option is not implemented on Windows.

http://www.lmdb.tech/doc/group__mdb.html

RickiNano:
@qwahzi I don't own a Mac or Linux computer so I can't test that unfortunately.

PR #4974, opened 2025-12-16
Remove stale test network

RickiNano:
The test network isn't being actively used anymore. There are also no active nodes on the network anymore
This PR removes the test network


dsiganos:
The test network was used to create private test networks. The live and beta network do not provide the necessary hooks to create adhoc private test networks. If that is still the case then the test network should not be removed.

gr0vity-dev:
The nano CI test framework https://ct.bnano.info also relies on the test network and the possibility to use ENV variables to define custom genesis blocks and PoW thresholds among other

RickiNano:
I wasn't aware of this. I assumed it was unused since there there were no peers on the test network anymore.
I'll close this pr then

PR #4981, opened 2026-01-05
Extended ledger

Exxenoz:
## Summary

This branch adds optional extended ledger indices to speed up RPC paths that previously required expensive ledger scans. The indices are persisted in the store, maintained by ledger runtime hooks, and can be populated or dropped through dedicated CLI commands.

The extended ledger indices added are:

- `account_block_by_height`
  - Indexes `{ account, height } -> block_hash`.
  - Speeds up `account_history`, especially offset-based lookups.

- `account_delegator_by_weight`
  - Indexes delegators by `{ representative, weight, delegator }`.
  - Speeds up `delegators` and `delegators_count`.

- `account_receivable_by_amount`
  - Indexes receivables by `{ account, amount, send_block_hash }`.
  - Speeds up `receivable` and `accounts_receivable` when sorted results are requested.

- `receive_block_by_send_block`
  - Maps a send/source block hash to the receive/open block that consumed it.
  - Speeds up `blocks_info` when `receive_hash=true`.

The branch also adds shared extended ledger lifecycle handling, including persisted initialized flags, startup population, drop behavior, runtime add/remove/update hooks, and logging for index population and maintenance.

**Note:** when the extended `account_delegator_by_weight` index is enabled, the `delegators` RPC now returns delegators in descending weight order instead of delegator account order.
**Note:** when the extended `account_receivable_by_amount` index is enabled, the `accounts_receivable` and `receivable` RPCs use that index only for sorted requests (`sorting: true`).

## CLI

New maintenance commands:

- `--populate_extended_ledger_indices`
  - Builds missing extended ledger indices for an existing ledger.

- `--drop_extended_ledger_indices`
  - Drops all extended ledger index data and marks the indices disabled.

The extended ledger index feature is controlled through the node configuration option `extended_ledger_index`.

## Tests

Adds focused core tests for extended ledger index population, runtime maintenance, rollback/delete behavior, drop behavior, disabled reopen behavior, and direct index-table verification.

The extended ledger index test cases were created with GPT-5.5 assistance (using Codex), and all changes were manually reviewed by me.


Exxenoz:
## Extended ledger indices populate performance observations

I tested the initial populate of all four extended ledger index tables with both LMDB and RocksDB.

### Test hardware

| Component | Specification |
|---|---|
| CPU | AMD Ryzen 9 5950X, 16 cores, 4.9GHz |
| SSD | Samsung SSD 970 EVO 1 TB |
| RAM | 32 GB DDR5 |

### Populate results

| Backend | `txn.refresh()` during populate | Populate duration | Ledger size before populate | Ledger size after populate | Size increase |
|---|---:|---:|---:|---:|---:|
| LMDB | No | ~3h 40m | 134 GB | 177 GB | +43 GB |
| LMDB | Yes | ~4h 50m | 134 GB | 218 GB | +84 GB |
| RocksDB | Yes | ~1h 30m | 88.3 GB | 126 GB | +37.7 GB |

### Open question

The main open question is how we should handle the LMDB behavior.

The additional LMDB file growth when using `txn.refresh()` could be caused by random insert patterns and LMDB page splits/page reuse behavior, but I am not sure whether this is expected or whether it indicates a bug or pathological behavior in the current populate strategy.

Should this be investigated further and potentially optimized for LMDB?

pwojcikdev:
I worked on this branch a bit with the latest fable model and here's a summary of my changes:


## Follow-up: RPC integration review, semantics fixes, and test suite

This comment covers the changes since the original PR description: a critical review of the
extended-index RPC integration, the resulting semantics decisions and fixes, the new `delegators`
cursor API, and the test suites pinning all of it.

### Problems found and how they were resolved

1. **`delegators` pagination lost entries on indexed nodes.** The indexed path iterated by
   weight but treated `start` as an account-number filter, so a client paging with account
   cursors would never see some delegators. Fixed: `start` is now a real cursor in weight order.
2. **Unifying the two `delegators` modes was rejected.** Making the non-indexed path
   weight-ordered would require collecting and sorting the whole account table on every call —
   too expensive and a DoS risk. They are now two documented APIs; the legacy path is unchanged.
3. **`accounts_receivable` with `sorting` returned the wrong entries.** It sorted only the first
   `count` entries found in hash order instead of the largest ones. It now scans everything,
   sorts, then cuts — matching `receivable` and the indexed path.
4. **Equal amounts were ordered differently per path.** The sort left ties hash-ascending, the
   index returns them hash-descending; with `count`/`offset` this changed which entries were
   returned. Both paths now use amount descending, hash descending on ties.
5. **`receivable` returned nothing for `sorting` + `offset` without `count`** due to an internal
   overflow. It now returns the tail after the offset.
6. **Account-based cursors drift.** Resuming by account re-reads its current weight, so weight
   changes between pages skip or repeat entries. Solved with the explicit `<weight>:<account>`
   cursor form and the `next` response field.

### RPC: `delegators`

The behavior of `delegators` depends on whether the extended delegator index is active on the
node (see *Index persistence* below). `delegators_count` is unaffected.

**Legacy API (index not active) — unchanged.** Delegators are enumerated in **ascending account
order**, at most `count` (default 1024) entries, `threshold` filters by minimum balance, and
`start` is an exclusive account iteration point (results begin strictly after it; the account
does not need to exist).

**Indexed API (index active).** Delegators are enumerated in **descending weight order**
(descending account on equal weights).

| Parameter | Type | Description |
|---|---|---|
| `account` | account | The representative to enumerate delegators for |
| `count` | number, optional (default 1024) | Maximum entries per response |
| `threshold` | amount, optional | Only delegators with weight **≥ threshold** are returned (inclusive lower cutoff) |
| `start` | string, optional | Resume position; accepts the forms below |

`start` forms:

| Form | Meaning | Equal entry included? |
|---|---|---|
| `<weight>:<account>` | Explicit cursor position; resumes strictly below it in (weight, account) order. Not affected by ledger changes; the referenced entry does not need to exist | No (exclusive) |
| `nano_…` or `:nano_…` | The position is that account's **current** weight. Unknown account → `account_not_found` error. Subject to drift if the weight changes between requests | No (exclusive) |
| `<weight>:` | Inclusive upper bound: results start at the given weight, **including its whole tie group** (the missing account side is treated as a virtual maximum) | Yes (inclusive) |
| `:` or otherwise malformed | `bad_account_number` error | — |

**Response.** In addition to the `delegators` map (account → weight, in response order), the
response contains a top-level `"next"` field **while more entries remain**. Its value is the
`<weight>:<account>` cursor of the last returned entry — pass it as `start` to fetch the next
page; when `"next"` is absent, enumeration is complete. The cursor can also be constructed from
the last response row (`value + ":" + key`).

```json
request:  { "action": "delegators", "account": "nano_rep…", "count": 2 }
response: {
  "delegators": {
    "nano_delegator_a…": "300…",
    "nano_delegator_b…": "300…"
  },
  "next": "300…:nano_delegator_b…"
}
```

Note for docs: sending a cursor-form `start` to a node without the index active fails with
`bad_account_number` (the legacy API only accepts accounts), and JSON responses across the two
modes differ in ordering by design.

### RPC: `receivable` / `accounts_receivable`

These behave **identically with and without the index** (the index is purely an accelerator
here); responses are byte-equal and covered by parity tests.

- With `sorting=true`, the response contains the top-`count` receivables by **amount descending,
  send-hash descending on equal amounts**, after `offset` entries are skipped
  (`receivable` only). This fixes the previous `accounts_receivable` behavior of sorting only
  the first `count` entries found in hash order.
- Confirmation filtering (`include_only_confirmed`, default true) is applied **before** `offset`
  and `count`: unconfirmed receivables never consume the offset.
- `offset` with `count` unset now returns the full tail after the offset (previously empty due
  to the overflow in item 6).
- Preserved quirk (unsorted queries only, pinned by test): `offset` consumes confirmed entries
  in hash order **before** the threshold filter, so a confirmed entry below `threshold` still
  eats offset.

### Index persistence vs. configuration — doc-critical

Enabling `node.extended_ledger_index` in the config populates any missing index tables at
startup and persists **per-index flags in the ledger meta table**. From then on the persisted
flags are authoritative, not the config option:

- Removing the option later does **not** remove the indices — they continue to be maintained on
  every processed block, and the RPCs above keep their indexed behavior.
- Removal requires `nano_node --drop_extended_ledger_indices` (offline); population without the
  config option is available via `--populate_extended_ledger_indices`.
- Consequence for the docs: **the `delegators` API switch is keyed to the persisted flag, not to
  the config file.** An operator who ever enabled the option serves the new API until the
  indices are explicitly dropped. Pruning remains incompatible (startup refuses with a pointer
  to the drop command).

### Implementation notes

- New store-view primitive `account_delegator_by_weight_view::rlower_bound (txn, key)` — reverse
  iterator strictly below an arbitrary composite key; this is what makes explicit cursors seek
  without any ledger lookup. Non-trivial view methods gained doc comments.
- `receivable`/`accounts_receivable` were deduplicated into a shared `collect_receivables`
  helper (one emit path, three modes: indexed walk, sorted fallback, unsorted scan);
  `delegators` was split into `parse_delegators_start` / `resolve_delegators_start` /
  `collect_delegators_indexed` / `collect_delegators_legacy`. Handlers are now parameter
  parsing + delegation.

### Tests

- `core_test/ledger_extended_index.cpp` (20 tests): store-level ordering/bounds incl. endianness
  canaries and `rlower_bound`; flag lifecycle (fresh enable, persistence across reopen,
  populate-existing, single-index populate with mixed-flag maintenance, drop + repopulate,
  stale-row clearing); incremental maintenance and rollbacks for state/legacy/epoch blocks;
  `find_block_hash_by_height` / `find_receive_block_by_send_hash` indexed-vs-walk parity.
- `rpc_test`: `delegators_legacy_order` pins the legacy API; `delegators_indexed_weight_order`
  and `delegators_indexed_cursor` pin the new one (all cursor forms, `next` round-trip,
  malformed inputs, and cursor stability under weight mutation). Receivable parity suite:
  `receivable_sorting_tie_parity`, `receivable_sorting_confirmed_offset_parity`,
  `receivable_unsorted_offset_before_threshold`, `accounts_receivable_sorting_top_parity` —
  each runs with the index on and off and asserts identical responses. Indexed-only tests carry
  an `indexed` name segment.

PR #5015, opened 2026-01-27
Update to C++ 23

RickiNano:


pwojcikdev:
Thanks for the PR. This upgrade will get merged, just not at this time. The last time we discussed upgrading to C++23 we decided it's best to postpone it after V29 release, especially since we don't need any of the C++23 features. There are services that require building the node from source and we don't want to force upgrades (eg. ubuntu 22 > 24, 22 is still LTS) at this time.

Issue #4053, opened 2023-01-14
Split Block table


clemahieu:
Currently the block table maps block_hash -> block data. Each block is keyed on random data which causes frequent database page splits, hurting performance and causing internal fragmentation. For highest performance and lowest internal fragmentation, the database should append data whenever possible.

Splitting the block table in to one table mapping block_hash -> index and another table mapping index -> block_data significantly reduces the amount of fragmented data. Initial tests show disk space saving up to 50%, a speed up for non-seeking disks, and a significant speed up for seeking disks (HDDs).

Organizing blocks this way has the side effect of providing a topology sort of blocks which can be used for incremental ledger snapshots.

My1:
> Initial tests show disk space saving up to 50%, a speed up for non-seeking disks, and a significant speed up for seeking disks (HDDs).

this is awesome to be honest especially for the smaller nodes.



> Organizing blocks this way has the side effect of providing a topology sort of blocks which can be used for incremental ledger snapshots.

this is also very cool.

qwahzi:
Related draft/work-in-progress PR for review & feedback: https://github.com/nanocurrency/nano-node/pull/4295

PR #5024, opened 2026-02-06
Migrate successor field from sideband to dedicated table

RickiNano:
This PR moves the successor field from the sideband data and into a new database table named successors. It is a rewrite of #4967 

It extracts the successor field from block_sideband into a new successor database table (block_hash -> successor_hash)
Block records are now immutable once written, eliminating predecessor read-modify-write cycles on every block insertion

This is phase 1 of the block table restructuring described in #4053 . It is a prerequisite for the full block table split which enables MDB_APPEND insertion giving 50% disk space savings

Migration (upgrade_v24_to_v25)

Rewrites every block entry in the database in a single pass:

  1. For each block, locates the 32-byte successor hash (always the first field in the old sideband serialization, immediately after the block data).
  2. If the successor is non-zero, inserts it into the new successor table.
  3. Reconstructs the block record without the 32 successor bytes and writes it back.
  4. Uses batch processing with transaction.refresh() (batch size: 2 for dev, 250K for production).
  5. Idempotent on crash recovery

I have tested with both LMDB and RocksDB
Claude code has been assisting in writing the code


pwojcikdev:
This looks good. The only sketchy part is the manual migration by copying buffers. A better approach would be to freeze block sideband class as it was in previous version and use that to deserialize > reserialize with newest version, similar to how it was done for:
```
/**
 * This is a snapshot of the account_info table at v22 which needs to be read for the v22 to v23 upgrade
 */
class account_info_v22 final
{ ...
```

nano/store/ledger_store.cpp
```diff
@@ -364,6 +374,93 @@ void ledger_store::upgrade_v23_to_v24 ()
...
+			{
+				// Read key as block_hash
+				nano::block_hash block_hash;
+				release_assert (key.size () == sizeof (block_hash));
+				std::memcpy (block_hash.bytes.data (), key.data (), sizeof (block_hash));
+
+				auto status = backend.put (transaction, nano::store::table::successor, block_hash, successor_hash);
+				backend.release_assert_success (status);
+			}
+
```
pwojcikdev: > Idempotent on crash recovery

How does this handle the case where block sideband table was only partially rewritten? Part of the entries could be using newer format already.

RickiNano: Good catch. I cleared the successor table to make it start over, but forgot about the blocks that had already been migrated. 
I have pushed a fix for this so the the successor table is not cleared and blocks are checked if they are in the new or old format, and skip new ones. It now resumes after a crash, skipping converted blocks. If a successor already exists but the block data is not modified, it will just overwrite the successor entry with the same data
I also added a progress indicator like in the old version

pwojcikdev: To avoid problems with the sideband versioning and make this indepotent, let's not modify the block sideband table. This upgrade should just copy successors to the extrnal table. We'll migrate block sideband table later together with some mechanism that allows for easier upgrades, since this will be needed for other improvements anyway

RickiNano: Just to be sure that I understand you correctly:
You want the migration to copy all the successors to the new successor table, but leave the successor field in the block sideband unchanged? After migration, the node will read and insert new succesors in the successor table. The successor data in the sideband is not updated.
This would cause 32 bytes of redundant data for each block in the ledger.
I can make the changes, I just wanted to verify that I'm understanding you correctly

pwojcikdev: Correct. The block sideband table will be updated as part of another upgrade.

RickiNano: PR updated

nano/store/ledger_store.cpp
```diff
@@ -364,6 +378,64 @@ void ledger_store::upgrade_v23_to_v24 ()
...
+				auto status = backend.put (transaction, nano::store::table::successor, hash, block_w_sideband.sideband.successor);
+				backend.release_assert_success (status);
+			}
+
+			processed++;
+			if (processed % batch_size == 0)
+			{
+				double const percentage = total_blocks > 0 ? (static_cast<double> (processed) / total_blocks * 100.0) : 0.0;
+				logger.info (nano::log::type::ledger_upgrade, "Processed {} blocks ({:.1f}%)", processed, percentage);
+				transaction.refresh ();
```
Exxenoz, line 427: I previously ran into issues with `transaction.refresh()` and LMDB when inserting large amounts of data. When using refresh, the affected tables consistently ended up consuming significantly more disk space than when inserting the same data without refreshing the transaction.

Because of that, I wanted to ask: after the successor upgrade, does the resulting ledger size roughly match the expected size on your end, or did you also notice a few extra gigabytes of overhead?

I initially suspected LMDB page splits caused by inserting unsorted keys, since LMDB needs to split pages more often in that case. This should not apply here, because the successor inserts block hashes in sorted order. I just wanted to double-check that assumption and make sure no unexpected LMDB behavior is at play here.

pwojcikdev: I noticed the same issue when working on a separate block sideband upgrade (adding a topological index to speed up bootstrap). Because the write pattern was effectively random, it was both slow and caused the ledger to grow to ~200GB.

@RickiNano could you run it against a copy of the current ledger and let us know how much the ledger grows afterward?

Even if we run the upgrade with LMDB in append mode, I suspect the DB will still grow over time once normal operation resumes, since subsequent inserts may not be append-only.

If the numbers look reasonable, we should merge it.

RickiNano: @pwojcikdev 
I just upgraded a recent LMDB production ledger now using this branch.
215345749 blocks in ledger.
Upgrade time: 15 minutes
Size before upgrade 127.860.494.336
Size after upgrade 141.541.904.384 byte
The new successor table adds 32 + 32 bytes for each block (but only if it has a successor). 
32 bytes can be removed from each block in the sideband (even if they don't have a successor). That is about 7 GB.
When/if block split is introduced the total size will be reduced by approx 50% for LMDB

PR #5034, opened 2026-02-22
POC: Split blocks table and remove successor from sideband

RickiNano:
POC of split block table and MDB_APPEND inserts.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,7 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(cmake --build:*)"
+    ]
+  }
+}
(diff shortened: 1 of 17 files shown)


Exxenoz:
Can we apply the same approach to the account table? This would reduce the required storage in the pending, rep_weight, confirmation_height and extended ledger tables.

RickiNano:
> Can we apply the same approach to the account table? This would reduce the required storage in the pending, rep_weight, confirmation_height and extended ledger tables.

Maybe. But saving would be small and it also adds complexity because reading an account will then require two reads after the table is split. The value size should also be larger than the key size to get any savings.
Blocks table is the most obvious candidate for splitting. The value size is quite big, and the table itself has a lot of rows. The main trade-off is the double read pattern.

PR #5035, opened 2026-02-22
Extend telemetry data

RickiNano:
Extends the telemetry message with three new fields: database_backend (uint8), confirmation_latency_ms (uint32), and bootstrap_status (uint8), adding 6 bytes to the payload (150 → 156 bytes)
Introduces versioned size constants (size_v1 = 150, size = 156)

**Forward compatibility**: An old node receives a 156-byte payload but only knows about 150 bytes of structured fields. The extra 6 bytes are read into unknown_data. Signature validation succeeds because serialize_without_signature appends unknown_data to the verification bytes, matching what the new node signed. No code changes are needed on old nodes

**Backward compatibility**: A new node receives a 150-byte payload. The deserialize check payload_length_a >= size (156) is false, so the three new fields remain at their zero-initialized defaults. data_size_ is set to 150 (the payload length), so serialize_without_signature skips the new fields, producing the same bytes the old node signed. Signature validation succeeds.

**Partial payload protection**: If a payload length falls between 151–155 bytes (more than v1 but less than the full v2 size), the new fields are not read

### DIFF
--- a/nano/messages/telemetry.cpp
+++ b/nano/messages/telemetry.cpp
@@ -67,9 +67,9 @@ telemetry_ack::telemetry_ack (nano::network_constants const & constants, telemet
 	message (constants, message_type::telemetry_ack),
 	data (telemetry_data_a)
 {
-	debug_assert (telemetry_data::size + telemetry_data_a.unknown_data.size () <= message_header::telemetry_size_mask.to_ulong ()); // Maximum size the mask allows
+	debug_assert (telemetry_data_a.serialized_size () <= message_header::telemetry_size_mask.to_ulong ()); // Maximum size the mask allows
 	header.extensions &= ~message_header::telemetry_size_mask;
-	header.extensions |= std::bitset<16> (static_cast<unsigned long long> (telemetry_data::size) + telemetry_data_a.unknown_data.size ());
+	header.extensions |= std::bitset<16> (static_cast<unsigned long long> (telemetry_data_a.serialized_size ()));
 }
 
 void telemetry_ack::serialize (nano::stream & stream_a) const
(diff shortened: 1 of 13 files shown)


nano/messages/telemetry.cpp
```diff
@@ -219,6 +233,9 @@ nano::error telemetry_data::serialize_json (nano::jsonconfig & json, bool ignore
 	json.put ("maker", maker);
 	json.put ("timestamp", std::chrono::duration_cast<std::chrono::milliseconds> (timestamp.time_since_epoch ()).count ());
 	json.put ("active_difficulty", nano::to_string_hex (active_difficulty));
+	json.put ("database_backend", database_backend);
+	json.put ("confirmation_latency_ms", confirmation_latency_ms);
+	json.put ("bootstrap_status", bootstrap_status);
```
pwojcikdev: Telemetry message should keep track which version it is, it seems currently to be done via `if (data_size_ > size_v1)` check, but an explicit `version > telemetry_v1` would be better. And only include these fields in json if they are actually present. It's mostly to avoid using 0 for confirmation_latency_ms if it's not present which I think is ugly.

nano/messages/telemetry.hpp
```diff
@@ -51,10 +71,14 @@ class telemetry_data
...
+	static auto constexpr size_v1 = sizeof (signature) + sizeof (node_id) + sizeof (block_count) + sizeof (cemented_count) + sizeof (unchecked_count) + sizeof (account_count) + sizeof (bandwidth_cap) + sizeof (peer_count) + sizeof (protocol_version) + sizeof (uptime) + sizeof (genesis_block) + sizeof (major_version) + sizeof (minor_version) + sizeof (patch_version) + sizeof (pre_release_version) + sizeof (maker) + sizeof (uint64_t) + sizeof (active_difficulty);
+	static auto constexpr size = size_v1 + sizeof (database_backend) + sizeof (confirmation_latency_ms) + sizeof (bootstrap_status);
 	static auto constexpr latest_size = size; // This needs to be updated for each new telemetry version
 
+	uint16_t serialized_size () const;
+
 private:
+	uint16_t data_size_{ latest_size };
```
pwojcikdev: Fields like this tend to cause problems. This can be done without it.

PR #5063, opened 2026-04-13
Enable sccache for Linux and Macos CI builds

RickiNano:
Github offers 10 GB cache storage and we are currently not using it.
This PR introduces compile caching of CI runs on unit tests and code sanitizer running Linux and MacOs. Most of the submodules in this project changes rarely (like Boost, Rocksdb, spdlog etc).
MSVC caching needs some additional changes, so caching on Windows builds are skipped in this PR. I can look into that later.
The first run will populate the cache and give zero cache hits. Subsequent pushes will use cached versions for any unchanged code. 
Each CI job will print cache hit rates to the logs.
From my testing, this reduces build times from 20-30 minutes to 3-10 minutes
https://github.com/mozilla/sccache

### DIFF
--- a/.github/workflows/code_sanitizers.yml
+++ b/.github/workflows/code_sanitizers.yml
@@ -53,7 +53,7 @@ jobs:
         id: build
         run: ci/build-tests.sh
 
-      - name: sccache stats
+      - name: Build cache stats
         if: always()
         run: sccache --show-stats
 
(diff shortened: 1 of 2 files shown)


.github/workflows/code_sanitizers.yml
```diff
@@ -42,10 +46,17 @@ jobs:
...
         run: sudo -E ci/prepare/linux/prepare.sh
 
+      - name: Setup sccache
+        uses: mozilla-actions/sccache-action@v0.0.9
+
       - name: Build Tests
         id: build
         run: ci/build-tests.sh
 
+      - name: sccache stats
```
pwojcikdev: Can we name these build steps "Build cache stats" for consistency

PR #5100, opened 2026-07-02
Fix false-positive 'Incorrect source epoch' in --validate_blocks

clemahieu:
## Summary

`--validate_blocks` flags every receive block where the receiving account had already epoch-upgraded past the sender's epoch, even though this is normal, valid ledger state.

## Problem

The check in `entry.cpp` compares `sideband.source_epoch` to the receiving block's own epoch with equality:

```cpp
if (sideband.source_epoch != node->ledger.version (*block))
```

At write time (`ledger_processor.cpp`), a receiving block's own epoch is set to `max(account's prior epoch, source_epoch)`. So whenever a receiving account is already on a higher epoch than the send it's receiving, `source_epoch < block epoch` by design — not an error. The equality check produces a false positive on every such block.

## Fix

Replace the equality check with the correct monotonic invariant: a block's epoch should never be *less than* the epoch of a send it received.

```cpp
if (node->ledger.version (*block) < sideband.source_epoch)
```

## Test plan

- [x] Ran `--validate_blocks` against a full live-network ledger copy before the fix: 219 blocks flagged, all receives where `source_epoch=epoch_0` and block epoch was `epoch_1`/`epoch_2`
- [x] Applied the fix, rebuilt, reran against the same ledger data: 0 errors, `Validation status: Ok`

### DIFF
--- a/nano/nano_node/entry.cpp
+++ b/nano/nano_node/entry.cpp
@@ -1596,9 +1596,13 @@ int main (int argc, char * const * argv)
 						print_error_message (boost::str (boost::format ("Incorrect sideband block details for block %1%\n") % hash.to_string ()));
 					}
 					// Check link epoch version
+					// A receiving block's own epoch is max(account's prior epoch, source_epoch)
+					// (see ledger_processor.cpp), so source_epoch <= block epoch is expected and
+					// normal whenever the receiving account had already epoch-upgraded past the
+					// sender. Only a block epoch *below* its source_epoch is an actual inconsistency.
 					if (sideband.details.is_receive && (!node->ledger.pruning || !node->store.pruned.exists (transaction, block->source ())))
 					{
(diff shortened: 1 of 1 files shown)


pwojcikdev (approved):

nano/nano_node/entry.cpp
```diff
@@ -1596,9 +1596,13 @@ int main (int argc, char * const * argv)
 						print_error_message (boost::str (boost::format ("Incorrect sideband block details for block %1%\n") % hash.to_string ()));
 					}
 					// Check link epoch version
+					// A receiving block's own epoch is max(account's prior epoch, source_epoch)
+					// (see ledger_processor.cpp), so source_epoch <= block epoch is expected and
```
pwojcikdev, line 1600: Referring to ledger_processor.cpp will get out of sync over time. Opus models like this overly verbose comment style, unless explicitly corrected in claude.md.

PR #5103, opened 2026-07-08
Move signature verification out of the write transaction and parallelize it

RickiNano:
Block signature verification currently runs serially inside the ledger write transaction: the block processor opens a write tx per batch and each ledger.process() call does its ed25519 checks while holding it. This PR moves that CPU work out of the transaction and onto a small worker pool, so batches are pre-validated in parallel before the write tx is opened.

Benefits

- Higher block throughput. Serial signature work now spread across up to 4 threads
- Shorter write transaction hold times. The write tx now spends its time on actual store reads/writes instead of signature validation
- Zero behavioral change. Results are bit-identical to today: same checks, same order of ledger application, same result codes, same notification order. A duplicate block with a bad signature still returns old, forks and rollbacks are untouched.

How it works

- block_context gains a nano::signature_verification result (unknown / valid / valid_epoch / invalid), defaulting to unknown.
- process_batch() fans the batch out across a dedicated nano::thread_pool between next_batch() and tx_begin_write(). Workers verify state blocks (against the account owner first, then the epoch signer for epoch links — the same order the ledger processor uses) and open blocks. Legacy send/receive/change blocks need a frontier lookup to determine the signer, so they stay unknown and are verified inline as today; they are effectively extinct on the live network.
- ledger.process() accepts the verification result (defaulting to unknown, so all existing callers keep today's behavior). Inside ledger_processor, a new validate_signature() helper consumes the pre-computed result where it is conclusive and falls back to validate_message() where it isn't. Debug builds re-verify every skipped check via debug_assert.
- Pre-verified invalid blocks are still rejected at the existing check site inside the transaction (not short-circuited earlier), preserving today's result-code precedence.

Co-authored with Claude Fable

### DIFF
--- a/nano/core_test/block_processor.cpp
+++ b/nano/core_test/block_processor.cpp
@@ -397,6 +397,141 @@ TEST (block_processor, add_many_empty)
 	ASSERT_EQ (node.ledger.block_count (), 1);
 }
 
+// Parallel signature pre-validation processes a valid block and increments the blocks_verified stat
+TEST (block_processor, verification)
+{
+	nano::test::system system;
+
+	nano::node_config node_config;
+	node_config.block_processor->verification_threads = 2;
+	auto & node = *system.add_node (node_config);
+
+	nano::keypair key;
+	nano::block_builder builder;
+	auto send = builder
+				.state ()
+				.account (nano::dev::genesis_key.pub)
+				.previous (nano::dev::genesis->hash ())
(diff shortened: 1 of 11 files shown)


RickiNano:
I need to rebase this branch on develop to fix the conflicts

RickiNano:
Rebased and ready for review @pwojcikdev

PR #5107, opened 2026-07-14
Add peering (networking-only) node mode

clemahieu:
Introduce a --peering_only mode for nodes that only participate in the peer-to-peer network and do not maintain a ledger. These are intended for the DNS bootstrap entries, where a node must be reachable and gossip peers but has no need to spend IO/CPU keeping the ledger current.

A peering node keeps a genesis-only store and disables the ledger-consuming subsystems (schedulers, active elections, bootstrap sync and server, vote/cementing/backlog processing, rebroadcasters, voting), and drops incoming publish/confirm_ack at the message boundary so no blocks or votes are processed.

To keep peering nodes from degrading peers that do serve a ledger, a node advertises a no_ledger capability in the v3 handshake (and the nano_node_peering telemetry maker). Peers that understand the capability exclude peering nodes from bootstrap candidate selection and rep-crawler targets, while still gossiping them normally so they remain useful for peer discovery.

The capability describes the property (no ledger); the mode name describes the role (peering). "light" was avoided as it collides with both pruned nodes and light wallets.

### DIFF
--- a/nano/messages/telemetry.hpp
+++ b/nano/messages/telemetry.hpp
@@ -18,7 +18,7 @@ enum class telemetry_maker : uint8_t
 {
 	nf_node = 0,
 	nf_pruned_node = 1,
-	nano_node_light = 2,
+	nf_peering_node = 2,
 	rs_nano = 3
 };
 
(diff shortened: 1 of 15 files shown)


nano/messages/telemetry.hpp
```diff
@@ -18,7 +18,7 @@ enum class telemetry_maker : uint8_t
 {
 	nf_node = 0,
 	nf_pruned_node = 1,
-	nano_node_light = 2,
+	nano_node_peering = 2,
```
pwojcikdev: There is strange inconsistency with naming here, what's the difference between nf and nano_node prefix? Perhaps we want to call it nf_peering_node. Or rename all enums to not have nf_ prefix

pwojcikdev (dismissed):
The code looks good. Two non breaking suggestions: change the cli option to --peering_only and add a testcase demonstrating that nodes A and B find each other by connecting only to a peering node C initially.

PR #5114, opened 2026-07-22
Add precompiled headers for RocksDB and kernel libraries (~33% faster build)

Omotola:
## Summary
Adds an opt-in `-DNANO_PCH` switch that introduces precompiled headers for the bundled **RocksDB** static library and for nano's first-party kernel libraries, cutting a full clean `nano_node` rebuild by **~33%**.

## Motivation
In a clean `nano_node` build, a large share of frontend parse time is spent on the bundled RocksDB library. Several of its core headers (`rocksdb/options.h`, `rocksdb/listener.h`, `port/win/port_win.h`, `db_impl.h`, `column_family.h`, ...) are included across most of RocksDB's translation units, yet the RocksDB static library ships with no precompiled header. `<chrono>` and nano's spdlog/fmt-based logging headers are the next most expensive.

## Change
New `-DNANO_PCH=ON` option (default OFF) that precompiles:

- **`nano/rocksdb_pch.hpp`** on the `rocksdb` target — its stable backbone headers (`rocksdb/options.h`, `rocksdb/listener.h`, `port/port.h`) plus heavy STL. This is the single largest lever.
- **`nano/nano_pch.hpp`** on nano's first-party libs (`nano_lib`, `secure`, `node`, `nano_store`, `nano_messages`, `rpc`, `nano_node`) — stable STL plus `boost.asio`, `boost.multiprecision`, `fmt`, `spdlog`.

Both headers contain only standard-library and stable third-party headers — no volatile project headers — so they rarely invalidate.

## Measured impact
Full clean rebuild, MSVC (VS 2022, 14.44), Release, `/m:16`, 5 runs each:

| Variant | Mean (s) | StdDev (s) |
|---|---|---|
| baseline (no PCH) | 595.62 | 10.05 |
| `NANO_PCH=ON` | 397.10 | 1.40 |

**~198 s faster (33.33%)**, statistically significant (post < baseline − stddev).

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -495,6 +495,19 @@ else()
 endif()
 add_subdirectory(submodules/rocksdb EXCLUDE_FROM_ALL)
 
+# Precompiled headers derived from a vcperf /jsonAnalysis trace of a clean
+# nano_node build. Enabled with -DNANO_PCH=ON. RocksDB's own backbone headers
+# (rocksdb/options.h, rocksdb/listener.h, port/win/port_win.h) are the hottest
+# included files in the whole build, and the bundled RocksDB static library
+# (~700 translation units) ships with no PCH -- so precompiling that backbone
+# once is the single largest build-time win.
+option(NANO_PCH "Enable precompiled headers derived from vcperf analysis" OFF)
+if(NANO_PCH AND COMMAND target_precompile_headers AND TARGET rocksdb)
+  target_precompile_headers(
+    rocksdb PRIVATE
+    "$<$<COMPILE_LANGUAGE:CXX>:${PROJECT_SOURCE_DIR}/nano/rocksdb_pch.hpp>")
+endif()
+
 # cpptoml
 include_directories(submodules/cpptoml/include)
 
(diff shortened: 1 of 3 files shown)


pwojcikdev:
There is already a precompiled header option `NANO_PRECOMPILED_HEADERS` in cmake, the existing pch.hpp is more complete. Not sure why would we need rocksdb pch? it's a fixed dependency, doesn't get rebuilt during incremental builds

Omotola:
@pwojcikdev  both fair points, I can drop the nano_pch.hpp and rebase onto the existing `NANO_PRECOMPILED_HEADERS` .

For RocksDB: agreed it won't help incremental builds, but it targets clean/CI rebuilds . 9 of the 15 hottest headers in a clean build are RocksDB's  internals, with no PCH today. That PCH is most of my clean-rebuild win (595.6s → 397.1s, ~33%). 

I can scope my PR to just the RocksDB PCH and frame it as a clean-build optimization if that's okay with you.

PR #5132, opened 2026-08-11
Fix representatives count limit off-by-one

Helemen7:
The non-sorting branch of the `representatives` RPC returned one entry more than the requested `count`, because the break check ran after adding the entry to the response. The sorting branch already enforces the limit exactly.

This surfaced during review of #5124 (Copilot flagged it); keeping it out of that PR per review feedback. No change to the sorting path, and the default (no `count`) behavior is unaffected.

Verification: build OK locally, `rpc.representatives_count_limit` added, 7 rpc representatives/count tests pass, clang-format check passes.

### DIFF
--- a/nano/node/json_handler.cpp
+++ b/nano/node/json_handler.cpp
@@ -3787,11 +3787,11 @@ void nano::json_handler::representatives ()
 				{
 					continue;
 				}
-				representatives.put (account.to_account (), amount.convert_to<std::string> ());
-				if (representatives.size () > count)
+				if (representatives.size () >= count)
 				{
 					break;
 				}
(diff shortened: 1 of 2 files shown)


Helemen7:
all failures are the same pre-existing upstream break, as rpc.cpp:736 accesses node->wallets.items, which is now private in wallet.hpp:403 after a recent refactor. This exact error reproduces on a clean checkout of develop without any changes on my end. my diff only touches json_handler.cpp and the new RPC test, and format/cmake-format/code_inspector all pass. should turn green once the wallet refactor fix lands upstream.

pwojcikdev:
Look good, the build error was because of some cross-pr incompatibility that landed on develop.

PR #5144, opened 2026-08-18
Reduce sccache bloat from sanitizer builds

RickiNano:
Sanitizer jobs are using 72% of the 10 GB cache available.

The cause is Debug defaulting to `-g`: full DWARF on top of sanitizer instrumentation dominates object size, and it buys nothing in CI where we only read the sanitizer's own stack traces. Since sccache caches object files directly, smaller objects mean smaller cache entries.

`ci/build.sh` now passes `-g1` when `SANITIZER` is set. Line tables and function names are kept, so reports stay symbolized with accurate file:line; variable and type info is dropped. Non-sanitizer builds are untouched.

Using the g1 flag could free up a few GB that could be used for Windows build caching

Separately, sccache was intermittently reporting 0% cache hits because the server timed out mid-job. Setting the idle timeout to "0" on the sanitizer and unit test jobs fixes this.

Note: `SCCACHE_GHA_RW_MODE` is `READ_ONLY` outside develop, so this PR's sanitizer jobs will be a full cache miss. 
The cache only shrinks once this lands on develop.


pwojcikdev:
> DWARF is the debug-information format that compilers like GCC and Clang put into object files and executables so debuggers/symbolizers can translate machine addresses into useful source-level information.
> For example, without debug info a sanitizer might report:
> 0x7f1234abcd
> With DWARF, it can turn that into something like:
> foo::parse() at src/foo.cpp:127
> DWARF can contain much more than just file and line numbers. Full -g debug info can include local variable names and locations, struct/class definitions, types, function parameters, scopes, inline-function information, and source line mappings. In a large C++ build, all of that can make .o files dramatically larger.
> That's why your change helps:
> -g → essentially full DWARF debug information.
> -g1 → much smaller/minimal debug information, primarily what is needed for stack traces and source locations.
> no -g → little/no source-level debug information.
> So in your sanitizer CI case, you mostly need:
> Address → function → source_file.cpp:line
> You don't need the richer information that lets someone open GDB and inspect things like:
> p my_local_variable
> ptype MyHugeTemplateType

not sure if we want to miss full debug info for santizer builds. we do full gdb dump if errors are found and that includes info about all the local variables etc, which are useful when tracking corruption bugs.

PR #5149, opened 2026-08-20
CI: Cache Windows builds

RickiNano:
Enables sccache for Windows CI builds, matching the caching already in place for Linux and macOS. The default MSVC generator ignores compiler launchers, so this switches Windows to Ninja and fixes what that switch broke along the way:

- RocksDB's generic ASM language was claiming .asm ahead of MASM, breaking Boost.Context — pinned CMAKE_ASM_SOURCE_FILE_EXTENSIONS before RocksDB enables the language.
- RocksDB hardcodes /Zi and /d2Zi+, which defeat compiler caching (shared PDB; sccache rejects /d2Zi+ outright). A new cmake/Modules/RocksDBMsvcCache.cmake, hooked in via CMAKE_PROJECT_rocksdb_INCLUDE, strips those after RocksDB sets its own flags; the rest of the project switches /Zi//ZI → /Z7 when a compiler launcher is active.
- Git's link.exe was shadowing the MSVC linker once bash put /usr/bin first on PATH — removed during prepare.

Tested on my own fork with warmed cache:: Windows lmdb (https://github.com/RickiNano/nano-node/actions/runs/32287902103) and Windows rocksdb (https://github.com/RickiNano/nano-node/actions/runs/32287902103) both went green with a 97%+ cache hit rate, and build times dropped ~30% or approximately 20 minutes.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -152,6 +152,15 @@ jobs:
       BUILD_TYPE: ${{ matrix.RELEASE && 'RelWithDebInfo' || 'Debug' }}
       TEST_USE_ROCKSDB: ${{ matrix.BACKEND == 'rocksdb' && '1' || '0' }}
       DEADLINE_SCALE_FACTOR: ${{ matrix.BACKEND == 'rocksdb' && '2' || '1' }}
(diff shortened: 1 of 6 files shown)


RickiNano:
Rebased on develop 21/09 to fix merge conflicts

